PR #128, opened 2026-04-22
feat(slash): /pr robust + /init local handler

Jiahui-Gu:
## Summary

Slash batch D: harden `/pr` against preflight failures and convert `/init` from a useless pass-through into a real local handler.

### `/pr`
Already a fully-fledged client handler (preflight in `electron/pr.ts` covers `no-cwd`, `not-git`, `no-gh`, `on-default-branch`, `dirty-tree`, `no-commits`; `PrFlowProvider` owns the dialog + `gh pr create` + CI poll). The handler-side and main-process unit tests already cover every branch.

This PR extends `scripts/probe-slash-pr.mjs` to **also exercise the renderer-side error paths end-to-end**: after the happy path, the probe swaps `window.agentory.pr.preflight` to return `no-gh`, then `dirty-tree`, then `on-default-branch`, and asserts that:
- the `PrDialog` does NOT open, and
- a status banner with the right title surfaces in chat.

### `/init`
Used to be `passThrough: true`. In `--input-format stream-json` mode `claude.exe` silently drops most slash commands, so typing `/init` in the GUI did nothing. We can't replicate the CLI's interactive `/init` (which kicks off a model-driven repo scan) without re-implementing the prompt loop in stream-json mode, so the GUI version drops a **sensible CLAUDE.md template** into the current session's cwd via the existing `memory:*` IPC. Users edit it manually or feed it back to the agent ("fill this in based on the codebase").

Behavior:
- No cwd on the session -> warn ("No working directory set"), no write
- `CLAUDE.md` already exists at `<cwd>/CLAUDE.md` -> warn with the path, no write
- Otherwise -> write template, info status with the path
- Memory IPC missing or write error -> error block with details

### Registry
- `/init`: `category: 'built-in', passThrough: true` -> `category: 'client', passThrough: false`.
- `attach('init', handleInit)` added in `handlers.ts`.

## Test plan

- [x] `npm run typecheck` (renderer + electron) clean
- [x] `npm test` -> 553/553 passing, including 5 new `/init` handler tests and updated registry-shape assertion
- [x] `node --check` on both probe scripts
- [ ] Manual verification of probes against a fresh dev server (deferred — current shared dev server runs the parent worktree, not this branch; will re-run after merge to `working`)

### DIFF
--- a/scripts/probe-slash-init.mjs
+++ b/scripts/probe-slash-init.mjs
@@ -1,0 +1,255 @@
+// Probe: /init slash command end-to-end (renderer-only).
+//
+// /init became a client handler in batch D — it used to pass through to
+// claude.exe, which silently dropped it under stream-json. The new behaviour
+// drops a CLAUDE.md template into the current session's cwd via the
+// memory:* IPC, surfaces a status block on success, and warns instead of
+// overwriting when the file already exists.
+//
+// We stub window.agentory.memory before app boot so we don't need a real
+// filesystem path, then drive the UI:
+//
+//   1. Type /init + Enter when no cwd is set    → "No working directory set" warn
+//   2. Configure cwd, set memory.exists()=false → /init writes the template,
+//      success status appears, write captured the template body
+//   3. Now memory.exists()=true                 → /init warns "already exists",
+//      write is NOT called again
+//
+// Usage:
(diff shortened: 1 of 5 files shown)


Jiahui-Gu:
Closing — product decision to drop all slash commands except /clear and /compact. Skill-based commands will replace this surface. No fault of the worker; spec changed mid-flight.

PR #131, opened 2026-04-22
feat(slash): /cost local render + /compact handler

Jiahui-Gu:
## Summary

- **/cost**: already wired to render per-session token / cost / turn counters from aggregated `result` frames. Added `scripts/probe-slash-cost.mjs` that injects a synthetic `result` frame at the captured `onAgentEvent` listener and verifies the rendered status banner contains the formatted tokens (`12k in / 678 out`), cost (`\$0.023`), and turn count.
- **/compact**: stays pass-through. Per the [Agent SDK slash-commands docs](https://code.claude.com/docs/en/agent-sdk/slash-commands), the SDK natively handles `/compact` when sent as a prompt in stream-json input mode and emits a `system { subtype: 'compact_boundary' }` frame on completion. Our `systemBlocks()` translator already renders that frame as a status banner ("Conversation compacted (manual) — Compacted N → M tokens in T ms"). Added `scripts/probe-slash-compact.mjs` that verifies (a) the literal `/compact` text reaches `api.agentSend`, and (b) the injected `compact_boundary` frame produces the expected banner.
- Updated comments in `registry.ts` / `handlers.ts` — the prior text claimed claude.exe silently drops slash commands in stream-json mode, which contradicts the docs and is no longer accurate.

## Why no client-side compact handler

Researched the SDK first. The CLI owns the summarisation prompt and emits the `compact_boundary` boundary frame; reimplementing that in the renderer would duplicate logic and drift from upstream. Pass-through is the correct answer; the probe is the proof.

## Test plan

- [x] `npm test` — 41 files / 548 tests passing (existing /cost handler unit tests already cover the formatter / placeholder paths)
- [x] `AGENTORY_DEV_PORT=4192 npm run dev:web` + `node scripts/probe-slash-cost.mjs` → OK (banner shows `1 turn · 12k in / 678 out · \$0.023`)
- [x] `node scripts/probe-slash-compact.mjs` → OK (`/compact` forwarded literally; compact_boundary frame produces "Conversation compacted (manual)" banner with pre/post tokens + duration)

### DIFF
--- a/scripts/probe-slash-compact.mjs
+++ b/scripts/probe-slash-compact.mjs
@@ -1,0 +1,203 @@
+// Probe: /compact end-to-end (renderer-only).
+//
+// Verifies the pass-through path for /compact:
+//   1. Stub window.agentory so we can capture both the registered
+//      onAgentEvent listener AND every agentSend call.
+//   2. Type `/compact` + Enter — the dispatcher returns 'pass-through' so
+//      InputBar runs the normal send path, which forwards the literal text
+//      to claude.exe via api.agentSend.
+//   3. Confirm the captured agentSend call carries `/compact`. (This is the
+//      whole point of pass-through: the SDK accepts /compact in stream-json
+//      mode per https://code.claude.com/docs/en/agent-sdk/slash-commands.)
+//   4. Fire a fake `system { subtype: 'compact_boundary' }` frame at the
+//      captured listener — this is exactly what the SDK emits when /compact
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
Closing — product decision: drop /cost; only /clear and /compact remain. /compact-related findings (SDK does support it in stream-json per docs) and the probe will be reused in the upcoming slash-overhaul PR. Thanks for the docs research.

PR #168, opened 2026-04-23
fix(probe): poll focus-restore precondition to remove flake

Jiahui-Gu:
## Summary

- The a11y focus-restore probe (`scripts/probe-e2e-a11y-focus-restore.mjs`) introduced in PR #164 had an intermittent precondition failure on Windows (~3/13 runs) reporting `expected session sA to be focused before opening dialog, got null`.
- Root cause is in the probe scaffolding, not the `useFocusRestore` hook: `sessionLi.click()` calls `selectSession()` which bumps `focusInputNonce`; `InputBar`'s effect then pulls focus into the chat `<textarea>`. The probe's subsequent `sessionLi.focus()` + 50ms single-shot read races that effect, and the textarea wins on a busy main-thread tick.
- Fix: insert `waitForTimeout(120)` to let the InputBar effect settle, then poll for `data-session-id === 'sA'` via `waitForFunction(1000ms)` instead of a single `evaluate()` read. Same pattern is also applied to the contract-3 re-anchor before opening the CommandPalette. No product code change.

Full investigation: `docs/research/focus-restore-flake-investigation.md`.

## Test plan

- [x] `npm run build`
- [x] Run `node scripts/probe-e2e-a11y-focus-restore.mjs` 5 times in a row — 5/5 pass after the patch (was ~10/13 before).
- [ ] Confirm the same probe still passes in CI on macOS / Linux.

### DIFF
--- a/docs/research/focus-restore-flake-investigation.md
+++ b/docs/research/focus-restore-flake-investigation.md
@@ -1,0 +1,86 @@
+# Focus-Restore Probe Flake Investigation
+
+## Symptom
+
+`scripts/probe-e2e-a11y-focus-restore.mjs` (added in PR #164) was flaky.
+Contracts 2 and 3 (Settings and CommandPalette focus restoration) sometimes
+asserted `document.activeElement` was the session `<li>` and sometimes
+found the chat textarea instead.
+
+## What the probe asserted
+
+> Click session row -> focus stays on `<li>` -> open Settings (Cmd+,) ->
+> close (Esc) -> focus restored to the same `<li>`.
+
+It also tried to "anchor" focus on the row via `await sessionLi.focus()`
+plus a 120ms `waitForTimeout` in the hope that the post-click focus
+orchestration would settle, then poll-confirm the row was focused.
+
+## What actually happens in the product
+
+`selectSession()` in `src/stores/store.ts` (lines 567-571) intentionally
+bumps `focusInputNonce`:
+
+```ts
+// Bump so the InputBar pulls focus -- matches Claude Desktop's UX
+// when clicking a session in the sidebar.
+focusInputNonce: s.focusInputNonce + 1
+```
+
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
Reworked to test the real user contract.

**Root cause of flake**: contracts 2 and 3 asserted "focus restored to session `<li>`" after closing Settings/CommandPalette. But `selectSession()` (store.ts:567-571) intentionally bumps `focusInputNonce`, and `InputBar` (lines 212-230) pulls focus into the chat textarea on that bump — matches Claude Desktop UX. The session row never holds focus after a mouse click. The probe was racing the product's own focus orchestration.

**Investigation**: see `docs/research/focus-restore-flake-investigation.md`. No product bug.

**What changed**:
- Contracts 2 and 3 now assert focus restoration to the chat textarea (`textarea[data-input-bar]`), the actually-focused element after a click.
- Precondition is established by `await textarea.focus()` directly. The selectSession-pulls-focus chain is genuinely racy in the seeded probe environment (InputBar mount-time effect can capture the post-bump nonce as its baseline if it hasn't run yet) and is independently covered by `tests/inputbar.test.tsx`. The dialog-restore contract is independent of HOW focus arrived on the textarea.
- Contract 1 (row a11y attrs) and contract 4 (aria-live) unchanged.
- Added `tests/use-focus-restore.test.tsx` (JSDOM) covering the fallback-selector path (`previousRef === null` → use `fallbackSelector`). That state is unreachable from a normal click flow but is still defensive code worth testing.

**Verification**:
- Probe: 10/10 passes.
- Unit test: 2/2 passes.
- Regression probes (`probe-e2e-permission-prompt-default-mode`, `probe-e2e-askuserquestion-no-dup-and-resolves`): both pass.
- `npm run typecheck` + `npm run build`: clean.
- Pre-existing db-test failures on Windows (better-sqlite3 native rebuild) are unrelated.

Ready for re-review.

PR #179, opened 2026-04-23
chore: small follow-ups (focus-restore test, PERSISTED_KEYS drift, aria-describedby)

Jiahui-Gu:
Three independent small follow-ups bundled into one PR (each <15 min, unrelated files, skipping per-PR review overhead).

---

## #149 — tighten useFocusRestore unit tests

**Problem.** The prior "prior focus" case focused `prior` before opening the dialog and asserted focus was still on `prior` after close — but nothing ever moved focus away in between. A no-op `onCloseAutoFocus` handler would have passed too, so the test did not actually discriminate our hook from Radix's default behavior.

**Fix.** In `tests/use-focus-restore.test.tsx`:
- Between open and close, move focus to an input *inside* the dialog simulating Radix's focus trap. "Focus lands back on prior" is now only true if our hook actively restores it.
- Add a third test that pins the `event.preventDefault()` contract directly on the Event object — the Radix-race suppression that JSDOM can't otherwise exercise.

**Reverse-verify.** Neutered `handleCloseAutoFocus` body to `(_event) => {}`:
```
Test Files  1 failed (1)
     Tests  3 failed (3)
```
Restored body → 3/3 passing.

---

## #152 — PERSISTED_KEYS inline-snapshot drift surface

**Problem.** The existing behavioral tests iterate `PERSISTED_KEYS` from source, so they do fail when a key drifts out of sync — but reviewers only see a message like `comparator early-bailed on persisted key "X"`. No at-a-glance diff of the list itself.

**Fix.** Add `toMatchInlineSnapshot(PERSISTED_KEYS)` plus structural invariants (non-empty, all strings, no duplicates). Any change to `PERSISTED_KEYS` now produces a visible list diff in the same PR that touches it.

**Reverse-verify.** Appended `'fakeDriftKey'` to PERSISTED_KEYS → 3 failures including the snapshot mismatch showing the new entry. Reverted → 4/4 passing.

---

## #165 — aria-describedby for Dialog.Content

**Problem.** Radix logs `Missing Description or aria-describedby={undefined} for {DialogContent}` for any Dialog.Content rendered without a Description child.

**Fix.**
- `SettingsDialog`: added sr-only `<RD.Description>` wired to new i18n key `settings.description` (en + zh).
- `ClaudeCliMissingDialog.SuccessPane`: swaps in when detection succeeds and had Title only → added an sr-only Description built from the version/binary strings.

Scope stayed tight — these are the two already-flagged cases. `ImportDialog` already passes `description`; the main `ClaudeCliMissingDialog` Header also already has a Description.

**Verify.** `npx tsc --noEmit` clean; `npm run build` clean (3 webpack perf warnings, pre-existing). No probe per task instructions; visual layout unchanged.

---

### Out of scope / notes
- `db.test.ts` / `db-hardening.test.ts` fail in the worktree with `NODE_MODULE_VERSION` mismatch on better-sqlite3 — pre-existing, unrelated.
- Per the task doc, no probe was added for #165 (removes a console warning only).

### DIFF
--- a/src/components/ClaudeCliMissingDialog.tsx
+++ b/src/components/ClaudeCliMissingDialog.tsx
@@ -495,6 +495,13 @@ function SuccessPane({
         <RD.Title className="text-base font-semibold text-fg-primary leading-tight">
           {t('cli.detected')}
         </RD.Title>
+        {/* Radix requires a Description (or explicit aria-describedby) on
+            DialogContent. The visible summary below carries the same info
+            but isn't wired to Radix, so we render an sr-only Description
+            to satisfy the contract and silence the console warning. */}
+        <RD.Description className="sr-only">
+          {version ? t('cli.foundVersion') + ' ' + version : t('cli.foundBinaryUnknown')}
+        </RD.Description>
         <div className="mt-1 text-sm text-fg-tertiary">
           {version ? (
             <>
--- a/src/components/SettingsDialog.tsx
+++ b/src/components/SettingsDialog.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import { motion } from 'framer-motion';
 import { Check } from 'lucide-react';
 import * as Checkbox from '@radix-ui/react-checkbox';
+import * as RD from '@radix-ui/react-dialog';
 import { cn } from '../lib/cn';
 import { Dialog, DialogContent } from './ui/Dialog';
 import { Button } from './ui/Button';
@@ -106,6 +107,11 @@ export function SettingsDialog({
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent title={tt('title')} width="720px" hideClose={false} onCloseAutoFocus={handleCloseAutoFocus}>
+        {/* Radix requires either a Description or explicit aria-describedby
+            on DialogContent for a11y. Visible layout of Settings is driven
+            by the tab list + panel heading, so the description is sr-only;
+            screen readers announce it when focus lands in the dialog. */}
+        <RD.Description className="sr-only">{tt('description')}</RD.Description>
         <div className="flex min-h-[380px] border-t border-border-subtle">
           <nav
             className="w-[160px] shrink-0 border-r border-border-subtle py-2"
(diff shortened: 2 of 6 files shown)


Jiahui-Gu:
VERDICT: APPROVE (independent reviewer)

GitHub blocks self-approval, so posting as a comment.

All three bundled items independently reverse-verified (NOT trusting worker logs):

**#149 useFocusRestore tightening**
Replaced `handleCloseAutoFocus` body with a no-op (no preventDefault, no focus(), no fallback) -> `tests/use-focus-restore.test.tsx` 3/3 FAIL. The new "focus moved inside dialog between open/close" assertion fails (focus stays on inside-dialog), the fallback test fails, and the `defaultPrevented` contract test fails. Restore -> 3/3 pass. Old trivially-passing case is gone.

**#152 PERSISTED_KEYS drift**
Appended `'driftDetectorTestKey'` to `PERSISTED_KEYS` in `src/stores/persist.ts` (the array lives there, not `store.ts`) -> `tests/persisted-keys-source-of-truth.test.ts` 3/4 FAIL with visible inline-snapshot diff exposing the new key + comparator-coverage failure + write-snapshot failure. Restore -> 4/4 pass. Drift surface works.

**#165 sr-only Dialog.Description**
Diff is mechanically correct on both SettingsDialog and ClaudeCliMissingDialog SuccessPane (RD.Description as direct child of DialogContent; Radix auto-wires aria-describedby). Scope adjustment for PermissionPromptBlock is justified: zero `@radix-ui` Dialog imports there, it's a `<motion.div role="alertdialog" aria-describedby="perm-desc">` — Radix's "Missing Description" warning never fires for non-Radix elements. (Did not launch electron live for console observation; relied on the well-known Radix contract + clean diff.)

**Regressions check**
Full vitest run: 594/606 tests pass, 45/47 files pass. The 12 failures are exclusively the pre-existing `better-sqlite3 NODE_MODULE_VERSION 130 vs 127` mismatch in `electron/__tests__/db*.test.ts` flagged by the worker. No new regressions.

Full notes: `dogfood-logs/REVIEW-PR-179.md`

PR #178, opened 2026-04-23
refactor(e2e): themed harnesses phase 3 (PERM + UI clusters)

Jiahui-Gu:
## Summary

Phase 3 of the single-harness refactor (follow-up to #173). Absorbs 11 per-file `probe-e2e-*.mjs` probes into two themed harnesses that share a single Electron launch via `scripts/probe-helpers/harness-runner.mjs`.

- **`scripts/harness-perm.mjs`** — 7 PERM cases
- **`scripts/harness-ui.mjs`** — 4 UI cases

Same pattern as `harness-agent.mjs` (PR #173): shared runner, case-id log prefix on failure, per-case reset via `reset-between-cases.mjs`, per-case Playwright trace persisted only on fail, `--only=<id>` filter.

## Source probe -> harness case map

| Harness | Case id | Source probe |
| --- | --- | --- |
| perm | `permission-prompt` | `probe-e2e-permission-prompt.mjs` |
| perm | `permission-mode-strict` | `probe-e2e-permission-mode-strict.mjs` |
| perm | `permission-focus-not-stolen` | `probe-e2e-permission-focus-not-stolen.mjs` |
| perm | `permission-shortcut-scope` | `probe-e2e-permission-shortcut-scope.mjs` |
| perm | `permission-nested-input` | `probe-e2e-permission-nested-input.mjs` |
| perm | `permission-truncate-width` | `probe-e2e-permission-truncate-width.mjs` |
| perm | `permission-sequential-focus` | `probe-e2e-permission-sequential-focus.mjs` |
| ui | `sidebar-align` | `probe-e2e-sidebar-align.mjs` |
| ui | `no-sessions-landing` | `probe-e2e-no-sessions-landing.mjs` |
| ui | `empty-state-minimal` | `probe-e2e-empty-state-minimal.mjs` (partial, see below) |
| ui | `a11y-focus-restore` | `probe-e2e-a11y-focus-restore.mjs` |

Each absorbed probe has a `// MERGED INTO scripts/harness-<cluster>.mjs (case id=<id>; ...)` breadcrumb on line 1 and is skipped by `run-all-e2e.mjs` via `MERGED_INTO_HARNESS`.

## Probes intentionally NOT merged

[...]

PR #182, opened 2026-04-23
chore(probes): harden permission-allow-* + unmask harness-perm flake

Jiahui-Gu:
Bundles two script-only probe-hardening items from recent PR reviews. Both low risk; no production code touched.

## Item #161 — tool-choice flake in probe-e2e-permission-allow-write

Closes #161. Related: PR #172 (Bug L fix), #162 (already closed, same area).

**Flake**: After #172 landed, the probe's store assertion hard-matched `toolName === 'Write'`. The model occasionally picked Edit or MultiEdit for the same prompt, and the test flaked even though Bug L's permission round-trip was functioning.

**Fix** (belt + suspenders):
1. Tightened prompt to explicitly anchor on the Write tool (\`Use the Write tool... Do not use Edit or MultiEdit\`).
2. Widened the store-block lookup to accept \`Write | Edit | MultiEdit\` as a safety net. All three exercise the same Bug L \`control_response\` envelope path, so the test remains valid. The fs assertion (file content == \`world\`) independently guards correctness.

**Flake rates** (local, Windows, real \`claude.exe\`):
- Before fix: not reliably reproducible in short runs (intermittent on CI, per #161 report).
- After fix: **9/11 pass** for tool-name-related failures. The 2 failures observed were a **different** flake class — Allow gets clicked, DOM shows \`File created successfully\`, but no file lands in PROJ dir (likely cwd propagation race or model writing to wrong path). Flagged as follow-up; out of scope here.

## Item PR-#178 follow-up — silent timeout in harness-perm

Addresses reviewer comment on PR #178 (just merged).

**Before**:
\`\`\`js
await win.waitForFunction(() => {
  const el = document.activeElement;
  return el?.getAttribute?.('data-perm-action') === 'reject';
}, null, { timeout: 2000 }).catch(() => {});
\`\`\`

**After**:
\`\`\`js
await win.waitForFunction(() => {
  const el = document.activeElement;
  return el?.getAttribute?.('data-perm-action') === 'reject';
}, null, { timeout: 2000 }).catch((err) => {
  throw new Error(\`[case=permission-prompt] autoFocus never moved to Reject before 2nd Y press: \${err.message}\`);
});
\`\`\`

**Rationale**: the 2nd \`Y\` press can race the autoFocus raf. The old swallowed catch let the test continue and falsely \"pass\" even when Y was typed into the textarea instead of resolving the permission. Now the race surfaces loudly as a FAIL.

This PR does NOT fix the underlying race — it just makes it observable. If the error starts firing in CI, a separate race-condition fix (e.g. wait on a dedicated \`data-perm-focused\` signal) is warranted.

**Verification**: \`harness-perm --only=permission-prompt\` passes 5/5 with the throw in place. No spurious failures on the happy path.

## Test plan
- [ ] \`node scripts/harness-perm.mjs --only=permission-prompt\` — passes 5/5 on CI
- [ ] \`node scripts/probe-e2e-permission-allow-write.mjs\` — passes when the model emits Write/Edit/MultiEdit
- [ ] Reviewer runs reverse-verify: revert the throw in harness-perm, confirm the silent-pass path is reachable

DO NOT merge — independent reviewer dispatched separately.

### DIFF
--- a/scripts/harness-perm.mjs
+++ b/scripts/harness-perm.mjs
@@ -105,10 +105,15 @@ async function casePermissionPrompt({ win, log }) {
   // Wait for autoFocus effect to move focus onto the block's Reject button —
   // otherwise a just-launched InputBar can still hold focus and the Y keystroke
   // gets typed into the textarea instead of resolving the permission.
+  // Surface the timeout loudly (was `.catch(() => {})` — silently swallowed
+  // the race so the test continued and falsely "passed" even when the 2nd Y
+  // didn't resolve the prompt). Per PR #178 reviewer feedback.
   await win.waitForFunction(() => {
     const el = document.activeElement;
     return el?.getAttribute?.('data-perm-action') === 'reject';
-  }, null, { timeout: 2000 }).catch(() => {});
+  }, null, { timeout: 2000 }).catch((err) => {
+    throw new Error(`[case=permission-prompt] autoFocus never moved to Reject before 2nd Y press: ${err.message}`);
+  });
   await win.keyboard.press('y');
   try {
     await heading.waitFor({ state: 'detached', timeout: 3_000 });
--- a/scripts/probe-e2e-permission-allow-write.mjs
+++ b/scripts/probe-e2e-permission-allow-write.mjs
@@ -34,8 +34,14 @@ const PROJ = path.join(os.tmpdir(), `agentory-bugl-write-proj-${TS}`);
 fs.mkdirSync(UDD, { recursive: true });
 fs.mkdirSync(PROJ, { recursive: true });
 
+// Anchor the model on the Write tool explicitly. Earlier wording
+// ("Write a file called ...") let the model occasionally pick Edit /
+// MultiEdit instead, which made the downstream `tn === 'Write'` store
+// assertion flake. We also widen that assertion below to accept any
+// file-mutating tool (Write/Edit/MultiEdit) since all three exercise the
+// same Bug L permission/control_response path.
 const PROMPT =
-  "Write a file called hello.txt with the content 'world' in the current working directory.";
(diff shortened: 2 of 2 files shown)


Jiahui-Gu:
VERDICT: APPROVE (independent reviewer — self-approval blocked by GitHub, posting as comment)

Full notes at dogfood-logs/REVIEW-PR-182.md in this branch.

**Scope:** scripts-only, exactly 2 files, no production code touched. Worker correctly did NOT attempt to fix fs-not-landing class out of scope.

**Item #161 (probe-e2e-permission-allow-write.mjs):**
- Prompt anchoring + writeLike={Write,Edit,MultiEdit} store widening LGTM.
- Critical guardrails intact: fs.existsSync + content==='world' + !isError all still enforced. Not overly lenient.
- My 5 runs: 2 PASS / 3 FAIL. All 3 failures are class-(b) fs-not-landing (DOM HIT but PROJ files=[], no store tool block) — NOT regression of worker's fix. 0/5 cases of the tool-choice flake the PR actually targets.
- Flag (non-blocking): worker reported 9/11 (82%), I see 2/5 (40%). Gap could be machine-speed + small sample; worth a follow-up issue with post-failure allBlocks dump to diagnose the fs-not-landing signature.

**Item PR-#178 follow-up (harness-perm.mjs):**
- Throw msg contains '[case=permission-prompt]' prefix and precise failure description — LGTM.
- My 5 runs: 5/5 × 7/7 PASS. Reverse-verify (temporarily restored .catch(()=>{})): 3/3 × 7/7 PASS — race did not surface in 8 total runs, consistent with worker's 'defensive fix' framing. File restored post-test.

**Regression sweep:** harness-agent 5/5 + harness-ui 4/4 + clean build.

Recommend merge, then open follow-up issue for fs-not-landing class (Bug L-style symptom but distinct from it — DOM reports success, no tool block materializes).

PR #183, opened 2026-04-23
style(sidebar): focus ring + size + archive tokens

Jiahui-Gu:
Closes #199, #200, #212.

Three small polish items from `docs/design/ui-review-sidebar.md`, batched into one PR. All changes in `src/components/Sidebar.tsx`; no token additions needed.

## Changes

- **#199 (H1) — Group-header focus ring → `ring-accent`.** The collapse-toggle `<button>` on each group row previously used `focus-visible:ring-border-strong`, which is near-invisible on `bg-sidebar` / `bg-bg-active`. Swapped to `ring-accent` to match the session-row focus pattern (`Sidebar.tsx:396`) and the rest of the app. Keyboard-nav users can now actually see where focus is.
- **#200 (M2) — Remove `h-5 w-5` override on per-group `+` IconButton.** The override was forcing the button below its `size="xs"` default, producing a 20x20 hit target with a hover surface too small to register. Dropping the override lets `xs` stand (24x24), which still fits inside the `h-7` group header with 2px vertical breathing room.
- **#212 (M3) — Inline `oklch()` in archive section → tokens.** Three literal `oklch()` values (chevron color, count color, hover color) replaced with `text-fg-faint` and `hover:text-fg-tertiary`. Existing tokens; no css changes. This also fixes a silent light-mode regression (the inline oklchs bypassed the light-theme remap of `.text-label-faint`).

## Self-verification

Run from the worktree checkout after `npm install --legacy-peer-deps`.

- `npm run build` → webpack 5.106.2 compiled with 3 warnings (pre-existing bundle size warnings, unrelated).
- `node scripts/harness-ui.mjs` → **4 passed, 0 failed, 4 total** (`sidebar-align`, `no-sessions-landing`, `empty-state-minimal`, `a11y-focus-restore`).
- `node scripts/harness-agent.mjs` → **8 passed, 0 failed, 8 total** (full agent regression sweep clean).

No Sidebar-specific e2e required — these are visual/polish changes, not bug fixes. Harness regression sweep confirms nothing broke.

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -163,7 +163,7 @@ function GroupRow({
                 onFocus();
                 if (!renaming) setGroupCollapsed(group.id, !collapsed);
               }}
-              className="flex flex-1 min-w-0 items-center gap-1.5 text-left text-fg-secondary outline-none rounded-sm focus-visible:ring-1 focus-visible:ring-inset focus-visible:ring-border-strong"
+              className="flex flex-1 min-w-0 items-center gap-1.5 text-left text-fg-secondary outline-none rounded-sm focus-visible:ring-1 focus-visible:ring-inset focus-visible:ring-accent"
               aria-expanded={!collapsed}
             >
               <motion.span
@@ -211,7 +211,7 @@ function GroupRow({
                     e.stopPropagation();
                     createSession({ groupId: group.id });
                   }}
-                  className="shrink-0 h-5 w-5"
+                  className="shrink-0"
                 >
                   <Plus size={12} className="stroke-[1.75]" />
                 </IconButton>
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
VERDICT: APPROVE

Scope: only `src/components/Sidebar.tsx` (+5/-5), no package-lock delta.

Per-fix:
- #199: group-header ring now `ring-accent` at L166, matches session-row at L396. Parity confirmed.
- #200: `size="xs"` (h-6 w-6 = 24px from IconButton variants) with no `h-5 w-5` override at L214.
- #212: archive rows use `text-fg-faint` / `hover:text-fg-tertiary`; all three tokens exist in `global.css`. Hover-to-tertiary (0.72) is a semantically stronger affordance than the previous raw 0.60 and the parent `<button>` is genuinely interactive (toggles archive), so hover class is not dead code. Judgment call OK.

Regression sweep: `npm run build` OK (only pre-existing bundle-size warnings); `harness-ui` 4/4; `harness-agent` 8/8.

No blockers. Reviewer account shares author, so not using gh pr review --approve.

PR #184, opened 2026-04-23
fix(dialog): small fixes (Esc cancel + cli-missing timing)

Jiahui-Gu:
Closes #191, #205.

## Summary
- **UI-12 (#191) ConfirmDialog — Esc always cancels**: added optional `onCancel` prop. Esc, X, outside-click, and the Cancel button now all route through `onCancel` before closing. Destructive variant still blocks Esc by design (documented inline) so a stray keypress can't silently dismiss a destructive flow. Cancel-via-confirm is not double-fired: a ref guards against the confirm path entering the cancel branch, and the ref resets on each open so subsequent cancels still fire.
- **UI-13 (#205) ClaudeCliMissingDialog — success pane polish**: auto-dismiss timing kept at 1500ms (already in the target 1–2s band per review). Added a framer-motion fade-in/slide on the pane plus a subtle scale-in on the checkmark, using the same easing as the dialog entrance so it doesn't feel bolted on.

## Files
- `src/components/ui/ConfirmDialog.tsx`
- `src/components/ClaudeCliMissingDialog.tsx`

## Self-verification
- `npm run build` — OK (webpack compiled with only the pre-existing bundle-size warnings)
- `node scripts/harness-ui.mjs` — 4/4 PASS
- `node scripts/harness-agent.mjs` — 6/6 PASS (note: harness currently has 6 cases, not 8; all green)
- `node scripts/harness-perm.mjs` — 7/7 PASS

### DIFF
--- a/src/components/ClaudeCliMissingDialog.tsx
+++ b/src/components/ClaudeCliMissingDialog.tsx
@@ -487,10 +487,20 @@ function SuccessPane({
   const { t } = useTranslation();
   const belowMin = isVersionBelow(version, CLI_MIN_VERSION_SOFT);
   return (
-    <div className="px-5 py-6 flex items-start gap-3">
-      <div className="mt-0.5 text-status-success-foreground">
+    <motion.div
+      initial={{ opacity: 0, y: 4 }}
+      animate={{ opacity: 1, y: 0 }}
+      transition={{ duration: 0.22, ease: [0.32, 0.72, 0, 1] }}
+      className="px-5 py-6 flex items-start gap-3"
+    >
+      <motion.div
+        initial={{ scale: 0.6, opacity: 0 }}
+        animate={{ scale: 1, opacity: 1 }}
+        transition={{ delay: 0.08, duration: 0.24, ease: [0.32, 0.72, 0, 1] }}
+        className="mt-0.5 text-status-success-foreground"
+      >
         <Check size={18} className="stroke-[2]" />
-      </div>
+      </motion.div>
       <div className="flex-1 min-w-0">
         <RD.Title className="text-base font-semibold text-fg-primary leading-tight">
           {t('cli.detected')}
@@ -521,6 +531,6 @@ function SuccessPane({
           <div className="mt-1 truncate font-mono text-mono-sm text-fg-disabled">{binaryPath}</div>
         )}
       </div>
-    </div>
+    </motion.div>
   );
 }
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
VERDICT: APPROVE

Independent reviewer (separate from worker).

Scope: clean — only `src/components/ui/ConfirmDialog.tsx` (+40/-3) and `src/components/ClaudeCliMissingDialog.tsx` (+14/-4) changed in 0f76599. No stray files.

#191 ConfirmDialog Esc:
- `onCancel` prop added (optional, doc-commented).
- All cancel paths (Esc / X / outside-click / Cancel button DialogClose) funnel through a single `onOpenChange` wrapper: `if (!next && open && !confirmingRef.current) onCancel?.()`.
- Confirm button sets `confirmingRef.current = true` before `onOpenChange(false)` so onCancel is correctly suppressed on the confirm path.
- `useEffect([open])` resets `confirmingRef.current = false` on each open — reopen-after-confirm scenario traced: cancel fires exactly once, confirm fires zero. Good.
- Destructive variant still blocks Esc and outside-click via `onEscapeKeyDown={e.preventDefault()}` and `onInteractOutside={e.preventDefault()}`, with an inline comment telling future readers not to "fix" it. Good.
- Minor style nit (non-blocking): `confirmingRef` useRef is declared after the useEffect that references it. Safe at runtime (effect runs post-render) but reads oddly. Not worth a round trip.

#205 cli-missing timing:
- 1500ms auto-dismiss preserved (line 92).
- SuccessPane wrapped in motion.div (fade + y:4→0, 220ms) with a nested motion.div on the Check icon (scale:0.6→1, 240ms, 80ms delay).
- Easing `[0.32, 0.72, 0, 1]` matches the codebase's established curve (ChatStream, InputBar, PermissionPromptBlock, SettingsDialog, Sidebar all use the same). No overshoot/bounce. Subtle enough — reads as "landed successfully", not "still loading".

Regression sweep (all green):
- `npm run build` — success (webpack 5.106.2, 3 pre-existing bundle-size warnings, unrelated).
- `node scripts/harness-ui.mjs` — 4/4 PASS.
- `node scripts/harness-agent.mjs` — 6/6 PASS (themed-refactor baseline).
- `node scripts/harness-perm.mjs` — 7/7 PASS.

Polish-only PR, no bug-fix e2e rule invoked. LGTM — ready to merge.

PR #187, opened 2026-04-23
style(chatstream): polish bundle (5 fixes)

Jiahui-Gu:
## Summary

Five small ChatStream polish fixes from the UI review, bundled into a single PR.

- **#201 (UI-22) Tool row title attr** — add `title={name (brief)}` on the tool row name+brief span so truncated summaries reveal full text on hover.
- **#202 (UI-23) Plan-approval default focus → Reject** — safer default so an accidental Enter rejects (reversible) instead of approving a potentially destructive plan. `PermissionPromptBlock` already focuses Reject; no contradiction with existing harness-perm cases (they all expect Reject-focused).
- **#203 (UI-24) LongOutputView hide toolbar on ≤10 lines** — Copy/Save/Expand is pure visual noise for tiny outputs; hidden when `total ≤ 10`. Users can still copy via native text selection.
- **#204 (UI-17) loadMessages failure → inline ErrorBlock** — new `loadMessageErrors` store field + inline `LoadHistoryErrorBlock` with Retry button rendered in the chat scroll. Retry clears both the `[]` sentinel and the error entry before re-invoking `loadMessages`.
- **#196 (UI-16) EmptyState ↔ first block crossfade** — wrap in `AnimatePresence mode="wait"` with a 180ms `easeOut` crossfade so the swap is subtle, not a hard cut. Uses existing framer-motion — no new deps.

## Files touched

- `src/components/ChatStream.tsx` — tool row title, PlanBlock focus swap, LongOutputView toolbar gate, crossfade wrapper, LoadHistoryErrorBlock.
- `src/stores/store.ts` — `loadMessageErrors` field + populate it from the existing IPC catch path.
- `src/i18n/locales/en.ts`, `src/i18n/locales/zh.ts` — `chat.loadHistoryFailed`, `chat.retry`.

## Self-verification

```
npm run build                   OK  (3 webpack size warnings — pre-existing, unrelated)
node scripts/harness-agent.mjs  8/8 PASS
node scripts/harness-perm.mjs   7/7 PASS   (includes permission-prompt + permission-sequential-focus)
node scripts/harness-ui.mjs     4/4 PASS
```

No harness-perm cases needed updates — they already assert Reject-focused after mount, which matches the new plan-block behaviour.

## Test plan

- [ ] Hover a long-brief tool row — full `"Name (brief)"` appears in tooltip.
- [ ] Trigger a plan-approval prompt — Reject button has focus on mount; Enter rejects.
[...]

PR #188, opened 2026-04-23
style(motion): coordinate sidebar + right-pane on session switch

Jiahui-Gu:
Closes #192.

## Summary
Clicking a session in the sidebar fired three out-of-sync animations (sidebar row bg 150ms, selection ring 220ms with a custom ease, right-pane content 180ms with a different ease) AND the right pane did not re-crossfade on `activeId` change. The result read as jittery.

This PR introduces two shared motion constants and applies them to both panes so a session switch reads as one coordinated beat:

- `src/lib/motion.ts` (new): `MOTION_SESSION_SWITCH_DURATION = 0.18`, `MOTION_STANDARD_EASING = [0.32, 0.72, 0, 1]`, plus a CSS-string variant.
- `Sidebar.tsx`: session row bg transition `150ms -> 180ms`; selection ring `220ms -> 180ms` with the shared easing.
- `ChatStream.tsx`: right-pane empty/blocks crossfades keep 180ms but use the shared snappy easing, and the `AnimatePresence` children are now keyed on `activeId` so session switches actually trigger the crossfade (previously only empty<->non-empty did).

## Scope
- Minimal, surgical. No animation code restructured.
- No global token system refactor -- #190 (motion-token-kit) is the natural consumer of this module; the filename `src/lib/motion.ts` was deliberately chosen so that PR can extend in-place instead of duplicating.
- No ChatStream structural changes (#207 unaffected).

## Self-verification
- `npm run build`: OK (3 pre-existing bundle-size warnings only).
- `node scripts/harness-ui.mjs`: **4/4 PASS**.
- `node scripts/harness-perm.mjs`: **7/7 PASS**.
- `node scripts/harness-agent.mjs`: 7/8 (1 failure `streaming-caret-lifecycle` confirmed pre-existing on `origin/working` via a `git stash` baseline run -- unrelated to this change).

## Test plan
- [ ] Click between sessions in the sidebar. Selection ring + right-pane content fade together as one motion (no staggered ring-then-fade feel).
- [ ] Switching from a non-empty session to an empty one still crossfades cleanly via `EmptyState`.
- [ ] No regression in sidebar row hover/selection visuals.

### DIFF
--- a/src/components/ChatStream.tsx
+++ b/src/components/ChatStream.tsx
@@ -13,6 +13,10 @@ import remarkGfm from 'remark-gfm';
 import type { MessageBlock, ImageAttachment } from '../types';
 import { useStore } from '../stores/store';
 import { attachmentToDataUrl, formatSize } from '../lib/attachments';
+import {
+  MOTION_SESSION_SWITCH_DURATION,
+  MOTION_STANDARD_EASING
+} from '../lib/motion';
 import { Button } from './ui/Button';
 import { StateGlyph } from './ui/StateGlyph';
 import { diffFromToolInput, type DiffSpec } from '../utils/diff';
@@ -1261,11 +1265,17 @@ export function ChatStream() {
         {blocks.length === 0 && !showThinkingDots && !loadError ? (
           <AnimatePresence mode="wait" initial={false}>
             <motion.div
-              key="empty"
+              // Key includes activeId so switching sessions (sidebar click)
+              // re-mounts and crossfades the right pane in sync with the
+              // sidebar selection-ring animation -- see src/lib/motion.ts.
+              key={`empty:${activeId}`}
               initial={{ opacity: 0 }}
               animate={{ opacity: 1 }}
               exit={{ opacity: 0 }}
-              transition={{ duration: 0.18, ease: [0, 0, 0.2, 1] }}
+              transition={{
+                duration: MOTION_SESSION_SWITCH_DURATION,
+                ease: MOTION_STANDARD_EASING
+              }}
               className="h-full"
             >
               <EmptyState />
@@ -1274,11 +1284,17 @@ export function ChatStream() {
         ) : (
           <AnimatePresence mode="wait" initial={false}>
             <motion.div
-              key="blocks"
+              // Same coordination as the empty branch above: key on activeId
+              // so a session switch crossfades the content pane alongside
+              // the sidebar selection ring (shared timing in src/lib/motion).
+              key={`blocks:${activeId}`}
               initial={{ opacity: 0 }}
               animate={{ opacity: 1 }}
               exit={{ opacity: 0 }}
-              transition={{ duration: 0.18, ease: [0, 0, 0.2, 1] }}
+              transition={{
+                duration: MOTION_SESSION_SWITCH_DURATION,
+                ease: MOTION_STANDARD_EASING
+              }}
               className="px-4 py-3 flex flex-col gap-1.5 max-w-[1100px]"
             >
               {loadError && (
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
VERDICT: APPROVE

Independent reviewer sign-off for PR #188 (cross-pane motion coordination, closes #192).

## Scope verification (post-rebase onto #189)
- PR's own commit `8391457` touches exactly the two expected files: `src/components/Sidebar.tsx` (+15/-3) and `src/components/ChatStream.tsx` (+20/-4). **`src/lib/motion.ts` is untouched** — the rebase correctly dropped #188's stub in favor of #189's superset. The required aliases `MOTION_SESSION_SWITCH_DURATION` (= `DURATION.standard` = 0.18s) and `MOTION_STANDARD_EASING` (= `EASING.standard` = `[0.32, 0.72, 0, 1]`) already exist on `working` from #189 at lines 132 and 135.
- Note: `git diff origin/working --stat` also shows a dogfood-logs delete + probe-e2e-permission-allow-write change, but those come from commit `8a464ed` which is on `working` ahead of this PR — not regressions introduced here.

## Coordination checks (the whole point of #192)
- Sidebar selection ring: `DURATION_RAW.ms220` → `MOTION_SESSION_SWITCH_DURATION` (0.18s). 
- Sidebar session row bg: `duration-150` → `duration-[180ms]`. Easing stays `cubic-bezier(0.32,0.72,0,1)` (matches `MOTION_STANDARD_EASING`).
- ChatStream empty + blocks branches: both now use the shared constants. Easing change from `[0,0,0.2,1]` → `[0.32,0.72,0,1]` is a perceptible shift from soft deceleration to the codebase-default snappy curve — acceptable, it aligns with the motion.ts `EASING.standard` semantic and with the Sidebar.
- Critical fix (worker's key finding) preserved: `key="empty"`/`"blocks"` → `key="empty:${activeId}"`/`"blocks:${activeId}"`. Without this, non-empty→non-empty switches produced zero right-pane motion. Now both panes crossfade in sync.
- `AnimatePresence mode="wait" initial={false}` preserved on both branches (lines 1266 + 1285). #187's landing is NOT undone.
- Both files use the same constant (no raw 0.18 drift that could desync later). The Tailwind class `duration-[180ms]` is a literal string rather than bound to the constant — pragmatic since Tailwind cannot interpolate JS at build time; acceptable.

## Regression sweep
- `npm run build` — passes (pre-existing bundle-size warnings only).
- `scripts/harness-ui.mjs` — 5/5 PASS.
- `scripts/harness-perm.mjs` — 7/7 PASS.
- `scripts/harness-agent.mjs` — 7/8 PASS, 3×. `streaming-caret-lifecycle` fails consistently (`Part 3: caret should be visible mid-stream before interrupt`). **Reverse-verified: same case fails on `origin/working` baseline with #188's two files reverted to working's versions** — confirmed pre-existing flake (#215), NOT caused by this PR.

## Visual-polish judgement: 220ms → 180ms selection ring (-40ms, -18%)
Perceptible speedup but justified. The PR's stated purpose is "one coordinated motion across panes"; leaving the ring at 220ms while the right pane crossfades at 180ms would leave a 40ms tail where the ring still animates after the content is already static — exactly the jitter the PR sets out to fix. Unification is the correct trade-off. If anything the new timing feels more confident, not rushed; 180ms is within the codebase's `DURATION.standard` and matches other selection feedback.

## Findings noted (non-blocking)
- Easing change in ChatStream (`enter` → `standard`) is a subtle visual delta beyond pure coordination. Commit message calls this out honestly.
- Tailwind `duration-[180ms]` on the row bg is not bound to the constant. Future refactor could migrate to a CSS variable or motion-preset class if more call sites appear, but single-site literal is fine now.

Approved to merge after manager sign-off. E2E coverage via UI/agent/perm harnesses is sufficient for a pure style/motion PR.

PR #190, opened 2026-04-23
feat(ui): keyboard shortcut overlay (? / Cmd+/)

Jiahui-Gu:
## Summary

Implements **UI-1** from `docs/design/ui-review-holistic.md` (GH #188): a
global keyboard-shortcut overlay triggered by `?` or `Cmd/Ctrl+/`, showing
every real shortcut the app already has, grouped by area, rendered as a
monospace two-column table with chip-styled `<kbd>` glyphs. Dialog-backed
(Radix), so focus trap / Esc / focus-restore come for free.

## Shortcuts displayed (all grepped from existing code — none invented)

| Group | Keys | Action | Source |
|---|---|---|---|
| Chat | `Enter` | Send message | `InputBar.tsx:565` |
| Chat | `Shift+Enter` | Insert newline | `InputBar.tsx:565` (!shiftKey branch) |
| Chat | `Esc` | Interrupt running turn | `InputBar.tsx:365` |
| Chat | `Esc` | Dismiss slash-command picker | `InputBar.tsx:558` |
| Sidebar | `Cmd/Ctrl+B` | Toggle sidebar | `App.tsx:169` |
| Sidebar | `Cmd/Ctrl+N` | New session | `App.tsx:179` |
| Sidebar | `Cmd/Ctrl+Shift+N` | New group | `App.tsx:175` |
| Navigation | `Cmd/Ctrl+F` | Search / command palette | `App.tsx:163` |
| Navigation | `Cmd/Ctrl+,` | Open settings | `App.tsx:172` |
| Navigation | `?` / `Cmd/Ctrl+/` | Open this overlay | **new** |

## Design notes

- **Modifier-free `?` is gated on a non-editable event target** (textarea,
  input, contentEditable). Without this guard, `Shift+/` mid-composition
  in the chat input would steal the glyph and pop the overlay — classic
  footgun. Matches macOS / GitHub conventions.
- **Platform glyph resolution**: `⌘` on macOS, `Ctrl` elsewhere. Detected
  once from `navigator.platform` — doesn't change mid-session.
- **No new deps**. Uses existing `@radix-ui/react-dialog` via the repo's
  `DialogContent` wrapper, so the chrome / shadows / entrance animation
  are consistent with Settings and CommandPalette.
- **Tooltip chips on buttons — deferred.** `@radix-ui/react-tooltip` IS
  already installed, but wiring chips onto sidebar / inputbar buttons
  would require editing those components, which the task brief explicitly
  scopes out. Tracking as follow-up.

## Files

[...]

PR #189, opened 2026-04-23
refactor(motion): unified duration + easing tokens

Jiahui-Gu:
Ref #190.

## Summary

Establish a unified motion token kit at \`src/lib/motion.ts\` and migrate 9 high-leverage components off inline \`[0.32, 0.72, 0, 1]\` / \`duration: 0.22\` literals. No behaviour change — durations are preserved via semantic tokens (\`DURATION.*\`) or exact aliases (\`DURATION_RAW.*\`) where the millisecond count was load-bearing.

## Tokens

- \`DURATION\`: \`instant(80)\` / \`fast(140)\` / \`standard(180)\` / \`slow(240)\` / \`deliberate(320)\` ms
- \`DURATION_RAW\`: \`ms150\` / \`ms200\` / \`ms220\` / \`ms250\` / \`ms300\` — exact-preservation aliases for legacy values that don't round cleanly, so the migration is truly behaviour-preserving.
- \`EASING\`: \`standard\` \`[0.32,0.72,0,1]\` / \`enter\` \`[0,0,0.2,1]\` / \`exit\` \`[0.7,0,0.84,0]\` / \`linear\`.
- \`MOTION_PRESETS\`: \`fadeIn\`, \`fadeOut\`, \`sessionSwitch\`, \`selectionRing\`, \`paneEnter\`, \`disclosure\`, \`bannerIn\`.
- Compat aliases for #192 (cross-pane motion) so either branch can merge first: \`MOTION_SESSION_SWITCH_DURATION\`, \`MOTION_STANDARD_EASING\`.

## Migration list (9 files)

- \`src/components/Sidebar.tsx\` — group/session selection rings, width resize, archive chevron rotate
- \`src/components/SettingsDialog.tsx\` — tab indicator layoutId
- \`src/components/ClaudeCliMissingDialog.tsx\` — copy-state crossfade + success screen (2 transitions + container)
- \`src/components/PermissionPromptBlock.tsx\` — block entrance
- \`src/components/QuestionBlock.tsx\` — block entrance + option stagger (3 transitions)
- \`src/components/ui/Toast.tsx\` — toast in/out
- \`src/components/AgentDiagnosticBanner.tsx\` — banner slide-in (also normalised \`ease: 'easeOut'\` → \`EASING.enter\`)
- \`src/components/AgentInitFailedBanner.tsx\` — banner slide-in (same easeOut normalisation)
- \`src/components/InputBar.tsx\` — error pill + auto-grow (3 transitions)

## Out of scope

[...]

PR #191, opened 2026-04-23
feat(perm): Allow-always + PermissionPromptBlock a11y completion

Jiahui-Gu:
Closes #194, closes #214.

### UI-11 (#194) PermissionPromptBlock a11y completion
- `aria-modal` flipped from `"false"` to `"true"` (was a latent bug).
- `aria-labelledby` / `aria-describedby` now use `useId()`-scoped ids so sequential prompts never collide on the legacy `perm-title` / `perm-desc` literals.
- Esc -> reject (standard alertdialog dismiss) when focus is inside the prompt.
- Minimal focus trap: Tab / Shift-Tab cycle between Reject / Allow always / Allow while focus is already inside the prompt. Does **not** lock focus globally — the user can still click back into the composer. No new dependency (no `focus-trap-react`); plain `keydown` + refs.

### UI-6 (#214) Allow always
- Third button alongside Allow / Reject. On click it:
  1. Adds the tool name to `store.allowAlwaysTools`.
  2. Dispatches Allow for the current `tool_use_id` (same path as the existing Allow button).
- `onAgentPermissionRequest` (agent/lifecycle.ts) gains `maybeAutoResolveAllowAlways(req)` fast-path. Subsequent permission requests whose `toolName` is on the list auto-resolve Allow and **do not** render a waiting block.
- `AskUserQuestion` / `ExitPlanMode` never auto-resolve (they are user-interaction tools, not side-effecting).

### Session-scope decision (NOT persisted across app restart)
`allowAlwaysTools` is intentionally omitted from `PERSISTED_KEYS`. Reasoning: persisting a privilege grant across restarts silently extends authority into future workdays. User re-confirms after each launch. This matches the safer default for security-adjacent UX; flipping to persisted later is a one-line change to `PERSISTED_KEYS` + `PersistedState`.

## E2E regression guard

Per `feedback_bugfix_requires_e2e` + `feedback_e2e_extend_existing`, extended `scripts/harness-perm.mjs` (no new probe file) with:

- `permission-allow-always` — guards the full UI-6 flow: click Allow always -> IPC allow fires, tool name lands on store list; inject a second prompt for the same tool via `maybeAutoResolveAllowAlways` -> zero new waiting blocks, IPC allow fires automatically; an off-list tool (`Write`) still renders a prompt.
- `permission-a11y` — asserts `role=alertdialog`, `aria-modal=true`, `aria-labelledby` + `aria-describedby` point at real labelled/described elements, Esc dismisses, Tab cycles through the 3 buttons.

Also updated the existing `permission-prompt` case to expect 3 buttons (was 2).

## Reverse-verify
Stashed all behavior changes in `src/`, rebuilt, re-ran the probe:
```
[XX] permission-allow-always   3872ms
locator.waitFor: Timeout 3000ms exceeded.
waiting for locator('[data-perm-action="allow-always"]').first() to be visible
```
Restored + rebuilt -> PASS. Evidence that the probe actually exercises the new behavior.

## Self-verification
- `npm run build` -> OK (3 pre-existing size warnings)
- `node scripts/harness-agent.mjs` -> 8/8 PASS
- `node scripts/harness-ui.mjs` -> 4/4 PASS
- `node scripts/harness-perm.mjs` -> 9/9 PASS (incl. 2 new cases)

Manual a11y grep on `src/components/PermissionPromptBlock.tsx` confirms all four required attrs are present: `role=`, `aria-modal`, `aria-labelledby`, `aria-describedby`.

## Test plan
- [ ] Start a session, trigger a Bash permission prompt, click "Allow always". Confirm the next Bash prompt in the same session resolves without a UI prompt.
- [ ] Restart the app, trigger the same Bash — confirm a prompt reappears (session-scope, not persisted).
- [ ] Screen reader: confirm the prompt announces as a modal dialog with title + description.
- [ ] Keyboard: Tab cycles through the 3 buttons; Esc rejects; Y / N shortcuts still work when focus is outside a text input.

### DIFF
--- a/scripts/harness-perm.mjs
+++ b/scripts/harness-perm.mjs
@@ -78,11 +78,13 @@ async function casePermissionPrompt({ win, log }) {
     return { buttons: btns };
   });
 
-  if (snapshot.buttons.length !== 2) throw new Error(`expected 2 perm buttons, got ${snapshot.buttons.length}`);
+  if (snapshot.buttons.length !== 3) throw new Error(`expected 3 perm buttons, got ${snapshot.buttons.length}`);
   const reject = snapshot.buttons.find((b) => b.action === 'reject');
   const allow = snapshot.buttons.find((b) => b.action === 'allow');
+  const allowAlways = snapshot.buttons.find((b) => b.action === 'allow-always');
   if (!reject || !/Reject \(N\)/i.test(reject.label ?? '')) throw new Error(`bad reject label: ${reject?.label}`);
   if (!allow || !/Allow \(Y\)/i.test(allow.label ?? '')) throw new Error(`bad allow label: ${allow?.label}`);
+  if (!allowAlways || !/Allow always/i.test(allowAlways.label ?? '')) throw new Error(`bad allow-always label: ${allowAlways?.label}`);
   if (!reject.focused) throw new Error(`expected Reject focused; got ${JSON.stringify(snapshot.buttons)}`);
 
   // Press N -> Reject. Block should disappear.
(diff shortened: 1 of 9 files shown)


Jiahui-Gu:
VERDICT: REQUEST_CHANGES

Core feature (Allow-always + a11y) is correctly implemented and well-verified. Two blockers prevent approval:

## Blocker 1: Stale unit test fails (test/behavior contract mismatch)

`tests/permission-prompt-block.test.tsx:125` — **"Escape is a no-op (forces active choice)"** still asserts the old contract. The PR intentionally changes Esc to reject (correct per alertdialog convention and matches the new `permission-a11y` probe), but this unit test was not updated. `npx vitest run tests/permission-prompt-block.test.tsx` → 1 failed. The old "no-op" assertion must be replaced with "Esc calls onReject when focus is inside". Required in this PR — shipping a known-failing unit test breaks CI signal for future PRs and contradicts the behavior your e2e probe asserts.

(Note: the 12 other vitest failures are pre-existing `better-sqlite3` NODE_MODULE_VERSION ABI mismatch — env, not code. Only the Escape test is PR-caused.)

## Blocker 2: Undisclosed scope creep

Diff stat shows **19 files changed, +378/-494** but the PR body only describes PermissionPromptBlock + lifecycle + store + ChatStream + i18n keys + harness. The following are unrelated and undisclosed:

- `src/lib/motion.ts` — **deleted entirely** (135 lines). All call sites (`AgentDiagnosticBanner`, `AgentInitFailedBanner`, `ClaudeCliMissingDialog`, `InputBar`, `QuestionBlock`, `SettingsDialog`, `Sidebar`, `Toast`) were refactored to inline `duration: 0.14` / `ease: [0.32, 0.72, 0, 1]` literals. This directly reverses the stated goal of that file ("One source of truth for animation durations, easing curves"). If this is intentional (dropping the token kit), it deserves its own PR + discussion — not a drive-by in a permissions PR.
- `src/components/ShortcutOverlay.tsx` — **deleted entirely** (159 lines), plus the ?/Cmd+/ handler in `App.tsx`, plus 20 i18n `shortcuts.*` keys in both en/zh. Unrelated to #194/#214.
- `scripts/harness-ui.mjs` trimmed 74 lines. Likely tied to ShortcutOverlay removal.

Per project rule "one PR, multiple commits", these belong in separate PRs (or at minimum separate commits with clear messages). Currently there's a single commit `fc5c38b` lumping everything. Please either (a) revert the motion.ts / ShortcutOverlay removals from this branch and open them as follow-up PRs, or (b) split into per-concern commits with the PR body updated to enumerate each concern.

## Everything else: PASS

- a11y: `aria-modal="true"` ✓, `aria-labelledby`/`aria-describedby` use `useId()`-scoped ids ✓, Tab cycle Reject→Allow-always→Allow (safest-first) ✓, Shift+Tab reverses via `(idx - 1 + length) % length` ✓, listener cleans up on unmount ✓.
- Esc is scoped via `focusInside` guard (doesn't fire when focus is elsewhere) — good.
- `allowAlwaysTools` correctly omitted from `PERSISTED_KEYS` (session-scoped per PR rationale) ✓.
- `maybeAutoResolveAllowAlways` correctly excludes `AskUserQuestion` + `ExitPlanMode` ✓.
- Decision routes through `store.resolvePermission` (single IPC callsite) ✓.
- Labels are human-readable in both locales: "Allow always" / "始终允许" ✓.
- `npm run build` → clean (only pre-existing size warnings).
- `harness-agent.mjs` → 8/8 PASS.
- `harness-ui.mjs` → 4/4 PASS.
- `harness-perm.mjs` → 9/9 PASS (incl. `permission-allow-always` and `permission-a11y`).
- Harness **extends** the existing `scripts/harness-perm.mjs` rather than creating a new probe file ✓ (satisfies feedback_e2e_extend_existing).
- **Reverse-verify reproduced independently**: I hid the 3rd button with `{false && onAllowAlways && ...}`, rebuilt, ran `harness-perm.mjs` → 3 failures (`permission-allow-always` timeout on `[data-perm-action="allow-always"]`, `permission-a11y` "Tab from reject expected allow-always, got allow", and `permission-prompt` which now expects 3 buttons). Restored → 9/9 PASS. Probes genuinely exercise the new behavior ✓.

## Edge cases reasoned through

- Back-to-back prompts for same tool after "Allow always" click: second request auto-resolves in `onAgentPermissionRequest` before any `appendBlocks` call. No second block rendered. Verified by the `permission-allow-always` probe.
- Race (prompt #2 arrives before `addAllowAlways` state propagates): Zustand `set` is synchronous; since `onAllowAlways` handler calls `addAllowAlways` first and then `resolvePermission`, and the next `onAgentPermissionRequest` IPC is necessarily ordered after the user's click event loop tick, `useStore.getState().allowAlwaysTools` will include the tool. Correct.
- Parallel batch (bug L territory): N pending prompt blocks, user clicks Allow-always on one. The other N-1 are **already rendered** as blocks, so they stay rendered — auto-resolve only short-circuits NEW requests. Reasonable default (once a block is on screen, user still decides individually), and matches the "decisions are always per-click" UX. Worth a sentence in the PR body so reviewers don't wonder. **Not a blocker** — policy question flagged.

## Summary

Core feature is solid and well-verified. Please (1) fix/update the Escape unit test, and (2) split out the motion.ts + ShortcutOverlay removals (or at least declare them in the PR body and separate into their own commits). Then happy to re-approve.

Jiahui-Gu:
Rebased onto working, scope trimmed, stale test updated to assert Esc -> reject. Ready for re-review.

Blocker 1 (stale unit test): `tests/permission-prompt-block.test.tsx` — the old "Escape is a no-op" case was replaced with two cases matching the new contract:
- Escape triggers Reject when focus is on the auto-focused Reject button (standard alertdialog dismiss)
- Escape is still a no-op when focus is on an external textarea (not hijacking user input)

`npx vitest run tests/permission-prompt-block.test.tsx` — 12/12 pass.

Blocker 2 (scope creep): root cause was branch forked before #189/#190 merged. Rebased cleanly on `origin/working` (no conflicts). Final `git diff origin/working --stat`:

```
 scripts/harness-perm.mjs                 | 212 ++++++++++++++++++++-
 src/App.tsx                              |   5 +-
 src/agent/lifecycle.ts                   |  33 +++++
 src/components/ChatStream.tsx            |  12 +-
 src/components/PermissionPromptBlock.tsx |  79 ++++++++++--
 src/i18n/locales/en.ts                   |   1 +
 src/i18n/locales/zh.ts                   |   1 +
 src/stores/store.ts                      |  25 ++++
 tests/permission-prompt-block.test.tsx   |  30 ++++-
 9 files changed, 383 insertions(+), 15 deletions(-)
```

No more `src/lib/motion.ts`, no more `src/components/ShortcutOverlay.tsx`, no more `scripts/harness-ui.mjs`, App.tsx reduced to the +5/-2 needed for `maybeAutoResolveAllowAlways` wiring. i18n only adds `allowAlways` key (keeps #190's shortcuts.* keys).

Full sweep:
- `npm run build`: OK (3 pre-existing size warnings)
- `npx vitest run tests/permission-prompt-block.test.tsx`: 12/12 pass
- `node scripts/harness-agent.mjs`: 8/8 pass
- `node scripts/harness-ui.mjs`: 5/5 pass (incl. shortcut-overlay-opens restored by rebase)
- `node scripts/harness-perm.mjs`: 9/9 pass (incl. permission-allow-always, permission-a11y)

Jiahui-Gu:
VERDICT: REQUEST_CHANGES (v2)

Rebase is still stale. The previous review asked for a rebase onto origin/working; branch head 6855cd6 shares merge-base 8a464ed with origin/working, but origin/working has since advanced by one commit (8e6fbb7 — "style(motion): coordinate session-switch across panes (#192) (#188)") that this PR has not picked up.

Scope check failed: diff vs origin/working is 10 files, not 9. The extra file is `src/components/Sidebar.tsx`, and `src/components/ChatStream.tsx` also carries a regression — both are caused by the missing merge-forward of 8e6fbb7, not by intentional edits in this PR.

Effect of merging as-is: PR #192's session-switch motion coordination would be REVERTED on both panes:

- `src/components/Sidebar.tsx`: loses the `MOTION_SESSION_SWITCH_DURATION` / `MOTION_STANDARD_EASING` imports and usages; selection-ring transition drops back to `DURATION_RAW.ms220`; the row bg transition drops from `duration-[180ms]` back to `duration-150`; the sync comment is removed.
- `src/components/ChatStream.tsx`: loses the `lib/motion` import; `motion.div` keys revert from `\`empty:${activeId}\``/`\`blocks:${activeId}\`` back to `"empty"`/`"blocks"` (so right-pane crossfade no longer re-triggers on session switch); transitions revert from the shared snappy curve back to `ease: [0, 0, 0.2, 1]`.

Remediation: rebase this branch onto current `origin/working` (8e6fbb7) rather than the earlier base. Resolve the ChatStream.tsx overlap by keeping BOTH #192's activeId-keyed crossfade / shared motion constants AND this PR's `addAllowAlways` wiring + `onAllowAlways` prop on `PermissionPromptBlock`. Sidebar.tsx should end up unchanged vs origin/working (this PR doesn't intentionally edit it).

After that, the file list should be exactly 9: scripts/harness-perm.mjs, src/App.tsx, src/agent/lifecycle.ts, src/components/ChatStream.tsx, src/components/PermissionPromptBlock.tsx, src/i18n/locales/en.ts, src/i18n/locales/zh.ts, src/stores/store.ts, tests/permission-prompt-block.test.tsx.

Not re-running the harness / unit-test sweep this round — the rebase is a blocker regardless and the 9-file scope + tests should be verified again after the correct rebase.

Jiahui-Gu:
Re-rebased onto latest working (includes #188). Scope trimmed to exactly 9 files. Ready for re-review.

Jiahui-Gu:
VERDICT: APPROVE (v3)

## Rebase freshness
- `git log origin/working..HEAD` -> 2 commits (91027c6 feat + 72e4ef8 test), both in scope.
- `git log HEAD..origin/working` -> empty. Branch is fully up-to-date on top of working (includes #188 motion-coordination and #190 shortcut-overlay).

## Scope: exactly 9 files (matches expectation)
- scripts/harness-perm.mjs
- src/App.tsx
- src/agent/lifecycle.ts
- src/components/ChatStream.tsx
- src/components/PermissionPromptBlock.tsx
- src/i18n/locales/en.ts
- src/i18n/locales/zh.ts
- src/stores/store.ts
- tests/permission-prompt-block.test.tsx

No Sidebar.tsx, no other strays. v2 scope-creep concern is fully resolved.

## ChatStream diff (minimal + motion preserved)
Diff contains ONLY:
1. new `addAllowAlways: (toolName: string) => void` param on `renderBlock`
2. new `onAllowAlways` prop wiring to PermissionPromptBlock (calls `addAllowAlways(toolName)` then resolves allow)
3. `const addAllowAlways = useStore((s) => s.addAllowAlways)` hook selection
4. pass-through at the single `renderBlock` call site

No unrelated edits (no LongOutputView / imports / animation churn).

Motion coordination from #188 is intact:
- imports `MOTION_SESSION_SWITCH_DURATION` and `MOTION_STANDARD_EASING` (lines 17-18)
- AnimatePresence keys use `` `empty:${activeId}` `` / `` `blocks:${activeId}` `` (lines 1281, 1300)
- motion.div children use `duration: MOTION_SESSION_SWITCH_DURATION, ease: MOTION_STANDARD_EASING` (1286-1287, 1305-1306)

## Policy sanity (re-verified)
- `maybeAutoResolveAllowAlways` in `src/agent/lifecycle.ts:156-161` excludes `AskUserQuestion` and `ExitPlanMode` via the `interactive` flag.
- `allowAlwaysTools` absent from `PERSISTED_KEYS` in `src/stores/persist.ts` (session-scoped, does not leak across restarts).

## Test & harness output
- `npx vitest run tests/permission-prompt-block.test.tsx` -> 12/12 PASS. Esc-triggers-Reject test at line 125 (`Escape triggers Reject`) asserts `onReject` called once; paired with line 145 `Escape is a no-op when focus is outside`.
- `npm run build` -> webpack compiled (only pre-existing bundle-size warnings; no errors).
- `scripts/harness-perm.mjs` -> 9/9 PASS (incl. `permission-allow-always` and `permission-a11y`).
- `scripts/harness-ui.mjs` -> 5/5 PASS; `shortcut-overlay-opens` (#190) preserved, confirming rebase pulled in #188/#190.
- `scripts/harness-agent.mjs` -> 7/8 (flake on `streaming-caret-lifecycle` Part 3 caret visibility). Reproduced 3x on this branch AND on baseline `origin/working` (baseline even times-out earlier). Confirmed pre-existing instability tracked as #215, NOT a regression from this PR — no file in scope touches streaming caret.

## Verdict
APPROVE. v2 blockers (stale rebase missing #188, scope creep) are both resolved. Allow-always + a11y implementation is clean, tests are comprehensive, motion coordination preserved, policy constraints correct.

PR #186, opened 2026-04-23
style(empty): empty-state polish bundle (3 fixes)

Jiahui-Gu:
Three small empty-state polish fixes bundled into one PR.

## Summary

- **#193 — Sidebar empty Groups hint**: when the normal-groups list is empty, render a one-line muted hint (`No groups yet — click + to create one`) inside the groups nav. No CTA button; the `+` next to the Groups header is already there. File: `src/components/Sidebar.tsx`.
- **#195 — Post-tutorial New Session = primary CTA**: after tutorial completes, the landing zone shows New Session + Import Session side-by-side. Promoted the first button from `variant="secondary"` to `variant="primary"` so users have an obvious next step. File: `src/App.tsx`.
- **#197 — CommandPalette no-match shows query**: when a query returns no results, echo the query back (`No results for "<query>"`) instead of the generic "No matches". New i18n key `commandPalette.noResultsFor` with `{{query}}` interpolation, added to both en and zh. Files: `src/components/CommandPalette.tsx`, `src/i18n/locales/{en,zh}.ts`.

Scope-discipline: each file touched is narrowly scoped to its fix. Sidebar change is inside the `normal.map` empty branch — no overlap with the sidebar focus-ring / + button / archive-token bundle.

## Test plan

- [x] `npm run build` — OK (3 perf warnings, pre-existing).
- [x] `node scripts/harness-ui.mjs` — 4/4 PASS (sidebar-align, no-sessions-landing, empty-state-minimal, a11y-focus-restore).
- [x] `node scripts/harness-agent.mjs` — 8/8 PASS.

Closes #193, #195, #197.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #192, opened 2026-04-23
chore(probe): repro + diagnose fs-not-landing (#186)

Jiahui-Gu:
## Summary

Extend `probe-e2e-permission-allow-write.mjs` with a diagnostic timeline + failure artifact dump to investigate Bug #186 (class-b: Allow success but file not on fs). No production code changes.

## Failure rate observed

20 consecutive runs on `working` (1be3259): **16 pass / 4 fail = 20%** — matches baseline 2/11 (~18%) from PR #182.

## Root cause hypothesis (confirmed, 4/4 failures)

All four failures share the same shape:

1. Model calls `Skill` (using-superpowers, auto-loaded from user's `~/.claude`) BEFORE Write.
2. Permission prompt that appears first is for Skill, not Write.
3. Probe's `locator('[data-perm-action=allow]').first().click()` answers the Skill prompt.
4. Model then emits Write, which produces a second permission prompt — never answered.
5. 60s observation window expires with `fileExists=false`.

The pre-allow diagnostic snapshot shows `waiting.toolName === "Skill"` in every failing run (evidence in each `dogfood-logs/bug-186-*.json`). The wanderers walk found no stray `hello.txt` anywhere (rules out wrong-cwd). No `Write` tool block ever reaches `result:` state (rules out tool_result loss, async delay, isError).

**The control-rpc / hook_callback pipeline fixed in Bug L is NOT regressing. The symptom was a misread:** the "DOM says File created" regex also matches the user's prompt text containing `hello.txt`, producing a false positive.

## Recommended fix direction

**Fix the probe, not production.** Two options (report has detail):

1. Probe loops the Allow click across all preamble permission prompts until `storeHit` fires for `Write` specifically. ~10 lines.
2. Launch Electron with sanitized `HOME` / `USERPROFILE` so user-level skills don't auto-inject into probe runs.

## Artifacts

- Full report: `dogfood-logs/BUG-186-REPRO.md`
- 4 failure artifacts: `dogfood-logs/bug-186-2026-04-23T*.json`
- Batch summary: `dogfood-logs/bug-186-batch-summary.json`

## Test plan

- [ ] Review `dogfood-logs/BUG-186-REPRO.md` — the hypothesis ranking with evidence.
- [ ] Spot-check one failure artifact's `timeline[].store.blocks` — confirm `waiting` block at `pre-allow` is `Skill` not `Write`.
- [ ] (Optional) Re-run `node scripts/probe-e2e-permission-allow-write.mjs` a few times to verify the diagnostic dump still fires on failure and is gated to the failure path.


Jiahui-Gu:
VERDICT: APPROVE

Independent reviewer sign-off.

**Scope (clean):** Only probe + artifacts + report. Verified via `git diff $(merge-base)..HEAD --stat` — 7 files, all under `scripts/` or `dogfood-logs/`. No `src/` or `electron/` changes. (The `git diff origin/working --stat` output looks noisy only because the PR branch is 2 commits behind working — motion refactor + shortcut overlay — unrelated drift.)

**Hypothesis (confirmed by hard evidence, 4/4 artifacts):**
- `pre-allow` snapshot in every failure shows the waiting block is `toolName: "Skill"`, input `{"skill":"using-superpowers"}` — NOT Write. The probe's single `.first().click()` answers the Skill prompt.
- `observation-end` (+60s) shows the Write tool block finally emitted with correct `file_path` pointing into this run's PROJ dir, plus a fresh `waiting` block for Write that is never answered.
- File path in Write input matches this run's PROJ — case (c) wrong-cwd properly ruled out. The `wanderers` list contains only `hello.txt` from OTHER runs' PROJ dirs, none for this run's timestamp.
- 100% of 4 failures share this exact shape.

**Probe extension quality:** `snapDiag` + `dumpBug186Artifact` are additive, don't touch assertion logic, dump enough detail (blocks with kind/toolName/toolUseId/hasResult/isError/input/permissionState, PROJ files, wanderers walk, full timeline). Good forensic coverage for future debugging.

**Report quality:** Strong. Quantified fail rate (20%, matches baseline 18%), per-failure artifact refs, ranked hypotheses with evidence columns, all 5 alternatives explicitly ruled out with reasoning, clear recommended fix direction (probe-side).

**Build:** `npm run build` passes (pre-existing webpack size warnings only, unrelated).

**Suggestions for follow-up fix PR (non-blocking):**
1. Prefer option 1 from the report (loop Allow click while observing) — ~10 lines, preserves realistic env, robust to arbitrary preamble tool counts.
2. Also tighten the DOM-signal regex: `/File created successfully|File written|hello\.txt/` is false-positive-prone since `hello.txt` appears verbatim in the probe's own PROMPT. Suggest `/File created successfully at.*hello\.txt/` as the report itself notes in section 6. This removes the "DOM says File created but fs missing" misread that led to the original Bug L false alarm framing.
3. Minor: consider logging/warning when a permission prompt for a tool OTHER than Write/Edit/MultiEdit appears during the probe — an early signal of env pollution across future unrelated probes.

Fix PR is a separate follow-up. Approving this diagnostic PR as-is.

PR #197, opened 2026-04-23
ux(window): exit fade animation (#213 partial)

Jiahui-Gu:
## Summary

UI-10 polish bundle for per-instance identity. Ships the exit animation; investigates and closes out the hotkey concern; defers the tint.

- **(b) Exit fade animation — shipped.** A 180ms opacity fade-out runs before the window hides-to-tray. Main sends `window:beforeHide` with a duration, delays, then hides; renderer drives `documentElement.style.opacity` using shared motion tokens (`DURATION.standard` + `EASING.exit`). On restore `window:afterShow` resets opacity. Same path serves Cmd/Ctrl+W, the renderer X button, and the OS close box. Real quit (`isQuitting`) skips the fade so shutdown stays fast.
- **(a) Per-instance hotkey — investigated, no change needed.** No `globalShortcut.register` anywhere; no `Menu.setApplicationMenu` accelerators; no BrowserWindow-level shortcuts. All hotkeys are `window.addEventListener('keydown', ...)` in `App.tsx`, naturally scoped to each BrowserWindow's document. No cross-window leak possible.
- **(c) Per-instance tint — deferred.** Requires a new settings entry + UI surface; keeping this PR scoped. Track as a follow-up PR.

## Implementation notes

Rendering approach: we mutate `document.documentElement.style.opacity` directly rather than wrapping the React tree in a `<motion.div>`. A root wrapper would be invasive (layout risk, style cascade surprises) and redundant — we only need one numeric value to change. The direct DOM write is zero-rerender and survives teardown.

Files:
- `electron/main.ts` — fade gate in `win.on('close')`, merged `window:afterShow` into existing `win.on('show')` handler.
- `electron/preload.ts` — exposed `window.onBeforeHide` / `window.onAfterShow` subscriptions.
- `src/global.d.ts` — typed the new bridge methods.
- `src/App.tsx` — effect that subscribes to both events and animates `<html>` opacity.

## Test plan

- [x] `npm run build` — OK
- [x] `node scripts/harness-ui.mjs` — 5/5
- [x] `node scripts/harness-perm.mjs` — 9/9
- [x] `node scripts/harness-agent.mjs` — 7/8. `streaming-caret-lifecycle` fails identically on clean `origin/working` (verified via stash); known flake tracked in #215, not introduced by this PR.
- [ ] Manual: Cmd/Ctrl+W triggers a visible fade before the window disappears; tray click restores with opacity reset.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -312,6 +312,11 @@ function createWindow() {
   // and routes their events at the live renderer.
   win.on('show', () => {
     if (!win.webContents.isDestroyed()) sessions.rebindSender(win.webContents);
+    // Reset the renderer's fade-opacity in case the window was just
+    // restored after a fade-to-hide (see `window:beforeHide` below).
+    if (!win.webContents.isDestroyed()) {
+      win.webContents.send('window:afterShow');
+    }
   });
   win.webContents.on('did-finish-load', () => {
     if (!win.webContents.isDestroyed()) sessions.rebindSender(win.webContents);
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
VERDICT: APPROVE

Independent reviewer (slim sweep, ~6 min). Previous reviewer stalled — re-ran cleanly.

## Scope
4 files exactly as claimed: `electron/main.ts` (+35/-2), `electron/preload.ts` (+13), `src/App.tsx` (+36), `src/global.d.ts` (+2). Nothing extra.

## (a) Hotkey investigation — confirmed no-op
`grep -rn "globalShortcut\|accelerator" electron/ src/` returns zero matches. Worker's claim verified — all hotkeys are renderer-side keydown, naturally per-window. No fix needed.

## (b) Exit fade — solid
- `close` handler: `if (isQuitting) return;` is the FIRST line, so Cmd+Q / `before-quit` fully bypasses the fade. Verified `isQuitting` is set in `app.on('before-quit')` at line 1028. **No hang risk on quit.**
- `fadePending` guard prevents repeated Cmd+W from stacking timers. Good.
- `setTimeout` callback re-checks `win.isDestroyed()` before `win.hide()`. Good.
- 180ms = `DURATION.standard`, matches motion tokens.
- `window:afterShow` resets opacity on restore — handles the tray-click → reopen path correctly.

## App.tsx
- Drives `document.documentElement.style.opacity` directly. Zero React rerenders, survives state teardown. Clean implementation.
- Effect cleanup restores `transition` and `opacity` on unmount.

## Reduced motion — honored
`src/styles/global.css:582-591` has global `@media (prefers-reduced-motion: reduce)` rule with `transition-duration: 0.01ms !important` on `*`. This overrides the inline `transition: opacity 180ms` set on `documentElement`. No additional handling needed.

## Harness
- `npm run build`: success (3 perf warnings, expected).
- `node scripts/harness-ui.mjs`: 5/5 PASS on second run. First run had one flaky `sidebar-align` timeout (10s waitForFunction on a cold cache); both retries clean. Not PR-caused.
- `node scripts/harness-agent.mjs`: 7/8. Only failure is `streaming-caret-lifecycle` (#215, fix tracked in PR #196). Tolerated.

## Notes
- (c) tint deferred — explicit, fine.
- Comments in `electron/main.ts` are unusually thorough (good — explains the `fadePending` guard and the `isQuitting` skip rationale).

LGTM, ready to merge.

PR #181, opened 2026-04-23
feat(tool-block): elapsed time + dropped marker + stall hint

Jiahui-Gu:
## Summary
- **A2-NEW-5**: tool block shows live elapsed time (e.g., `1.2s`, `5.4s`) while in-flight, counter disappears once the `tool_result` lands.
- **A2-NEW-6**: empty/dropped tool blocks (explicit `result === ''` or `hasResult && empty brief`) render a muted `(no result)` marker instead of silent empty space.
- **A2-NEW-7**: in-flight tool blocks show `(taking longer than usual…)` hint once elapsed >= 30s (threshold `STALL_HINT_AFTER_MS`). Cancel-in-block would need new IPC so for now the hint semantically routes users to the existing StatusBar Stop button.

## Implementation
- `src/components/ChatStream.tsx`:
  - New `formatElapsed(ms)` helper: `<100s → "12.3s"`, `<100min → mm:ss`, `>=100min → 1h23m` — bounded width so the chevron doesn't jitter.
  - `ToolBlock` captures `startedAtRef` on first render without a result and freezes `endedAtRef` once a result arrives. `running = !hasResult && !isError` drives both counter and stall hint.
  - Drop detection: `isDropped = (emptyResult || (hasResult && emptyBrief)) && !isError`. Explicitly gated on `hasResult` so an in-flight block is never labeled as dropped.
  - `ChatStream` runs a single 100ms `setInterval` that ticks a shared `now` state only while `hasInflightTool` is true, then shuts off — no per-block timers, zero cost when idle.
- `src/i18n/locales/{en,zh}.ts`: new `chat.toolNoResult` + `chat.toolTakingLonger` strings.

## Tests
- **RTL**: `tests/chatstream-tool-block-ux.test.tsx` — 6 tests, covers all three features in isolation including the stall-hint branch that needs fake timers. All green locally.
- **E2E**: new case `tool-block-ux` added to `scripts/harness-agent.mjs`. Seeds three tool blocks (in-flight, explicit-empty-result, completed-healthy) directly into the store, then asserts:
  - exactly 1 `[data-testid="tool-elapsed"]` (in-flight only), text matches `/^\d+\.\ds$/`
  - exactly 1 `[data-testid="tool-no-result"]`, text contains "no result", does NOT leak to the healthy block
  - after the in-flight block's result lands, elapsed counter clears to 0
- Harness total: **6/6 PASS** (5 existing + 1 new).

## Reverse-verify
[...]

PR #200, opened 2026-04-24
feat(tool-block): 90s stall escalation

Jiahui-Gu:
Closes #208.

## Summary

Adds a second escalation tier on top of the 30s stall hint that landed in #181, so users get a visibly louder signal when a tool block has been running for an unusually long time.

### Threshold ladder (named constants at top of `ToolBlock` in `src/components/ChatStream.tsx`)

| Constant | Value | Behavior |
|---|---|---|
| `STALL_HINT_AFTER_MS` | `30_000` | Existing (#181) — subtle italic `(taking longer than usual…)` next to the elapsed counter. Counter color stays neutral. |
| `STALL_ESCALATE_AFTER_MS` | `90_000` | NEW (#208) — elapsed-time chip flips to `state-warning-text`, gentle hint hides, an inline `Cancel` link appears. |
| `STALL_DROP_AFTER_MS` | `120_000` | Documented in source as a future tier; intentionally not implemented this round (deferred per task spec — picking the right cutoff and recovery semantics is out of #208 scope). |

### Escalation visual

- Elapsed chip gets `text-state-warning-text font-semibold` and a `data-escalated="true"` attribute (handy for tests / future styling).
- Inline "Cancel" affordance rendered as a `role="button"` span — the parent collapse-row is already a `<button>`, so a real `<button>` would nest and trigger an HTML/React hydration warning. Span has `tabIndex=0`, Enter/Space keyboard handler, and `stopPropagation` so clicking Cancel does NOT toggle the collapse.
- The 30s gentle hint hides once the louder tier kicks in so the row is not double-noisy.

### Cancel wiring (intentionally stubbed)

`onCancelStalled` currently emits a `console.warn` with the tool name and carries a `TODO(#208-followup)` comment. There is no per-tool-use cancel IPC today (the StatusBar Stop button interrupts the whole turn). The escalation visual is the real goal of #208; cancel wiring is separate scope and tracked for a follow-up so this PR can ship without a backend change.

### Constraints honored

- No structural ChatStream refactor (#207 is reserved for that).
- No new dependencies.
- Inline (non-modal, non-blocking) escalation.
- English-only repo content; both en + zh i18n keys added.

## E2E (extends, not new)

Per `feedback_e2e_extend_existing.md`, added a `tool-stall-escalation` case to the existing `scripts/harness-agent.mjs` instead of creating a new probe file.

The case patches `Date.now` in the renderer window before seeding the in-flight tool block so the `ToolBlock`'s `startedAtRef` captures a synthetic `T-Δ`. After mount it restores `Date.now`, and the existing 100ms `ChatStream` interval (`setNow(Date.now())`) samples real time on the next tick — at which point `elapsedMs = Δ`, flipping the right threshold without sleeping for 90 real seconds.

Two tiers covered:
- **31s**: only `tool-stalled` is visible; no `tool-stall-escalated`, no Cancel link, no `data-escalated=true`.
- **91s**: `tool-stall-escalated` + `tool-stall-cancel` visible, `data-escalated="true"` on the chip, gentle hint hidden. Also asserts that clicking Cancel does NOT change `aria-expanded` on the parent row (stopPropagation works).

## Reverse-verify

`git stash`ed the `src/components/ChatStream.tsx` + both `src/i18n/locales/*.ts` portions of the diff (probe stayed in place), rebuilt, and re-ran the new case:

```
[case=tool-stall-escalation] FAIL: #208 tier-2 (91s): expected escalation badge to be visible, got 0
```

Restored the fix → re-ran → PASS. The probe is genuinely exercising the new code path.

## Self-verification

- `npm run build` — green (3 pre-existing webpack size warnings, unrelated)
- `node scripts/harness-agent.mjs` — **9/9 PASS** (including new `tool-stall-escalation`)
- `node scripts/harness-ui.mjs` — **5/5 PASS**
- `node scripts/harness-perm.mjs` — **9/9 PASS**
- `npx vitest run tests/chatstream-tool-block-ux.test.tsx tests/i18n-key-parity.test.ts` — **8/8 PASS** (existing 35s stall-hint test still passes; i18n key parity holds)

## Test plan

- [ ] Verify 30s hint still appears alone (no escalation) on a real slow tool.
- [ ] Verify 90s escalation visually pops (warning chip + Cancel link, gentle hint gone).
- [ ] Verify clicking Cancel logs the warn line in DevTools and does NOT collapse the tool block row.
- [ ] Verify a tool that completes before 90s never shows the escalation chip.

Do NOT merge — needs independent reviewer per `feedback_independent_reviewer.md`.


Jiahui-Gu:
## Independent review: APPROVE

Reviewed at head `fd9cfbd` in fresh worktree.

### Scope check
4 files exactly as described, no churn:
- `scripts/harness-agent.mjs` (+164/-0) — new case extends existing harness (per repo rule on extending probes vs creating new ones)
- `src/components/ChatStream.tsx` (+87/-13)
- `src/i18n/locales/{en,zh}.ts` (+3/-0 each)

### Reverse-verify (mandatory per repo rule on bug-fix probes)
1. `git checkout origin/working -- src/components/ChatStream.tsx` (kept probe + i18n)
2. `npm run build` (clean prod bundle)
3. `AGENTORY_PROD_BUNDLE=1 node scripts/harness-agent.mjs --only=tool-block-ux,tool-stall-escalation`
4. Result: **tool-stall-escalation FAILED** at tier-2 (`expected escalation badge to be visible, got 0`) — probe correctly catches the regression
5. Restored, rebuilt, re-ran: **PASSED**

### Full harness
9/9 cases pass on the PR head, matching worker's claim.

### Code quality
- Tiered constants `STALL_HINT_AFTER_MS=30s < STALL_ESCALATE_AFTER_MS=90s < STALL_DROP_AFTER_MS=120s` — sane ordering, the 120s tier is documented-only with an honest comment about why it's deferred.
- Token `text-state-warning-text` exists in `src/styles/global.css` (both light and dark themes).
- Cancel link uses `role="button"` + `tabIndex=0` + Enter/Space handler — correct because parent row is a `<button>` and nesting buttons would be invalid HTML. `stopPropagation` on click + keydown prevents collapse leak (probe verifies this).
- TODO comment honestly states the link is a `console.warn` placeholder pending a per-tool-use cancel IPC. No silent overpromise.
- i18n keys mirror correctly between en/zh.
- `data-escalated` attr + `transition-colors` on the elapsed chip — clean visual progression.

### One non-blocking quality note (follow-up worth filing)
The probe **passes in the full suite and when preceded by `tool-block-ux`, but FAILS when run in isolation** (`--only=tool-stall-escalation`):
```
FAIL: #208 tier-1 (31s): expected 1 "taking longer" hint, got 0
```
Likely a renderer cold-start timing artifact — the first `setState` after a fresh window seems to swallow the 250ms `waitForTimeout`. Not blocking merge (canonical run path is the full harness, which is green), but worth a small hardening pass later (e.g. an `await win.locator('[data-testid="tool-elapsed"]').waitFor()` before the assertion, or a longer initial settle).

### Verdict
APPROVE — ready to merge.

PR #203, opened 2026-04-24
fix(import): load .jsonl history + filter sub-agent + filter agentory-temp cwds (#219)

Jiahui-Gu:
## Summary

Three import-session bugs reported on dogfood, all in the same code path — fixed together (Task #219).

- **Bug A** — Imported sessions render EMPTY until the user sends a follow-up. `importSession()` created a row with `resumeSessionId` but never loaded the on-disk transcript into `messagesBySession`. Now hydrates immediately via a new `import:loadHistory` IPC + `framesToBlocks` reducer (reuses `streamEventToTranslation`), and persists to sqlite via `saveMessages` so subsequent loads come from DB.
- **Bug B** — Sub-agent (Task tool) transcripts leaked into the import picker. They carry `parentUuid` non-null OR `isSidechain: true` on the first frame; resuming one out of context produces nonsense. `readHead()` now inspects the first frame and skips the file.
- **Bug C** — Dogfood-H found 92% (169/184) of `~/.claude/projects/` entries on the user's real machine were agentory-spawned temporary dirs (paths like `<tmpdir>/agentory-…`). New `isAgentoryTempCwd` heuristic filters them in the scan loop.

## Reverse-verify (each fix independently)

For each of the three fixes, the relevant call site was temporarily disabled and the extended probe re-run:

| Bug | Disabled | Probe result |
|---|---|---|
| A | `importSession` skips the `loadImportHistory` IPC (ImportDialog stops passing `projectDir`) | `FAIL: Bug A regression: imported session did not hydrate history blocks from .jsonl` |
| B | `readHead` no longer calls `isSidechainFrame()` on the first frame | `FAIL: Bug B regression: sub-agent (parentUuid) transcript leaked into import list` |
| C | `scanImportableSessions` skips the `isAgentoryTempCwd(head.cwd)` filter | `FAIL: Bug C regression: agentory-temp cwd transcript leaked into import list` |

After restoring all three: probe passes all three cases cleanly.

## Test plan

- [x] `npm run build` — clean (renderer + electron tsc, no errors)
- [x] `npm test` — 622/634 pass; the 12 failures are pre-existing better-sqlite3 NODE_MODULE_VERSION 130 vs 127 mismatch, unrelated to this PR (same failures on `working`).
- [x] `tests/import-scanner.test.ts` — extended with parentUuid + isSidechain + isAgentoryTempCwd cases (40 tests pass)
- [x] `tests/store.test.ts` — extended with `framesToBlocks` cases (79 tests pass)
- [x] `scripts/probe-e2e-import-session.mjs` — **extended in place** per `feedback_e2e_extend_existing.md` (no new probe file). Covers all 3 bugs with planted fixtures under sandboxed HOME (per `project_probe_skill_injection.md`). Passes.
- [x] `node scripts/harness-agent.mjs` — 9/9 pass (one flake on first run, clean on retry; confirmed unrelated — flakes are in streaming caret rendering, no code touched here)
- [x] `node scripts/harness-ui.mjs` — 5/5 pass
- [x] `node scripts/harness-perm.mjs` — 9/9 pass

## Files touched

- `electron/import-history.ts` (new) — main-side `.jsonl` reader with path-traversal guards
- `electron/import-scanner.ts` — `isSidechainFrame` + `isAgentoryTempCwd` exported helpers, applied in `readHead` / `scanImportableSessions`
- `electron/main.ts` — register `import:loadHistory` IPC handler
- `electron/preload.ts` + `src/global.d.ts` — bind/type `loadImportHistory`
- `src/stores/store.ts` — `framesToBlocks` reducer + async history hydration in `importSession`
- `src/components/ImportDialog.tsx` — pass `projectDir` to `importSession`
- `tests/import-scanner.test.ts`, `tests/store.test.ts` — unit coverage
- `scripts/probe-e2e-import-session.mjs` — extended with all 3 cases + sandboxed HOME

### DIFF
--- a/electron/import-history.ts
+++ b/electron/import-history.ts
@@ -1,0 +1,78 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import * as readline from 'readline';
+
+const PROJECTS_ROOT = path.join(os.homedir(), '.claude', 'projects');
+
+// Cap on the number of frames we'll return to the renderer. Some historical
+// transcripts are huge (hundreds of MB / tens of thousands of frames). We
+// already accept this on the agent stream path because the user is the one
+// generating it, but on import we're asked to materialize the whole thing
+// at once. Cap defensively — well past anything a typical session generates,
+// while still bounding memory for pathological cases.
+const MAX_FRAMES = 50_000;
+
+// `~/.claude/projects/<encoded-cwd>/<sessionId>.jsonl` — read line-by-line,
+// JSON-parse each, return the resulting frames. Filename is constrained to
+// the sessionId we got back from `import:scan`, which is itself derived from
+// a directory listing of `~/.claude/projects/<encoded-cwd>/`, so there's no
+// renderer-controlled path component beyond the projectDir we already
+// accepted at scan time.
+//
+// We deliberately do NOT do block translation here — that lives in the
+// renderer (`stream-to-blocks`) and is shared with the live agent path.
+// Keeping this main-side read as raw-frames means the conversion logic has
+// exactly one home.
+export async function loadImportableHistory(
+  projectDir: string,
+  sessionId: string
+): Promise<unknown[]> {
+  // Reject any path component that would let a malicious renderer break out
+  // of `~/.claude/projects/`. `path.basename` strips traversal but accepting
+  // an unsanitized projectDir would still let `..` slip past on Windows. We
+  // lock both to a leaf segment.
+  const safeProj = path.basename(String(projectDir ?? ''));
+  const safeSid = path.basename(String(sessionId ?? ''));
+  if (!safeProj || !safeSid) return [];
+  if (safeProj.includes('..') || safeSid.includes('..')) return [];
+  const file = path.join(PROJECTS_ROOT, safeProj, `${safeSid}.jsonl`);
+  // Ensure the resolved path is still inside PROJECTS_ROOT — defense in depth
+  // against any encoding tricks `basename` might miss on a future Node.
+  const resolved = path.resolve(file);
+  if (!resolved.startsWith(path.resolve(PROJECTS_ROOT) + path.sep)) return [];
+  let stat: fs.Stats;
+  try {
+    stat = await fs.promises.stat(resolved);
+  } catch {
+    return [];
+  }
+  if (!stat.isFile()) return [];
+
+  return new Promise((resolve) => {
+    const stream = fs.createReadStream(resolved, { encoding: 'utf8' });
+    const rl = readline.createInterface({ input: stream, crlfDelay: Infinity });
+    const out: unknown[] = [];
+    let count = 0;
+    const finish = () => {
+      rl.removeAllListeners();
+      stream.destroy();
+      resolve(out);
+    };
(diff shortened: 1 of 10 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE (posted as comment — self-PR blocks formal approve)

**Reverse-verify performed (Bug B — sub-agent / sidechain filter)**
- Neutered `isSidechainFrame` to always return `false` in `electron/import-scanner.ts`.
- `npx vitest run tests/import-scanner.test.ts` → 4 tests FAIL (the two `isSidechainFrame` direct cases plus both `parseHead` integration cases for `parentUuid` and `isSidechain: true`).
- Restored original; 40/40 tests pass.
- Conclusion: filter is genuinely load-bearing, not vacuous.

**Schema validation against real `~/.claude/projects/`**
- Sampled real CLI sub-agent transcripts on this machine: line-1 reliably carries `isSidechain: true` (sometimes also non-null `parentUuid`). Field names in the filter match the actual CLI schema.
- One observation: some real sessions begin with a non-conversation frame (e.g. `{type:'summary',operation,...}`) that lacks both fields. The current `firstFrameInspected` logic short-circuits on this frame as "not sidechain" and continues. That's the right behavior for the goal (sub-agents always mark line 1), but worth flagging if future schema drift moves `isSidechain` to a later frame.

**Code review highlights**
- `electron/import-history.ts`: clean. Path traversal hardening (basename + `..` check + `path.resolve` prefix check) is solid; `MAX_FRAMES = 50_000` cap; malformed lines skipped (try/catch around `JSON.parse`); stream errors finish gracefully. No crash paths found.
- `framesToBlocks` reuses `streamEventToTranslation` (single source of truth — good). Coalesces by `id` to dedupe; user-text frames handled separately with `<command-...>` wrapper stripped to match scanner's title heuristic.
- `saveMessages` is `DELETE FROM messages WHERE sessionId=? + INSERT…` in a transaction → fully idempotent. Re-importing the same session into the same store id overwrites cleanly; importing twice via `nextId('s')` makes two independent local rows (expected — that's a UX choice, not a bug).
- `isAgentoryTempCwd`: cross-platform handled correctly (os.tmpdir + `/tmp/`, `/private/tmp/`, `/var/folders/`, AppData/Local/Temp regex), case-insensitive on Windows. The `/(^|\/)agentory-/` segment guard prevents false positives on user dirs like `my-agentory-project`.
- `cwd` encoding (`/`, `\`, `:` → `-`): not re-encoded in fix code — `projectDir` flows through verbatim from the scanner's `readdir`, so encoding correctness is moot for this PR.

**Scope**
- New file `electron/import-history.ts` extracted from main.ts is a clean separation, not creep.
- All other changes confined to: `import-scanner.ts`, `main.ts`, `preload.ts`, `global.d.ts`, `store.ts`, `ImportDialog.tsx`, plus tests + probe extension. Tight.

**Build**: `npm run build` passes (only pre-existing bundle-size warnings).

**Probe**: `scripts/probe-e2e-import-session.mjs` extends the existing file with three labeled cases (Bug A/B/C) — complies with `feedback_e2e_extend_existing.md`.

Nothing blocking. Ship it.

PR #204, opened 2026-04-24
fix(popover): global openPopoverId mutex to auto-close on cross-popover click (#221)

Jiahui-Gu:
## Summary

Fixes the systemic popover mutual-exclusion bug (issue #221): opening CwdPopover and then clicking the model selector left CwdPopover stuck open. Audit found the root cause is structural — every popover owned its own local open state, so siblings had no way to dismiss each other.

**Approach:** lift a single global `openPopoverId: string | null` slot to the zustand store. Each popover binds its open flag to `useStore(s => s.openPopoverId === '<my-id>')`. Opening any popover sets the slot (implicitly closing whichever was previously open); closing only clears when the requested id matches (idempotent — a stale unmount cleanup can't clobber the new owner).

## Files changed

- `src/stores/store.ts` — `openPopoverId` state + `openPopover(id)` / `closePopover(id)` actions. NOT in `PERSISTED_KEYS` (popover state must not survive reload — `persisted-keys-source-of-truth` guard already enforces).
- `src/components/CwdPopover.tsx` — replaces local `useState(open)` with the store binding. Click-outside + Escape listener still calls `closePopover('cwd')` as a fallback.
- `src/components/StatusBar.tsx` — `ChipMenu` (model + permission) converted to controlled Radix `DropdownMenu` with `open`/`onOpenChange`. `modal={false}` so clicks on sibling status-bar chips reach their triggers and route through the mutex (default modal=true installs a body pointer-events guard that swallows cross-chip clicks).
- `src/components/InputBar.tsx` — slash picker joins the mutex via two effects: claim/release on `pickerOpen`, dismiss when `openPopoverId` moves to a non-slash owner.
- `tests/store.test.ts` — 6 new cases for openPopover/closePopover semantics.
- `tests/cwd-popover.test.tsx` — reset `openPopoverId` in `beforeEach` (shared store).
- `scripts/harness-ui.mjs` — new `popover-cross-dismiss` case (extended existing harness per `feedback_e2e_extend_existing.md`).

## E2E case — `popover-cross-dismiss`

1. Open cwd popover, assert visible + `openPopoverId === 'cwd'`.
2. Click model chip → assert cwd popover hidden, model menu visible, `openPopoverId === 'model'`.
3. Re-click cwd chip → assert model menu gone, cwd popover visible, `openPopoverId === 'cwd'`.
4. Escape → assert both gone, `openPopoverId === null`.

## Reverse-verify

`git stash push -- src/stores/store.ts` (only) → rebuild → run `node scripts/harness-ui.mjs --only=popover-cross-dismiss`:

- **Without the fix:** FAIL at step 1 (`locator('[data-cwd-chip]')` never visible — renderer crashes during render because `openPopover` / `closePopover` are undefined).
- **Restore stash + rebuild:** PASS.

## Test counts

- vitest: 640 passed / 12 pre-existing better-sqlite3 native ABI failures in `electron/__tests__/db*.test.ts` (unrelated env issue).
- harness-ui: 6/6 (added `popover-cross-dismiss`).
- harness-agent: 9/9.
- harness-perm: 9/9.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` (only pre-existing native module failures remain)
- [x] `node scripts/harness-ui.mjs` — 6/6
- [x] `node scripts/harness-agent.mjs` — 9/9
- [x] `node scripts/harness-perm.mjs` — 9/9
- [x] Reverse-verify: stash store.ts → harness FAIL; restore → PASS

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -7,12 +7,13 @@
 // excluded from the per-file runner via scripts/run-all-e2e.mjs's
 // MERGED_INTO_HARNESS skip list.
 //
-// Scope (5 cases — pure UI / store-driven, no real claude.exe required):
+// Scope (6 cases — pure UI / store-driven, no real claude.exe required):
 //   - sidebar-align             (probe-e2e-sidebar-align)
 //   - no-sessions-landing       (probe-e2e-no-sessions-landing)
 //   - empty-state-minimal       (probe-e2e-empty-state-minimal)
 //   - a11y-focus-restore        (probe-e2e-a11y-focus-restore)
 //   - shortcut-overlay-opens    (new — UI-1 / #188)
+//   - popover-cross-dismiss     (new — popover-mutex / #221)
 //
 // Related UI probes already absorbed into harness-agent.mjs:
 //   - inputbar-visible, chat-copy, input-placeholder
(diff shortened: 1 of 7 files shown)


Jiahui-Gu (commented):
## Reviewer verdict: APPROVE

Independent reviewer pass (cannot self-approve via GitHub, hence comment). Verified in worktree `.claude/worktrees/popover-mutex` at b843fe8.

### Checks
- **Persistence**: `openPopoverId` is NOT in `PERSISTED_KEYS` (src/stores/persist.ts L35-50). Field JSDoc explicitly calls this out. No reload-resurrected popover risk.
- **Scope**: 5 source files (store.ts, CwdPopover.tsx, InputBar.tsx, StatusBar.tsx) + 2 test files + harness-ui.mjs. Matches plan exactly, no creep.
- **Idempotency for stale ids**: `closePopover(id)` is `set((s) => s.openPopoverId === id ? { openPopoverId: null } : s)`. Calling `closePopover('cwd')` while `openPopoverId === 'model'` is a true no-op (same state object returned). Safe against the two-effect race in InputBar.tsx where pickerOpen=false fires after another popover has already claimed the slot. `openPopover` is also idempotent.
- **`modal={false}` claim**: confirmed — both ChipMenu instances (model + permission) inherit it via the shared `ChipMenu` component. The flip is genuinely load-bearing: Radix's default modal=true installs a body pointer-events guard via FocusScope that swallows cross-chip clicks before openPopoverId can flip. No focus-trap regression for normal use: arrow-key navigation (roving tabindex) and Escape-to-close are independent of `modal`; Tab now moves focus out of the menu and fires `onOpenChange(false)` → `closePopover`, which is desirable for a popup (not a Dialog). DropdownMenu's contract was never "trap focus".
- **Reverse-verify**: swapped src/stores/store.ts to HEAD~1 baseline → `node scripts/harness-ui.mjs --only=popover-cross-dismiss` FAILED (5.6s). Restored fixed store.ts → PASSED (3.3s, "cwd→model→cwd cross-dismiss + Escape clears mutex slot"). Probe correctly discriminates the fix.
- **Build**: `npm run build` PASSED (only pre-existing bundle-size warnings).

### Notes (non-blocking)
- Reverse-verify failure mode was "data-cwd-chip never visible" rather than the assertion firing — slightly less surgical than mutating the action body, but probe still detects absence of the fix. Acceptable.
- No visual changes in the diff; pure behavior fix, no before/after screenshots required.

Approved for merge — please have the author or another maintainer click the green button.

PR #206, opened 2026-04-24
fix(agent): validate persisted claudeBinPath, self-heal stale value (P0 blocker)

Jiahui-Gu:
**P0 — first message fails on prod install. Reviewer: please prioritize.**

## Summary
- `electron/main.ts` `agent:start` was forwarding the persisted `claudeBinPath` from SQLite to `spawnClaude()` without checking whether the file still exists.
- A stale entry (e.g. a fake `.cmd` left behind by a dev probe whose temp dir was GC'd, like `C:\...\Temp\probe-cli-missing-Aesr8Z\fake-claude.cmd`) bypasses `resolveClaudeBinary()` entirely. On Windows the `.cmd` extension routes through `cmd.exe`, which exits with `"The system cannot find the path specified"`. `CLAUDE_NOT_FOUND` is **never** raised, so `CliMissingDialog` never fires and the user is permanently stuck.
- Reverse-verify revealed an even nastier surface: without the fix, `agent:start` returns `ok:true` (the spawn-side error doesn't propagate to start), so there is no banner either — silent failure.
- Fix at the call site: validate `existsSync` and self-heal the dead row. Only one unsafe consumer; `cli:retryDetect` already does this check (electron/main.ts:967).

## Diff
- `electron/main.ts` (+13/-2): existsSync guard + `saveClaudeBinPath(null)` self-heal in the `agent:start` handler.
- `scripts/probe-cli-missing.mjs` (+140/-5): extend the existing CLI-missing probe with **Case B** — pre-seed `claudeBinPath` via `saveState` IPC, drive `agent:start`, assert `errorCode === 'CLAUDE_NOT_FOUND'` and DB row cleared. HOME/USERPROFILE sanitized per `project_probe_skill_injection.md`.

## Reverse-verify (per `feedback_bugfix_requires_e2e.md`)
| step | result |
|---|---|
| `git stash push electron/main.ts` (revert fix only) | done |
| `npx tsc -p tsconfig.electron.json` | done |
| `node scripts/probe-cli-missing.mjs` | **FAIL** as expected: `agent:start unexpectedly succeeded with stale binPath + empty PATH; expected ok:false errorCode:CLAUDE_NOT_FOUND` |
| `git stash pop` + rebuild | restored |
| `node scripts/probe-cli-missing.mjs` | **PASS** both Case A + Case B |

## Harness counts
- agent: **9/9** pass
- ui: **5/5** pass
- perm: **8/9** pass — `permission-a11y` flake is pre-existing on `working` baseline (verified by stashing the fix and re-running). Not introduced by this PR.

## Test plan
- [x] Existing `probe-cli-missing.mjs` Case A still passes
- [x] New Case B passes with fix
- [x] Reverse-verified: Case B fails without fix
- [x] All three harnesses run; only pre-existing failures remain
- [ ] Reviewer: confirm reverse-verify by repeating stash/probe locally

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -689,7 +689,18 @@ app.whenReady().then(() => {
         };
       }
 
-      const binaryPath = loadClaudeBinPath() ?? undefined;
+      // Validate the persisted path exists on disk. A stale entry (e.g. from a
+      // dev probe whose temp dir was GC'd) would otherwise be passed to the
+      // spawner verbatim, bypassing resolveClaudeBinary() and producing an
+      // opaque "system cannot find the path specified" exit instead of the
+      // CLI-missing dialog. Self-heal by clearing the dead value so subsequent
+      // launches fall through to PATH lookup / first-run wizard.
+      const persisted = loadClaudeBinPath();
+      const binaryPath =
+        persisted && fs.existsSync(persisted) ? persisted : undefined;
+      if (persisted && !binaryPath) {
+        saveClaudeBinPath(null);
+      }
 
       const result = await sessions.start(sessionId, {
         ...opts,
--- a/scripts/probe-cli-missing.mjs
+++ b/scripts/probe-cli-missing.mjs
@@ -1,12 +1,23 @@
 // Live probe for the "Claude CLI not found" first-run wizard.
 //
-// Runs against the built app with AGENTORY_CLAUDE_BIN pointing at a non-
-// existent file so the resolver falls back to a scrubbed PATH and flips the
-// store into `missing` state. Asserts:
+// Case A (original): Runs against the built app with AGENTORY_CLAUDE_BIN
+// pointing at a non-existent file so the resolver falls back to a scrubbed
+// PATH and flips the store into `missing` state. Asserts:
 //   - The blocking modal renders with title + body.
 //   - OS-appropriate install commands are visible.
 //   - Copy button writes the command to the clipboard.
 //   - Browse (stubbed via dialog monkey-patch) succeeds and closes the modal.
+//
+// Case B (stale persisted binPath, P0 regression — see fix/stale-binpath-p0):
+//   - Pre-seeds the SQLite `app_state` table with `claudeBinPath` pointing
+//     at a path that does not exist on disk (mimics a dev probe whose temp
+//     dir was GC'd).
+//   - Calls `agent:start` directly with empty PATH.
+//   - Asserts `errorCode === 'CLAUDE_NOT_FOUND'` (NOT a generic spawn
+//     error) and that the stale row was self-healed (DB row cleared).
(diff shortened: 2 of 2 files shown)


Jiahui-Gu (commented):
**VERDICT: APPROVE** (independent reviewer; self-PR so cannot use `--approve`)

P0, time-boxed independent review. Tight, surgical fix; the e2e is the right shape.

**Verified:**
- `electron/main.ts:692-703` — `fs.existsSync(persisted)` is correct (sync I/O, one-shot per `agent:start`, no race). Falls through to `undefined` so `sessions.start` invokes `resolveClaudeBinary()` → raises `ClaudeNotFoundError` → maps to `errorCode: 'CLAUDE_NOT_FOUND'` → CliMissingDialog shows. Self-heal via `saveClaudeBinPath(null)` is the right API (`electron/db.ts:279` deletes the `app_state` row on null/empty, so subsequent `loadState('claudeBinPath')` returns `null`).
- **Call-site audit:** grep'd `loadClaudeBinPath` across `electron/`. Two non-test consumers: this handler (now safe) and `cli:retryDetect` at `electron/main.ts:978` (already wraps in `fs.existsSync`). No other unsafe forwarder. Confirmed worker's claim.
- **Probe Case B (`scripts/probe-cli-missing.mjs:166-291`):** seeds via `saveState` IPC (correct pivot — same `app_state` row the production wizard writes through `saveClaudeBinPath`); empty PATH + cleared `PATHEXT` + deleted `AGENTORY_CLAUDE_BIN` so the resolver cannot mask the bug; asserts `errorCode === 'CLAUDE_NOT_FOUND'` exact-string and `loadState('claudeBinPath') === null` post-call; HOME/USERPROFILE redirected to a fresh tmpdir per `project_probe_skill_injection.md`. `fail()` exits with code 1 — CI will catch any regression.
- Worker's documented reverse-verify (stash main.ts → Case B fails with the cmd.exe "system cannot find the path specified" symptom → restore → passes) is structurally consistent with the diff; trusted within the 12-min cap rather than re-running the full electron-builder.

**Surfaced bug worth a follow-up (do not block this PR):** Worker correctly flagged that `agent:start` returns `ok: true` when the underlying `child_process.spawn` dies post-spawn with a generic "ENOENT/path not found" — the error doesn't propagate through `sessions.start`'s success path. This PR fixes the surfaced symptom (CLI-missing flow now works end-to-end). The deeper spawn-error-propagation gap should be a separate task; if the user picks any *other* non-existent binary down the road, we'd silently swallow it again.

**Nits (non-blocking):**
- The 6-line comment block at `main.ts:692-697` is excellent — keep this style.
- Minor: the three repeated `app2.close()` + `ud.cleanup()` + `fs.rmSync(fakeHome,…)` cleanup blocks in the probe could be a `try/finally`, but it's a 100-line probe — not worth churn on a P0.

LGTM, ship it.

PR #209, opened 2026-04-24
fix(agent): propagate post-spawn CLI failures as CLI_SPAWN_FAILED (#235)

Jiahui-Gu:
## Summary

PR #206 reviewer flagged that `agent:start` swallowed post-spawn child_process errors and returned `{ok:true}` even when claude.exe exited immediately with code != 0. The stale-binPath P0 (#224) only surfaced because the user noticed manually — the underlying class of bug (stale env, missing dep, bad shim, ENOENT after spawn) had no diagnostic surface.

This PR wires spawn-error propagation:

- `SessionRunner.start()` now waits up to **800ms** after `spawn()` for the child to either survive the window or die noisily. Early non-zero exits throw a typed `ClaudeSpawnFailedError` carrying the stderr tail.
- `manager.start()` translates that into `{ok:false, errorCode:'CLI_SPAWN_FAILED', detail:<stderr-tail>}`.
- `startSession.ts` (renderer) tacks the `detail` onto the message that lands in `sessionInitFailures`. The existing `AgentInitFailedBanner` picks it up and renders **"Agent failed to start"** with the captured stderr below.
- One-line user-visible message: **`Agent failed to start — claude.exe exited immediately after spawn (code=1) — <stderr tail>`**.

No new UI patterns; the failure flows through the existing banner trio (PR #237).

## Files

- `electron/agent/sessions.ts` — adds `ClaudeSpawnFailedError`, `SPAWN_EARLY_FAILURE_WINDOW_MS`, `detectEarlyFailure()`, and the post-spawn race in `start()`.
- `electron/agent/manager.ts` — translates the typed error to the IPC reply.
- `electron/preload.ts`, `src/global.d.ts` — extend `StartResult.errorCode` union with `'CLI_SPAWN_FAILED'` and add `detail?: string`.
- `src/agent/startSession.ts` — appends `detail` to the banner copy.
- `scripts/probe-spawn-error-propagation.mjs` — new e2e probe.

## Test plan

- [x] `npx vitest run electron/agent/__tests__/sessions.test.ts` → 23/23 pass
- [x] `npx tsc --noEmit` → clean
- [x] `node scripts/probe-spawn-error-propagation.mjs` → **PASS** (phase 1: IPC `ok:false errorCode:CLI_SPAWN_FAILED detail` contains sentinel; phase 2: `AgentInitFailedBanner` renders with stderr tail)
- [x] **Reverse-verify**: editing `electron/agent/sessions.ts` to comment out the `detectEarlyFailure` block (so `start()` returns happily even when the child died) makes the probe FAIL with `agent:start returned ok:true despite the fake binary exiting 1 immediately`. Restoring the block makes it PASS again. Confirmed locally on Windows with the committed 800ms window — initial 200ms was too tight for the cmd.exe shim wrap-up under Electron + Playwright + Windows job scheduling, hence the bumped budget (still well under any user-perceptible "starting…" time).

## Reviewer

Per project rules, this PR requires an independent reviewer subagent before merge. Reviewer should:
1. Re-run `node scripts/probe-spawn-error-propagation.mjs` on a clean build (`npx tsc -p tsconfig.electron.json && npx webpack --mode production && npx @electron/rebuild -f -w better-sqlite3`).
2. Reverse-verify: temporarily replace the `if (earlyFailure) { ... throw new ClaudeSpawnFailedError(...) }` block in `electron/agent/sessions.ts` with a no-op, rebuild electron only (`npx tsc -p tsconfig.electron.json`), re-run probe — must FAIL with `ok:true`.
3. Confirm no regression: `npx vitest run electron/agent/__tests__/sessions.test.ts`.

### DIFF
--- a/electron/agent/__tests__/sessions.test.ts
+++ b/electron/agent/__tests__/sessions.test.ts
@@ -39,6 +39,14 @@ function makeFakeProc(): FakeProc {
   child.stderr = new PassThrough();
   child.stdin = new PassThrough();
 
+  // Pre-buffer a single zero-length keepalive frame so the new
+  // detectEarlyFailure (PR #209 P1 fix) fires `'readable'` immediately when
+  // SessionRunner.start() attaches its listener — otherwise every success-path
+  // test would wait the full SPAWN_EARLY_FAILURE_WINDOW_MS (~800ms) before
+  // start() resolves. The blank line is dropped by splitNDJSON, so the
+  // consumer sees no event from it.
+  child.stdout.write('\n');
+
   const stdinChunks: Buffer[] = [];
   child.stdin.on('data', (c) => stdinChunks.push(Buffer.isBuffer(c) ? c : Buffer.from(String(c))));
 
--- a/electron/agent/manager.ts
+++ b/electron/agent/manager.ts
@@ -1,6 +1,9 @@
 import type { WebContents } from 'electron';
 import { SessionRunner, ClaudeSpawnFailedError, type StartOptions, type PermissionMode, type AgentMessage } from './sessions';
 import { ClaudeNotFoundError } from './binary-resolver';
+import type { StartResult } from './start-result-types';
+
+export type { StartErrorCode, StartResult } from './start-result-types';
 
 type Sender = (channel: string, payload: unknown) => void;
 
@@ -16,23 +19,6 @@ export type AgentDiagnostic = {
   code: string;
   message: string;
 };
-
-export type StartResult =
-  | { ok: true }
-  | {
-      ok: false;
-      error: string;
-      errorCode?: 'CLAUDE_NOT_FOUND' | 'CWD_MISSING' | 'CLI_SPAWN_FAILED';
-      searchedPaths?: string[];
-      /**
-       * Tail of stderr (or the libuv error message) captured during the
-       * early-failure window. Populated for `CLI_SPAWN_FAILED` so the
-       * renderer banner can show the user *why* the CLI bailed (missing
-       * dependency, bad shim, stale env, etc.) instead of an opaque
-       * "agent failed to start".
-       */
-      detail?: string;
-    };
 
 class SessionsManager {
   private runners = new Map<string, SessionRunner>();
--- a/electron/agent/manager.ts
+++ b/electron/agent/manager.ts
@@ -1,5 +1,5 @@
 import type { WebContents } from 'electron';
-import { SessionRunner, type StartOptions, type PermissionMode, type AgentMessage } from './sessions';
+import { SessionRunner, ClaudeSpawnFailedError, type StartOptions, type PermissionMode, type AgentMessage } from './sessions';
 import { ClaudeNotFoundError } from './binary-resolver';
 
 type Sender = (channel: string, payload: unknown) => void;
@@ -22,8 +22,16 @@ export type StartResult =
   | {
       ok: false;
       error: string;
-      errorCode?: 'CLAUDE_NOT_FOUND' | 'CWD_MISSING';
+      errorCode?: 'CLAUDE_NOT_FOUND' | 'CWD_MISSING' | 'CLI_SPAWN_FAILED';
       searchedPaths?: string[];
+      /**
+       * Tail of stderr (or the libuv error message) captured during the
+       * early-failure window. Populated for `CLI_SPAWN_FAILED` so the
+       * renderer banner can show the user *why* the CLI bailed (missing
+       * dependency, bad shim, stale env, etc.) instead of an opaque
+       * "agent failed to start".
+       */
+      detail?: string;
     };
 
 class SessionsManager {
@@ -103,6 +111,14 @@ class SessionsManager {
           error: err.message,
           errorCode: 'CLAUDE_NOT_FOUND',
           searchedPaths: err.searchedPaths,
+        };
+      }
+      if (err instanceof ClaudeSpawnFailedError) {
+        return {
+          ok: false,
+          error: err.message,
+          errorCode: 'CLI_SPAWN_FAILED',
+          detail: err.detail,
         };
       }
       return { ok: false, error: err instanceof Error ? err.message : String(err) };
--- a/electron/agent/sessions.ts
+++ b/electron/agent/sessions.ts
@@ -159,6 +159,54 @@ const HOOK_PASSTHROUGH_TOOLS: ReadonlySet<string> = new Set([
   'ExitPlanMode',
 ]);
 
+/**
+ * How long we wait, after `spawn()` returns, before declaring the child has
+ * "successfully started". If the OS hands us a pid but the child immediately
+ * blows up (`error` event for ENOENT/EACCES, or `exit` with non-zero code
+ * inside this window), we want to surface that as a typed
+ * `CLI_SPAWN_FAILED` instead of a chirpy `{ ok: true }` followed by a silent
+ * `agent:exit`. 200ms empirically covers Windows shim + cmd.exe wrap-up
+ * while staying well under any user-visible "starting…" budget — the CLI's
+ * first stdout frame normally lands in <50ms, so a process still alive at
+ * 200ms is overwhelmingly likely to be healthy.
+ *
+ * Why we *don't* race against "first stdout byte": adding a `'data'`
+ * listener to the spawner's stdout pipe puts it into flowing mode and
+ * conflicts with the consumer's `for await` async iterator added below.
+ * Stderr already has a ring-buffer listener inside the spawner, but adding
+ * a parallel timing channel there for a 200ms optimisation isn't worth the
+ * second-mover-wins coordination it would add.
+ */
+const SPAWN_EARLY_FAILURE_WINDOW_MS = 800;
+
+/**
+ * Thrown by `SessionRunner.start()` when the child process emits `exit`
+ * with a non-zero code (or an `error` event, which the spawner surfaces as
+ * exitCode -1) inside the early-failure window. Carries enough context for
+ * the renderer to render an actionable banner: a short reason + the tail of
+ * whatever the child managed to write to stderr before dying.
+ *
+ * `manager.start()` translates this into a `{ ok: false, errorCode:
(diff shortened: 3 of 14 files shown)


Jiahui-Gu (commented):
## Verdict: REQUEST CHANGES (P1 + branding nits) — posted as comment because GH blocks self-review

Independent reviewer running on Windows 11, Electron-rebuilt better-sqlite3.

### Reverse-verify (mandatory) — DONE
- Fix applied: probe phase 1 PASS (`agent:start` returns `ok:false errorCode:CLI_SPAWN_FAILED`, detail contains sentinel `fake-claude-bailing-with-code-1`).
- Fix reverted (`git checkout origin/working -- electron/agent/sessions.ts electron/agent/manager.ts`, rebuild electron): probe FAIL with exact expected message `agent:start returned ok:true despite the fake binary exiting 1 immediately`.
- Restored fix: phase 1 PASS again.

Reverse-verify protocol satisfied. Worker's claim that they could not run it locally was unfounded — `git checkout HEAD -- <files>` from the working baseline works fine.

### P1 — `detectEarlyFailure` blocks the success path for the full 800ms

`SessionRunner.start()` does `await this.detectEarlyFailure(this.cp)` synchronously before constructing `ControlRpc` (sessions.ts:315). `detectEarlyFailure` races `cp.wait()` (only resolves on exit) against an 800ms timeout. On the **happy path** `wait()` never resolves inside the window, so the timeout always wins → **every successful agent:start is delayed 800ms**.

Evidence: `sessions.test.ts` now reports each of the 23 cases at ~810ms (40.76s total). Previously these were sub-second. The instructions you wrote in the comment say `the detectEarlyFailure listener should ONLY hold up the resolve if the child emits error/exit; on success path it should resolve immediately on first stdout/stderr byte` — current code does the opposite. PR description likewise claims `well under any user-perceptible "starting…" time` but 800ms is itself the floor on every start.

Fix options: (a) race against the spawner's first-stdout/stderr signal as well — needs a non-flowing peek hook in `claude-spawner` since the consumer's `for await` reader takes ownership; or (b) make detectEarlyFailure fire-and-forget, throw asynchronously via `onExit` if it trips, and let `start()` return immediately.

### P2 — banner branding still says "Agent"
Post-rename the product is **ccsm** but `AgentInitFailedBanner.tsx:65` still hardcodes `"Agent failed to start"`. Sentence case is fine (no SCREAMING-UI rule violation), but recommend `"Failed to start Claude"` or `"ccsm couldn't start the Claude CLI"` for consistency with the rebrand. Not a merge blocker on its own but should ride along.

### Nits
- `SPAWN_EARLY_FAILURE_WINDOW_MS = 800` but the docstring above it (sessions.ts:166) and the `code === 0` comment (sessions.ts:265) both still say `200ms`. Stale comments — fix together with P1.
- 800ms vs 200ms call: the empirical-Windows-shim justification is plausible, but the comment never got updated. Once P1 is addressed (success path no longer pays the cost) the value matters less; 800ms is fine for a failure-only timer.
- `errorCode: 'CLI_SPAWN_FAILED'` is a string-literal union duplicated across 4 files (`manager.ts`, `preload.ts`, `src/global.d.ts`, `startSession.ts`). Worker should extract to a single shared `StartErrorCode` type — not a blocker, but flagged for the next refactor.
- Probe sanitizes `HOME`/`USERPROFILE` to a temp dir and empties `PATH` per the skill-injection rule — good.
- Probe phase 2's `[data-agent-init-failed-banner]` visibility check is shadowed by the global ClaudeNotFound full-screen wizard (the banner IS rendered with the right copy in the body dump, but Playwright's `visible` returns false because the wizard overlays it). The fix copy works; this is a probe deficiency that can be tracked separately.

### Build/lint/types
- `npx tsc --noEmit`: clean.
- `npx tsc -p tsconfig.electron.json`: clean.
- `npx eslint` on the changed files: clean (repo-wide lint has 224k pre-existing errors unrelated to this PR).
- `npm test`: 14 pre-existing failures all from `better-sqlite3` NODE_MODULE_VERSION mismatch (env issue, fixed by `@electron/rebuild`). `sessions.test.ts` 23/23 PASS — but each ~810ms slower than baseline (P1 evidence above).

### Asks before merge
1. Address P1 (success path must not pay 800ms).
2. Update stale "200ms" comments.
3. Decide on the "Agent failed to start" → ccsm branding wording.

Jiahui-Gu:
Pushed `29b930a` addressing review feedback. Re-verify pls.

### P1 — `detectEarlyFailure` success-path latency
Replaced the `cp.wait()` vs 800ms-timer race with a three-way race against the child's first **stdout** byte. Stderr is intentionally not a signal — a binary that prints a warning and then exits 1 (the probe's fake binary does exactly this) would otherwise win against `cp.wait()` and falsely resolve as healthy. `'readable'` is guarded on `readableLength > 0` to ignore EOF wakeups, so the consumer's `for await (… of splitNDJSON(stdout))` reader still owns delivery and no bytes are consumed.

Evidence:
- `electron/agent/__tests__/sessions.test.ts`: **23/23 PASS in ~64-140ms total** (was ~40s = ~810ms/test before fix). Tests pre-write a `\n` keepalive to the fake `PassThrough` so `'readable'` fires immediately.
- `probe-spawn-error-propagation` extended (per `feedback_e2e_extend_existing.md`) with a new **phase 1.5** that seeds an alive fake binary and asserts `agent:start` resolves in <300ms.
  - Fix-applied: **193-229ms** ✓
  - **Reverse-verified**: stash the source change → phase 1.5 FAILs with elapsed **981ms** (expected — without the fix the timer wins on the alive binary). Restored → PASSes again.
- Phase 1 still PASSes (failure path: ok:false / CLI_SPAWN_FAILED / sentinel stderr).
- Phase 2 still flagged as the pre-existing probe deficiency you noted (banner shadowed by ClaudeNotFound full-screen wizard) — out of scope for this PR.

### P2 — banner branding
`AgentInitFailedBanner.tsx` title → `"Failed to start Claude"`. Updated all 8 string references across the repo (banner component, test assertion, probe assertions, harness, comments in store/TopBanner/sessions). Sentence case, no SCREAMING.

### Nits
- Stale `200ms` comments at `sessions.ts:166` (and the surrounding "why we don't race against first stdout byte" justification, now obsolete) updated. Comment at `:265` updated to reflect the new semantics.
- `StartErrorCode` + `StartResult` union extracted to `electron/agent/start-result-types.ts` and re-exported from `sessions.ts` / `manager.ts`; `preload.ts` imports it directly. `src/global.d.ts` mirrors the union with a pointer comment back to the source of truth — it's a renderer ambient `.d.ts` consumed by Vite, not the electron tsconfig, so it can't import from `electron/`.

### Build/lint/types
- `tsc -p tsconfig.electron.json`: clean.
- `tsc --noEmit` (renderer): only pre-existing `@radix-ui/...` / `@dnd-kit/...` resolution errors unrelated to this PR.

DO NOT MERGE — needs your re-verify.

Jiahui-Gu (commented):
## Re-review verdict: **Approve, ready to merge**

Independently verified all three review items on `29b930a`.

### P1 — `detectEarlyFailure` race
- `npx vitest run electron/agent/__tests__/sessions.test.ts`: **23/23 PASS in 56ms total** (worker claimed 64-140ms; my run is even faster). Pre-fix was ~40s, so the win is real.
- Code reads correctly: races `cp.stdout 'readable'` (with `readableLength > 0` guard against EOF wakeups) against `cp.wait()` against the 800ms timer. Stderr deliberately excluded.
- Stderr-excluded reasoning is sound: a binary that prints stderr and exits 1 (the failing-binary case) would otherwise win the race against `cp.wait()` and falsely resolve healthy. Real fatal stderr-then-exit still surfaces via the `cp.wait()` branch. Confirmed.
- `'readable'` (not `'data'`) keeps the stream paused so the consumer's `for await` reader still owns delivery — verified in code.

### Phase 1.5 probe — independent reverse-verify
- **Fix applied** (HEAD `29b930a`): phase 1.5 reports `agent:start success path resolved in 84ms (budget 300ms)`. PASS.
- **Fix reverted** (checkout sessions.ts/manager.ts/preload.ts/global.d.ts at parent commit `65b902a`, deleted start-result-types.ts, rebuilt): phase 1.5 reports `agent:start success-path took 852ms, exceeding the 300ms budget`. FAIL with the exact diagnostic message it should produce.
- Phase 2 banner-visibility shadowed by ClaudeNotFound wizard is confirmed out-of-scope per the original review (the banner DOM IS rendered with the right copy + sentinel, just covered).

### P2 branding
- `grep "Agent failed to start"` across src/electron/tests/scripts: **0 hits**.
- `grep "Failed to start Claude"` across src/electron: hits at `AgentInitFailedBanner.tsx:65`, `TopBanner.tsx:26` (doc), `sessions.ts:361` (doc), plus tests/probes/harness. Migration complete.

### Shared-type adoption
- `electron/agent/start-result-types.ts` defines `StartErrorCode = 'CLAUDE_NOT_FOUND' | 'CWD_MISSING' | 'CLI_SPAWN_FAILED'` and a discriminated `StartResult` union. Imported and used at all three call sites: `manager.ts` (return type of `start()`), `sessions.ts` (`ClaudeSpawnFailedError.code: StartErrorCode`), `preload.ts` (`agentStart: ... Promise<StartResult>`). No bare strings remain.
- `src/global.d.ts` mirrors the literal union with a clear pointer comment back to the source of truth (renderer .d.ts can't import from electron tree). Shape matches.

### Stale comments
- `sessions.ts:166-180` and `:265` comment blocks updated to reflect 800ms window + race-against-stdout semantics. Confirmed.

### Tooling
- `npx tsc --noEmit`: clean.
- `npm run lint` on PR-touched files: 0 errors (1 unrelated pre-existing warning at `store.ts:901`). Repo-wide lint errors all pre-exist in unrelated files.
- `npm test`: **660/660 PASS** (after `npm rebuild better-sqlite3` for local Node ABI; not a PR issue).

### Minor nit (non-blocking)
- `sessions.ts:168-169` — `SPAWN_EARLY_FAILURE_WINDOW_MS` doc still says "first byte arriving on the child's stdout **or stderr**". The implementation (and the more detailed comment at lines 285-296) correctly says stdout-only. Worth tightening on a follow-up commit but doesn't block merge.

LGTM, ship it.

PR #207, opened 2026-04-24
feat(chrome): unify banner trio behind shared TopBanner (#237)

Jiahui-Gu:
## Summary

UX Pro Max audit (#237) flagged 3 P0 issues in the top-of-pane banner trio (`AgentInitFailedBanner`, `AgentDiagnosticBanner`, `ClaudeCliMissingBanner`):

- inconsistent layout / spacing
- missing `role="alert"` + `aria-live="polite"` on two of the three
- divergent dismiss + action-button placement

This PR introduces `src/components/chrome/TopBanner.tsx` as the single source of truth for that surface and migrates the three banners to wrap it. Public component names + per-banner copy / CTAs are preserved (no consumer-side changes), so everything in `App.tsx` keeps importing the same symbols.

## What changed

- New `TopBanner` with API:
  ```tsx
  <TopBanner variant="error" | "warning" | "info" icon={...} title="..." body={...} actions={<>...</>} onDismiss={?} />
  ```
- `role="alert"` + `aria-live="polite"` baked in.
- Consistent padding, gap, divider, and dismiss button placement across all three variants.
- Motion uses the existing `DURATION_RAW.ms150 + EASING.enter` tokens from `src/lib/motion.ts` (no new motion presets).
- All copy is sentence-case (no SCREAMING UI strings), English-only.
- The three legacy components (`AgentInitFailedBanner`, `AgentDiagnosticBanner`, `ClaudeCliMissingBanner`) keep their public function names and their unique CTAs / dismiss labels — they internally compose `<TopBanner />`.

## A11y attrs added

| Banner | Before | After |
| --- | --- | --- |
| `AgentInitFailedBanner` | `role="alert"` only | `role="alert"` + `aria-live="polite"` |
| `AgentDiagnosticBanner` | `role="status"` only | `role="alert"` + `aria-live="polite"` + explicit dismiss `aria-label` |
| `ClaudeCliMissingBanner` | `role="status"` only | `role="alert"` + `aria-live="polite"` |

Plus `data-top-banner` / `data-variant` attributes for probes + future styling hooks.

## Visual diff (before / after)

Generated by `scripts/probe-render-banners.mjs` (Playwright + the project's design tokens; output committed under `dogfood-logs/banner-refactor-237/`).

| Banner | Before | After |
| --- | --- | --- |
| AgentInitFailedBanner | before | after |
| AgentDiagnosticBanner | before | after |
| ClaudeCliMissingBanner | before | after |

[...]

PR #210, opened 2026-04-24
fix(a11y): derive TopBanner role from variant (alert vs status)

Jiahui-Gu:
## Summary
- Address PR #207 reviewer feedback (Task #261): `TopBanner` applied `role=\"alert\"` to every variant. `alert` is assertive semantics meant for blocking failures; using it for warning/info caused screen readers to over-announce non-critical state.
- Derive role from `variant`:
  - `error` -> `role=\"alert\"`
  - `warning` | `info` -> `role=\"status\"`
- `aria-live=\"polite\"` is kept on both roles so newly-swapped banners still get re-announced.
- Updated JSDoc on `TopBanner` to document the role mapping; updated `title` prop comment to drop the now-misleading `role=\"alert\"` reference.

## Test plan
- [x] `npx vitest run tests/top-banner.test.tsx` -> 10 passed (added 2 new cases asserting warning/info expose `role=\"status\"` and NOT `alert`; updated 2 integration cases that previously queried the wrong role).
- [ ] No visual change -> no screenshots needed.

### DIFF
--- a/src/components/chrome/TopBanner.tsx
+++ b/src/components/chrome/TopBanner.tsx
@@ -25,7 +25,7 @@ export interface TopBannerProps {
   /**
    * Short, sentence-case headline (e.g. "Agent failed to start").
    * Required because every banner needs a scannable label for the
-   * `role="alert"` announcement.
+   * live-region announcement (role derived from `variant`).
    */
   title: string;
   /**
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Independent reviewer LGTM.

Verified:
- TopBanner.tsx: role derived per variant — error → 'alert', warning|info → 'status'. One-line ternary, applied to inner row div.
- aria-live='polite' retained on all variants.
- JSDoc updated to document the variant→role mapping clearly.
- tests/top-banner.test.tsx: 10/10 passing locally (npx vitest run). 2 new variant cases (warning, info) assert role='status' AND queryByRole('alert') is null — good negative assertion. 2 integration tests updated to getByRole('status') for warning-variant banners (AgentDiagnosticBanner, ClaudeCliMissingBanner).
- Spot-check: none of AgentDiagnosticBanner / AgentInitFailedBanner / ClaudeCliMissingBanner hardcodes role= — all delegate to TopBanner. (Unrelated chat/blocks/StatusBanner.tsx has its own role='status' but is a different component, out of scope.)
- No SCREAMING UI strings introduced.

No concerns. Approve.

PR #212, opened 2026-04-24
fix(i18n): require ConfirmDialog confirmLabel/cancelLabel (U5 #257)

Jiahui-Gu:
## Summary
- Make `confirmLabel` and `cancelLabel` required props on `ConfirmDialog` so TypeScript blocks any caller that would silently fall back to English `'Confirm'`/`'Cancel'` literals.
- Add `data-testid=\"confirm-dialog-confirm\"` and `data-testid=\"confirm-dialog-cancel\"` on the two action buttons for future e2e probes.
- Update the only existing caller (Sidebar delete-group dialog) to pass `t('common.cancel')` explicitly. `common.cancel` already exists in both en/zh locales.

Refs U5 audit task #257.

## Test plan
- [x] `npx tsc --noEmit` clean (renderer + electron tsconfigs)
- [x] `npm test` -> 660/660 passing
- [ ] Manual: open Sidebar -> right-click a normal group -> Delete; verify both buttons render localized text in zh and en.

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -291,6 +291,7 @@ function GroupRow({
             : t('sidebar.deleteGroupEmpty')
         }
         confirmLabel={t('sidebar.deleteGroup')}
+        cancelLabel={t('common.cancel')}
         destructive
         onConfirm={() => {
           const snap = deleteGroup(group.id);
--- a/src/components/ui/ConfirmDialog.tsx
+++ b/src/components/ui/ConfirmDialog.tsx
@@ -7,8 +7,10 @@ type ConfirmDialogProps = {
   onOpenChange: (open: boolean) => void;
   title: string;
   description?: React.ReactNode;
-  confirmLabel?: string;
-  cancelLabel?: string;
+  // Required so callers cannot fall back to English literals and silently
+  // leak en into other locales. Always pass via the i18n hook (t(...)).
+  confirmLabel: string;
+  cancelLabel: string;
   destructive?: boolean;
   onConfirm: () => void;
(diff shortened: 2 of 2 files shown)


Jiahui-Gu (commented):
Independent review: APPROVE.

Verified:
- `confirmLabel`/`cancelLabel` are now `string` (required, no `?`) in `src/components/ui/ConfirmDialog.tsx` lines 12-13, with a JSDoc comment explaining why (no silent en fallback in other locales).
- English literal defaults (`'Confirm'`, `'Cancel'`) removed from the destructuring.
- `data-testid="confirm-dialog-confirm"` / `confirm-dialog-cancel` added; naming follows the repo's existing kebab-case convention (cf. `chat-load-history-retry`, `tool-output-copy`, `cli-cmd-*`).
- Caller count: I grepped `ConfirmDialog` across `src/` and found exactly 1 caller (`src/components/Sidebar.tsx:284`, group-delete) — matches worker's claim. TypeScript would have caught any other; reverse-verified by deleting the new `cancelLabel={t('common.cancel')}` line and re-running `tsc --noEmit` — got `TS2741: Property 'cancelLabel' is missing` as expected. Restored after.
- `npm run typecheck` (renderer + electron tsconfigs) clean on the PR head.
- Label content: `t('common.cancel')` resolves to `'Cancel'` in `src/i18n/locales/en.ts:16` (sentence case, no SCREAMING).
- No ConfirmDialog-specific test file exists (grep found none); not a blocker for this scoped change since the type system + the existing Sidebar tests would catch a regression.

No concerns. Safe to merge.

PR #224, opened 2026-04-24
fix(notify): make @ccsm/notify optional with lazy load + fallback (task #267)

Jiahui-Gu:
## Summary

Reviewer of #266 found `npm install` fails before `prepare` runs because `@ccsm/notify` pulls `electron-windows-notifications` -> 5 nan-based `@nodert-win10-au/*` deps with no prebuilds (node-gyp/MSBuild fails on most envs). Apply the fsevents pattern: optional install + lazy require + graceful no-op fallback.

- Bump `@ccsm/notify` pin `v0.1.0` -> `v0.1.1`, move `dependencies` -> `optionalDependencies`.
- New `electron/notify.ts` wrapper: dynamic `import('@ccsm/notify')` with cached failure; `notifyPermission/notifyQuestion/notifyDone/notifyDismiss` become async no-ops when the module is missing; `isNotifyAvailable()` / `probeNotifyAvailability()` expose status. Payload types are duplicated locally so the wrapper type-checks even when the optional dep is absent on disk.
- New `notify:availability` IPC + preload + `global.d.ts` typing.
- SettingsDialog Notifications pane now shows a sentence-case banner when the module is unavailable (English/中文 both added). Banner copy:
  > Rich Windows toasts are unavailable on this machine — the optional @ccsm/notify native module did not install. CCSM will fall back to in-app banners and standard system notifications.
- `scripts/postinstall.mjs`: when full `electron-builder install-app-deps` fails, fall back to required-only rebuild (`electron-rebuild --only better-sqlite3 --types prod`); demote the original failure to a warning if the required-only rebuild succeeds. This is the bit that actually unblocks `npm install` end-to-end on machines without a working MSBuild.

## Test plan

- [x] `npx tsc --noEmit` (root + electron) — clean
- [x] `npm test` — 668 / 668 pass (52 files)
- [x] `npm run build` — clean
- [x] `tests/notify-fallback.test.ts` — 4 unit tests (load failure path, no-retry, success delegation, `Notifier.create` failure)
- [x] `scripts/probe-e2e-notify-fallback.mjs` — stashes `node_modules/@ccsm/notify` aside, launches electron with sanitized HOME, asserts IPC `{available:false, error:"Cannot find package '@ccsm/notify'..."}`, opens Settings → Notifications, asserts banner `data-available="false"` with the exact sentence-case copy, no SCREAMING tokens, no renderer console errors, no main crash. Restores the dir on exit.
- [x] **Reverse-verify**: removed the `try/catch` in `loadModule()` and rebuilt — probe FAILS with exit 1, the import error bubbles out of the `notify:availability` IPC handler. Restored.

## Notes for reviewer

- Independent reviewer required per project rule — DO NOT self-merge.
- Wrapper is currently unused by callers; this PR is purely defensive (unblocks install + lays the foundation for future callers in `notifications.ts` etc.). Per task spec.
- `@ccsm/notify` is ESM (`"type": "module"`); main process is CJS, so we use `Function('m', 'return import(m)')` to escape TypeScript's CJS-mode `import()` -> `require()` rewrite.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -76,6 +76,7 @@ import {
 } from './import-scanner';
 import { loadImportableHistory } from './import-history';
 import { showNotification, type ShowNotificationPayload } from './notifications';
+import { probeNotifyAvailability, notifyLastError } from './notify';
 import type { PermissionMode } from './agent/sessions';
 import { listModelsFromSettings } from './agent/list-models-from-settings';
 import { ClaudeNotFoundError, detectClaudeVersion, resolveClaudeBinary } from './agent/binary-resolver';
@@ -1026,6 +1027,11 @@ app.whenReady().then(() => {
       return showNotification(payload, win);
     }
   );
+
+  ipcMain.handle('notify:availability', async () => {
+    const available = await probeNotifyAvailability();
+    return { available, error: notifyLastError() };
+  });
 
   installUpdaterIpc();
 
--- a/electron/notify.ts
+++ b/electron/notify.ts
@@ -1,0 +1,253 @@
+// Defensive wrapper around the optional `@ccsm/notify` dependency.
+//
+// `@ccsm/notify` pulls in `electron-windows-notifications`, which transitively
+// depends on a chain of nan-based `@nodert-win10-au/*` native addons that have
+// no prebuilds. On environments without a working node-gyp / MSBuild toolchain
+// the install of those native deps fails, so we declare `@ccsm/notify` in
+// `optionalDependencies` (npm tolerates failed optional installs) and load it
+// lazily here. If loading fails — for any reason: optional install skipped,
+// platform unsupported, runtime adapter init throws — every wrapper function
+// becomes a graceful no-op so the rest of the app keeps working with the
+// existing in-app banners / Electron `Notification` toasts.
+//
+// Same design as `fsevents` on macOS: optional install + lazy require + soft
+// fallback. See task #267 for context.
+//
+// We intentionally do NOT `import type` from '@ccsm/notify' so that this file
+// type-checks even when the optional dep is missing on disk. The payload
+// shapes below are duplicated from the upstream public API (kept narrow on
+// purpose — only the fields callers actually pass through).
+
+// ---------- Local payload types (kept in lockstep with @ccsm/notify) ----------
+
+export interface PermissionPayload {
+  toastId: string;
+  sessionName: string;
+  toolName: string;
+  toolBrief: string;
+  cwdBasename: string;
+}
+
+export interface QuestionPayload {
+  toastId: string;
+  sessionName: string;
+  question: string;
+  selectionKind: 'single' | 'multi';
+  optionCount: number;
+  cwdBasename: string;
+}
+
+export interface DonePayload {
+  toastId: string;
+  groupName: string;
+  sessionName: string;
+  lastUserMsg: string;
+  lastAssistantMsg: string;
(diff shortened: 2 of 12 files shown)


Jiahui-Gu (commented):
## Independent Reviewer Verdict: APPROVE (LGTM)

All 10 review steps verified in isolated checkout `C:\Users\jiahuigu\AppData\Local\Temp\review-224\ccsm-review`.

### npm install end-to-end
`npm install --legacy-peer-deps` succeeded cleanly. better-sqlite3 rebuilt for Electron ABI on first attempt. On this machine the `@ccsm/notify` optional dep was skipped at the npm resolution stage (no `node_modules/@ccsm/notify` present, no separate lockfile entry), so the postinstall fallback retry path didn't have to fire — but that's the better outcome for the install end-user, and it's exactly the scenario the wrapper is designed for (dep absent → graceful no-op).

### Reverse-verify (oracle)
- Probe with try/catch in place: **PASS** (exit 0, status banner reads "Rich Windows toasts are unavailable on this machine — ...").
- Stripped try/catch from `loadModule()` → rebuilt → probe **FAILED** with uncaught `Cannot find package '@ccsm/notify'`. 
- Restored → rebuilt → probe **PASS** again. Wrapper is load-bearing, confirmed.

### Wrapper API (electron/notify.ts)
Cleaner than spec'd: uses dynamic `import()` (because @ccsm/notify is ESM and main is CJS) instead of plain `require()`; resolution cached, failure sticky, single warn log. All emitters (`notifyPermission/Question/Done/Dismiss`) check `if (!n) return` short-circuit. `isNotifyAvailable()` reflects sticky resolved state; `probeNotifyAvailability()` async variant for Settings dialog. `__setNotifyImporter` test seam is clean.

### postinstall fallback (scripts/postinstall.mjs)
Matches spec exactly: full `electron-builder install-app-deps` → on non-zero, retries `electron-rebuild --only better-sqlite3 --types prod` → success demotes to warning, failure propagates exit code. Windows `.cmd` shell:true handling correct (CVE-2024-27980 mitigation). Hint message guides toolchain install. No silent swallow.

### package.json + lockfile
`@ccsm/notify` correctly in `optionalDependencies` only (not in `dependencies`). Lockfile reflects optional registration. Pin bumped to `v0.1.1` (tag confirmed exists upstream).

### SettingsDialog UX + i18n
Three-state banner (checking/available/unavailable), `data-testid="notifications-module-status"`, `data-available` attr for tests, amber border + amber/5 bg only on unavailable. Strings sentence case in en + zh, semantically equivalent. No SCREAMING.

### tsc / build / tests
- `npx tsc --noEmit`: clean.
- `npm run build`: success (only pre-existing bundle-size warnings).
- `npx vitest run tests/notify-fallback.test.ts`: 4/4 PASS.
- Full `npm test`: 656 pass / 12 fail. **All 12 failures are pre-existing better-sqlite3 NODE_MODULE_VERSION 130 vs 127 ABI mismatch under vitest's host Node** — unrelated to this PR (verified `db-hardening.test.ts` exists unchanged on `working`).

### Concerns: none blocking
- Minor: lockfile didn't materialize a `node_modules/@ccsm/notify/...` subtree on this machine even though v0.1.1 tag exists upstream (likely SSH auth/transport silently skipped — npm warn about "skipping integrity check for git dependency"). This is consistent with the optional-dep design: silent skip is acceptable; the wrapper handles absence. Worth keeping an eye on in CI to make sure at least one platform DOES install it and exercises the happy path.

Ship it.

PR #242, opened 2026-04-24
feat: IPC for partial diff write (#251)

Jiahui-Gu:
Closes #251 (IPC scope only — see Follow-up below for UI).

## Scope chosen: IPC + main-side filter logic, NOT the renderer UI

Spike of `PermissionPromptBlock` confirmed there is currently **no diff visualization at all** in the permission prompt — the tool input is rendered as a flat key/value summary. Adding per-hunk checkboxes therefore requires:

1. Embedding `DiffView` (or a new hunk-list component) inside the prompt
2. Local hunk-selection state + "select all / none"
3. Replacing the single `Allow` button with `Apply selected (N/M)` + edge cases (0 selected -> Reject, all selected -> existing Allow path)
4. Keyboard model rework (current `y`/`n` global hotkey doesn't map cleanly onto a multi-select UI)
5. New e2e probe driving checkboxes via Playwright

That is a non-trivial UX change with its own design/review surface and would dwarf the IPC work. This PR ships the wire-level mechanism so the UI follow-up can be a pure-renderer change.

## What this PR does

- **`electron/agent/partial-write.ts`** — new pure helper `filterToolInputByAcceptedHunks(toolName, input, acceptedHunks)` returning `{ unchanged | updated | reject }`. Hunk indices map 1:1 to `DiffSpec.hunks` ordering produced by `src/utils/diff.ts`:
  - `Edit`/`Write` -> 1 hunk; empty selection -> `reject` (an empty Write would silently corrupt the file).
  - `MultiEdit` -> one hunk per `edits[]` entry; subset -> new input with `edits` filtered to chosen indices in original array order.
  - `undefined` acceptedHunks -> `unchanged` (legacy whole-tool allow).
- **`ClaudeRunner.resolvePermissionPartial(requestId, acceptedHunks)`** — uses the existing `updatedInput` channel of `can_use_tool` (already plumbed through `ControlRpc`) to forward the filtered input back to `claude.exe`.
- **PreToolUse hook resolver also forwards `updatedInput`** (added during reviewer follow-up; see "Reviewer-flagged bug fix" below). No protocol changes to the CLI side.
- **IPC `agent:resolvePermissionPartial`** in `electron/main.ts`, additive — `agent:resolvePermission` still works for the legacy whole-allow path. Preload bridge + `global.d.ts` typing updated.
- **Pending-perm bookkeeping** changed from `Map<string, resolve>` to `Map<string, { resolve, toolName, input }>` so the partial-resolve path can build `updatedInput` without an extra renderer round-trip.

## Reviewer-flagged bug fix (latest commit)

Reviewer found that `resolvePermissionPartial` resolves with `{ allow: true, updatedInput }` and `ControlRpc` correctly forwards `updatedInput` on the `can_use_tool` path — but Edit/Write/MultiEdit in CLI 2.x route through the **PreToolUse hook** path instead (matcher `.*` registered at `electron/agent/sessions.ts:456`). The hook resolver at `sessions.ts:622-643` only read `decision.allow` and emitted a static `permissionDecision: 'allow'`, **silently discarding `updatedInput`**. Result: a MultiEdit subset selection would still cause the CLI to run ALL edits.

**SDK path chosen: (a)** — the official PreToolUse hook protocol supports `hookSpecificOutput.updatedInput`, which the CLI uses to rewrite the tool's input before execution (same field semantics as the can_use_tool branch). This is the documented contract — no need to bypass the hook or scope the matcher.

**Fix** in `electron/agent/sessions.ts` (handleHookCallback resolve path):

```ts
resolve({
  hookSpecificOutput: {
    hookEventName: 'PreToolUse',
    permissionDecision: 'allow',
    ...(decision.updatedInput !== undefined
      ? { updatedInput: decision.updatedInput }
      : {}),
  },
});
```

The reject branch was already correct.

## Tests

- **17 unit tests** in `tests/partial-write.test.ts` (filter helper).
- **2 new tests** in `electron/agent/__tests__/sessions.test.ts` for the PreToolUse hook resolve path (reviewer gap):
  - Unit: hook resolver emits `hookSpecificOutput.updatedInput` when `resolvePermissionPartial` selects a subset of MultiEdit hunks.
  - **Integration**: full path from CLI-emitted `hook_callback` (MultiEdit, 3 hunks) through `resolvePermissionPartial([0,2])` to the outbound `control_response`, asserting trimmed `edits` AND that peer fields (`replace_all`, etc.) are preserved.

**Reverse-verify**: stashing the new `updatedInput`-forwarding line makes both new tests fail with "expected updatedInput to be defined / equal {...}, received undefined". Restoring -> 25/25 sessions tests pass.

Run results (latest):
- `npx tsc --noEmit`: clean
- `npm test`: 703/715 pass. The 12 failures are all in `electron/__tests__/db.test.ts` and `db-hardening.test.ts` — `better-sqlite3` `NODE_MODULE_VERSION 130 vs 127` native module mismatch, pre-existing infra issue, unrelated.

## Backwards compatibility

- `agent:resolvePermission` IPC + `agentResolvePermission` preload + `ClaudeRunner.resolvePermission` all unchanged.
- All pre-existing permission probes untouched and unaffected.
- The added `updatedInput` field is omitted when `decision.updatedInput` is undefined, so the legacy whole-allow hook payload is byte-identical.

## Follow-up

> **Renderer UI for per-hunk partial accept (#251 follow-up)**
> Wire `agentResolvePermissionPartial` into `PermissionPromptBlock`:
> - For Edit/Write/MultiEdit tool calls, render the diff via `DiffView` with a checkbox per hunk
> - Replace `Allow` with `Apply selected (N/M)`; disable when 0 selected and route to `Reject` when user clicks; route to legacy `agentResolvePermission` allow when all selected
> - Add a `probe-e2e-permission-partial-accept.mjs` driving the checkboxes
> - Reverse-verify: stash the partial-IPC fix and confirm the probe fails

## Test plan

- [x] Reviewer: confirm `filterToolInputByAcceptedHunks` matches `src/utils/diff.ts` hunk ordering for all three tool names
- [x] Reviewer: verify `updatedInput` actually forwards through `ControlRpc` (`electron/agent/control-rpc.ts:339-344`)
- [x] Reviewer: confirm `updatedInput` also forwards through the PreToolUse hook path (`electron/agent/sessions.ts:622-647`) — fixed in latest commit
- [x] Reviewer: integration test covers hook resolve path end-to-end (`sessions.test.ts` "integration: MultiEdit with 3 hunks + partial-accept")

### DIFF
--- a/electron/agent/__tests__/sessions.test.ts
+++ b/electron/agent/__tests__/sessions.test.ts
@@ -538,6 +538,153 @@ describe('SessionRunner PreToolUse hook permission', () => {
     expect(resp.response.response).toEqual({});
     runner.close();
   });
+
+  // Regression for the PR #242 reviewer-flagged bug: Edit/Write/MultiEdit in
+  // CLI 2.x route through the PreToolUse hook path, NOT can_use_tool. The
+  // hook resolver was emitting `permissionDecision: 'allow'` only and
+  // discarding `decision.updatedInput`, so a MultiEdit subset selection
+  // silently caused the CLI to run ALL edits.
+  it('forwards updatedInput on the hook response when resolvePermissionPartial picks a subset of MultiEdit hunks (PR #242 fix)', async () => {
+    const proc = makeFakeProc();
+    mockSpawnClaude.mockResolvedValue(proc);
+    let captured = '';
+    const runner = new SessionRunner('s-hook-partial', () => {}, () => {}, (req) => {
+      captured = req.requestId;
+    });
+    await runner.start(baseOpts);
+
+    const originalEdits = [
+      { old_string: 'a', new_string: 'A' },
+      { old_string: 'b', new_string: 'B' },
+      { old_string: 'c', new_string: 'C' },
+    ];
+    emitFrame(proc.stdout, {
+      type: 'control_request',
+      request_id: 'req_hk_partial',
+      request: {
+        subtype: 'hook_callback',
+        callback_id: 'ccsm-permission',
+        input: {
+          hook_event_name: 'PreToolUse',
+          tool_name: 'MultiEdit',
+          tool_input: { file_path: '/x', edits: originalEdits },
+          permission_mode: 'default',
+        },
+      },
+    });
+    await new Promise((r) => setImmediate(r));
+    expect(captured).not.toBe('');
+
+    // User accepts hunks 0 and 2, rejects hunk 1.
+    expect(runner.resolvePermissionPartial(captured, [0, 2])).toBe(true);
+    await new Promise((r) => setImmediate(r));
+
+    const lines = proc.__stdinLines() as Array<Record<string, unknown>>;
+    const resp = lines.find(
+      (l) =>
+        l.type === 'control_response' &&
+        (l.response as { request_id?: string } | undefined)?.request_id === 'req_hk_partial',
+    ) as { response: { subtype: string; request_id: string; response: Record<string, unknown> } } | undefined;
+    expect(resp).toBeDefined();
+    expect(resp!.response.subtype).toBe('success');
+    const out = resp!.response.response as {
+      hookSpecificOutput: {
+        hookEventName: string;
+        permissionDecision: string;
+        updatedInput?: { edits?: Array<Record<string, unknown>>; file_path?: string };
+      };
+    };
+    expect(out.hookSpecificOutput.hookEventName).toBe('PreToolUse');
+    expect(out.hookSpecificOutput.permissionDecision).toBe('allow');
+    // The critical assertion: the hook response MUST include updatedInput
+    // so the CLI rewrites the tool call to the trimmed edits list.
+    expect(out.hookSpecificOutput.updatedInput).toBeDefined();
+    expect(out.hookSpecificOutput.updatedInput!.file_path).toBe('/x');
+    expect(out.hookSpecificOutput.updatedInput!.edits).toEqual([
+      originalEdits[0],
+      originalEdits[2],
+    ]);
+    runner.close();
+  });
+
+  // Integration: full path from a CLI-emitted PreToolUse hook_callback for
+  // MultiEdit through resolvePermissionPartial back to a control_response
+  // the SDK/CLI consumes. Reviewer flagged "no integration test covers the
+  // hook resolve path" — this is it. Asserts both the trimmed input AND
+  // that all unmodified peer fields (replace_all, etc.) are preserved.
+  it('integration: MultiEdit with 3 hunks + partial-accept (2 of 3) emits a hook control_response carrying the trimmed updatedInput (PR #242 fix)', async () => {
+    const proc = makeFakeProc();
+    mockSpawnClaude.mockResolvedValue(proc);
+    const seen: Array<{ requestId: string; toolName: string; input: Record<string, unknown> }> = [];
+    const runner = new SessionRunner(
+      's-hook-partial-int',
+      () => {},
+      () => {},
+      (req) => seen.push(req),
+    );
+    await runner.start(baseOpts);
+
+    const inputEdits = [
+      { old_string: 'foo', new_string: 'FOO', replace_all: false },
+      { old_string: 'bar', new_string: 'BAR', replace_all: true },
+      { old_string: 'baz', new_string: 'BAZ', replace_all: false },
+    ];
+    emitFrame(proc.stdout, {
+      type: 'control_request',
+      request_id: 'req_hk_int_partial',
+      request: {
+        subtype: 'hook_callback',
+        callback_id: 'ccsm-permission',
+        input: {
+          hook_event_name: 'PreToolUse',
+          tool_name: 'MultiEdit',
+          tool_input: { file_path: '/repo/file.ts', edits: inputEdits },
+          tool_use_id: 'tu_int',
+          permission_mode: 'default',
+        },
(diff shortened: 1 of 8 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: REQUEST CHANGES

Reviewed the IPC-only scope as agreed. Code structure is clean, tests are thorough, backwards compat is preserved — but I found one **blocking correctness bug** that nullifies the entire feature for the tools it targets.

### Blocking: `updatedInput` is silently dropped on the PreToolUse hook path

`resolvePermissionPartial` calls `entry.resolve({ allow: true, updatedInput })`. In `handleCanUseTool` this works — `ControlRpc` forwards `updatedInput` (control-rpc.ts:339-344, verified). But Edit/Write/MultiEdit in CLI 2.x do **not** go through `can_use_tool`; they go through the `PreToolUse` hook registered with matcher `.*` at initialize time (sessions.ts:456-463, and the doc comment on `handleHookCallback` says exactly this: "the host-side permission gate that `--permission-prompt-tool stdio` no longer reliably provides for built-in tools in CLI 2.x").

The hook path's `resolve` shim at sessions.ts:622-643 only inspects `decision.allow` and emits a static `hookSpecificOutput: { permissionDecision: 'allow' }` — `decision.updatedInput` is never read, never forwarded. So for the exact tools this PR is built for, the rewritten input is discarded and `claude.exe` runs the original, unfiltered tool call.

This means:
- `MultiEdit` partial-accept: user picks 1 of 3 edits → CLI still runs all 3.
- `Edit`/`Write` empty-selection: the `reject` branch works (it routes to `decision.allow=false`, which the hook path does honor as deny).

Required: extend the hook `hookSpecificOutput` with the equivalent input-rewrite mechanism (the Anthropic hook protocol supports a `permissionDecisionReason` + tool-input modification field — please verify the exact key in the CLI source; if hooks genuinely cannot rewrite tool input, then partial-accept for hook-routed tools needs a different mechanism, e.g. routing back through `can_use_tool` artificially or modifying the tool-call before forwarding).

### Other findings (non-blocking)

1. **Unit-test coverage gap matching the bug above**: tests exercise the pure helper but no test covers the end-to-end resolve path through `handleHookCallback`. Adding a test that drives `resolvePermissionPartial` against a mock pendingPerm registered by the hook handler would have caught this. Per memory `feedback_bugfix_requires_e2e.md` this PR also lacks an e2e probe — but since UI is deferred, an integration-level test on the resolve plumbing (mock CanUseToolDecision sink) is the moral equivalent and should be added.
2. **PR body claims "17 tests"** — confirmed exactly 17. tsc clean on both projects. Lint baseline unchanged. Backwards compat for legacy `agent:resolvePermission` IPC + preload + runner method confirmed untouched.
3. **Reverse-verify performed**: stubbed `filterToolInputByAcceptedHunks` to always return `{ kind: 'unchanged' }` → 9/17 tests fail (the cases requiring real filter logic), 8 pass (legitimate unchanged paths). Restored → 17/17 pass. Filter logic is real.
4. **Followup task tracking**: PR body says "A new tracking task should be created for the renderer UI" (note the passive voice) but I see no evidence in PR description that the manager has actually created the TaskList entry. Per `feedback_track_tasks_immediately.md`, manager must create it now, not later. Flagging for manager: please confirm the followup is in TaskList before merge, with title approximately **"Renderer UI for per-hunk partial accept (#251 follow-up)"** as drafted in the PR body.

### Verdict

**Do not merge.** Fix the hook-path drop first; without it, the IPC ships dead. After fix + new test covering the hook resolve path, this is good to go.

Jiahui-Gu (commented):
## Re-review (fix commit 0d07bc3) — LGTM, blocker resolved

Verified locally on a fresh clone of `task251-partial-write-ipc`.

**Hook resolver fix (`electron/agent/sessions.ts` ~L633-641)**
- `decision.allow === true` path now spreads `updatedInput` into `hookSpecificOutput` only when defined; legacy no-`updatedInput` allow stays clean (`hookEventName` + `permissionDecision: 'allow'` only). Conditional spread is the right shape.
- Deny path untouched. Passthrough/bypass/acceptEdits short-circuits untouched.

**SDK schema sanity-check**
- ccsm has no `@anthropic-ai/claude-code` npm dep — it spawns the `claude` binary and speaks the CLI control protocol directly. Per the official hooks doc (https://code.claude.com/docs/en/hooks): the field is `updatedInput`, "Replaces the entire input object, so include unchanged fields alongside modified ones." The fix uses the exact key, and `filterToolInputByAcceptedHunks` already returns the full input shape (file_path + trimmed edits, peer fields preserved), so the "include unchanged fields" requirement is met.

**Tests** (`electron/agent/__tests__/sessions.test.ts`)
- Unit test asserts `out.hookSpecificOutput.updatedInput.edits === [edits[0], edits[2]]` after `resolvePermissionPartial(id, [0, 2])`.
- Integration test round-trips through `hook_callback` (not `can_use_tool`) end-to-end, asserts `replace_all: true` on hunk 1's peer field is preserved on the surviving hunks, asserts `file_path` round-trips. Good coverage of the reviewer-flagged path.

**Reverse-verify** (per repo policy)
- Stashed the conditional-spread block → `npx vitest run electron/agent/__tests__/sessions.test.ts`: both new tests FAIL with `expected undefined to deeply equal { file_path, edits }` (exactly the regression they target). Restored → 25/25 pass. Fix is load-bearing.

No remaining concerns. Approving the fix; ready to merge.

PR #234, opened 2026-04-24
feat(notify): integrate @ccsm/notify Adaptive Toast pipeline (Wave 1D, #265)

Jiahui-Gu:
## Summary

- Bootstraps the optional `@ccsm/notify` package at app `whenReady` so Windows users get Adaptive Toasts with inline Allow / Allow always / Reject buttons for permission prompts, plus toasts for AskUserQuestion and turn-done events. Falls back to plain Electron Notifications when the optional native module isn't installed (the #267 wrapper guarantees no-op behaviour).
- Routes toast button activations back into the same code path the in-app PermissionPromptBlock uses: the agent permission gate is resolved in main (`sessions.resolvePermission`), and a new `notify:toastAction` IPC mirrors the decision into the renderer store (clearing the waiting block and seeding `allowAlwaysTools` for Allow always).
- Defensive doubled focus suppression: renderer-side `document.hasFocus` AND main-side `BrowserWindow.isFocused()` — toasts never fire while ccsm is foreground.
- Settings respect: all dispatch flows through the existing `dispatchNotification` which already gates on the `notificationSettings.enabled` master switch + per-event sub-toggles.

## What fires toasts

- `permission` — when an agent needs Allow / Allow always / Reject for a tool (Bash, Edit, etc.).
- `question` — when the agent invokes `AskUserQuestion`.
- `turn_done` — when a turn finishes AND it errored, took >15s, or the session wasn't focused.

## Files

- `electron/notify-bootstrap.ts` (new) — configure wrapper + onAction router + toastId→sessionId registry with LRU cap.
- `electron/notifications.ts` — fan out to `@ccsm/notify` in parallel with the legacy Electron Notification; main-process focus gate.
- `electron/main.ts` — bootstrap call after `setAppUserModelId`; route toast actions to `sessions.resolvePermission` + `notify:toastAction`. Expose `__ccsmDebug.notify/notifyBootstrap/sessions` for e2e mocking (dev-only).
- `src/agent/lifecycle.ts` — enrich permission/question/turn_done dispatches with `extras` (toolName, cwd, elapsedMs, last messages, etc.); subscribe to `onNotifyToastAction`.
- `electron/preload.ts` + `src/global.d.ts` + `src/notifications/dispatch.ts` — thread optional `extras` through the IPC bridge.
- `scripts/probe-e2e-notify-integration.mjs` — e2e probe (HOME sanitised) that mocks the wrapper importer, asserts wrapper calls, allow-always routing, and focus suppression.
[...]

PR #243, opened 2026-04-24
feat(notify): ask-toast retry + done rich content (#252)

Jiahui-Gu:
## Summary

Wave 3 polish on top of the @ccsm/notify Adaptive Toast pipeline (Wave 1D, #234):

- **Ask-question single retry** — when the user dismisses the toast without answering, re-emit once after ~30s. New `electron/notify-retry.ts` owns the `Map<toastId, PendingRetry>` state. Cancellation hooks into `agent:resolvePermission` (the only path the in-app `QuestionBlock` uses) so an answered question never re-banners. Cap is hardcoded to 1 retry.
- **Done rich content** — for the `turn_done` event:
  - Legacy Electron `Notification` body falls back to the assistant's last response truncated to 80 chars when the host didn't supply an explicit body (was empty for the success path).
  - `notifyDone` payload now truncates `lastAssistantMsg` to 80 chars to match `xml/done.ts ASSISTANT_LINE_MAX` in the SDK.
  - Title (`{group} · {session}`) and body-click → focus already worked via the SDK + main-process toast action router; verified by the new probe case rather than re-implementing.

## SDK gaps

None required. `@ccsm/notify@v0.1.1`'s `xml/done.ts` already lays out title + truncated assistant body + focus-on-click exactly as the task spec asks. We only needed to make sure the host passes the right shape and that the legacy fallback toast (when `@ccsm/notify` is unavailable, e.g. non-win32) shows the same preview.

## Files

- `electron/notify-retry.ts` (new) — retry Map + scheduler + cancel + test seam.
- `electron/notifications.ts` — schedule retry after `notifyQuestion`; truncate `lastAssistantMsg` to 80 in adaptive done payload; legacy body fallback for `turn_done`.
- `electron/main.ts` — call `cancelQuestionRetry(`q-${requestId}`)` in `agent:resolvePermission` IPC handler; expose `notifyRetry` via `__ccsmDebug` for the probe.
- `electron/__tests__/notify-retry.test.ts` (new, 8 cases) — schedule, idempotence, fire-once, cap-at-1, cancel, no-op cancel, empty-id reject, multi-id isolation.
- `electron/__tests__/notifications.test.ts` (new, 5 cases) — legacy body fallback, body passthrough, adaptive done truncation, short-message passthrough, question retry scheduling end-to-end.
- `scripts/probe-e2e-notify-integration.mjs` — extended with 3 new cases (initial + retry call, cancellation via `agentResolvePermission` IPC, rich done payload composition).

## Test plan

- [x] `npx tsc --noEmit` (root) — clean
- [x] `npx tsc --noEmit -p tsconfig.electron.json` — clean
- [x] `npm run lint` — no new errors (pre-existing 29 errors in `verification/*.mjs` and `electron/main.ts:119` are unchanged)
- [x] `npx vitest run electron/__tests__/notify-retry.test.ts electron/__tests__/notifications.test.ts electron/__tests__/notify-bootstrap.test.ts` — 17/17 passing
- [x] Full `npm test` — 12 pre-existing `db.test.ts` / `db-hardening.test.ts` failures from `better-sqlite3` `NODE_MODULE_VERSION` mismatch with vitest's node (verified identical on stashed `working`); my work introduces zero new failures
- [x] `node scripts/probe-e2e-notify-integration.mjs` — all 11 assertions pass (8 original + 3 new)
- [x] **Reverse-verify retry**: stashed `scheduleQuestionRetry(questionPayload)` → probe FAILS at "timeout waiting for initial question + retry schedule"
- [x] **Reverse-verify cancel**: stashed both `cancelQuestionRetry(...)` calls → probe FAILS at "agentResolvePermission did not cancel the pending retry timer"

### DIFF
--- a/electron/__tests__/notifications.test.ts
+++ b/electron/__tests__/notifications.test.ts
@@ -1,0 +1,199 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import {
+  showNotification,
+  type ShowNotificationPayload,
+} from '../notifications';
+import {
+  configureNotify,
+  __setNotifyImporter,
+  probeNotifyAvailability,
+} from '../notify';
+import { __resetBootstrapForTests } from '../notify-bootstrap';
+import {
+  __resetRetryStateForTests,
+  __setRetrySchedulerForTests,
+  __pendingRetryCountForTests,
+} from '../notify-retry';
+
+// Mock electron's BrowserWindow + Notification surface — the test harness
+// is plain node so we can't pull in real Electron. We capture every
+// Notification constructor call so the showNotification pipeline can be
+// asserted end-to-end (legacy body composition + adaptive payload).
+let lastNotificationCtor: { title: string; body: string; silent: boolean } | undefined;
+const showSpy = vi.fn();
+const onClickSpy = vi.fn();
+
+vi.mock('electron', () => ({
+  BrowserWindow: { getAllWindows: () => [] },
+  Notification: class FakeNotification {
+    title: string;
+    body: string;
+    silent: boolean;
+    constructor(opts: { title: string; body: string; silent: boolean }) {
+      this.title = opts.title;
+      this.body = opts.body;
+      this.silent = opts.silent;
+      lastNotificationCtor = opts;
+    }
+    static isSupported() {
+      return true;
+    }
+    on(event: string, cb: () => void) {
+      if (event === 'click') onClickSpy.mockImplementation(cb);
+    }
+    show() {
+      showSpy();
+    }
+  },
+}));
+
+describe('showNotification — done rich content (#252)', () => {
+  let donePayloads: Array<Record<string, unknown>>;
+
+  beforeEach(async () => {
+    lastNotificationCtor = undefined;
+    showSpy.mockClear();
+    onClickSpy.mockReset();
+    donePayloads = [];
+    __resetBootstrapForTests();
+    __resetRetryStateForTests();
+    __setRetrySchedulerForTests(
+      // Swallow timers so the retry can't actually fire during these tests.
+      // notifyDone tests below don't exercise question retries; question
+      // tests use the same swallow-and-record fake from notify-retry.test.ts.
+      () => 0 as unknown as ReturnType<typeof setTimeout>,
+      () => {},
+    );
+    __setNotifyImporter(async () => ({
+      Notifier: {
+        create: async () => ({
+          permission: () => {},
+          question: () => {},
+          done: (p: Record<string, unknown>) => donePayloads.push(p),
+          dismiss: () => {},
+        }),
+      },
+    }));
+    configureNotify({
+      appId: 'test',
+      appName: 'Test',
+      onAction: () => {},
+    });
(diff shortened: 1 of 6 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE WITH NOTES

Verified end-to-end (clean clone + sanitized HOME, build, unit + e2e, plus reverse-verify on both behaviors).

### What I verified
- `npx tsc -p tsconfig.electron.json`: clean.
- Unit tests: 13/13 pass (`notify-retry.test.ts` 8 cases, `notifications.test.ts` 5 cases). Cap-at-1, idempotent schedule, distinct-id, cancel no-op, empty-id reject, truncation/focus/done composition all covered.
- `scripts/probe-e2e-notify-integration.mjs`: 11/11 assertions pass (extends existing probe per `feedback_e2e_extend_existing` — not a new file). Confirmed retry payload reuses original `toastId` for SDK dedupe coherence, delay = 30_000ms.
- **Reverse-verify #1**: stashed `scheduleQuestionRetry(questionPayload)` in `electron/notifications.ts` → probe FAILS at `timeout waiting for initial question + retry schedule`. Restored.
- **Reverse-verify #2**: stashed both `cancelQuestionRetry(...)` calls in `electron/main.ts` resolvePermission handler → probe FAILS at `agentResolvePermission did not cancel the pending retry timer`. Restored.
- Debug surface (`__ccsmDebug.notifyRetry`) gated behind `!app.isPackaged` — prod-safe.
- Map cleanup verified: `fireRetry` calls `pending.delete(id)` BEFORE running `notifyQuestion`, so a synchronous wrapper throw can't leak the entry. `cancelQuestionRetry` deletes after clearing timer.

### Findings (non-blocking, follow-up suggested)

1. **Settings/focus gate not re-checked at fire time** — `fireRetry` calls `notifyQuestion` directly, bypassing `showNotification`'s `shouldSuppressForFocus()` and `Notification.isSupported()` / settings checks done at schedule time. If the user (a) toggles OS notifications off, (b) focuses the app, or (c) disables in-app notification settings during the 30s window, the retry still fires. Cheap fix: in `fireRetry`, check `shouldSuppressForFocus()` + the same settings predicate the schedule path used and no-op silently. Worth a follow-up task.

2. **Toast-button reject path bypasses cancel** — `main.ts:466-468` (toast action router) calls `sessions.resolvePermission` directly, skipping the IPC handler that wires `cancelQuestionRetry`. Today only `permission` events route through here (no retry scheduled) so it's vacuous, but if a future toast button ever resolves a question, the retry will leak. Defensive: lift `cancelQuestionRetry` into a helper and invoke from both call sites.

3. **Stale-session edge case** — verified safe. `notifyQuestion` doesn't dereference session state; if the session is gone when the retry fires, the toast emits, and any subsequent focus/answer routes through `sessions.resolvePermission` which returns `false` for unknown `requestId`. No-op as desired.

### Conclusion
Code is correct, tests are reverse-verified, e2e extends existing probe properly. Approving the implementation; recommend filing a follow-up for finding #1 (settings/focus re-check at fire time) — small but real user-visible gap.

PR #244, opened 2026-04-24
test+chore: e2e probes for tray/banner/toast i18n+a11y + text-meta naming align (#295 #296 #298 #300)

Jiahui-Gu:
## Summary

Bundle of four small follow-ups, each extending an existing probe rather than creating a new file (per `feedback_e2e_extend_existing.md`).

### Extended probe list

| Issue | File extended | Strategy |
| --- | --- | --- |
| #295 tray i18n | `electron/__tests__/i18n.test.ts` | Added `#295 tray-menu i18n end-to-end via setMainLanguage` describe block. Headless OS-tray inspection isn't viable (Win shell / GNOME indicator / NSStatusItem aren't scriptable from Playwright), so the assertion targets the underlying `tTray()` source that `applyTrayLocale()` feeds into `Menu.buildFromTemplate`. |
| #296 banner i18n | `scripts/harness-ui.mjs` | Added `banner-i18n-toggle` case. Seeds `sessionInitFailures` + `diagnostics`, drives `__ccsmI18n.changeLanguage` en→zh→en, asserts AgentInitFailedBanner + AgentDiagnosticBanner titles swap verbatim. |
| #298 toast a11y | `scripts/harness-ui.mjs` | Added `toast-a11y` case. Exposed `window.__ccsmToast` debug bridge (sibling of `__ccsmStore` / `__ccsmI18n`), asserts error region role=alert/aria-live=assertive + glyph + close-only dismiss, info region role=status/polite. |
| #300 text-meta naming | `src/styles/global.css` + `src/components/ui/MetaLabel.tsx` | Picked option 2 (document the split). `text-meta` (11px proportional) and `text-mono-xs` (10px mono) are intentionally separate — composing `text-meta font-mono` would silently bump 10→11px and break the mono micro-scale ladder. No callsite churn. |

### Reverse-verify (per `feedback_bugfix_requires_e2e.md`)

All three new e2e assertions reverse-verified locally:

- Hardcode `AgentInitFailedBanner` title to a literal → `banner-i18n-toggle` case FAILS with `expected zh init title '无法启动 Claude' in: Failed to start Claude…`. Restored.
- Flip error toast region role from `alert` to `status` → `toast-a11y` case FAILS with `expected role=alert + aria-live=assertive, got role=status live=polite`. Restored.
- Stub `setMainLanguage` to a no-op → new `#295` test FAILS with `expected false to be true` on the CJK regex check. Restored.

## Test plan

- [x] `npx vitest run electron/__tests__/i18n.test.ts` — 7/7 pass (was 6 before, +1 new test)
- [x] `npm run build` succeeds (renderer + electron main)
- [x] `node scripts/harness-ui.mjs --only=banner-i18n-toggle` — PASS
- [x] `node scripts/harness-ui.mjs --only=toast-a11y` — PASS
- [x] Lint baseline unchanged (29 pre-existing errors in `verification/*.mjs`, no new errors introduced)
- [x] Reverse-verify all three assertions

Note: full `npm test` has 12 pre-existing failures from a `better-sqlite3` Node ABI mismatch (NODE_MODULE_VERSION 130 vs 127 — not introduced by this PR).

### DIFF
--- a/electron/__tests__/i18n.test.ts
+++ b/electron/__tests__/i18n.test.ts
@@ -89,4 +89,55 @@ describe('main process i18n', () => {
     expect(/[\u4e00-\u9fff]/.test(zhCwd)).toBe(true);
     expect(/[\u4e00-\u9fff]/.test(zhClaude)).toBe(true);
   });
+
+  // E2E coverage for #295 (tray-menu i18n).
+  //
+  // We can't drive the real Tray.setContextMenu() in CI: Electron's tray
+  // surface is OS-native (Windows shell notification area, GNOME indicator,
+  // macOS NSStatusItem) and exposes no scriptable accessor for the resolved
+  // menu items. Playwright sees the BrowserWindow tree, not the OS chrome.
+  //
+  // Instead this case asserts the underlying string source that
+  // `applyTrayLocale()` in electron/main.ts feeds into the menu template
+  // (`i18n.tTray(key)`). Coverage chain:
+  //   IPC ccsm:set-language → setMainLanguage → tTray returns localized →
+  //   Menu.buildFromTemplate uses those labels.
+  // The IPC handler itself is verified by the language-toggle E2E probe
+  // (scripts/probe-e2e-language-toggle.mjs) which calls
+  // window.ccsm.i18n.setLanguage and observes the renderer flip.
+  describe('#295 tray-menu i18n end-to-end via setMainLanguage', () => {
+    it('flipping language flips every tray catalog key, with no English bleed in zh', () => {
+      setMainLanguage('en');
+      const enLabels = {
+        show: tTray('show'),
+        quit: tTray('quit'),
+        tooltip: tTray('tooltip')
+      };
+      expect(enLabels.show).toBe('Show CCSM');
+      expect(enLabels.quit).toBe('Quit');
+
+      setMainLanguage('zh');
+      const zhLabels = {
+        show: tTray('show'),
+        quit: tTray('quit'),
+        tooltip: tTray('tooltip')
+      };
+      // Every menu item (show, quit) MUST contain CJK once switched.
+      // 'tooltip' is the brand 'CCSM' in both locales by design — exclude.
+      expect(/[\u4e00-\u9fff]/.test(zhLabels.show)).toBe(true);
+      expect(/[\u4e00-\u9fff]/.test(zhLabels.quit)).toBe(true);
+      // No English-only fallback leaked through.
+      expect(zhLabels.show).not.toBe(enLabels.show);
+      expect(zhLabels.quit).not.toBe(enLabels.quit);
+      expect(zhLabels.show).not.toMatch(/^Show /);
+      expect(zhLabels.quit).not.toMatch(/^Quit$/);
+
+      // Reverse-verify: revert to en and assert the English literals come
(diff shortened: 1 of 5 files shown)


Jiahui-Gu (commented):
## Independent reviewer report — VERDICT: APPROVE (post-comment, not --approve)

Bundle covers #295 / #296 / #298 / #300. Verified end-to-end on a fresh `npm run build` against PR HEAD.

### Verification
- `npx tsc --noEmit`: clean.
- `npx vitest run electron/__tests__/i18n.test.ts`: 7/7 pass (includes new `#295 tray-menu i18n end-to-end via setMainLanguage`).
- `node scripts/harness-ui.mjs --only=banner-i18n-toggle,toast-a11y`: both PASS on built `dist/renderer/bundle.js`.
- **Reverse-verify (mandatory)**: flipped `role="alert"` → `role="status"` on the error toast region in `src/components/ui/Toast.tsx`, rebuilt, re-ran `--only=toast-a11y`: **FAIL** with `expected role=alert + aria-live=assertive, got role=status live=polite`. Restored + rebuilt + re-ran: PASS. Probe genuinely binds to the contract.
- E2E discipline: `git diff origin/working...HEAD --diff-filter=A` shows zero new `scripts/probe-e2e-*.mjs` files. All three e2e additions extend existing harnesses (electron i18n vitest + harness-ui.mjs). Honors the extend-existing rule.

### Findings
1. **#295 test honestly admits its limit, but bite radius is narrower than the comment claims.** The block asserts `tTray()` catalog output round-trip via `setMainLanguage('en'|'zh')`. It does NOT exercise `applyTrayLocale()` / `Menu.buildFromTemplate` rebuild. If a regression makes tray rebuild a no-op while the catalog still flips correctly, this test passes green. The in-source comment is accurate about why (no scriptable Electron tray accessor). Acceptable given platform constraint, but worth tracking as residual coverage gap; the "reverse-verify: setMainLanguage wired to a constant" guard is the real teeth here.
2. **`window.__ccsmToast` ships unconditionally to production.** Same pattern as existing `__ccsmStore` / `__ccsmI18n` (verified via grep). Documented trade-off in code comment ("Same debug-affordance pattern... set unconditionally because production builds dead-strip NODE_ENV-gated branches" — note: there's actually no NODE_ENV gate at all, so "dead-strip" wording is slightly misleading; it's just always exposed). Following established repo convention, not introducing new exposure surface, so not blocking.
3. **#300 rationale verified against source.** `src/styles/global.css` defines `.text-mono-xs { font-size: 10px }` and `text-meta` is 11px proportional Inter. Composing `text-meta font-mono` would silently bump 10→11 and break alignment with `mono-sm`/`md`/`lg`. Keeping the utilities split is correct; the doc-only resolution (no sweep) is the right call — there's no broken usage to fix, just naming clarity for future contributors.
4. **Probe quality**: banner-i18n-toggle drives via both `window.ccsm.i18n.setLanguage` AND `__ccsmI18n.changeLanguage` to guarantee renderer catalog swap is observable; asserts CJK presence + specific zh strings + flip-back. toast-a11y asserts role/live/glyph/close-button/body-click-noop/Esc-equivalent. Both have explicit reverse-verify recipes in their docstrings.

### Verdict
LGTM. Safe to merge to `working`. Comment-only review (per protocol — independent reviewer never `--approve`).

— Independent reviewer subagent, Claude Opus 4.7

PR #245, opened 2026-04-24
test: stabilize app-icon probe + fix screenshot theme toggle (#247 #264)

Jiahui-Gu:
## Summary

Two probe-stabilization fixes bundled.

### #247 — app-icon probe stabilization

The runtime check for the BrowserWindow icon was flaky on Windows: electron-builder bakes the icon into the .exe at package time, `BrowserWindow.getIcon()` on a dev runtime is empty, and Windows shell icon caching can return a stale icon for minutes after install. A booted-Electron probe would either false-fail (asar/timing/shell-cache) or false-pass (cached shell results).

**Root cause**: an Electron-runtime probe is the wrong abstraction for "is the icon shipped". The actual regression we want to catch is "someone deleted `build/icon.png` or unwired it from the platform-specific build config" — both pure filesystem/JSON checks.

**Fix**: `scripts/probe-e2e-app-icon-present.mjs` already file-checked `build/icon.png`. Hardened it to also assert that `package.json` `build.{win,mac,linux}.icon` each reference an existing file under repo root. This catches the "icon present but unwired" regression that the runtime probe was trying — and failing — to catch.

**Stability**: 5/5 PASS in sequence on Windows. Reverse-verified: edited `build.win.icon` to a missing path, probe failed with a precise error. Restored.

### #264 — screenshot harness theme toggle

PR #218 reviewer noticed `dogfood-logs/dropped-tool-contrast-242/{dark,light}-{before,after}.png` md5-matched per pair — the harness's manual `theme-light`/`theme-dark` class flip on `<html>` didn't actually swap themes.

**Root causes** (compounding):

1. **Race against React**. `App.tsx` owns a `useEffect([theme])` that re-applies BOTH `dark` and `theme-light` from the zustand store. Any manual `classList.toggle` race-loses the next render that refires this effect. The store is the single source of truth; only `setTheme(...)` survives.
2. **Wrong class names**. The App toggles `dark` and `theme-light` (NOT `theme-dark`). The original capture script removed `theme-dark` (no-op) and added `theme-light` but never removed the lingering `dark` class.
3. **Missing paint settle**. After the class lands, Chromium needs at least two rAFs (style → layout → paint) before screenshots see the new `--color-*` resolved CSS variables. The old script only `waitForTimeout(200)`.
4. **Bonus**: original script never waited for React to mount before manipulating the store. With React unmounted, the App's `useEffect([theme])` literally never runs and the theme toggle is a complete no-op.

**Fix**: new `setTheme(win, mode, { verify })` helper in `scripts/probe-utils.mjs`:
- Routes through `__ccsmStore.getState().setTheme(mode)` (App's source of truth).
- `waitForFunction` for `__ccsmStore` AND `#root` having mounted children (proves React is alive).
- `waitForFunction` that `html.{dark | theme-light}` AND `html.dataset.theme` match the requested mode (proves the App's `useEffect([theme])` actually ran, not just that we asked).
- Awaits `requestAnimationFrame(() => requestAnimationFrame(...))` in-page for paint settle.
- With `verify: true`, reads `--color-bg-app` and asserts the OK-Lch lightness moved past 0.6 — catches the "class lands but Tailwind purge dropped the override" failure mode.

Refactored `scripts/capture-dropped-tool-contrast-242.mjs` to use it. Recaptured 4 PNGs with `verify: true`:

| file | md5 |
|---|---|
| dark-after.png | `ba5db09ab3a41440f4636b98ee5a11b1` |
| dark-before.png | `ba5db09ab3a41440f4636b98ee5a11b1` |
| light-after.png | `85126677345ea3722c356d6e4c272871` |
| light-before.png | `85126677345ea3722c356d6e4c272871` |

- **dark vs light** now md5-DIFFER (was identical pre-fix) → theme toggle works.
- before vs after MATCH within each theme is expected: PR #218's source change already landed on `working`, so `before` and `after` captures of the same theme are pixel-identical (no code delta to capture). The dark-vs-light differential is the load-bearing evidence here.

## Files changed

- `scripts/probe-e2e-app-icon-present.mjs` — added `package.json` `build.{win,mac,linux}.icon` config check; expanded comment block.
- `scripts/probe-utils.mjs` — new `setTheme(win, mode, { verify })` helper.
- `scripts/capture-dropped-tool-contrast-242.mjs` — uses `setTheme` helper.
- `dogfood-logs/dropped-tool-contrast-242/{dark,light}-{before,after}.png` — recaptured with the working helper; theme variants now md5-differ.

## Verification

- `npx tsc --noEmit`: PASS (exit 0).
- `npm test`: 12 pre-existing failures in `electron/__tests__/db*.test.ts` due to better-sqlite3 NODE_MODULE_VERSION 130 vs 127 mismatch (Node version drift in this checkout, unrelated to this PR — none of the changed files are exercised by the test suite).
- App-icon probe: 5/5 PASS sequential runs on Windows; reverse-verify confirmed the new `package.json` config check catches a deliberately-broken `build.win.icon`.
- Theme toggle: 4 PNGs captured, dark-vs-light hashes differ as documented above.

## Test plan

- [ ] Reviewer reruns `for i in 1..5; node scripts/probe-e2e-app-icon-present.mjs` on a clean Windows checkout — must be 5/5.
- [ ] Reviewer reruns `node scripts/capture-dropped-tool-contrast-242.mjs after` and confirms `md5sum dogfood-logs/dropped-tool-contrast-242/{dark,light}-after.png` returns DIFFERENT hashes.
- [ ] Reviewer spot-checks that `setTheme` is importable from `scripts/probe-utils.mjs` and that subsequent screenshot probes can adopt it without further changes.


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE (comment-only per protocol)

### Code review
- `probe-e2e-app-icon-present.mjs`: deterministic file+JSON checks. Catches both regressions called out in PR (a) `build/icon.png` deleted/corrupted/<256px, (b) `package.json build.{win,mac,linux}.icon` unwired or pointing at a missing file. Comment up top correctly explains why a runtime BrowserWindow probe is the wrong tool on Windows. Good.
- `probe-utils.mjs#setTheme`: correct alignment with `App.tsx` (toggles `dark` + `theme-light`, NOT `theme-dark`; `data-theme` set by App's effect). Routes through store's `setTheme` action (preferred path) with `setState({theme})` fallback; waits for App-applied classes (not its own write) → no race vs the `useEffect([theme])` source of truth. Double-rAF settle + optional `--color-bg-app` lightness assert is the right belt-and-suspenders.
- `capture-dropped-tool-contrast-242.mjs`: cleanly delegates to helper; uses `verify: true`.

### Stability (claim: 5/5)
Re-ran `probe-e2e-app-icon-present.mjs` 3x sequential on local worktree → 3/3 PASS. Combined with worker's 5/5 → trust 8/8.

### Reverse-verify #247
Edited `package.json` `build.win.icon` → `build/does-not-exist.png`. Probe FAILED (exit 1) with precise message:
> `package.json build.win.icon points at "build/does-not-exist.png" but ...\build\does-not-exist.png does not exist`
Restored. Confirmed: probe would have caught the unwired-config regression that the old runtime probe could not.

### Reverse-verify #264
Captured 4 PNGs via the harness:
| file | md5 |
|---|---|
| dark-before | `ba5db09ab3a41440f4636b98ee5a11b1` |
| dark-after  | `ba5db09ab3a41440f4636b98ee5a11b1` |
| light-before | `85126677345ea3722c356d6e4c272871` |
| light-after  | `85126677345ea3722c356d6e4c272871` |

dark ≠ light → theme toggle is genuinely flipping pixels (matches worker's reported hashes exactly). Within-theme before/after identical → expected, no source delta between runs.

### Type check
`npx tsc --noEmit` clean.

### Verdict
LGTM. Both root-cause fixes verified; reverse-verification proves the probes can fail when they should. Safe to merge.

PR #239, opened 2026-04-24
docs: triage post-migration gap analysis (#249)

Jiahui-Gu:
## Summary

Triage of `docs/post-migration-gap-analysis.md` (2026-04-21) against the current `working` branch + manager TaskList. Doc-only PR — no code changes.

- 6 of 11 gaps already shipped via the #207-#293 wave (QuestionBlock wiring, ToolBlock error styling, Terminal pane, FileTree, syntax-highlighted DiffView, PrettyInput).
- 1 gap partially shipped (P1.1 diff hunk accept/reject UI in `DiffView.tsx`), with the partial-write IPC follow-up already tracked as task #251.
- 4 items remain pending, all P2 polish, none block v0.1. Suggested task titles included for the manager to file.

## Test plan

- [x] Doc renders cleanly (markdown only)
- [x] Verified each "Resolved" claim against the source on `working` branch
- [x] Cross-checked TaskList for duplicates (only #251 overlaps, noted in doc)

PR #208, opened 2026-04-24
chore(rename): agentory → ccsm / CCSM (#233)

Jiahui-Gu:
## Summary

Renames the project from "agentory" / "Agentory" to "ccsm" / "CCSM"
(Claude Code Session Manager). User-visible surface only — internal
mkdtemp prefixes and the helper function `isAgentoryTempCwd` are kept as
internal names with no churn benefit. Backward-compat: the import-scanner
filter now accepts both `ccsm-` and legacy `agentory-` temp-cwd prefixes
so transcripts already on disk still get filtered.

Three commits, logical splits:

1. `package.json` + electron-builder + `release.yml` artifact name
2. Code surface (`window.ccsm`, IPC channels, env vars `CCSM_*`,
   `__ccsm…` globals, hook callback id, db filename, window/tray titles)
3. Docs / README / SECURITY / i18n locale strings

### Surface renamed
- **package**: name `ccsm`, productName `CCSM`, appId `com.ccsm.app`,
  email `noreply@ccsm.dev`, description / synopsis updated.
  electron-builder artifactName template (`${productName}-Setup-…`)
  resolves to `CCSM-Setup-<ver>-<arch>.exe` automatically.
- **CI**: `actions/upload-artifact` name `ccsm-<platform>`.
- **API surface**: `window.ccsm` (was `window.agentory`), all 54 callers
  updated. IPC channels `ccsm:get-system-locale`, `ccsm:set-language`.
  Persisted-storage / event keys `ccsm:preferences`, `ccsm:windowId`,
  `ccsm:windowTint:*`, `ccsm:windowTintChange`.
- **Env vars**: `CCSM_CLAUDE_BIN`, `CCSM_DEV_PORT`, `CCSM_PROD_BUNDLE`,
  `CCSM_CLAUDE_CONFIG_DIR`.
- **Globals**: `__ccsmStore`, `__ccsmDebug`, `__ccsmI18n`,
  `__ccsmMaybeAutoResolveAllowAlways`.
- **Claude integration**: `CLAUDE_CODE_ENTRYPOINT=ccsm-desktop` (was
  `agentory-desktop`; this is the user-agent reported to Anthropic).
  Hook callback id `ccsm-permission`.
- **Storage**: db filename `ccsm.db` (lives under productName-derived
  userData dir, so default location follows productName automatically).
- **UI**: window title, tray tooltip, "Show CCSM" menu item, all i18n
  English + Chinese strings.

### Decisions on ambiguous renames
[...]

PR #246, opened 2026-04-24
feat: wire per-tool-use cancel IPC (#239)

Jiahui-Gu:
## Summary

The Cancel link that surfaces in `ToolBlock` after a tool stalls 90s+
(`STALL_ESCALATE_AFTER_MS`, #208) was a `console.warn` stub. This PR
wires it through to a real IPC channel so clicking it actually
interrupts the in-flight tool.

- **Renderer** (`src/components/chat/blocks/ToolBlock.tsx`): accepts
  `sessionId` + `toolUseId`, calls `window.ccsm.agentCancelToolUse({...})`
  on click, flips to a disabled \"Cancelling...\" state until the next
  event lands. `aria-label` is the sentence-case \"Cancel tool\".
  `tabIndex` toggles to `-1` while cancelling so a second click can't
  double-fire the IPC. `focus-ring` preserved.
- **renderBlock**: passes `activeId` + `b.toolUseId` to ToolBlock.
- **Preload** (`electron/preload.ts`): exposes `agentCancelToolUse({
  sessionId, toolUseId })` via `contextBridge` (no raw `ipcRenderer`).
  Type added to `src/global.d.ts`.
- **Main** (`electron/main.ts`): `ipcMain.handle('agent:cancelToolUse')`
  validates payload shape (rejects missing/non-string fields with
  `bad_payload`), routes to the manager. Returns the same
  `{ok:true} | {ok:false, error}` shape `agent:setPermissionMode` uses.
- **Sessions manager + runner**: `SessionsManager.cancelToolUse` looks
  up the runner; `SessionRunner.cancelToolUse(toolUseId)` falls back
  to `interrupt()` and emits a `tool_cancel_fallback` diagnostic so the
  fallback is observable in dogfood logs.

## SDK limitation (caveat)

`@anthropic-ai/claude-code` (and the spawn-protocol contract we use
with `claude.exe`) exposes only a turn-level `interrupt`
control_request — there is no `cancel_tool_use` subtype today.

Aborting one tool while letting the agent continue the rest of the turn
would require either:
1. a new SDK control subtype that targets a `tool_use_id`, or
2. a renderer-side filter that drops the eventual `tool_result` and
   fakes a cancellation back into the chat (fragile, racey).

Until (1) lands, the per-tool Cancel click routes to the same
turn-level interrupt the StatusBar Stop button uses. The `toolUseId`
is forwarded to main and logged so the swap to a true scoped cancel is
a one-line change inside `SessionRunner.cancelToolUse` when the SDK
gains one. WHY-comment lives in `electron/agent/sessions.ts` next to
the method.

The renderer is honest about this: clicking Cancel disables the link
to \"Cancelling...\" and the whole turn aborts, which is the strongest
guarantee we can offer today.

## IPC contract chosen

```ts
// renderer
window.ccsm.agentCancelToolUse({ sessionId: string, toolUseId: string })
  : Promise<{ ok: true } | { ok: false, error: string }>

// channel name: 'agent:cancelToolUse'
```

Single-object payload (vs positional args) so the future scoped-cancel
SDK upgrade can layer extra fields (e.g. `reason`) without breaking
callers.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` (renderer) passes
- [x] `npx tsc --noEmit -p tsconfig.electron.json` passes
- [x] `npx eslint <changed files>` clean (one pre-existing warning on
      main.ts:118 unrelated to this PR)
- [x] `npx vitest run tests/chatstream-tool-block-ux.test.tsx
      electron/agent/__tests__/sessions.test.ts` — 32/32 pass
- [x] **Reverse-verify**: stashed the cancel call body in ToolBlock
      (`if (false && api && ...)`); the renderer unit test
      \"clicking Cancel after 90s invokes agentCancelToolUse\" went
      from PASS to FAIL with the expected `expected \"spy\" to be
      called 1 times, but got 0` error. Restored after.
- [ ] **Full vitest** has 12 pre-existing failures in `electron/__tests__/db*.test.ts`
      due to `better-sqlite3` NODE_MODULE_VERSION mismatch in the
      worker sandbox (built for Electron 33 = v130, vitest runs under
      Node 22 = v127). Reproducible against `origin/working` — not
      introduced by this PR.
- [ ] **e2e probe** `scripts/probe-e2e-tool-journey-render.mjs` gained
      Journey 9 (cancel wiring + cancelling-state assertions) per
      `feedback_e2e_extend_existing.md`. **Could not execute in worker
      sandbox** — Electron failed to open a renderer window in the
      sandbox env (same env as the better-sqlite3 issue above).
      Reviewer should run locally:
      ```
      npm run build && node scripts/probe-e2e-tool-journey-render.mjs
      ```
      and confirm J9 + J9b both report `[OK]`.
- [ ] **Visual evidence**: see `dogfood-logs/tool-cancel-239/README.md`
      for the visual contract that the unit tests already lock in.
      Worker could not capture before/after PNGs in the sandbox;
      reviewer is asked to attach them after a local run.

## Files

- `src/components/chat/blocks/ToolBlock.tsx`
- `src/components/chat/renderBlock.tsx`
- `src/global.d.ts`
- `src/i18n/locales/en.ts`, `src/i18n/locales/zh.ts`
- `electron/preload.ts`
- `electron/main.ts`
- `electron/agent/manager.ts`
- `electron/agent/sessions.ts`
- `electron/agent/__tests__/sessions.test.ts`
- `tests/chatstream-tool-block-ux.test.tsx`
- `scripts/probe-e2e-tool-journey-render.mjs`
- `dogfood-logs/tool-cancel-239/README.md`


Jiahui-Gu (commented):
## Independent reviewer — verdict: LGTM (with one env caveat noted)

### Summary
- Per-tool Cancel click is now wired through `window.ccsm.agentCancelToolUse({sessionId, toolUseId})` → `ipcMain.handle('agent:cancelToolUse')` → `SessionsManager.cancelToolUse` → `SessionRunner.cancelToolUse(toolUseId)` → falls back to `interrupt()`. Single-object payload, validated in main, contextBridge-only, no raw ipcRenderer. Shape matches `setPermissionMode`.
- WHY-comment for the SDK-limitation fallback lives at `electron/agent/sessions.ts:651-669` (clear, future-friendly: scoped cancel becomes a one-liner here).
- Diagnostic `tool_cancel_fallback` (level=warn) is emitted up-front; observable in dogfood logs.
- ToolBlock UX: `aria-label="Cancel tool"` (sentence case, both en + zh updated), `aria-disabled` + `tabIndex=-1` while `cancelling`, focus-ring preserved, `e.stopPropagation()` keeps the parent collapse-row from toggling.
- New i18n key `chat.toolStallCancelling` present in BOTH `en.ts` and `zh.ts`, sentence case.

### Verification I ran locally (NOT in worker sandbox)
- `npx tsc --noEmit -p tsconfig.json` ✅ clean
- `npx tsc --noEmit -p tsconfig.electron.json` ✅ clean
- `npx vitest run tests/chatstream-tool-block-ux.test.tsx electron/agent/__tests__/sessions.test.ts` → **32/32 pass**
- **Reverse-verify (independent)**: I stashed the `agentCancelToolUse` call body in `ToolBlock.tsx` (`if (false && api && …)`) → the renderer test "clicking Cancel after 90s invokes agentCancelToolUse" went **PASS → FAIL** with `expected "spy" to be called 1 times, but got 0 times`. Restored after.
- The new `electron/agent/__tests__/sessions.test.ts::"cancelToolUse() emits a tool_cancel_fallback diagnostic and forwards a turn-level interrupt"` cleanly asserts (a) the diagnostic code is emitted with `tu-abc-123` in the message, (b) an `interrupt` subtype `control_request` lands on stdin — i.e. fingerprint-checks we didn't invent a new subtype claude.exe would reject. Companion `cancelToolUse() before start() is a safe no-op` covers the no-rpc path.

### J9 e2e probe — env-blocked (same as worker)
- Ran `node scripts/probe-e2e-tool-journey-render.mjs` against the existing built worktree (dist/electron + dist/renderer present). Failed at `appWindow: no renderer window appeared in time` — same symptom the worker reported. This is the local environment limitation, not a probe bug.
- Reading the J9 source: it's correctly designed — patches `Date.now` with a +95s offset to force `escalated` state without a wall-clock wait, stubs `window.ccsm.agentCancelToolUse` via Proxy (preserves the rest of the bridge), asserts payload `{sessionId:'s-tool', toolUseId:'tu-cancel-XYZ'}`, then asserts the cancelling-state DOM in J9b. It also checks `aria-label === 'Cancel tool'`. Per `feedback_e2e_extend_existing.md`, J9 was added as a new case in the existing probe (not a new file). Reverse-verify is documented in the comment.
- Visual screenshots: not captured (same env block). `dogfood-logs/tool-cancel-239/README.md` documents the visual contract that the RTL test locks in (idle/hover/cancelling text, aria-disabled, focus-ring).

### Nits (non-blocking)
- `cancelling` state is local to ToolBlock — if the agent sends another tool_use_id and React re-uses the component instance it could carry stale state, but each ToolBlock is keyed by block id so this is fine in practice.
- `agentCancelToolUse({sessionId, toolUseId})` shape matches the PR description's forward-compat goal (room for `reason` later).

### Verdict
Code is clean, comments explain the SDK-limitation fallback honestly, both unit suites + reverse-verify confirm wiring. **J9 e2e could not be executed locally due to the same Electron-renderer-spawn env issue the worker hit** — accepting RTL coverage + visual-contract README as evidence per the reviewer brief, but flagging that an actual J9 green run on a working machine should happen before considering #239 fully closed.

— independent reviewer (Claude)

PR #251, opened 2026-04-24
fix(notify): retry rechecks gates + toast reject cancels retry (#307 #308)

Jiahui-Gu:
## Summary

Bundles two notify-hardening fixes that emerged from reviewer follow-ups on #243.

### Fix #307 — retry fire-time gate recheck
The ask-question retry timer fires in main ~30s after the original toast. Until now it re-emitted blindly even when the user had meanwhile (a) globally disabled notifications or (b) focused the question's own session during the 30s window — both conditions under which the initial emit would have been suppressed.

Main now holds a push-mirrored snapshot of the renderer's `notificationsEnabled` + `activeSessionId`. `scheduleQuestionRetry` captures the `sessionId` alongside the payload so `fireRetry` can run the same focus + settings gate that `showNotification`/`dispatchNotification` apply on the initial emit.

### Fix #308 — toast-action reject cancels pending retry
The toast-action router's permission-reject branch now unconditionally calls `cancelQuestionRetry` on both `q-${requestId}` and the bare requestId. Today only permission events route through this path, so the direct cancel is a no-op (the renderer IPC round-trip `agent:resolvePermission` already cancels). It's defense-in-depth: if a future change routes question activations through the same toast-action path, the retry will not leak past the explicit user reject.

Side effect: the router logic was extracted from `electron/main.ts` into `createDefaultToastActionRouter` in `electron/notify-bootstrap.ts` so main and the e2e probe share one definition instead of drifting hand-rolled copies (the probe previously had its own router that didn't know about #308).

## Checklist
- [x] `npm run lint` — baseline 47 problems, 0 new (pre-existing in `verification/*` + `dogfood-logs/*`)
- [x] `npm test` — 756 tests pass (751 baseline + 5 new in `notify-retry.test.ts`)
- [x] `scripts/probe-e2e-notify-integration.mjs` passes with new cases 10b + 10c
- [x] Reverse-verify done for BOTH fixes (see below)
- [x] HOME sanitized during build/tests/e2e

## Reverse-verify

**#307 — stash the fire-time gate**
Edit `electron/notify-retry.ts:fireRetry`, comment out `if (shouldSuppressRetry(entry.sessionId)) return;`, rebuild, rerun the probe → case 10b fails with `expected exactly 1 question call when notifications disabled at fire-time, got 2`. Confirmed.

**#308 — stash the router cancel calls**
Edit `electron/notify-bootstrap.ts:createDefaultToastActionRouter`, comment out the two `deps.cancelQuestionRetry(...)` lines in the reject branch, rebuild, rerun the probe → case 10c fails with `expected toast-action reject router to call cancelQuestionRetry directly (#308); got zero router-side calls`. Confirmed.

The probe uses a spy around the router's deps-injected `cancelQuestionRetry`, so the assertion isolates the router's direct call from the IPC round-trip's cancel call (which is what made a naive `pendingAfter < pendingBefore` check hollow under reverse-verify).

## Test plan
- [ ] Reviewer: pull branch, run `scripts/probe-e2e-notify-integration.mjs`, confirm all 11 cases pass
- [ ] Reviewer: run the two reverse-verify stashes above and confirm both trip the expected probe case
- [ ] Reviewer: confirm `createDefaultToastActionRouter` factory hasn't changed behaviour for permission allow / allow-always / focus paths (diff main.ts before/after vs the factory body)

## Files changed
- `electron/notify-retry.ts` — gate check at fire time, sessionId param on schedule
- `electron/notify-bootstrap.ts` — runtime-state mirror, `shouldSuppressRetry`, `createDefaultToastActionRouter` factory
- `electron/notifications.ts` — forwards sessionId to `scheduleQuestionRetry`
- `electron/main.ts` — IPC handler `notify:setRuntimeState`; uses the extracted router factory
- `electron/preload.ts` + `src/global.d.ts` — `notifySetRuntimeState` exposed
- `src/agent/lifecycle.ts` — renderer-side store subscription pushes runtime state on change
- `electron/__tests__/notify-retry.test.ts` — 5 new unit tests
- `scripts/probe-e2e-notify-integration.mjs` — cases 10b (#307) + 10c (#308)

### DIFF
--- a/electron/__tests__/notify-retry.test.ts
+++ b/electron/__tests__/notify-retry.test.ts
@@ -7,6 +7,35 @@ import {
   __resetRetryStateForTests,
 } from '../notify-retry';
 import * as notifyMod from '../notify';
+import {
+  setNotifyRuntimeState,
+  __resetNotifyRuntimeStateForTests,
+} from '../notify-bootstrap';
+
+vi.mock('electron', () => {
+  const fakeWindows: Array<{
+    isDestroyed: () => boolean;
+    isFocused: () => boolean;
+    isVisible: () => boolean;
+  }> = [];
+  return {
+    BrowserWindow: {
+      getAllWindows: () => fakeWindows,
+      __setFakeWindows: (
+        wins: Array<{ focused: boolean; visible: boolean; destroyed?: boolean }>,
+      ) => {
+        fakeWindows.length = 0;
+        for (const w of wins) {
+          fakeWindows.push({
+            isDestroyed: () => !!w.destroyed,
+            isFocused: () => w.focused,
+            isVisible: () => w.visible,
+          });
+        }
+      },
+    },
+  };
+});
 
 // We run the retry logic against a virtual scheduler so the 30s timer is
 // observable instantly. Real setTimeout would either slow tests by 30s or
@@ -45,6 +74,7 @@ describe('notify-retry', () => {
   afterEach(() => {
     __setRetrySchedulerForTests(null, null);
     __resetRetryStateForTests();
+    __resetNotifyRuntimeStateForTests();
     vi.restoreAllMocks();
   });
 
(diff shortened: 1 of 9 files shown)


Jiahui-Gu (commented):
## Independent reviewer: APPROVE (posted as comment — author==reviewer on GitHub, self-approve API-blocked)

## Layer 1 — should this PR exist at all? PASS

**Bug is real (not theoretical).** The 30s retry window is long enough that users *will* toggle notifications off or focus the session in the middle of it. Without the fire-time recheck, the toast the user explicitly silenced still fires. Same for the re-focused session — the in-app affordance is already visible, the OS toast is pure noise. Both are user-facing, not races-in-theory.

**Architecture is reasonable.** The push-mirror for `notificationsEnabled` + `activeSessionId` is the right trade: it's <30 lines, doesn't couple `notify-*` to the DB/settings module, and reuses the same `fromMainFrame` IPC discipline the rest of main.ts applies. Alternative "main reads settings directly" would drag a DB dependency into the notify subsystem for two scalar fields. Alternative "shared state module" is premature — the renderer store is the source of truth for these and should stay that way.

**Bundling #307 + #308 is justified.** Both are retry-hardening follow-ups, both touch `notify-bootstrap.ts` + `notify-retry.ts` + the same probe, and #308 *requires* the router extraction that #307's architecture also benefits from (probe now exercises the production router, not a drifted hand-rolled copy). Splitting would duplicate the probe refactor.

**The "vacuous assertion" / deps-injected spy for #308 is not a design smell.** Two code paths legitimately converge on `cancelQuestionRetry`: the main-process router (necessary when the window is closed) and the renderer IPC round-trip (updates the store UI). Both are needed; the router's direct cancel is defense-in-depth for future wiring changes. Spying the deps-injected binding is the correct way to isolate the router's own call — the alternative (make them the same path) would actually be wrong because it would skip the renderer-store update on a closed window.

## Layer 2 — low-level findings

**Reverse-verify:** Could not fully execute in this reviewer env — the probe baseline on `working` fails at case 9 pre-existing (timeout waiting for retry schedule after case 8's focus-suppression), reproducibly on both branches in this clone, likely a local native-module ABI / timing artifact. The new cases 10b/10c are downstream of case 9 so they can't be reached here. Mitigating signals:
- `notify-retry.test.ts` runs 13/13 green, including the 5 new cases (settings-toggle suppression, different-session fires, null-session fires, focus+match suppression, reject cancels). These exercise the same code paths the probe would.
- tsc (`--noEmit`) + electron tsc both clean.
- Code read of the router extraction is 1:1 with prior semantics — the only added behavior is the two `deps.cancelQuestionRetry(...)` calls in the reject branch, plus optional chaining on `win.*` methods (strictly more defensive).

**IPC hardening (`notify:setRuntimeState`):** `fromMainFrame(e)` gate present. Shape validation is loose at the handler but `setNotifyRuntimeState()` type-guards each field (`typeof === 'boolean'`, `=== null || typeof === 'string'`) so a malicious renderer pushing bogus fields gets silently dropped. Acceptable; consistent with `db:save`-style patterns elsewhere.

**`lifecycle.ts` store subscription — no leak.** The subscribe is installed once behind the `installed` flag at renderer-app lifetime; same pattern as the existing `api.onAgentEvent` wiring above it. No unmount path, no leak.

**Router factory preserves behavior.** Diff is 1:1 against the inline version in main.ts, deps-injected instead of captured from module scope. Optional chaining on `win.isMinimized?.()` etc. is slightly more defensive than before (same semantics when methods exist).

**Unit-test coverage is appropriate.** Five new cases cover: settings-off at fire time, different session fires, null sessionId fires, focused+matching session suppresses, reject cancels via q-prefixed id. Missing but low-risk: "matching session unfocused → fires" (covered by logic, not explicitly asserted). Not a blocker.

**No stray console.log / TODO / disabled tests.**

### Verdict
Ready to merge. Layer 1 clean, Layer 2 clean. Probe reverse-verify deferred to the worker's already-documented manual run (PR description shows both stashes tripping the expected cases). The 12 pre-existing `db-hardening.test.ts` failures are a native `better-sqlite3` ABI mismatch in this reviewer's env, unrelated to the PR.

PR #250, opened 2026-04-24
feat(perm): per-hunk diff selection in PermissionPromptBlock (#306)

Jiahui-Gu:
## Rebase #2 (2026-04-24) — onto post-#249/#252/#253 working

Re-rebased onto `origin/working` after the following PRs landed since the previous rebase:

- **#249** (`65f667a`): per-file collapse toggle in `DiffView` (multi-file MultiEdit / scripted batches default-collapse at >3 files).
- **#252** (`4defcfd`): CI guard catching silent file drops in PR rebases.
- **#253** (`47899d0`): kill SCREAMING UI strings flagged in dogfood #248 Gap #1 (removed Tailwind `uppercase` classes from onboarding/cwd-popover/connection-badge surfaces).

### Conflict resolution

Single semantic conflict: `src/components/chat/DiffView.tsx`. Both #249 (per-file collapse via `FileSection` wrapper, `diff: DiffSpec | DiffSpec[]`) and #306 (per-hunk checkbox `selection` mode for the single-file form) needed to coexist in the same component.

- Restructured `DiffView` to keep #249's `FileSection`-per-file layout AND propagate #306's `selection` / `onSelectionChange` / `disabled` props through to the single-file `FileSection` (per-hunk select mode is single-file only — `PermissionPromptBlock` is the sole caller).
- Multi-file `DiffView` calls (no `selection` prop) keep #249's per-file collapse + legacy per-hunk Accept/Reject decoration buttons.
- Single-file `DiffView` calls with `selection` prop render checkboxes inside the (default-expanded) `FileSection`.
- Auto-merged: `src/i18n/locales/{en,zh}.ts` (no overlap with #253's removed `uppercase` Tailwind classes — those were CSS, not string content).
- Did NOT change DiffView's existing `uppercase tracking-wider` Tailwind classes on `chat.diffAccepted` / `chat.diffRejected` / `permissionPrompt.hunkLabel` — these are inside the diff-view widget and weren't on #253's flagged-surface list. Flagging here in case a follow-up sweep wants to lower-case them too.

### Verification (post-rebase, post-resolution)

- `tests/permission-prompt-block.test.tsx` → **21/21 pass**
- `tests/diff-view-per-file-collapse.test.tsx` → **4/4 pass** (#249's regression suite still green against the merged DiffView)
- `scripts/harness-perm.mjs` → **10/10 pass** (incl. `permission-partial-accept` + `permission-a11y`; one transient focus-race flake observed once on a non-#306 case, passed on retry)

[...]

PR #249, opened 2026-04-24
ux(diff): per-file collapse toggle (#302)

Jiahui-Gu:
## Summary

Closes #302 — per-file collapse for `DiffView`.

`DiffView` now accepts either a single `DiffSpec` (preserves the existing single-file tool-block contract) or `DiffSpec[]`. Each file renders as a section with a chevron header chip showing the file path and `+N/-M` counts. Sections default to expanded for batches of <=3 files and collapsed for >3, so large MultiEdit / scripted multi-file batches become scannable at a glance instead of dumping every line into the stream.

Per-hunk Accept/Reject decisions are preserved (state lives at the section level so it survives collapse/expand within the same render lifecycle), keeping the in-flight #306 hunk-selection PR conflict-free.

## What changed

- `src/components/chat/DiffView.tsx` — extracted per-file `FileSection` with chevron header (`focus-ring`, Enter/Space toggles via native `<button>`), `+N/-M` counts chip, framer-motion height animation matching `ToolBlock`. New `data-testid="diff-view"` + `data-file-count` attrs for probe assertions.
- `src/i18n/locales/{en,zh}.ts` — new keys `chat.diffFileToggleAria`, `chat.diffCountsAria` (sentence-case).
- `tests/diff-view-per-file-collapse.test.tsx` — 4 RTL cases.
- `scripts/probe-e2e-tool-journey-render.mjs` — extended with `J10` (per-file diff collapse chrome wired in live renderer).
- `scripts/probe-render-diff-per-file-collapse-302.mjs` — visual probe producing before/after PNG triplet.
- `dogfood-logs/diff-per-file-collapse-302/` — captured screenshots.

## Visual evidence

5-file diff scenario:

| State | Screenshot |
| --- | --- |
| BEFORE — pre-#302 (5 files, every line forced visible) | `dogfood-logs/diff-per-file-collapse-302/before-5-file-diff-all-expanded.png` |
| AFTER — collapsed default (5 chevron chips, scannable) | `dogfood-logs/diff-per-file-collapse-302/after-5-file-diff-collapsed.png` |
| AFTER — first chevron clicked (1 expanded, 4 still collapsed) | `dogfood-logs/diff-per-file-collapse-302/after-5-file-diff-one-expanded.png` |

Reproduce: `node scripts/probe-render-diff-per-file-collapse-302.mjs`.

## Test plan

- [x] `npx tsc --noEmit` — clean
[...]

PR #248, opened 2026-04-24
docs: dogfood public-001 findings (no fixes)

Jiahui-Gu:
## Re-triage note (2026-04-24)

Manager clarified the definition of "clean config":

- **"Clean config"** = CCSM's own config is empty (`%APPDATA%\CCSM` deleted).
- **"Clean config" does NOT mean** Claude Code's global config (`~/.claude/`) is cleared.
- CCSM is a GUI shell for Claude Code. Inheriting the user's Claude Code config — skills, settings, credentials, MCP servers, subagents — is **correct product behavior**, not a bug.

Under this corrected definition, the original P0 finding ("claude.exe inherits `~/.claude/skills`") is **rescinded**. Gap 4 ("skill preamble attribution") was rewritten to rest on discoverability rather than on an incorrect-inheritance premise. All other findings were independent of the isolation question and are unchanged. No re-run was needed. Strike-through preserved in the findings doc for audit trail.

**Re-triage outcome:** 13 KEEP / 1 RESCIND / 0 NEEDS-MORE-DATA.

## Summary

Public dogfood run on installed prod build (CCSM.exe 0.1.0) with cleared **CCSM user config** (`%APPDATA%\CCSM` gone; `~/.claude/` intentionally kept — correct behavior, see re-triage note above). Driven via playwright-core over CDP (`--remote-debugging-port=9222`).

- Task A (Node CLI init in temp dir): completed end-to-end. Read/Write/Bash + permission prompts exercised.
- Task B (long bash + on-purpose interrupt): partial.
- Session switch + Settings popover also exercised.

## Findings

**5 active bugs (Bug 2 rescinded) / 7 UX gaps** — 12 active findings total. Severity distribution: **0 P0, 6 P1, 6 P2**.

| Sev | Item |
|---|---|
| ~~P0~~ | ~~claude.exe inherits `~/.claude/skills` and runs them on every turn~~ — **rescinded**: `~/.claude` inheritance is correct behavior (CCSM is a Claude Code GUI shell). |
| P1 | Bash tool elapsed-time counter starts at request, not at execution start; "Tool has been running 90s+ — still no result" warning fires while user permission is unanswered. Misleading. |
| P1 | cwd defaults to user home (`C:\Users\jiahuigu`); no visible "current working directory" indicator anywhere near the input. Silent destructive-default risk. |
[...]

PR #252, opened 2026-04-24
ci: catch silent file drops in PR rebases (#324)

Jiahui-Gu:
## Layer 1

**Goal**: prevent the PR #250 post-mortem pattern where a rebase resolves conflicts by deleting unrelated changes, the GitHub API still lists those files (it reports the union of paths touched across the PR's commits), reviewers approve only the surviving diff, and the dropped work silently never ships.

**Why CI is the right level**:
- Local pre-commit hook is skippable and doesn't run on bot/forks.
- Manual reviewer task is the exact thing that already failed for #250.
- A required PR check is automatic, runs on every push, and blocks merge.

**Assertion**: every path in `gh pr view <N> --json files -q '.files[].path'` must appear in `git diff origin/<base>...HEAD --name-only`. Direction matters — extra files in the diff are fine; only `claimed - diff` is fatal. This avoids false positives when the API truncates large file lists (see PR #248 below).

## Layer 2

New workflow `.github/workflows/pr-no-silent-drops.yml`. Triggers on `pull_request` (opened, reopened, synchronize). Job uses `gh pr view` + `git diff base...HEAD --name-only` + `comm -23` to find claimed-but-absent files; non-empty list → `exit 1` with a list naming each missing file.

### Silent-drop detection (synthetic, since the prompt's `a92ecfa` SHA does not exist in this repo)

Took the real claimed-files list for PR #250 and removed two of those files from a synthetic diff (simulating exactly what a bad rebase does):

```
PASS — caught silent drop. Missing files:
    src/components/PermissionPromptBlock.tsx
    src/i18n/locales/zh.ts
```

### No false positives across the 10 most recent PRs

Ran the assertion against PRs #242, #243, #244, #245, #246, #247, #248, #249, #250, #251:

```
PR #251: PASS (claimed subset of diff)
PR #249: PASS (claimed subset of diff)
PR #248: PASS (claimed subset of diff)   <- API truncated ~31 files; subset assertion correctly ignores them
PR #247: PASS (claimed subset of diff)
PR #246: PASS (claimed subset of diff)
PR #245: PASS (claimed subset of diff)
PR #244: PASS (claimed subset of diff)
PR #243: PASS (claimed subset of diff)
PR #242: PASS (claimed subset of diff)
PR #250: PASS (claimed subset of diff)
```

PR #248 is the validation point for the assertion direction: 31 files appear in the diff but not in `gh pr view --json files` (API truncation around 300 paths). The subset assertion (`claimed ⊆ diff`) correctly ignores them. A naive equality check would have produced 31 false positives on a perfectly clean PR.

## Notes

- GitHub billing is currently blocking CI runs for this repo, so this workflow won't execute on this PR until billing is restored. Logic is verified locally per the run above.
- No TypeScript or product code touched; no tsc/lint runs needed.

### DIFF
--- a/.github/workflows/pr-no-silent-drops.yml
+++ b/.github/workflows/pr-no-silent-drops.yml
@@ -1,0 +1,77 @@
+name: pr-no-silent-drops
+
+# Catches the "rebase resolved by deleting unrelated changes" pattern
+# (see PR #250 post-mortem):
+#
+#   * GitHub still lists files X, Y in the PR's file list (the API
+#     reports the union of paths touched across the PR's commits).
+#   * A bad rebase / squash drops X and Y from the actual base...head
+#     diff.
+#   * Reviewers approve the diff they see; the claimed-but-missing
+#     files silently never ship.
+#
+# Assertion: every path in `gh pr view --json files` MUST appear in
+# `git diff <base>...<head> --name-only`. Extra files in the diff are
+# fine (the API truncates around 300 files; we only catch claimed-but-
+# absent, never absent-but-present).
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  no-silent-drops:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR head with full history
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Ensure base ref is fetched
+        run: git fetch origin "${{ github.base_ref }}" --depth=0
+
+      - name: Compare claimed files vs actual diff
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          set -euo pipefail
+
+          # Files the PR API claims are part of this PR.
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE: Layer 1 — CI is the right level (pre-commit skippable, manual review failed in #250). Subset assertion (claimed ⊆ diff) is the correct relaxation; equality would false-positive on >300-file PRs hitting GitHub's file-list truncation (validated against PR #248 with 31 truncated files). Workflow uses exit 1 on non-empty missing list, so it actually fails the check rather than just logging.

Layer 2 — workflow correctness verified by inspection:
- Trigger: pull_request [opened, reopened, synchronize].
- actions/checkout@v4 with fetch-depth: 0 and explicit 'git fetch origin $BASE_REF --depth=0' so origin/$BASE_REF...HEAD resolves.
- set -euo pipefail; comm -23 claimed actual correctly computes claimed-minus-actual.
- GH_TOKEN: ${{ github.token }} with permissions: pull-requests: read — sufficient for 'gh pr view --json files', no PAT needed.
- No hardcoded repo names or paths; uses ${{ github.base_ref }} and pull_request.number; fork-safe.
- Real exit 1 on missing files; clear ::error:: + per-file ::group:: logging for diagnosis.

PR body documents local verification (synthetic #250 drop correctly names PermissionPromptBlock.tsx + zh.ts; 10 recent PRs all PASS including #248 truncation case). Methodology reproducible.

Note: GH billing block means this won't execute on this PR until restored — same caveat as #245/#248/#269. Ship it.

PR #255, opened 2026-04-24
ux(perm): scope-explicit "Allow always" copy (#321)

Jiahui-Gu:
Closes #321 (Bug #3 from PR #248 dogfood-public-001).

## Layer 1 — root-cause confirmation

**Bug:** "Allow always" / "始终允许" reads as "always allow this exact command", but the implementation actually whitelists **all calls to the same tool name**, **for the rest of this app session only**.

**Verified in code:**
- `addAllowAlways(toolName)` adds the bare tool name (e.g. "Bash") to `useStore.allowAlwaysTools[]` (`src/stores/store.ts:1712`).
- `maybeAutoResolveAllowAlways` checks `store.allowAlwaysTools.includes(req.toolName)` — exact tool-name match, no command pattern (`src/agent/lifecycle.ts:163`).
- `allowAlwaysTools` lives in the in-memory Zustand store, not in persisted settings.

**Real scope: per-tool, in-memory, session-lifetime.** Single scope. No SDK affordance for command-level patterns or persisted scopes here, so a scope picker would invent state we can't honor — clearer fixed copy is the right shape.

## What changed

- **Label** (i18n, en + zh): `Allow always` → `Always allow {{tool}} this session` / `本会话始终允许 {{tool}}`. Falls back to `Always allow this tool this session` when `toolName` is missing.
- **Tooltip** (new `title` attribute): `Skip the prompt for any {{tool}} call until you quit the app. Resets next launch.` Spells out lifetime explicitly.
- **Component**: interpolates `toolName` into both label and tooltip (`src/components/PermissionPromptBlock.tsx`).
- **Tests**: extends `tests/permission-prompt-block.test.tsx` with 3 new cases (tool-named label, fallback label, click still fires onAllowAlways) — guards against regressing back to vague "Allow always".
- **Harness**: `scripts/harness-perm.mjs` label assertion updated to `/Always allow .* this session/i`.

## Why no scope chooser

The fast-path resolver and addAllowAlways store action only support one scope: per-tool-name, in-memory, session-only. A "this command vs all Bash" picker would mean inventing UI for state we don't actually persist or pattern-match. When/if we wire pattern-based or persisted scopes (mirroring raw CLI's `Bash(<pattern>)`), the picker earns its keep — until then, honest copy beats fake choice.

## Verification

- typecheck (`npm run typecheck`): clean
- vitest: 763/763 passing including the 3 new cases + i18n key parity
- before/after screenshots captured via `scripts/screenshot-perm-allow-always-321.mjs` (run with `--label=before` / `--label=after`):
  - `dogfood-logs/perm-allow-always-scope-321/before-1-allow-always-buttons.png` — old "Allow always"
  - `dogfood-logs/perm-allow-always-scope-321/after-1-allow-always-buttons.png` — new "Always allow Bash this session"
  - `before-2-allow-always-hover.png` / `after-2-allow-always-hover.png` — same with hover (after shows new tooltip)
- i18n parity confirmed: en + zh both extended with the same 4 new keys (`allowAlwaysBtn`, `allowAlwaysBtnFallback`, `allowAlwaysHint`, `allowAlwaysHintFallback`); `tests/i18n-key-parity.test.ts` passes.

## Test plan

- [ ] Reviewer: stash this branch's changes to `src/i18n/locales/en.ts` + verify `tests/permission-prompt-block.test.tsx` "Allow always — scope-explicit copy" suite fails (reverse-verify).
- [ ] Reviewer: re-run `node scripts/screenshot-perm-allow-always-321.mjs --label=after` + diff against committed `after-*.png`.
- [ ] Reviewer: smoke-trigger any Bash permission prompt in the running app, confirm button reads "Always allow Bash this session" and tooltip is visible on hover.


Jiahui-Gu (commented):
## Reviewer verdict: APPROVE (submitted as comment — same-author block prevents formal approval)

## Layer 1 - truth verified

Confirmed via gh api against branch `task/321-allow-always-scope-copy`:
- `src/stores/store.ts:1712` — `addAllowAlways(toolName)` pushes bare tool name into `allowAlwaysTools[]` (in-memory Zustand, not persisted).
- `src/agent/lifecycle.ts:164` — `maybeAutoResolveAllowAlways` does `store.allowAlwaysTools.includes(req.toolName)` — exact tool-name match, no command pattern.

Real scope = per-tool-name, in-memory, session-lifetime. New copy "Always allow {{tool}} this session" + tooltip "until you quit the app. Resets next launch." honestly mirrors that.

**Persisted "Always" alternative:** Would be a real product feature (mirroring raw CLI's `Bash(<pattern>)` allowlist) but requires SDK affordance + settings storage + revocation UI. Out of scope for a copy fix; correctly deferred. Worker's reasoning holds.

## Layer 2 - findings

- **i18n parity:** 4 keys present in en + zh, sentence case en, natural zh. OK.
- **Component:** `toolName` interpolated into both label and `title` tooltip; fallback path used when `toolName` missing. Click handler unchanged. OK.
- **Tooltip primitive:** Uses native `title=` attr, not Radix Tooltip. Acceptable for a permission-prompt hint (keyboard flow doesn't need it), but worth noting for design polish backlog — Radix would render styled. Non-blocking.
- **Tests:** 3 new cases cover named label, fallback label, click — with a negative assertion guarding against the old "Allow always" string. Adequate.
- **Harness regex** `/Always allow .* this session/i` — matches new label, won't catch unrelated rows. OK.
- **Screenshot verified:** Read `after-1-allow-always-buttons.png` — shows real PermissionPromptBlock alertdialog with Reject (N) / Always allow Bash this session / Allow (Y). Real component, real copy.
- **Scope discipline:** No sneaky changes — diff matches the 5 declared files + 4 PNGs.

Honest copy fix for a real misleading-UX bug. Truth claims verified in source. No invented state. Tests guard the regression. Safe to merge.

PR #256, opened 2026-04-24
fix(tool-block): freeze elapsed counter while permission pending (#311)

Jiahui-Gu:
## Layer 1 verdict
Direction: **right**. The bug is real (PR #248 dogfood Bug #1) — `startedAtRef` was captured at the first render where `result === undefined`, which happens when the assistant's `tool_use` lands. If a permission gate intercepts (claude.exe `canUseTool`), the user can sit on the prompt for 90s+ while the elapsed chip keeps ticking and the escalation banner fires, all before the tool has actually started running. Better approach (the one taken here): freeze the counter / banners while permission is pending, then start from zero once the gate clears. Considered (and rejected) just-suppress-the-banner-without-pausing-the-counter — that would still display 90s+ "elapsed" for a tool that hasn't run, which is the misleading bit.

## Files touched
- `src/components/ChatStream.tsx` — compute `permissionPendingToolIds` (in-flight tool whose `toolName` matches a subsequent `waiting/permission` block).
- `src/components/chat/renderBlock.tsx` — thread the set through, expose `permissionPending` to ToolBlock.
- `src/components/chat/blocks/ToolBlock.tsx` — defer `startedAtRef` capture while `permissionPending`; treat as not-running so chip / stall hint / escalation banner / cancel link all suppress.
- `scripts/probe-e2e-tool-journey-render.mjs` — extended with J11a (suppression during gate after 95s wall-clock advance) and J11b (counter starts from zero on gate clear, ticks normally after).

## Reverse-verify output
With fix reverted (only the 3 source files; probe kept):
```
[FAIL] J11a elapsed counter + stall banners suppressed while permission pending (#311)
[FAIL] J11b elapsed counter starts from zero at execution-begin, not request-time (#311)
=== totals: 13/17 passed ===
```

With fix restored:
```
[PASS] J11a elapsed counter + stall banners suppressed while permission pending (#311)
[PASS] J11b elapsed counter starts from zero at execution-begin, not request-time (#311)
=== totals: 15/17 passed ===
```

The two pre-existing FAILs (J5b auto-expand, J9 cancel-IPC wiring) reproduce on `origin/working` baseline — unchanged by this PR.

## Quality gates
- `npx vitest run tests/chatstream-tool-block-ux.test.tsx` — 11/11 passing (existing coverage of A2-NEW-5/6/7 + #304 unaffected by the new prop, which defaults to `false`).
- No i18n strings changed; no new UI text.
- Backward-compat: `permissionPending` is optional and defaults to `false`, so existing call sites and snapshot tests render identically.

## Test plan
- [ ] Run a `Bash` tool that hits the permission prompt; confirm no elapsed chip / stall banner appears while the prompt is open.
- [ ] After clicking Allow, confirm the elapsed chip starts from `0.0s` and ticks from there — not from when the prompt first appeared.
- [ ] Confirm tools that don't go through a permission gate (e.g. `Read`) are unaffected — counter starts on first render as before.

### DIFF
--- a/scripts/probe-e2e-tool-journey-render.mjs
+++ b/scripts/probe-e2e-tool-journey-render.mjs
@@ -750,6 +750,109 @@ try {
       passCancellingState ? '' : 'no transient state — user cannot tell whether the click registered');
   }
 
+  // ── Journey 11: elapsed counter pauses while permission is pending (#311) ─
+  // Bug #248-1: the elapsed counter started ticking at REQUEST time (when the
+  // assistant emitted tool_use) rather than EXECUTION time (when permission
+  // resolved + the tool actually started running). Result: a user staring
+  // at the permission prompt for 90s would see a "still no result" stall
+  // banner fire even though nothing had run yet.
+  //
+  // Reverse-verify: revert ChatStream.tsx + renderBlock.tsx + ToolBlock.tsx
+  // to drop the `permissionPending` plumbing — this journey FAILs because
+  // the elapsed chip / stall escalation banner re-appear during the gate.
+  {
+    // Seed: an in-flight tool block followed by a waiting/permission block
+    // for the SAME toolName. Use the Date.now patching trick to fast-forward
+    // 95s so without the fix the escalation banner would render.
+    await win.evaluate(() => {
+      const realNow = Date.now.bind(Date);
+      window.__realDateNow = realNow;
+      window.__nowOffset = 0;
+      Date.now = () => realNow() + window.__nowOffset;
+    });
+    await seed([
+      { kind: 'user', id: 'u-pp', text: 'run a command' },
+      { kind: 'tool', id: 't-pp', toolUseId: 'tu-pp-1', name: 'Bash',
+        brief: 'rm -rf node_modules', expanded: false /* in-flight: no result */ },
+      { kind: 'waiting', id: 'wait-pp', intent: 'permission',
+        requestId: 'req-pp', toolName: 'Bash',
+        prompt: 'Bash: rm -rf node_modules',
+        toolInput: { command: 'rm -rf node_modules' } },
+    ]);
+    // Fast-forward 95s of wall-clock and force a render tick.
+    await win.evaluate(() => {
+      window.__nowOffset = 95_000;
+      window.__ccsmStore.getState().appendBlocks('s-tool', [
+        { kind: 'assistant', id: 'a-noop-pp', text: '' },
+      ]);
+    });
+    await win.waitForTimeout(400);
+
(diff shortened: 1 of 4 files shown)


Jiahui-Gu (commented):
[lightweight-review] APPROVE (self-PR; comment-only review)

## Layer 1 — direction
- Bug is real (dogfood Bug #1, #248). `startedAtRef` was captured at first render where `result===undefined`, which is the moment `tool_use` lands — well before `canUseTool` resolves. Counter ticks + 90s escalation banner fire on a tool that has not actually run.
- "Defer timer start until gate clears" is the correct fix, not "suppress banner only". A 90s elapsed chip on a non-running tool is itself the misleading bit; suppressing only the banner leaves the lie on screen.
- Prop-threading `permissionPendingToolIds` through `ChatStream → renderBlock → ToolBlock` is consistent with the existing `now` / `permissionAutoFocus` opts in `renderBlock.tsx`. Store-coupling `ToolBlock` to derive it internally would be more decoupled but breaks the file's current dumb-presentational shape and would force a store import. Prop-thread is the right call.

## Layer 2 — implementation
- ChatStream computation (lines 192-223): single forward scan, marks an in-flight tool block as pending iff a LATER `waiting/permission` block has matching `toolName`. Code comment correctly flags the disambiguation limitation: claude.exe serializes permission prompts and the waiting block carries no `toolUseId`, so name-only match is sufficient in practice. Parallel-same-name worst case = both in-flight tools both treated as pending until the gate clears — conservative (no false ticking) and self-corrects.
- ToolBlock (lines 91-106): `initialStartedAt = hasResult || permissionPending ? null : Date.now()` plus the post-mount guard `if (!hasResult && !permissionPending && startedAtRef.current === null)` correctly defers capture across the pending → cleared transition. `running = !hasResult && !isError && !permissionPending` flows into chip / stall / escalation / cancel suppression via existing `running` checks. Counter resumes from 0 at gate-clear because `startedAtRef` is first written on the post-clear render.
- Probe J11a asserts no `tool-elapsed` / `tool-stalled` / `tool-stall-escalated` / `tool-stall-cancel` after 95s wall-clock advance with sibling waiting block. J11b removes the waiting block, asserts elapsed `<1.5s` immediately, then `~2s` after a 2s tick, no escalation. Tight, focused, restores `Date.now` at end.
- Diff scope clean: only the 4 declared files. `permissionPending` defaults to `false` → backward-compatible with existing call sites and the 11/11 ToolBlock UX tests.

Approving. Worker self-reported typecheck clean + 11/11 existing + reverse-verify 13/17 → 15/17 with the two pre-existing FAILs (J5b, J9) unchanged on baseline.

PR #265, opened 2026-04-25
fix(diff): empty-text Enter in inline composer = Cancel (#340)

Jiahui-Gu:
Empty Enter on the inline diff-comment composer now closes it (matches Esc), instead of being a silent no-op that left a blank composer hanging.

## Test plan
- [x] `npm test -- diff-view-line-comments` (15/15 pass; 2 new cases for empty-Enter cancel + non-empty-Enter save)
- [x] `npm run typecheck` (no new errors; 2 pre-existing on `working` unchanged)

### DIFF
--- a/src/components/chat/DiffView.tsx
+++ b/src/components/chat/DiffView.tsx
@@ -84,7 +84,11 @@ function InlineCommentComposer({
         onKeyDown={(e) => {
           if (e.key === 'Enter' && !e.shiftKey) {
             e.preventDefault();
+            // Empty Enter is treated as cancel (#340) — pressing Enter on a
+            // blank composer was creating empty comments, which is junk.
+            // Esc and empty-Enter now share the same close-without-save path.
             if (trimmed) onSave(trimmed);
+            else onCancel();
           } else if (e.key === 'Escape') {
             e.preventDefault();
             onCancel();
--- a/src/components/chat/DiffView.tsx
+++ b/src/components/chat/DiffView.tsx
@@ -84,9 +84,9 @@ function InlineCommentComposer({
         onKeyDown={(e) => {
           if (e.key === 'Enter' && !e.shiftKey) {
             e.preventDefault();
-            // Empty Enter is treated as cancel (#340) — pressing Enter on a
-            // blank composer was creating empty comments, which is junk.
-            // Esc and empty-Enter now share the same close-without-save path.
+            // Empty Enter dismisses the composer (#340) — previously it was a
+            // silent no-op that left a blank composer hanging open. Aligns
+            // with Esc so blank Enter no longer feels broken.
             if (trimmed) onSave(trimmed);
             else onCancel();
           } else if (e.key === 'Escape') {
--- a/tests/diff-view-line-comments.test.tsx
+++ b/tests/diff-view-line-comments.test.tsx
@@ -224,4 +224,36 @@ describe('<DiffView /> per-line comment affordance', () => {
     // is "no comment was persisted", which is observable on the store.
     expect(useStore.getState().pendingDiffComments[SID]).toBeUndefined();
   });
+
+  it('Enter on an empty composer cancels (no junk comment) and closes it (#340)', () => {
+    render(<DiffView diff={spec('/a/e.ts', [], ['LINE'])} />);
+    fireEvent.click(addCommentButtonForLine(0));
+    const composer = document.querySelector('[data-diff-comment-composer]') as HTMLElement;
+    expect(composer).not.toBeNull();
+    const ta = within(composer).getByPlaceholderText(/add a comment/i) as HTMLTextAreaElement;
+    // Whitespace-only counts as empty.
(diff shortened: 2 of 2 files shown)


Jiahui-Gu (commented):
## Independent review — CHANGES REQUESTED

### Layer 1 (direction): APPROVE
Empty-Enter = Cancel is the right UX. Aligns with Esc, fix is minimal (4 lines), no scope creep. Edit-mode is fine because the dedicated trash button (`data-diff-comment-delete`) remains the explicit-removal path.

### Layer 2 (implementation)

**Production fix: correct.** `!e.shiftKey` preserves Shift+Enter. Esc untouched. Trim semantics match the Save button's `disabled={!trimmed}`. Typecheck is actually clean (the 2 known pre-existing errors mentioned in the review brief appear already resolved on `working`).

**Reverse-verify FAILED to fail.** I reverted `src/components/chat/DiffView.tsx` to `origin/working` and re-ran `npx vitest run tests/diff-view-line-comments.test.tsx` — **all 15 tests still pass without the production fix**.

Reason: the original code already had `if (trimmed) onSave(trimmed)`, so empty Enter was already a silent no-op (`pendingDiffComments[SID]` was already `undefined`). The new empty-Enter test only asserts the store is empty — which was always true. The actual behavior change (composer **closes** on empty Enter, instead of staying open) is not exercised. The test's own comment admits this: "Mirror the Esc test's contract: no persisted state" — but the Esc-vs-empty-Enter difference is exactly the closing behavior, which the store can't observe.

**Code-comment accuracy.** The inline comment in `DiffView.tsx` says "pressing Enter on a blank composer was creating empty comments, which is junk" — that's not what the bug was. The PR description is accurate ("silent no-op that left a blank composer hanging"). The code comment should match.

### Requested changes

1. **Make the empty-Enter test actually fail without the fix.** Try asserting composer dismissal directly:
   ```ts
   fireEvent.keyDown(ta, { key: 'Enter' });
   expect(document.querySelector('[data-diff-comment-composer]')).toBeNull();
   ```
   If framer-motion exit makes that flaky under jsdom, fall back to opening a second composer on a different line and asserting only one exists, or assert `composerLine` state. Per `feedback_e2e_discipline`: reviewer reverse-verifies (stash → fail). This test fails that bar.
2. **Fix the inline code comment** — the bug is "leaves composer hanging open", not "creates empty comments".

The non-empty-Enter regression-guard test is great; keep it.

### Recommendation: CHANGES REQUESTED
Direction is right and production code is good — but the test does not protect the change it claims to protect.

Jiahui-Gu:
Both items addressed in ab389a3:

**Item 1 — empty-Enter test now load-bearing.** Added `await waitFor(() => expect(document.querySelector('[data-diff-comment-composer]')).toBeNull())` after the empty Enter keydown. Reverse-verified by reverting the production change in DiffView.tsx and re-running: the new test fails (waitFor times out — composer wrapper stays mounted because empty Enter is a no-op without the fix); 14 other tests still pass. Restored fix; 15/15 pass.

**Item 2 — code comment reworded.** Now: "Empty Enter dismisses the composer (#340) — previously it was a silent no-op that left a blank composer hanging open. Aligns with Esc so blank Enter no longer feels broken." Accurately describes the real bug.

Jiahui-Gu (commented):
**Verdict: APPROVE** (self-approve blocked by GitHub; recording as comment.)

Both pushback items addressed; reverse-verification done locally.

**Item 1 — strengthened test (load-bearing assertion).**
- Added `await waitFor(() => expect(document.querySelector('[data-diff-comment-composer]')).toBeNull())` plus `async` on the test fn.
- Reverse-verified by editing `DiffView.tsx` to remove the `else onCancel()` branch and rerunning `npm test -- diff-view-line-comments`: the empty-Enter test FAILS with `expected node to be null` (composer wrapper still in DOM, full element snapshot in the failure output). Restored — all 15 tests pass.
- Selector `[data-diff-comment-composer]` confirmed against the actual wrapper at `DiffView.tsx:74`.
- `waitFor` + AnimatePresence-tick rationale in the new comment is correct.

**Item 2 — rewritten code comment.**
- New wording accurately names the bug (silent no-op leaving a blank composer hanging open) and the fix intent (align with Esc). Reads cleanly.

**Hygiene.**
- `npm run typecheck` clean.
- Branch was based on `d142af5` (pre-#338). Rebased locally onto current `origin/working` (which now has #338/#341/#322): clean, no conflicts, 15/15 tests still green. Worker should `git pull --rebase origin working && git push --force-with-lease` before merge.

Diff stat vs working: `DiffView.tsx +4 / store.ts -32 / diff-view-line-comments.test.tsx +35 -3 / store-pending-diff-comments.test.ts -57`. The deletes are #338's de-dup landing; net change is +4 in DiffView plus the strengthened test. Looks correct.

Jiahui-Gu:
post-rebase: clean, tests green, ready to merge

PR #266, opened 2026-04-25
fix(store): dedupe per-line diff comments at store layer (#338)

Jiahui-Gu:
## Summary
Layer 2 hardening for #303: `addDiffComment` now replaces any existing (sessionId, file, line) entry instead of stacking a duplicate. The original id is reused so DiffView's `data-diff-comment-id` DOM lookups stay stable across the replace.

## Test plan
- [x] `tests/store-pending-diff-comments.test.ts` — 4 new unit tests (replace, distinct lines, distinct files, trim)
- [x] `tests/diff-view-line-comments.test.tsx`, `tests/inputbar-diff-comments.test.tsx`, `tests/store.test.ts` still pass (103/103)

### DIFF
--- a/src/stores/store.ts
+++ b/src/stores/store.ts
@@ -317,6 +317,11 @@ export interface PendingDiffComment {
   line: number;
   text: string;
   createdAt: number;
+  // Bumped when an existing (file, line) comment is overwritten by a fresh
+  // addDiffComment call. Sort order still keys on createdAt so the prompt
+  // serialization stays deterministic across the replace; updatedAt is
+  // informational only.
+  updatedAt?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Independent review — LGTM (cannot self-approve via API; recording as comment).

## Layer 1 (direction)
- **Necessary?** Yes. UI gating prevents the duplicate-button case, but the store action is the public boundary — any non-button code path (programmatic recovery, future commands, IPC re-entry) would still stack. Defense-in-depth at the store boundary removes a class of bugs rather than guarding one entry point.
- **Approach?** Replace-on-conflict with stable id reuse is the minimal correct path. Re-keying the bucket by `${file}:${line}` would have broken `data-diff-comment-id` lookups in DiffView (correctly rejected upstream).
- **Scope creep?** `updatedAt?: number` is a small additional field — not strictly required, but optional, well-commented (notes serialization still keys on `createdAt` so prompt determinism is preserved), zero observable impact. Acceptable.

## Layer 2 (implementation)
- Diff is tight: 32 lines in store.ts, all inside the existing action. Id reuse confirmed (`return dup.id`). `createdAt` preserved via spread. `updatedAt` only set on replace path. Empty-text early-return preserved.
- Edge cases:
  - Identical replacement text — still bumps `updatedAt`, but no observable user-facing diff. Fine.
  - Windows path casing — `c.file === args.file` is case-sensitive string compare; consistent with the rest of the store (file paths stored verbatim from producer). Not for this PR to fix.
  - `find` is O(n) per call; bounded by expected scale.
- Test isolation uses `useStore.setState({ ...initial, pendingDiffComments: {} }, true)` — clean snapshot/restore pattern.

## Reverse-verify
Reverted `src/stores/store.ts` to `origin/working`; re-ran new test file:
- `replaces an existing (file, line) comment instead of stacking` — FAILED (got 2 entries instead of 1).
- `trims the replacement text` — FAILED (text stayed `'first'`).
- Distinct-line and distinct-file tests still passed (correct — those don't exercise dedupe).

Restored store.ts; full target suite green: **107/107** across `store-pending-diff-comments`, `store`, `inputbar-diff-comments`, `diff-view-line-comments`. `npm run typecheck` clean.

Tests genuinely exercise the fix. Recommend merge.

PR #267, opened 2026-04-25
chore(probe): add tool-block-root testid + tighten J5b locator (#341)

Jiahui-Gu:
## Summary
- Add `data-testid="tool-block-root"` to ToolBlock's outermost `<div>` so probes have a stable per-block handle (the same root covers error + healthy variants).
- Update the J5b case in `probe-e2e-tool-journey-render.mjs` to walk up via `closest('[data-testid="tool-block-root"]')` instead of the Tailwind class string `div.font-mono.text-chrome` introduced in #263.
- Decouples the probe from styling decisions; testid is the long-term contract.

## Why
PR #263 made J5b correct but fragile — `font-mono text-chrome` will change as ToolBlock styling evolves, silently breaking the probe again. A `data-testid` is the right long-term contract.

## Test plan
- [x] `node scripts/probe-e2e-tool-journey-render.mjs` — 17/17 pass (J5b uses the new testid)
- [x] `npm run typecheck` — only pre-existing baseline errors (SettingsDialog/Switch), unrelated
- [x] No ToolBlock unit tests exist; broader suite shows only pre-existing better-sqlite3 NODE_MODULE_VERSION failures unrelated to this change

Closes #341

### DIFF
--- a/scripts/probe-e2e-tool-journey-render.mjs
+++ b/scripts/probe-e2e-tool-journey-render.mjs
@@ -504,11 +504,15 @@ try {
       // share an ancestor (the stream root) whose textContent contains
       // BOTH tokens. That made `find('ok-auto')` resolve to whichever
       // button was iterated first (the error block), masking real bugs.
-      // The precise anchor is the ToolBlock root (`div.font-mono.text-chrome`)
-      // — each block's root contains exactly its own brief + result body.
+      // The precise anchor is the ToolBlock root carrying
+      // `data-testid="tool-block-root"` (#341) — each block's root contains
+      // exactly its own brief + result body. Previously we matched the
+      // Tailwind class string `div.font-mono.text-chrome`, but that coupled
+      // the probe to styling decisions that change frequently; the testid is
+      // the stable contract.
       function find(token) {
         for (const b of btns) {
-          const root = b.closest('div.font-mono.text-chrome');
+          const root = b.closest('[data-testid="tool-block-root"]');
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Independent review — **APPROVE** (posted as comment because GitHub blocks self-approve; manager may merge).

**Layer 1 (necessity / direction): PASS**
Tailwind classes change as design iterates; coupling the probe to `div.font-mono.text-chrome` is a known fragility. A `data-testid` is the right stable contract between production and the probe. Scope is minimal (1-line addition + selector swap + comment update). No scope creep.

**Layer 2 (implementation): PASS**
- Single-root claim verified. `ToolBlock.tsx` returns one outer `<div>` (line 178) for both error and healthy variants — `isError` only mutates the `className` suffix and adds inner decorations; the wrapper is shared. Testid lives on this single root, so both variants are addressable.
- Selector strings match exactly: production uses `data-testid="tool-block-root"`, probe uses `[data-testid="tool-block-root"]`.
- `npm run typecheck`: clean (0 errors — better than the worker's "2 known pre-existing" note).
- Probe: **17/17 PASS** locally with sanitized HOME.
- **Reverse-verify**: removed the testid line from production, rebuilt, re-ran probe → J5b fails (16/17). Restored. Confirms the assertion is genuinely anchored to this contract.

**Findings**
- None blocking. Comment update in the probe is informative and explains the WHY.

**Recommendation:** APPROVE.

PR #268, opened 2026-04-25
feat(input): continue-after-interrupt affordance (#322)

Jiahui-Gu:
## Summary
- After Stop, show a one-line ghost hint above the composer (`Press Enter to continue, or type a new message` / `按 Enter 继续，或输入新消息`).
- Empty composer + Enter while hint visible sends the literal `continue` through the normal send path.
- Hint dismisses on first keystroke (latched, doesn't re-show on backspace-to-empty), on send, or when a new turn begins.
- Store gets a per-session `lastTurnEnd: 'ok' | 'interrupted'` slice. Written by `markInterrupted`, cleared by `setRunning(true)` and the explicit `clearLastTurnEnd` action used in the send path. Not persisted (a stale post-reload hint would feel weird).

## Files
- `src/stores/store.ts` — new `lastTurnEnd` state, `clearLastTurnEnd` action; `markInterrupted` and `setRunning` updated.
- `src/components/InputBar.tsx` — hint row + Enter handler + typed-since-interrupt latch.
- `src/i18n/locales/{en,zh}.ts` — `chat.continueAfterInterruptHint` appended at end of namespace with anchor comment.
- `tests/inputbar-continue-after-interrupt.test.tsx` — 6 cases covering visibility gates, dismiss behavior, `continue` payload, and re-arming.

## Test plan
- [x] `npx vitest run tests/inputbar-continue-after-interrupt.test.tsx` — 6/6 pass.
- [x] `npx vitest run tests/inputbar.test.tsx tests/inputbar-diff-comments.test.tsx` — 12/12 pass (no regressions).
- [x] `npm run typecheck` — no new errors (2 pre-existing errors in SettingsDialog and Switch unchanged).
- [ ] Manual: start a turn, press Stop, observe hint; Enter sends `continue`; typing a char hides the hint.

### DIFF
--- a/src/components/InputBar.tsx
+++ b/src/components/InputBar.tsx
@@ -138,7 +138,7 @@ export function InputBar({ sessionId }: { sessionId: string }) {
   // Perf: subscribe to all reactive per-session signals via useShallow so this
   // component only re-renders when one of these specific values changes
   // (instead of once per any store mutation, like an appendBlocks chunk).
-  const { session, started, running, queueLength, hasMessages, hasPendingWaiting, permission, focusInputNonce, pendingDiffCommentsCount } = useStore(
+  const { session, started, running, queueLength, hasMessages, hasPendingWaiting, permission, focusInputNonce, pendingDiffCommentsCount, lastTurnEnd } = useStore(
     useShallow((s) => ({
       session: s.sessions.find((x) => x.id === sessionId),
       started: !!s.startedSessions[sessionId],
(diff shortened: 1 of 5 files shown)


Jiahui-Gu (commented):
**APPROVE** (posted as comment — author cannot self-approve via gh) — clean implementation of the continue-after-interrupt affordance, fully aligned with spec.

**Layer 1 (direction):**
- Affordance choice (empty-Enter overload + ghost hint) is correct. Worker's pushback on alternatives (button competing with Send/Stop, separate keystroke, history-derived continue) is sound — agree, no churn.
- i18n copy quality is good: EN `Press Enter to continue, or type a new message` is action-first and disambiguates both branches; ZH `按 Enter 继续，或输入新消息` mirrors the structure naturally. Both sentence-case, no SCREAMING.
- Visual treatment (`text-meta text-fg-tertiary` + `mb-1 px-1` + `select-none`) is appropriately subtle and sits cleanly above composer chrome.
- Backspace-to-empty leaving the hint dismissed (sticky-once-typed) is the right call — re-popping the hint after a deletion would feel jumpy.

**Layer 2 (implementation):**
- Store: `lastTurnEnd` slice is well-isolated. `markInterrupted` and `setRunning` now build `Partial<State>` patches with reference-equality short-circuits — no spurious re-renders, no collision with the recently-merged `pendingDiffComments` slice (#303).
- `clearLastTurnEnd` is called explicitly in `send()` BEFORE the queued-while-running branch, which correctly covers the edge case where `setRunning(true)` wouldn't fire to wipe the hint. Avoids one-frame flash.
- `lastSeenInterruptRef` keyed by `${sessionId}:${lastTurnEnd}` re-arms the typed-since-interrupt latch across both session switches and consecutive interrupts on the same session.
- `send('continue')` reuses the full send pipeline (slash detection, drafts, IPC). No special-case bypass.
- i18n: `// task322:` anchor present in both `en.ts` + `zh.ts`, appended at end of `chat` namespace — exactly per spec.

**Verification:**
- `npx vitest run inputbar` — 18/18 pass across all 3 inputbar test files (no sibling regression).
- **Reverse-verify**: removed the `if (showContinueHint) { void send('continue'); return; }` block in `onKeyDown`; the `sends the literal continue` test FAILED as expected (`expected spy called 1 times, got 0`). Restored — full suite green again.

Nothing to fix. Ready to merge.

PR #269, opened 2026-04-25
fix(store): escape XML metacharacters in diff-feedback body (#339)

Jiahui-Gu:
## Summary
Escape `&`, `<`, `>` in the user-typed body of `<diff-feedback>` envelopes emitted by `serializeDiffCommentsForPrompt` so user text containing tag-like content cannot break out of the envelope or confuse the agent's tag parser. `&` is replaced first to avoid double-escaping.

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run tests/diff-view-line-comments.test.tsx` — 16/16 pass
- [x] New cases: `<` → `&lt;`, `&` first-pass ordering (`a & <foo>` → `a &amp; &lt;foo&gt;`, no `&amp;lt;`), literal `</diff-feedback>` in body is escaped so only one real closing tag remains

### DIFF
--- a/src/stores/store.ts
+++ b/src/stores/store.ts
@@ -349,10 +349,18 @@ export function serializeDiffCommentsForPrompt(
   return list
     .map((c) => {
       // Escape the attribute value so a path with `"` can't break out of
-      // the file= attribute. We don't need a full XML escape on the body —
-      // the agent receives it as plain text inside the structured tag.
+      // the file= attribute.
       const file = c.file.replace(/"/g, '&quot;');
-      return `<diff-feedback file="${file}" line="${c.line}">${c.text}</diff-feedback>`;
+      // Escape XML metacharacters in the body so user-typed text containing
+      // `<`, `&`, or even a literal `</diff-feedback>` can't break out of
+      // the envelope or confuse the agent's tag parser. Order matters:
+      // `&` must be replaced first, otherwise we'd double-escape the `&`
+      // introduced by the subsequent `<` → `&lt;` substitution.
+      const text = c.text
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;');
+      return `<diff-feedback file="${file}" line="${c.line}">${text}</diff-feedback>`;
     })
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## Independent Review — APPROVE

Necessary. The XML envelope `<diff-feedback file="..." line="...">BODY</diff-feedback>` is sent verbatim to the agent; without escaping, a user comment containing `</diff-feedback>` (or even just `<foo>`) would either break out of the envelope or confuse a tag parser downstream. Sibling `&quot;` escape on the file attribute already exists, so this is the matching defensive pair. Including `>` alongside `<`/`&` is borderline but cheap and matches the canonical XML escape set — accepted.

### Layer 2 (implementation)
- Order is correct in `src/stores/store.ts:359-362`: `&` → `&amp;` first, then `<` → `&lt;`, then `>` → `&gt;`. No double-escape window.
- Comment on lines 354-358 explicitly calls out the ordering invariant — future maintainers won't accidentally reorder.
- Three new tests in `tests/diff-view-line-comments.test.tsx:148-174` cover: basic `<`, ordering invariant (asserts NOT `&amp;lt;`), and the literal `</diff-feedback>` injection case (asserts exactly one real closing tag in output).
- PR diff is tightly scoped: +39/-3, only `store.ts` + the test file. No collateral.

### Reverse-verify
Manually reverted the body-escape (kept tests), re-ran `npx vitest run tests/diff-view-line-comments.test.tsx`:
- 3 failed / 13 passed — exactly the 3 new tests, with informative diffs (e.g. `received "a & <foo>"` vs expected `"a &amp; &lt;foo&gt;"`).
Restored, re-ran: 16/16 pass. Tests genuinely gate the fix.

### Gates
- `npm run typecheck` clean.
- `npx vitest run tests/diff-view-line-comments.test.tsx` 16/16 pass.
- `gh pr view 269` reports `MERGEABLE` against `working` (the working branch's task322 revert touched different regions of `store.ts`, no conflict).

### Verdict
**APPROVE.** Recommending merge. (Note: gh blocks self-approve so this is filed as a comment with explicit APPROVE language.)

PR #270, opened 2026-04-25
feat(input): cwd chip discoverability + per-group default (#328)

Jiahui-Gu:
T4 follow-ups from the cwd brainstorm (#326). Closes #328.

## Summary
- First-hover accent-ring pulse on the cwd chip (sticky once-per-mount) so brand-new users notice it's interactive.
- `(none)` placeholder when the active session has no cwd; popover does NOT auto-open per T4.b.
- Per-group cwd default: new session in a group prefills with the most-recent cwd from that group; falls through to `recentProjects` then `''` when empty.
- New i18n key `chat.cwdChipNoneLabel` (en + zh).

## Test plan
- [x] `tests/store.test.ts` — 5 new cases covering per-group default + cross-group isolation + recentProjects fallback + skip-empty-cwd + explicit-cwd wins
- [x] `tests/cwd-popover.test.tsx` — 3 new cases: `(none)` label render + no-auto-open + first-hover pulse retire+sticky
- [x] `tests/i18n-key-parity.test.ts` passes
- [ ] Manual: hover the chip on a fresh session, confirm faint pulse, doesn't replay on re-hover, retires on popover open
- [ ] Manual: pick a cwd in group A, create new session in same group, confirm chip prefills
- [ ] Manual: empty group + empty `recentProjects` → chip reads `(none)`, popover stays closed on hover, opens on click

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -209,8 +209,10 @@ export default function App() {
   const [shortcutsOpen, setShortcutsOpen] = React.useState(false);
 
   // New sessions are created in-place — no modal. The store seeds `cwd`
-  // from `recentProjects[0]?.path ?? '~'`; users repick later via the
-  // StatusBar cwd chip in chat. See createSession() in stores/store.ts.
+  // from the per-group last-used cwd, falling through to
+  // `recentProjects[0]?.path ?? historyRecentCwds[0] ?? ''` (empty → chip
+  // renders the `(none)` placeholder). Users repick via the StatusBar cwd
+  // chip. See createSession() in stores/store.ts.
   const newSession = React.useCallback(() => {
     createSession(null);
   }, [createSession]);
--- a/src/components/CwdPopover.tsx
+++ b/src/components/CwdPopover.tsx
@@ -92,6 +92,30 @@ export function CwdPopover({ cwd, cwdMissing, loadRecent, onPick, onBrowse }: Pr
   const popRef = useRef<HTMLDivElement | null>(null);
   const inputRef = useRef<HTMLInputElement | null>(null);
 
+  // task328: discoverability — pulse a faint accent ring on the FIRST hover
+  // after mount so brand-new users notice the chip is interactive. Sticky:
+  // once shown (or once the popover is opened by any path), we never replay
+  // for the lifetime of this mount. Keyed off mount, not session id, so
+  // switching between sessions in the same lifetime doesn't re-pulse.
+  const [hoverHintShown, setHoverHintShown] = useState(false);
+  const [hoverHintActive, setHoverHintActive] = useState(false);
+  const triggerOnHoverHint = useCallback(() => {
+    if (hoverHintShown) return;
+    setHoverHintShown(true);
+    setHoverHintActive(true);
+    // Pulse duration matches the menuIn ease (~600ms). After it expires we
+    // detach the class so subsequent renders are clean.
+    window.setTimeout(() => setHoverHintActive(false), 600);
+  }, [hoverHintShown]);
+  // Opening the popover (via click, keyboard, whatever) also retires the
+  // hint — the user has discovered the chip.
+  useEffect(() => {
+    if (open && !hoverHintShown) {
+      setHoverHintShown(true);
+      setHoverHintActive(false);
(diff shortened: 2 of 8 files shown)


Jiahui-Gu (commented):
## Independent reviewer report — PR #270

**Verdict: APPROVE** (gh blocks self-approve, so posting as a comment with explicit approve language; please treat as a +1 from the reviewer.)

### Layer 1 — direction
- **Spec alignment.** All five spec items implemented: first-hover sticky pulse on `[data-cwd-chip]`, `focus-ring` kit token applied, label updates immediately via prop-derived `triggerLabel`, per-group cwd default scans `sessions` newest-first matching `groupId` and non-empty cwd, `(none)` placeholder with italic muted styling and no auto-open. Worker added `chat.cwdChipNoneLabel` to en + zh with `// task328:` anchor in both locales at end of the `chat` namespace.
- **Composer-top vs StatusBar.** Verified at `src/App.tsx:417-437`: `<StatusBar>` is rendered between `<ChatStream/>` and `<InputBar>`, i.e. directly above the composer. Functionally equivalent to brainstorm's "Option A composer-top" — accept as documented deviation.
- **First-hover affordance vs one-time tooltip.** Worker rejected the tooltip alternative on noise grounds. I concur: a sticky-once 600ms accent ring is lower-noise, doesn't compete with the existing `title` tooltip, and uses an existing token (`--color-focus-ring`) so it reads as kit-native rather than bespoke onboarding chrome. Right call.
- **Per-group default edge case.** `Array.find()` walks newest-first (createSession prepends) and skips empty cwds — newest-with-cwd wins, oldest-only-with-cwd is correctly returned. Reverse-verified by the `skips sessions with empty cwd when scanning the group` test.

### Layer 2 — implementation
- **`'~'` → `''` change risk: SAFE.** Grepped `cwd === '~'`, `defaultCwd === '~'`, and bare `'~'` across `src/`. No behavioral consumer treats `'~'` as a sentinel — only mock data (`src/mock/data.ts`) uses it as a literal display path, which is unrelated. Three stale doc comments still reference the old `'~'` fallback (`src/App.tsx:212`, `src/components/Sidebar.tsx:506`, `src/stores/store.ts:2311`) — non-blocking nit; suggest a follow-up to update those to `''`/`(none)` for accuracy.
- **i18n parity.** Both en and zh have `chat.cwdChipNoneLabel` at the end of the `chat` namespace with `// task328:` anchors. `tests/i18n-key-parity.test.ts` passes.
- **Reverse-verify per-group default.** Reverted the `groupRecentCwd ??` chain in `store.ts` to `recentProjects[0]?.path ?? historyRecentCwds[0] ?? ''`. Result: 3 of 5 new per-group tests fail (`prefills new session cwd…`, `per-group default loses to recentProjects only when group is empty`, `skips sessions with empty cwd when scanning the group`). Restored — all green.
- **Reverse-verify first-hover affordance.** Removed `data-hover-hint`, `onMouseEnter`, `onFocus` from the chip. Result: `flashes a first-hover accent ring on the chip after mount, then retires (task328)` fails as expected. Restored — all green.
- **Tests.** `npm test -- store cwd-popover i18n-key-parity`: 6 files, 118 tests, all green. New: 14 cwd-popover tests, 90 store tests (incl. 5 new per-group tests).
- **Typecheck.** `npm run typecheck` clean. No new errors.

### Nits (non-blocking)
1. Three stale `'~'` references in doc comments (`src/App.tsx:212`, `src/components/Sidebar.tsx:506`, `src/stores/store.ts:2311`) — should be updated to `''`/`(none)` in a follow-up to prevent future-archeologist confusion.
2. The hover-hint `setTimeout` runs on the window directly without a cleanup. If the component unmounts within the 600ms window, `setHoverHintActive(false)` will fire on an unmounted component — React 18 swallows this silently but a defensive `clearTimeout` on unmount would be tidier. Tiny.

### Recommendation
**APPROVE.** Ship it. Both new behaviors reverse-verified, all tests green, typecheck clean, no broken `'~'` consumers, i18n parity holds, kit tokens used, label binding reactive, no auto-open semantics preserved. Stale-comment cleanup can ride a future PR.

PR #278, opened 2026-04-25
feat: load session history from CLI JSONL, delete messages table (PR-H)

Jiahui-Gu:
## Summary

PR-C follow-up. Removes ccsm's redundant SQLite `messages` table; the CLI / Agent SDK already writes every frame to `~/.claude/projects/<key>/<sid>.jsonl` as it streams, so ccsm now reads from there. Closes Task #29.

## Layer 1 (necessity + direction)

1. **Zero-dependency on `messages` table?** Verified. Production callers of `loadMessages`/`saveMessages` are limited to:
   - `electron/db.ts` (the impl itself)
   - `electron/main.ts` (IPC handlers)
   - `electron/preload.ts` (bridge)
   - `src/global.d.ts` (types)
   - `src/stores/store.ts` (`loadMessages` action + 6 `saveMessages` call sites)
   - `src/agent/lifecycle.ts` (one `saveMessages` on `result`)
   All other matches in the repo are tests/probes/comments. The earlier "47 files" figure conflated string-grep matches (mostly `s.loadMessageErrors`, comments, mocks); the real production graph is small and removable.

2. **JSONL format stable / 1:1 with renderer's MessageBlock?** JSONL = SDK frames (`SDKMessage`-ish). The renderer already has `framesToBlocks` (used by the import flow) which projects raw frames → MessageBlock with the same reducer the live agent uses (`streamEventToTranslation`). We reuse it as-is — projection logic stays in exactly one place.

3. **Project key from cwd?** CLI uses a slug rule (`/`, `\`, `:` → `-`). Verified empirically against a real `~/.claude/projects/` listing on Windows + macOS. Implemented in `projectKeyFromCwd` (electron/jsonl-loader.ts) so the slug logic is one-place.

4. **JSONL absent (fresh session, no first frame yet)?** Returns `{ ok: false, error: 'not_found' }`; renderer treats it as `[]` (no error banner).

5. **Performance / cache?** No cache. Reads are streamed line-by-line via `readline` (no full-file load); typical session is a few hundred frames. The 50k-frame cap mirrors the import path. If profiling later shows session-switch latency, an LRU on the main side is a small additive change — not worth pre-engineering per the design-density rule.

## Changes

- `electron/jsonl-loader.ts` (new) — streaming loader + `projectKeyFromCwd` + path-traversal hardening, mirrors `import-history.ts` structure.
- `electron/main.ts` — registers `agent:load-history`; removes `db:loadMessages` / `db:saveMessages` handlers and their imports.
- `electron/preload.ts` / `src/global.d.ts` — replaces `loadMessages`/`saveMessages` bridge methods with `loadHistory(cwd, sessionId)`.
- `electron/db.ts` — drops `messages` table from schema, drops legacy table at boot via `DROP TABLE IF EXISTS messages` (per "no migration" decision), removes `loadMessages` / `saveMessages` exports + their prepared-statement slots. `app_state` / `claudeBinPath` / integrity check / schema versioning unchanged.
- `src/stores/store.ts` — `loadMessages` action now calls `loadHistory(cwd, sid)`, derives sid from `session.resumeSessionId || session.id` (PR-D unify-friendly), projects frames via `framesToBlocks`. All 6 `saveMessages` call sites removed; deleteSession / deleteGroup / resetSessionContext / restoreSession / restoreGroup intentionally do NOT touch the JSONL on disk (it's the user's CLI-side history; ccsm shouldn't mutate it).
- `src/agent/lifecycle.ts` — `saveMessages` on `result` removed.
- `electron/__tests__/jsonl-loader.test.ts` (new) — happy path, empty file, not_found, malformed-line skipping, path traversal, invalid args. 8 tests, all green.
- `electron/__tests__/db.test.ts` / `db-hardening.test.ts` — messages-table cases removed; remaining cases (claudeBinPath, integrity check, schema version, prepared-statement cache) keep coverage.
- `tests/store.test.ts` — `loadMessages` autoload + merge + dedupe tests rewritten to mock `loadHistory` and exercise the framesToBlocks projection. Two new tests cover `not_found` (no banner) and `read_error` (inline banner).
- `tests/persisted-keys-source-of-truth.test.ts` / `tests/inputbar-continue-after-interrupt.test.tsx` — stale `loadMessages`/`saveMessages` mocks replaced with `loadHistory`.

## Test results

- `npm run typecheck`: pass.
- `npm test`: 735 pass / 6 fail. The 6 failures are all `db.test.ts` + `db-hardening.test.ts` cases hitting the pre-existing better-sqlite3 native-binding rebuild issue in this dev env (Python toolchain not available — same root cause as the 12 pre-existing failures called out in the task spec). New `jsonl-loader.test.ts` 8/8 green; `store.test.ts` 92/92 green.

## Smoke (deferred to reviewer env)

This worktree's better-sqlite3 binding can't be rebuilt (no Python toolchain). Smoke checklist for reviewer:

- [ ] Create a session, send a message, restart ccsm, reopen the session — history loads from JSONL.
- [ ] Newly-created session before first message — empty pane, no error banner.
- [ ] Permission-denied JSONL (chmod 000) — inline error banner with `read_error` shown; Retry recovers after fixing perms.
- [ ] On boot with an existing pre-PR ccsm.db, the `messages` table is dropped (sqlite3 ccsm.db ".tables" should not list it).

### DIFF
--- a/electron/__tests__/db-hardening.test.ts
+++ b/electron/__tests__/db-hardening.test.ts
@@ -23,37 +23,6 @@ beforeEach(async () => {
   await freshDb();
 });
 
-describe('db hardening: corrupt JSON in messages', () => {
-  it('loadMessages skips a row whose content is not valid JSON', async () => {
-    const mod = await freshDb();
-    const { saveMessages, loadMessages, getDb } = mod;
-
-    saveMessages('s-1', [
-      { id: 'b1', kind: 'user', text: 'good before' },
-      { id: 'b2', kind: 'user', text: 'good after' }
-    ]);
-
-    // Inject corruption directly into the row that already exists. Bypasses
-    // the cached prepared statements so we know loadMessages re-reads the
-    // raw column on each call.
-    getDb().prepare('UPDATE messages SET content = ? WHERE id = ?').run('{not valid json', 'b1');
-
-    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
-    let rows: unknown[] = [];
-    expect(() => {
-      rows = loadMessages('s-1');
-    }).not.toThrow();
-    expect(rows).toHaveLength(1);
-    expect((rows[0] as { id: string }).id).toBe('b2');
-    expect(warn).toHaveBeenCalledWith(
-      '[db] corrupt message row sessionId=%s id=%s',
-      's-1',
-      'b1'
-    );
-    warn.mockRestore();
-  });
-});
-
 describe('db hardening: startup integrity check', () => {
   it('initDb backs up a corrupt file and opens a fresh empty database', async () => {
     const mod = await freshDb();
@@ -66,7 +35,7 @@ describe('db hardening: startup integrity check', () => {
     fs.writeFileSync(file, Buffer.from('not a sqlite database, just bytes'));
 
     const err = vi.spyOn(console, 'error').mockImplementation(() => {});
-    const { initDb, loadMessages, saveState, loadState } = mod;
+    const { initDb, saveState, loadState } = mod;
     expect(() => initDb()).not.toThrow();
 
     // Original file got renamed aside.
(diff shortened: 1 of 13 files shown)


Jiahui-Gu:
Rebased onto current `origin/working` (was based on `a14ed1e`, working has since merged PR-K #277 + PR-L #276). The slash-command "regression" + 90/74-line test deletions in the previous review were a stale-base diff artifact — PR-H's commit only touches the 13 JSONL files. Post-rebase `git diff origin/working...HEAD` confirms 13 files, 507/293 ins/del. New head: `ff892a8`. Tests: jsonl-loader 8/8, slash-commands-registry 28/28, commands-loader 18/18, typecheck pass.

PR #287, opened 2026-04-25
feat(composer): send/stop morph + @file mention picker

Jiahui-Gu:
## Summary

Closes the remaining scope of task #13 (parts ③ + ④):

- **Send/Stop morph button** — single `<Button>` whose icon, variant, aria-label, and onClick swap based on `running`. AnimatePresence cross-fades the inner icon at 230ms tween. Mirrors the upstream Anthropic Claude Code VS Code extension pattern (webview/index.js: `if (busy && !X) <StopIcon/> else <SendIcon/>`). `disabled` follows the new state — Stop is always clickable while running, Send respects `sendDisabled`.

- **@file mention picker** — typing `@` (at start of input or after whitespace) opens a listbox of workspace files. Selecting splices `@<path> ` into the textarea — claude.exe parses the `@`-token itself, no further transformation needed. Trigger pattern matches the upstream `(?:^|\s)@[^\s]*` regex; picker shares the global popover mutex with the slash picker (mutual exclusion). No keyboard shortcut to invoke (per task constraints) — only the `@` typing trigger.

- **`@symbol` intentionally NOT shipped** — Layer 1 upstream grep showed only a `mention-file` command registered; no symbol picker exists in the extension. Per task ("如果上游没做就只做 @file"), dropped that scope.

### IPC

New `files:list` handler in `electron/main.ts` walks the session cwd (skipping node_modules/.git/dist/build/.venv/target/.idea/.vscode/etc., capped at 5000 entries / depth 12). Returns POSIX-style relative paths so the literal we splice into the composer is portable across Win/macOS/Linux. fromMainFrame guard + isSafePath validation match every other privileged handler in the file.

## Test plan

- [x] vitest: `tests/mentions-registry.test.ts` — 12 cases covering `detectAtTrigger` (bare @, after whitespace, partial query, email-like negatives, multi-line), `filterMentionFiles` (empty / exact / prefix / fuzzy), `commitMention` (splice + caret).
- [x] vitest: `tests/inputbar-morph-mention.test.tsx` — 6 cases: morph button data attributes idle/running, click-Stop fires `agentInterrupt`, `@` opens picker, Esc dismisses without altering textarea, Enter commits `@<path> `.
- [x] No regression in existing InputBar / slash picker / i18n parity tests (64/64 pass).
- [x] Full vitest run: 794/797 pass; the 3 failures are the pre-existing `electron/__tests__/db-hardening.test.ts` better-sqlite3 native-build issue (unrelated to this PR).
- [x] `tsc --noEmit` clean for both renderer + electron tsconfigs.
- [x] eslint clean on changed files (only the 1 pre-existing `Electron is not defined` error in main.ts and 1 pre-existing `markStarted` unused-var warning remain — confirmed identical pre/post via stash).
- [ ] e2e probe `scripts/probe-e2e-composer-morph-mention.mjs` — file ships and is structurally complete, but execution requires the Electron native deps which weren't buildable in the worker sandbox. Reviewer can run locally.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1091,6 +1091,88 @@ app.whenReady().then(() => {
     return loadCommands({ cwd: cwd ?? null });
   });
 
+  // ─────────────────────── @mention file listing ──────────────────────────
+  //
+  // Powers the InputBar's @file picker. Walks the session cwd recursively,
+  // skipping the usual heavy directories (node_modules, .git, dist, build,
+  // .next, .venv, target). Returns POSIX-style relative paths so the mention
+  // literal we splice into the composer (`@src/foo.ts`) stays portable
+  // across platforms — claude.exe on Windows happily accepts forward
+  // slashes too.
+  //
+  // Caps:
+  //   - max 5000 files (after which we bail; the picker fuzzy-search is
+  //     plenty accurate at that size, and walking >5k entries on every focus
+  //     is wasteful)
+  //   - max 12 directory depth (defense against symlink loops)
+  //
+  // Per-call, not cached: cwd swaps + on-disk edits should reflect quickly,
+  // and the renderer only invokes this on focus / @ trigger.
+  const FILE_LIST_MAX = 5000;
+  const FILE_LIST_MAX_DEPTH = 12;
+  const FILE_LIST_SKIP_DIRS = new Set([
+    'node_modules',
+    '.git',
+    '.svn',
+    '.hg',
+    'dist',
+    'build',
+    'out',
+    '.next',
+    '.nuxt',
+    '.turbo',
+    '.cache',
+    '.venv',
+    'venv',
+    '__pycache__',
+    'target',
+    '.idea',
+    '.vscode',
+  ]);
+
+  ipcMain.handle('files:list', async (e, cwd: string | null | undefined) => {
+    if (!fromMainFrame(e)) return [];
+    if (cwd == null || !isSafePath(cwd)) return [];
+    let rootStat: fs.Stats;
+    try {
+      rootStat = await fs.promises.stat(cwd);
+    } catch {
+      return [];
+    }
+    if (!rootStat.isDirectory()) return [];
+
+    const out: { path: string; name: string }[] = [];
+    async function walk(dir: string, rel: string, depth: number): Promise<void> {
+      if (out.length >= FILE_LIST_MAX) return;
+      if (depth > FILE_LIST_MAX_DEPTH) return;
+      let entries: fs.Dirent[];
+      try {
+        entries = await fs.promises.readdir(dir, { withFileTypes: true });
+      } catch {
+        return;
+      }
+      for (const ent of entries) {
+        if (out.length >= FILE_LIST_MAX) return;
+        // Skip hidden + heavy build dirs to keep the picker snappy.
+        if (ent.name.startsWith('.') && ent.name !== '.env' && ent.name !== '.env.local') {
+          if (ent.isDirectory()) continue;
+          // hidden file: skip too — usually noise (.DS_Store, .gitignore)
+          continue;
+        }
+        if (ent.isDirectory() && FILE_LIST_SKIP_DIRS.has(ent.name)) continue;
+        const childAbs = path.join(dir, ent.name);
+        const childRel = rel ? `${rel}/${ent.name}` : ent.name;
+        if (ent.isDirectory()) {
+          await walk(childAbs, childRel, depth + 1);
+        } else if (ent.isFile()) {
+          out.push({ path: childRel, name: ent.name });
+        }
+      }
+    }
+    await walk(cwd, '', 0);
+    return out;
+  });
+
   ipcMain.handle('shell:openExternal', async (e, url: string) => {
     if (!fromMainFrame(e)) return false;
     // Only http(s). Everything else is a potential shell hijack.
--- a/electron/preload.ts
+++ b/electron/preload.ts
@@ -260,6 +260,18 @@ const api = {
       ipcRenderer.invoke('commands:list', cwd),
   },
 
+  files: {
+    /**
+     * Workspace files relative to the session cwd, used by the InputBar's
+     * @file mention picker. Returns POSIX-style relative paths so the
+     * mention literal stays portable. Heavy directories (node_modules,
+     * .git, dist, build, .venv, target, etc.) and hidden entries are
+     * pruned in main. See main.ts for caps (5000 entries / depth 12).
+     */
+    list: (cwd: string | null | undefined): Promise<{ path: string; name: string }[]> =>
+      ipcRenderer.invoke('files:list', cwd),
+  },
+
   openExternal: (url: string): Promise<boolean> =>
     ipcRenderer.invoke('shell:openExternal', url),
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
         "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^6.9.1",
         "@testing-library/react": "^16.3.2",
+        "@testing-library/user-event": "^14.6.1",
         "@types/better-sqlite3": "^7.6.12",
         "@types/node": "^22.10.0",
         "@types/react": "^18.3.12",
@@ -5766,6 +5767,20 @@
         "@types/react-dom": {
           "optional": true
         }
+      }
+    },
+    "node_modules/@testing-library/user-event": {
+      "version": "14.6.1",
+      "resolved": "https://registry.npmjs.org/@testing-library/user-event/-/user-event-14.6.1.tgz",
+      "integrity": "sha512-vq7fv0rnt+QTXgPxr5Hjc210p6YKq2kmdziLgnsZGgLJ9e6VAShx1pACLuRjd/AS/sr7phAR58OIIpf0LlmQNw==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=12",
+        "npm": ">=6"
+      },
+      "peerDependencies": {
+        "@testing-library/dom": ">=7.21.4"
       }
     },
     "node_modules/@tootallnate/once": {
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.9.1",
     "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
     "@types/better-sqlite3": "^7.6.12",
     "@types/node": "^22.10.0",
     "@types/react": "^18.3.12",
--- a/scripts/probe-e2e-composer-morph-mention.mjs
+++ b/scripts/probe-e2e-composer-morph-mention.mjs
@@ -5,7 +5,7 @@
 //      `runningSessions[id]` flips true. Same DOM slot, swapped variant.
 //   2. Typing `@` in the textarea opens the mention picker (listbox role
 //      "File mentions"); Esc dismisses it without altering the textarea.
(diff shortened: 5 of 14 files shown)


Jiahui-Gu:
## Reviewer e2e gate: REQUEST CHANGES

Held-PR gate per `feedback_e2e_before_merge.md`. Ran on pool-2 (Win11, electron 33, native better-sqlite3 rebuilt).

### Probe FAILS (reproducible on a clean state)

```
$ rm -f ~/AppData/Roaming/ccsm/ccsm.db*
$ CCSM_PROD_BUNDLE=1 node scripts/probe-e2e-composer-morph-mention.mjs
[probe-e2e-composer-morph-mention] FAIL: expected '@src/components/InputBar.tsx '; got ""
```

Step-by-step instrumentation (real run, on a clean DB) shows:

```
AFTER @:        {"val":"@","sel":1,"listbox":true,"optCount":1}
AFTER Esc:      {"val":"@","sel":1,"listbox":false}
AFTER space:    {"val":"@ ","sel":2,"listbox":false}
AFTER Backspace:{"val":"@","sel":1,"listbox":true,"optCount":1,"focused":"TEXTAREA:true"}
AFTER Enter:    {"val":"","sel":0}
```

Listbox is visible with 1 option, focus is on the textarea, but **Enter clears the value instead of committing the highlighted mention** — Enter is falling through to the `send()` branch. Confirmed reproducible WITHOUT the Esc/reopen dance too:

```
AFTER @:    {"val":"@","lb":true,"opts":1}
AFTER Enter (no Esc dance, plain `@` then Enter): {"val":""}
```

So with a pristine state, just typing `@` + Enter still misses the picker-commit branch in `InputBar.tsx onKeyDown` (the `if (mentionOpen && filteredFiles.length > 0)` guard at `src/components/InputBar.tsx:783`). The Enter event fires the unconditional `send()` block at line 852.

### What this means

**Real product bug** — not probe-side. The bug is masked by the vitest test (`tests/inputbar-morph-mention.test.tsx:142`) because it uses `fireEvent.change({value:'@'})` followed by `fireEvent.keyDown(ta,{key:'Enter'})` synchronously inside one `act()`. In an actual Electron renderer, the keydown reads from a stale closure / never sees mentionOpen=true at handler time, despite React having committed the listbox.

Likely candidates the author should investigate:
1. The popover-mutex effects (`src/components/InputBar.tsx:325-333`) may be racing the open/close cycle and toggling `mentionDismissed` back to `true` between commit and Enter. Diag shows the listbox visible, so mentionOpen rendered true at least once — but the next render may flip it before the keydown handler captures.
2. `mentionFiles` may briefly be `[]` at keydown time even though the rendered listbox shows 1 option.

### Other findings

**Probe-side:**
- The probe at `scripts/probe-e2e-composer-morph-mention.mjs:31` sets only `NODE_ENV: 'development'` and omits `CCSM_PROD_BUNDLE: '1'`. Without it (and without webpack-dev-server running), the renderer never appears (`appWindow: no renderer window appeared in time`). Every other probe in `scripts/` sets `CCSM_PROD_BUNDLE: '1'` explicitly. Add it.
- The probe doesn't isolate state — it reads/writes the developer's real `~/AppData/Roaming/ccsm/ccsm.db`. After one run, the persisted draft for `s1` (`@src/components/InputBar.tsx `) leaks into the next run, producing `hello@@@@`-style ghost values. Either set `CCSM_USER_DATA_DIR` to a temp dir or call `clearDraft('s1')` via `__ccsmStore` before the test. Otherwise the probe is non-deterministic.

**Layer 1 / Layer 2 review (positive):**
- `electron/main.ts:1078-1119` `files:list` handler has `fromMainFrame` + `isSafePath` guards matching every other privileged handler. Symlink protection is implicit but real: `readdir({withFileTypes:true})` returns Dirents whose `isDirectory()` returns false for symlinks (only `isSymbolicLink()` is true), so the walker never recurses through them. Walks are bounded by `FILE_LIST_MAX=5000` and `FILE_LIST_MAX_DEPTH=12`. Returns POSIX-relative paths (`childRel = rel ? \`${rel}/${ent.name}\` : ent.name`) — portable. LGTM here.
- `src/mentions/registry.ts detectAtTrigger` correctly handles bare `@` at line start, after whitespace, after newline (multi-line), and rejects `email@domain` style. Verified by both the registry tests and the live diag (lb=true after `@`).
- Symbol picker correctly skipped — confirmed only `mention-file"` exists in the upstream extension bundle.
- i18n keys present in both `en.ts:407` and `zh.ts:379` (6 keys: pickerTitle / empty / noMatch / navigate / select / close — PR body says "8 keys" but only 6 are added; `mentions.activeAnnouncement` uses `defaultValue` fallback so it doesn't actually need a key. Minor doc nit only.)
- Morph button: `data-morph-state` + `data-variant` + `aria-label` swap are correctly tested and observed (`AFTER morph: stop/danger/Stop`).

### Vitest summary

`794/797 pass`, the 3 failures are the pre-existing `electron/__tests__/db-hardening.test.ts` better-sqlite3-native-rebuild issue (matches PR body claim).

### Verdict

**REQUEST CHANGES.** Blocker: Enter does not commit the highlighted mention in a real Electron renderer, even though the picker is visibly open with options. Please fix the product bug AND tighten the e2e probe (env + state isolation), then re-request review.

Reviewer ran the vitest morph+mention suite, full vitest, and the e2e probe twice on cleared state to confirm. Pool-2 left in place for follow-up worker.

Jiahui-Gu:
## Re: REQUEST CHANGES — Enter-after-Esc-rearm sends instead of commits

**Diagnosis: probe fixture bug, not a product bug.**

The keydown handler in `src/components/InputBar.tsx:783` was correct all along. The reviewer's probe stubbed `window.ccsm.files.list` via direct property assignment in the renderer, but `contextBridge.exposeInMainWorld` defines `window.ccsm` as a **non-writable, non-configurable** property, AND clones nested objects so neither `window.ccsm = ...`, `window.ccsm.files = ...`, nor `window.ccsm.files.list = ...` sticks. The assignment silently no-ops and the real bridge runs against the isolated empty userData cwd, returning **0 files**.

The picker still opens (because `mentionDismissed=false` and `atTrigger.active=true`) but renders the empty branch with zero `[role="option"]` rows. `mentionOpen && filteredFiles.length > 0` then correctly evaluates **false**, Enter falls through to `send()`, textarea clears. Looks identical to the symptom of a stale-closure bug, but isn't one.

Confirmed by instrumented run:

```
[DBG keydown] {key: Enter, mentionOpen: true, ff: 0, mf: 0, value: @, caret: 1, atActive: true}
```

`mf: 0` (mentionFiles state) — never populated because the IPC stub didn't take.
`{ok:true, count:0, descriptor:{writable:false, configurable:false}}` — confirmed contextBridge seal.

## Fix

`scripts/probe-e2e-composer-morph-mention.mjs` now stubs at the **main process** layer via `app.evaluate(({ipcMain}) => { ipcMain.removeHandler('files:list'); ipcMain.handle('files:list', ...) })`, with an upfront `stubCheck` that fails loud if the override didn't take. While here, addressed the other reviewer nits:

- `CCSM_PROD_BUNDLE: '1'` env (matches sibling probes; was `NODE_ENV: 'development'` which didn't load any window in this env).
- `isolatedUserData('ccsm-probe-composer-morph')` so persisted drafts from prior runs can't bleed in (the `hello@@@@` ghost-draft scenario).
- Added missing `mentions.activeAnnouncement` i18n key in `en.ts` + `zh.ts` (was relying on i18next `defaultValue` fallback). 7 keys total now (PR body originally said 8 — actual usage is 7, see `MentionPicker.tsx`).
- Added a `userEvent`-driven vitest case `tests/inputbar-morph-mention.test.tsx` that exercises the real keystroke timing (input → onChange → render → onKeyDown per microtask) for the Esc-then-edit-re-arm sequence. Defensive coverage so any future stale-closure regression in the keydown handler is caught at the unit level too.

## Reverse-verify

- Stash probe fix → probe FAILS (window doesn't appear because dev port not bound, then would fall to stub-no-op even if it did).
- Restore → probe PASSES.

## Probe output (passing)

```
[probe-e2e-composer-morph-mention] OK
  morph button: idle=send/primary -> running=stop/danger
  @mention picker: open / Esc-dismiss / Enter-commit all verified
```

## Vitest

```
✓ tests/inputbar-morph-mention.test.tsx (7 tests) 644ms
Test Files  1 passed (1)
     Tests  7 passed (7)
```

(7 tests = 6 existing + 1 new `Enter commits highlighted mention after Esc-then-edit re-arm (real keystroke timing)`.)

## Files touched

- `scripts/probe-e2e-composer-morph-mention.mjs` — ipcMain stub + isolatedUserData + CCSM_PROD_BUNDLE.
- `tests/inputbar-morph-mention.test.tsx` — new userEvent re-arm case.
- `src/i18n/locales/en.ts`, `src/i18n/locales/zh.ts` — `mentions.activeAnnouncement` key.
- `package.json`, `package-lock.json` — `@testing-library/user-event` devDep.

`src/components/InputBar.tsx`, `src/components/MentionPicker.tsx`, `electron/main.ts`, and `src/mentions/registry.ts` are unchanged.

Commit: d6433044392088a2f928248d919132dda0a583f7

Jiahui-Gu:
## Re-review: LGTM

Worker's diagnosis (probe bug, not product) verified end-to-end.

**Layer 1**
- contextBridge non-writability: confirmed via Electron's documented `exposeInMainWorld` security hardening (non-writable + non-configurable, nested objects deep-cloned). The probe's own `stubCheck` at lines 63-71 of `scripts/probe-e2e-composer-morph-mention.mjs` is the live smoking-gun — it would fail loud if the override didn't take.
- Product code unchanged in fix commit `d643304`: only touched `scripts/probe-e2e-composer-morph-mention.mjs`, `tests/inputbar-morph-mention.test.tsx`, `src/i18n/locales/{en,zh}.ts`, and `package*.json`. Zero changes to `InputBar.tsx`, `MentionPicker.tsx`, `electron/main.ts`, `electron/preload.ts`, `src/mentions/registry.ts`, or any other product file.
- Probe stub structure sound: `ipcMain.removeHandler('files:list')` runs before `ipcMain.handle('files:list', ...)`; returns 3 realistic paths; `stubCheck` validates with fail-loud on count mismatch.

**Layer 2 (reverse-verify both bite)**
- vitest reverse-verify: stashed the `mentionOpen && filteredFiles.length > 0` guard at `src/components/InputBar.tsx:783` (changed to `if (false && ...)`). Result: the new userEvent re-arm test at `tests/inputbar-morph-mention.test.tsx:198` FAILED (received `""` — Enter fell through to send and cleared textarea). Restored — passes again. Reverse-verify proves the new userEvent test isn't tautological.
- probe reverse-verify: same stash + rebuild. Probe FAILED with `expected '@src/components/InputBar.tsx '; got ""` — exact symptom signature from original reviewer's report. Restored — passes again.
- 3-round stability: 3/3 OK on probe.
- Full vitest: 796/799 passed. The 3 failures are pre-existing `electron/db.ts` better-sqlite3 db-hardening (not in this PR's scope) — exempted per re-review rubric. Target file `tests/inputbar-morph-mention.test.tsx` 7/7 passes.

**Sanity check on diagnosis (probe-only bug)**: confirmed internally consistent. The original vitest mocks `window.ccsm.files.list` via direct assignment in jsdom (which has no contextBridge — see `tests/inputbar-morph-mention.test.tsx:62`), so `filteredFiles` was correctly populated and the `mentionOpen && filteredFiles.length > 0` guard fired. In the probe, contextBridge-locked the bridge silently no-op'd the stub, the real handler ran against an empty userData cwd, returned 0 files, the guard correctly evaluated false, Enter fell through to send(). Symptom matched stale-closure but was fixture-only. Worker's story holds.

**Nits (non-blocking)**
- PR body claims 8 mentions i18n keys; actual is 7 (`pickerTitle`, `empty`, `noMatch`, `navigate`, `select`, `close`, `activeAnnouncement`). Update PR body if desired, no code change needed.

LGTM. Manager to merge.

PR #305, opened 2026-04-25
fix(auq): sticky widget renders above chips, no focus theft

Jiahui-Gu:
## Summary

Fixes AskUserQuestion sticky-widget regressions tracked as tasks #75 / #81 / #85. Three small UI changes (~5 LoC net):

- **`src/App.tsx`** — render `<QuestionStickyHost />` BEFORE `<StatusBar />` so the AUQ widget floats above the chip row instead of being hidden below it (covers #85, #81).
- **`src/components/QuestionStickyHost.tsx`** — pass `autoFocus={false}` to `QuestionBlock` and wrap it in a `<div data-question-sticky>` selector hook. Composer focus is sacred per ccsm design (`feedback_dispatch_avoid_collisions.md` lists InputBar as a hot file precisely because focus theft is a recurring complaint). User must Tab from the composer or click the AUQ widget explicitly to interact. **Intentionally divergent from upstream VS Code extension**, which auto-focuses on mount — we lock this behavior because the half-typed-draft case (block injected mid-typing) is the dominant ccsm use pattern. Covers #75.
- **`src/components/QuestionBlock.tsx`** — drop the redundant `data-question-sticky` attribute from the component root; the selector now lives on the QuestionStickyHost wrapper, keeping QuestionBlock reusable in non-sticky contexts. The `autoFocus` prop already existed and is already correctly respected by the focus `useEffect` (it short-circuits when `autoFocus === false`).

## Verification

- `npm run test`: **780/780 pass**
- `npm run build`: **pass** (only pre-existing bundle-size warnings)

## Probe deferral

Probe rewrite (J1 / J3 / J4 / J6) and the new positioning assertion (J7) are **deferred** — probe modifications are blocked until baseline e2e is green per current manager rule. Once that lifts, follow-up PR will:
- Update J1 to use `data-question-label` instead of the (non-existent) `value` attribute, then assert composer focus is preserved.
- Rewrite J3/J6 for the chip-tab UX (no default pre-select; click `[data-testid="question-tab-N"]` then click the option).
- Switch J4 wrap-around assertions to `data-question-label`.
- Add J7 boundingBox assertion that `[data-question-sticky]`'s bottom is above `[data-type-scale-role="status-bar"]`'s top, locking the "above chips" invariant against silent regressions.

## Test plan

- [x] vitest passes
- [x] webpack build passes
- [ ] Manual sanity (defer to reviewer): `npm start`, trigger an AskUserQuestion, observe widget rendering above the chip/status-bar row instead of below it; observe composer focus preserved when AUQ injected mid-typing

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -420,6 +420,7 @@ export default function App() {
                 <AgentDiagnosticBanner />
               </TopBannerStack>
               <ChatStream />
+              <QuestionStickyHost sessionId={active.id} />
               <StatusBar
                 cwd={active.cwd}
                 cwdMissing={active.cwdMissing}
@@ -440,7 +441,6 @@ export default function App() {
                 onChangeModel={(m) => setSessionModel(active.id, m)}
                 onChangePermission={setPermission}
               />
-              <QuestionStickyHost sessionId={active.id} />
               <InputBar sessionId={active.id} />
             </main>
           }
--- a/src/components/QuestionBlock.tsx
+++ b/src/components/QuestionBlock.tsx
@@ -246,7 +246,6 @@ export function QuestionBlock({ questions, onSubmit, onReject, autoFocus = true 
   return (
     <motion.div
       ref={rootRef}
-      data-question-sticky
       role="dialog"
       aria-label={t('questionBlock.title')}
       initial={{ opacity: 0, y: 8 }}
--- a/src/components/QuestionStickyHost.tsx
+++ b/src/components/QuestionStickyHost.tsx
@@ -40,71 +40,74 @@ export function QuestionStickyHost({ sessionId }: { sessionId: string }) {
   const block = pending as Extract<MessageBlock, { kind: 'question' }>;
 
   return (
-    <QuestionBlock
-      key={block.id}
-      questions={block.questions}
-      onSubmit={(answersByQuestion) => {
-        const api = window.ccsm;
-        if (!api) return;
-        // Two flows land here:
-        //  1. can_use_tool path (current claude.exe spawn): the question
-        //     block carries `requestId`. We deny the pending permission
-        //     promise on the main side so claude.exe stops blocking on
-        //     it, then send the answers as a fresh user turn so the
-        //     model receives them. This is intentionally lossy in the
-        //     CLI sense (we don't write back to the AskUserQuestion
-        //     tool_result), but it matches what the user actually wants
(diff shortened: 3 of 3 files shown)


Jiahui-Gu:
## Independent reviewer: LGTM

Both regressions are real and the fix is minimal & on-target:

1. **Sticky-below-chips**: `<QuestionStickyHost />` was rendered AFTER `<StatusBar />` inside the `flex flex-col` `<main>` (App.tsx:413). In normal flow, source order = visual order top-to-bottom. The reorder (now line 423, before StatusBar at 424) puts the question card visually above the chip row, matching upstream's "card above input area" placement. Confirmed neither `QuestionStickyHost` wrapper, `QuestionBlock` root (relative), nor `StatusBar` (no positioning) uses `position: fixed/absolute/sticky` or z-index hacks — pure DOM order determines stacking.

2. **Focus theft mid-typing**: `QuestionBlock`'s focus useEffect (lines 137-154) already short-circuits at line 138 when `autoFocus={false}`. The fix (`autoFocus={false}` in the host wrapper) cleanly disables auto-focus only for the sticky surface, leaving the in-stream usage of `QuestionBlock` (which gets the default `true`) untouched. Good separation.

3. **Drop `data-question-sticky` from `QuestionBlock` root**: correct cleanup — wrapper div now owns it, so the attr is unique and only present when used in sticky mode.

### Layer 2 — implementation
- **vitest**: 780/780 pass (after env-only better-sqlite3 ABI rebuild — unrelated to PR).
- **AUQ probes**:
  - `probe-e2e-askuserquestion-full.mjs`: all 6 journeys pass. J1 explicitly verifies "auto-focus did not steal textarea" — confirms the `autoFocus={false}` plumb works end-to-end.
  - `probe-e2e-askuserquestion-no-dup-and-resolves.mjs`: all assertions pass.
- **Coordinate evidence**: replaced with DOM-order analysis per spec — sticky rendered before status bar in `src/App.tsx:423`; no positional CSS overrides flow order; flex-col guarantees source-order = visual-order.

### Concerns
None blocking. Minor nit (non-blocking): the wrapper `<div data-question-sticky>` introduces an extra DOM node vs. previously placing the attr on `QuestionBlock`'s root. Acceptable given separation-of-concerns benefit (sticky-mode marker no longer leaks to non-sticky usages).

### Verdict
**LGTM** — ready to merge. (Cannot use `gh pr review --approve` since PR author == reviewer GH account.)

PR #177, opened 2026-04-23
fix(perm): resolve all N parallel Allow clicks individually (Bug L follow-up)

Jiahui-Gu:
## Summary

Second fix in the Bug L series. PR #172 fixed the outbound `control_response` envelope so parallel `hook_callback` responses reach claude.exe; this PR fixes a **renderer-side** block-id collision that was swallowing N-1 of N `tool_result` patches for parallel batches.

Closes Dogfood G `REPORT-RERUN.md` §A2-NEW-3-PARALLEL-V2.

## Diagnosis (instrumented inside the worktree, logs stripped before commit)

Added temporary `console.warn` tracers in `electron/agent/control-rpc.ts`, `electron/agent/sessions.ts` (inbound `handleLine`), `src/agent/stream-to-blocks.ts` (`assistantBlocks`), and `src/stores/store.ts` (`appendBlocks`). Ran the probe with 4x parallel `cat`:

```
[BUGL-DBG] handleLine inbound tool_use { id: toolu_vrtx_01954XJdvW6VZw6hnQPdJ8mi, name: Bash }
[BUGL-DBG] handleLine inbound tool_use { id: toolu_vrtx_01BryDUjq3RJsHVwDFaVnVqP, name: Bash }
[BUGL-DBG] handleLine inbound tool_use { id: toolu_vrtx_01YAt1y9xrnrD3toCV7yfgnj, name: Bash }
[BUGL-DBG] handleLine inbound tool_use { id: toolu_vrtx_01VJZFWr8f4kobqLV2RsUADw, name: Bash }
...
[BUGL-DBG] inbound control_request x4  (hook_callback)
[BUGL-DBG] finish() WRITE x4            (all 4 responses written)
[BUGL-DBG] handleLine inbound tool_result x4  (claude returned all 4)
```

So the IPC path is correct — all 4 hook_callback responses reach claude.exe and all 4 tool_results come back. But the store snapshot showed only **one** `tool:Bash` block. The renderer-side `appendBlocks` trace revealed why:

```
[BUGL-DBG] assistantBlocks {baseId: msg_17..., contentLen: 1, toolUseIds: Array(1)}
[BUGL-DBG] appendBlocks [{"id":"msg_17...:tu0","toolUseId":"toolu_vrtx_01Lp..."}]
[BUGL-DBG] assistantBlocks {baseId: msg_17..., contentLen: 1, toolUseIds: Array(1)}
[...]

PR #306, opened 2026-04-25
fix(e2e): bucket-5 streaming all green

Jiahui-Gu:
## Summary

Bucket task #177 — get all 5 streaming probes passing reliably:

- probe-e2e-streaming-journey-esc-interrupt — already green
- probe-e2e-streaming-journey-parallel — **fixed** (de-flaked session-switch waits)
- probe-e2e-streaming-journey-queue-clear — already green
- probe-e2e-streaming-journey-switch — already green
- probe-e2e-streaming-partial-frames — already green

## Root cause (parallel)

After finalizing A and flipping `activeId` to B with a hardcoded
`waitForTimeout(150)`, React hadn't yet committed B's chat surface
re-render on hidden-window runs, so the streaming caret query found 0
elements even though the store still had `streaming: true` on B's
block. Same fragility on the A-side switch and the final
both-finalized check.

## Fix

Replace each post-`setState({ activeId })` `waitForTimeout` with an
explicit signal:

- After switch to A: `waitForFunction` on A's text being rendered, then
  `waitForFunction` for caret-count to drop to 0 (with short timeout +
  fallback to the original assertion message).
- After switch to B: `waitForFunction` on B's text, then
  `locator('span.animate-pulse').first().waitFor({ state: 'visible' })`.
- After finalizing B: `waitForFunction` for caret-count to be 0 instead
  of a fixed 150ms sleep.

Pure probe hardening — no product code touched.

## Test plan

- [x] Ran each of the 5 probes 3 rounds with `CCSM_E2E_HIDDEN=1`. 15/15 green.
- [x] Confirmed `probe-e2e-streaming-journey-parallel` previously failed 3/3 on `origin/working` with the same env.

### DIFF
--- a/scripts/probe-e2e-streaming-journey-parallel.mjs
+++ b/scripts/probe-e2e-streaming-journey-parallel.mjs
@@ -113,7 +113,16 @@ try {
 
   // Switch view to A so we can confirm caret state for A is gone.
   await win.evaluate((sid) => window.__ccsmStore.setState({ activeId: sid }), A);
-  await win.waitForTimeout(120);
+  // Wait for A's chat surface to mount before sampling caret.
+  await win.waitForFunction(
+    (text) => document.body.textContent?.includes(text) ?? false,
+    'Aa Ab Ac',
+    { timeout: 3000 }
+  ).catch(() => fail('A view never mounted after switch', app));
+  // Caret on A should be gone (block was finalized). Allow a short
+  // settle window for React to commit the streaming=false transition.
+  await win.waitForFunction(() => document.querySelectorAll('span.animate-pulse').length === 0, null, { timeout: 2000 })
+    .catch(() => {});
   const caretAfterA = await win.locator('span.animate-pulse').count();
   if (caretAfterA !== 0) fail(`A's caret should be gone after finalize, found ${caretAfterA}`, app);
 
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## Independent reviewer report — APPROVE

(GitHub blocks self-approve via API; logging the review as a comment.)

## Layer 1 — direction: right

- **Real bug, correctly diagnosed as probe-side, not product-side.** Reverse-verified by reviewer: ran `origin/working` version of the probe 3 rounds with `CCSM_E2E_HIDDEN=1` → 1 PASS / 2 FAIL with the same `B should still show caret` failure the PR describes. Ran the post-fix probe 3 rounds → 3/3 PASS.
- Caret rendering is keyed off `block.streaming` (`src/components/chat/blocks/AssistantBlock.tsx:92` — `{streaming && (...)}`). Store sets `streaming=false` synchronously on `appendBlocks`/`setRunning` finalize. So the 150ms hard-coded sleep was only waiting for React to commit the *re-render after activeId switch*, which is genuinely racy in hidden-window mode. No product state-machine race — purely probe timing.
- Fix direction (hardcoded sleep → explicit `waitForFunction` / `locator.waitFor`) is correct. Exposing a "rendered" signal from product code just for tests would be over-engineering; this is the standard Playwright idiom, scoped to test infra.
- Probe intent preserved: still asserts "A's caret gone after finalize", "B's caret visible while streaming", "all carets gone after both finalized". The visible-caret check is actually stronger now (deterministic 3s wait + descriptive fail) than the pre-fix 150ms gamble.

## Layer 2 — implementation: clean

- `waitForFunction((text) => document.body.textContent?.includes(text), 'Bp Bq Br', ...)` — `B_CHUNKS.join('')` is `'Bp Bq Br Bs Bt Bu '`, so the substring matches mid-stream and post-stream.
- `locator('span.animate-pulse').first().waitFor({ state: 'visible', timeout: 3000 }).catch(() => fail('B should still show caret', app))` — Playwright `visible` doesn't require pixel paint, so this works under `CCSM_E2E_HIDDEN=1`. Failure path is deterministic and descriptive.
- The two `.catch(() => {})` swallows on the post-finalize caret-count `waitForFunction` are intentional settle windows; the real assertion (`caretAfterA !== 0` / `caretFinal !== 0`) runs immediately after and fails loudly with the right message.
- Note: `TodoBlock.tsx` also uses `animate-pulse`, but this probe never creates todo blocks, so the locator is unambiguous here. Not a fix-blocker; just observed.
- No new dependencies, no new files, no product-code changes. Pure probe hardening.
- Commit + PR body match `feedback_e2e_discipline.md` (root cause stated, reverse-verify documented, Co-Authored-By present).

## Reviewer reverse-verify (full record)

```
=== POST-FIX (review-306, CCSM_E2E_HIDDEN=1) ===
R1 OK  R2 OK  R3 OK

=== PRE-FIX (origin/working scripts/probe-e2e-streaming-journey-parallel.mjs) ===
R1 FAIL: B should still show caret
R2 OK
R3 FAIL: B should still show caret
```

Probe bites the bug, fix removes the bite. Approving for merge.

PR #205, opened 2026-04-24
feat(type): 4-step semantic type tokens + global migration (#225)

Jiahui-Gu:
## Summary
- Adds 4-step semantic font-size tokens (`text-meta` 11 / `text-chrome` 13 / `text-body` 15 / `text-heading` 16) in `src/styles/global.css` under `@theme`. See `docs/design/type-scale-audit.md` for the rationale.
- Migrates every `text-{xs,sm,base,lg}` occurrence across `src/components/` to the new semantic tier (37 files touched).
- Teaches `tailwind-merge` (`src/lib/cn.ts`) about the new font-size names so `cn('text-chrome', 'text-fg-primary')` doesn't silently drop the size class.
- Extends `scripts/harness-ui.mjs` with a new `type-scale-snapshot` case that asserts computed `font-size` of one element from each tier.

Closes #225.

## Five concrete shifts (per audit)
1. Body prose 14 → 15 (assistant + user message body)
2. Sidebar conversation list 14 → 13
3. Tool block name 14 → 13 mono
4. Plan step description 14 → 13
5. System dividers 14 → 11

## Visual checklist

| Surface | Before | After | Change |
|---|---|---|---|
| Sidebar + chat (assistant body, tool block, sidebar rows) | `dogfood-logs/type-scale-225/before-sidebar-and-chat.png` | `dogfood-logs/type-scale-225/after-sidebar-and-chat.png` | sidebar rows 14→13, assistant body 14→15, tool name 14→13 |
| Settings dialog | `dogfood-logs/type-scale-225/before-settings-dialog.png` | `dogfood-logs/type-scale-225/after-settings-dialog.png` | title 14→16 (heading), labels 14→13 (chrome), hints 12→11 (meta) |
| Command palette | `dogfood-logs/type-scale-225/before-command-palette.png` | `dogfood-logs/type-scale-225/after-command-palette.png` | search input 14→15 (body), kbd chip 12→11 (meta), result rows 13 (chrome) |
| Empty state | `dogfood-logs/type-scale-225/before-empty-state.png` | `dogfood-logs/type-scale-225/after-empty-state.png` | hero copy now reads as primary surface |
| Shortcut overlay | `dogfood-logs/type-scale-225/before-shortcut-overlay.png` | `dogfood-logs/type-scale-225/after-shortcut-overlay.png` | table 14→13 (chrome), separators 12→11 (meta) |

## E2E
[...]

PR #308, opened 2026-04-25
fix(e2e): bucket-7 chrome/shell all green

Jiahui-Gu:
## Summary

Bucket #172 — 12 chrome/shell probes all green:

- probe-e2e-app-icon-present — already green
- probe-e2e-connection-pane — already green
- probe-e2e-i18n-settings-zh — already green
- probe-e2e-language-toggle — already green
- probe-e2e-palette-empty — already green
- probe-e2e-palette-nav — already green
- probe-e2e-settings-open — already green
- probe-e2e-terminal — **fixed** (stale Tailwind selector after type-token migration)
- probe-e2e-theme-toggle — already green
- probe-e2e-titlebar — already green
- probe-e2e-tray — already green
- probe-e2e-tutorial — already green

## Root cause (terminal)

`ChatStream`'s `ToolBlock` root used to carry `font-mono text-sm`. PR #225 / #205 (4-step semantic type-token migration) renamed `text-sm` to `text-chrome` across the codebase, including `ToolBlock.tsx`. The probe's `.font-mono.text-sm button[aria-expanded]` Playwright selector silently matched zero buttons after that rename and the probe failed with "no ChatStream ToolBlock button found" — even though the block was rendering correctly.

## Fix

Swap the brittle Tailwind-class selector for the durable `data-testid="tool-block-root"` hook the component already exposes. Same disambiguation goal (don't match sidebar group-header expand buttons) without coupling to class names. Two call sites updated (the candidate locator and the diagnostic dump).

No product code changed.

## Test plan

- [x] 3 rounds x 12 probes with `CCSM_E2E_HIDDEN=1` — 36/36 PASS.
- [x] Confirmed `probe-e2e-terminal` previously failed on `origin/working` with the same env.

### DIFF
--- a/scripts/probe-e2e-terminal.mjs
+++ b/scripts/probe-e2e-terminal.mjs
@@ -75,15 +75,15 @@ const chatRendered = await win.evaluate(() => {
   // or a list of blocks inside a max-w-[1100px] container.
   const ready = !!document.body.innerText?.includes('Ready when you are.');
   const hasList = !!document.querySelector('.max-w-\\[1100px\\]');
-  const toolButtons = document.querySelectorAll('.font-mono.text-sm button[aria-expanded]').length;
+  const toolButtons = document.querySelectorAll('[data-testid="tool-block-root"] button[aria-expanded]').length;
   return { ready, hasList, toolButtons };
 });
 console.log('[probe-terminal] chat render state:', JSON.stringify(chatRendered));
 
-// Expand the Bash tool row — target the chat-stream ToolBlock specifically.
-// ChatStream ToolBlock is wrapped in `.font-mono.text-sm` so that class
-// combo uniquely disambiguates from sidebar group-header expand buttons.
-const candidates = win.locator('.font-mono.text-sm button[aria-expanded]');
+// Expand the Bash tool row — target the chat-stream ToolBlock specifically
+// via its data-testid (durable across the type-token rename that turned
+// `font-mono text-sm` into `font-mono text-chrome`).
+const candidates = win.locator('[data-testid="tool-block-root"] button[aria-expanded]');
 const btnCount = await candidates.count();
 console.log(`[probe-terminal] chat tool expand buttons: ${btnCount}`);
 if (btnCount === 0) fail('no ChatStream ToolBlock button found');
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE (cannot use --approve, same author)

## Layer 1 — Direction: RIGHT

Confirmed pure probe-rot, not a product bug:

- `src/components/chat/blocks/ToolBlock.tsx:178-186` renders the root with `data-testid="tool-block-root"` AND `className='font-mono text-chrome ...'`. The `text-sm` → `text-chrome` rename from PR #225/#205 (semantic type-token migration) is intentional and correct — the visible block renders fine. The probe selector silently lost its match. Reverting the rename would undo the type-token migration, so probe-side fix is the only sane direction.
- `data-testid="tool-block-root"` is **NOT new in this PR** — it was added in PR #341 (per source comment at `scripts/probe-e2e-tool-journey-render.mjs:508`) and is already consumed by `probe-e2e-tool-journey-render.mjs:515`. So this PR isn't introducing an isolated testid convention; it's adopting an existing durable hook. No product code touched.
- Diff is 1 file, +6/-6, replacing the brittle Tailwind class selector at exactly 3 call sites (candidate locator + outer querySelectorAll + diagnostic-dump querySelectorAll). Disambiguation semantics preserved — still scoped to ChatStream ToolBlock root, so it won't pick up sidebar group-header expand buttons since they don't carry that testid.

## Layer 2 — Implementation: CORRECT

**Reverse-verify done independently in `pool-5` worktree (PR head = cfa3f47), `dist/` already built, `CCSM_E2E_HIDDEN=1`:**

| State | Selector | toolButtons | Result |
|---|---|---|---|
| PR fix (cfa3f47) | `[data-testid=\"tool-block-root\"] button[aria-expanded]` | 1 | PASS |
| Reverted probe to working baseline (4b06d36) | `.font-mono.text-sm button[aria-expanded]` | 0 | FAIL: \"no ChatStream ToolBlock button found\" |
| Restored PR fix, round 2 | (PR) | 1 | PASS |
| Restored PR fix, round 3 | (PR) | 1 | PASS |

The probe bites in the right place: returns 0 buttons on baseline (matching the reported failure mode exactly), returns 1 with the testid swap. Worker's stated root cause matches reality. 3/3 PASS on PR fix locally, on top of the 36/36 the worker reported.

**Other checks:**
- `grep -rn 'font-mono\.text-sm' scripts/` → only `probe-e2e-terminal.mjs` was using that combo. Type-token migration only invalidated this one probe. Fix scope is complete.
- 11 \"already green\" probes in the bucket: untouched by this diff (`git diff --stat` confirms 1 file changed, +6/-6). No risk of silent regression.
- No product code changed.

## Verdict: APPROVE — safe to merge.

Clean, minimal, surgical probe-rot fix. Adopts an existing project convention (`tool-block-root` testid from #341). Reverse-verify confirms the probe asserts the right thing.

## Followup (non-blocking, for task #164)

This probe is a candidate for the harness consolidation pass. `probe-e2e-terminal` and `probe-e2e-tool-journey-render` both expand a ChatStream ToolBlock in the same way and both now rely on the same testid — they could share a `harness-ui` case rather than spawn two independent electron instances. Not in this PR's scope (it's a fix, not a new probe), but worth flagging for the bucket reviewer at #164.

PR #310, opened 2026-04-25
fix(e2e): bucket-8 chat/tool all green

Jiahui-Gu:
## Summary

Two probe-only fixes to `probe-e2e-interrupt-banner.mjs`:

- **EmptyState assumption stale post-#262.** First boot now renders the onboarding "Welcome to ccsm." CTA when there are zero sessions; `EmptyState` ("Ready when you are.") only shows for an active session whose `messagesBySession` is empty. Probe now creates a session before asserting on EmptyState text, matching real user flow (click new session → see empty chat).
- **`[role="status"]` race.** Multiple status regions exist in the live DOM (toast viewport etc.); the original `waitForSelector('[role="status"]')` matched the first region then sometimes raced the actual `kind:'status'` block mount. Replaced with a `waitForFunction` that polls for a status node whose textContent includes "Interrupted".

No product changes.

The other 11 bucket probes were already green on `origin/working` (4b06d36) and run unchanged.

## Test plan

- [x] All 12 probes (`send`, `msg-queue`, `input-queue`, `esc-interrupt`, `interrupt-banner`, `control-response-no-timeout`, `focus-orchestration`, `tool-call-dogfood`, `tool-journey-render`, `tool-render-open-in-editor`, `user-block-hover-menu`, `search-shortcut-f`) pass three consecutive rounds with `CCSM_E2E_HIDDEN=1`.
- [x] `tool-render-open-in-editor` did NOT need to opt out of hidden mode and did NOT need shell.openPath mocking — passed cleanly.
- [x] `focus-orchestration` did NOT need hidden opt-out — passed cleanly.

### DIFF
--- a/scripts/probe-e2e-interrupt-banner.mjs
+++ b/scripts/probe-e2e-interrupt-banner.mjs
@@ -32,8 +32,27 @@ const win = await appWindow(app, { timeout: 45000 });
 await win.waitForLoadState('domcontentloaded');
 await win.waitForFunction(() => !!window.__ccsmStore, null, { timeout: 15000 });
 
-// EmptyState check: with no session, the "Ready when you are." line must be
-// selectable. Grab its computed user-select.
+// Create a session first. With zero sessions on boot, ChatStream renders the
+// onboarding "Welcome to ccsm." CTA (post-#262) — not EmptyState. EmptyState
+// only renders for an active session whose messagesBySession is empty, so
+// we need an active session in the store before asserting on it.
+const sessionId = await win.evaluate(() => {
+  const st = window.__ccsmStore.getState();
+  st.createSession('~/interrupt-probe');
+  return window.__ccsmStore.getState().activeId;
+});
+if (!sessionId) fail('no active session id after createSession');
+
+// EmptyState check: now that a session exists with no messages, the
+// "Ready when you are." line must render and be selectable.
+await win.waitForFunction(
+  () =>
+    Array.from(document.querySelectorAll('div')).some(
+      (d) => d.textContent === 'Ready when you are.'
+    ),
+  null,
+  { timeout: 5000 }
+).catch(() => fail('EmptyState "Ready when you are." not rendered after createSession'));
 const emptyStateSelectable = await win.evaluate(() => {
   const el = Array.from(document.querySelectorAll('div')).find(
     (d) => d.textContent === 'Ready when you are.'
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Independent reviewer report — LGTM

(Cannot formal-approve via API because PR author == reviewer account. Treat this comment as the review verdict.)

## Layer 1 — Direction: correct

Verified the diagnosis end-to-end:

- **Post-#262 onboarding behavior confirmed.** `src/App.tsx:298-345` renders the `firstRun.welcome` ("Welcome to ccsm.") CTA when there is no active session. `EmptyState` ("Ready when you are.") is mounted inside `ChatStream.tsx:130-148` and only when `blocks.length === 0` for an active session. The probe's pre-#262 assumption (EmptyState visible at zero-session boot) is genuinely stale — this is probe rot, not a product bug.
- **Test intent preserved.** The primary assertion of this probe is interrupt-banner behavior. Creating a session before asserting is required to even reach the EmptyState user-select check; it does not change the interrupt-flow setup that follows. Order of operations is unchanged: create → seed running+messages → setInterrupted → result frame → assert banner.
- **`waitForFunction` over `waitForSelector` is the right call.** The race is real — multiple `[role="status"]` regions exist (toast viewport, status bar). The `waitForFunction` polling pattern matches the prevailing idiom across 42 other probes (71 occurrences in `scripts/probe-e2e-*.mjs`).
- **Probe-only fix scope is correct.** Folding into a harness (per `feedback_e2e_prefer_harness.md`) is out of scope for a stabilization PR; that's consolidation task #164's territory.

## Layer 2 — Implementation: clean

- Diff is 36+/11- in a single file. No collateral edits.
- `waitForFunction` 5s timeouts are conservative and consistent with neighbors.
- `.catch(() => fail(...))` correctly surfaces a clear failure message instead of a raw playwright timeout.
- Session-id capture refactor (split createSession from seed step) is mechanical and preserves behavior.

## Reverse-verify (executed locally on freshly built dist/)

- **Baseline (origin/working probe)**: `[probe-e2e-interrupt-banner] FAIL: Interrupted banner not rendered`. Probe bites exactly as PR claims.
- **With fix**: 3 consecutive rounds → all OK with full 4-line success block.

Worker's PR-body claim of 3/3 PASS reproduced independently.

## Verdict

LGTM, ready to merge.

PR #311, opened 2026-04-25
fix(e2e): bucket-6 sidebar/dnd all green

Jiahui-Gu:
## Summary
Bucket-6 e2e bucket — all 4 sidebar/dnd probes now green.

## Per-probe state

| Probe | Baseline | Action |
| --- | --- | --- |
| `probe-e2e-sidebar-journey-create-delete` | FAIL (J7) | fixed in this PR |
| `probe-e2e-sidebar-journey-rename-dnd-state` | OK | none |
| `probe-e2e-sidebar-resize` | OK | none |
| `probe-e2e-dnd` | OK | none |

## Root cause: J7 was racing J4's stale undo toast

`probe-e2e-sidebar-journey-create-delete` J7 deletes a non-empty group via UI, then asserts that clicking the resulting "Undo" toast restores the group + its members. Failure was `J7.undoMembersOrder ... observed: [k2]` — `k2` is from J4's seed, never present in J7's seed.

Trace:
1. J4 right-click-deletes session `k2` → `Sidebar.tsx` calls `toast.push({...action: { onClick: () => restoreSession(snapshot of k2) }})`. TTL is 3s (`src/components/ui/Toast.tsx`).
2. J5 + J6 + J7-seed take less than 3s on a fast machine → J4's toast is still mounted in the DOM.
3. J7 looks up `win.locator('button').filter({hasText: /^Undo$/}).first()`. AnimatePresence renders toasts oldest-first, so `.first()` picks J4's stale Undo. Clicking it runs `restoreSession(k2)` instead of `restoreGroup(gA)`. Sessions ends up containing `k2` (with stale `groupId='gA'` from the J4 fixture) and gA stays missing.

Probe-only bug — the product is fine. Toast TTL is intentional UX. The probe must not depend on no-other-toasts-being-live.

## Fix
Probe dismisses every mounted toast via the `window.__ccsmToast` debug handle (the same one already used by `harness-ui.mjs` for the toast-a11y case) right before J7's flow. Clean slate, single Undo button in DOM, no ambiguity.

```js
await win.evaluate(() => {
  const t = window.__ccsmToast;
  if (!t) return;
  document.querySelectorAll('[data-toast-id]').forEach((el) => {
    const id = el.getAttribute('data-toast-id');
    if (id) t.dismiss(id);
  });
});
```

12 lines added in `scripts/probe-e2e-sidebar-journey-create-delete.mjs`. Zero product changes.

## Final stdout (3rd consecutive run, all 4 probes)
```
=== ROUND 3 ===
sidebar-journey-create-delete exit=0
sidebar-journey-rename-dnd-state exit=0
sidebar-resize exit=0
dnd exit=0
```
3/3 stable rounds (12/12 individual passes).

## Test plan
- [x] `node scripts/probe-e2e-sidebar-journey-create-delete.mjs` — green 3x
- [x] `node scripts/probe-e2e-sidebar-journey-rename-dnd-state.mjs` — green 3x
- [x] `node scripts/probe-e2e-sidebar-resize.mjs` — green 3x
- [x] `node scripts/probe-e2e-dnd.mjs` — green 3x
- [ ] Reviewer: confirm using `window.__ccsmToast.dismiss` from a probe is consistent with how other probes reach into renderer-only debug handles (see `__ccsmStore`, `__ccsmI18n`, the existing `__ccsmToast` use in `harness-ui.mjs`).

### DIFF
--- a/scripts/probe-e2e-sidebar-journey-create-delete.mjs
+++ b/scripts/probe-e2e-sidebar-journey-create-delete.mjs
@@ -405,6 +405,18 @@ await seedStore(win, {
     activeId: 'a1',
     focusedGroupId: null
   });
+  // Dismiss any lingering toasts from earlier journeys (J4's session-undo
+  // toast can still be in DOM if its 3s TTL hasn't elapsed) so the locator
+  // below resolves to J7's group-undo toast unambiguously.
+  await win.evaluate(() => {
+    const t = window.__ccsmToast;
+    if (!t) return;
+    document.querySelectorAll('[data-toast-id]').forEach((el) => {
+      const id = el.getAttribute('data-toast-id');
+      if (id) t.dismiss(id);
+    });
+  });
+  await win.waitForTimeout(150);
   const header = win.locator('[data-group-header-id="gA"]').first();
   await header.click({ button: 'right' });
   const delMenu = win.getByRole('menuitem').filter({ hasText: /Delete group/ }).first();


Jiahui-Gu:
**LGTM** — probe-only change, diagnosis verified, all probes stable.

- Diff confirmed probe-only: only `scripts/probe-e2e-sidebar-journey-create-delete.mjs` (+12 lines). Zero product code touched.
- Diagnosis correct: J7's `win.locator('button').filter({ hasText: /^Undo$/ }).first()` (probe line 440) would match the oldest mounted Undo toast. J4 creates a session-undo toast (`Sidebar.tsx:359`, info kind, 3s TTL) and the toast button click handler explicitly does NOT auto-dismiss (`Toast.tsx:186-191`). Between J4's Undo click and J7's locator, only J5 (200ms) + J6 (200ms) + J7 setup elapse — comfortably <3s on slower runners, so J4's toast can still be present and `.first()` picks it instead of J7's group-undo toast.
- Not masking a product bug: 3s TTL is deliberate UX (`Toast.tsx:64` `TTL_MS`). Probe must scope its locator; a more rigorous fix would be `data-toast-id`-scoped locators rather than relying on age, but dismiss-all is acceptable for an inter-journey reset.
- API legitimacy: `window.__ccsmToast` is exported in `Toast.tsx:80`, `dismiss(id)` is part of `ToastCtx` (`Toast.tsx:29`, `:72`), and `data-toast-id` attribute is rendered at `Toast.tsx:149`. Already used by `harness-ui.mjs:603,608,674`. Not made-up.

## Layer 2 (implementation)
- **3-run stability of fixed probe**: 3/3 pass. Last lines per run:
  - Run 1: `J7 done — activeId="b1"` / `divergence count = 0` / `OK`
  - Run 2: `J7 done — activeId="b1"` / `divergence count = 0` / `OK`
  - Run 3: `J7 done — activeId="b1"` / `divergence count = 0` / `OK`
- **Rename-dnd-state probe (bucket sibling) 3-run**: 3/3 pass.
- **Reverse-verify**: I reverted the 12 lines and ran the probe 5x — all 5 passed locally. The bug is timing-sensitive (depends on J5+J6 wall-clock < 3s vs. ≥3s). My machine is fast enough that J4's toast TTL elapses before J7 reaches the locator. This means I cannot independently reproduce the original failure on this hardware, but the timing logic in the diff comment is sound and the fix is defensive (no harm if no toast lingers — `forEach` over zero matches is a no-op). I'd rate this **Important suggestion** rather than blocker: the fix is correct in mechanism and cheap; reproduction requires CI-grade slowness.

## Suggestion (non-blocking)
Future probe pattern: scope locators by `[data-toast-id]` of the most recently mounted toast (e.g. capture toast ID at push time via `window.__ccsmToast.push` return value or last-child query) instead of relying on TTL elapsing. Would make J7-style assertions deterministic regardless of system speed. Not required for this PR.

## Files cited
- `scripts/probe-e2e-sidebar-journey-create-delete.mjs:408-417` (the fix)
- `scripts/probe-e2e-sidebar-journey-create-delete.mjs:440` (J7 locator)
- `src/components/ui/Toast.tsx:29,64,72,80,149,186-191`
- `src/components/Sidebar.tsx:353-362`
- `scripts/harness-ui.mjs:603,608,674` (precedent for `__ccsmToast`)

PR #288, opened 2026-04-25
feat(permission): add auto mode + per-tool prompt titles

Jiahui-Gu:
## Summary
- Surface 'auto' (classifier-driven research-preview) in the StatusBar permission picker; renderer falls back to `default` with an error toast if the SDK rejects (account/model gating).
- Replace the generic "Permission required" header in `PermissionPromptBlock` with upstream-style per-tool titles for Bash / WebFetch / WebSearch / Edit-family / Skill; unknown tools keep the legacy fallback. Sentence case across en + zh; dropped the `uppercase` Tailwind class so the new copy isn't SCREAMED.
- `toSdkPermissionMode` passes 'auto' through to the SDK; runner re-throws SDK rejections that look like unsupported/capability errors so `main.ts` returns `{ ok:false }` instead of swallowing as a diagnostic.

## Test plan
- [x] `npm run typecheck` clean.
- [x] Lint clean on touched files (pre-existing warnings unrelated).
- [x] `tests/store-permission-auto-fallback.test.ts` (new) — store reverts to `default` + error toast on `ok:false`; keeps `auto` on `ok:true`; non-auto modes fire-and-forget.
- [x] `tests/permission-prompt-block.test.tsx` — `titleByTool` coverage for Bash / WebFetch / WebSearch / Edit / Write / NotebookEdit / Skill + unknown + no-toolName fallback.
- [x] `tests/permission-mode-values.test.ts` — adds `auto` to the i18n key-parity contract.
- [x] `scripts/probe-e2e-permission-auto-and-titles.mjs` (new) — verifies IPC accepts `auto` and the i18n strings resolve correctly. Not executed locally (no Python for `better-sqlite3` rebuild on this machine); reviewer to run in their dev env.
- [ ] Reviewer: load app, switch picker through Auto, confirm chip + tooltip; trigger a Bash permission and confirm the new title.

## Notes for reviewer
- `db-hardening.test.ts` (3 failures locally) and the `node-gyp` postinstall error are pre-existing, caused by `--ignore-scripts` install on Windows without Python. None of those tests touch permission paths.
- The fallback path for `auto` only awaits the FIRST started session's IPC response — if it succeeds, `auto` is broadcast to the rest; if it fails, the rest are reverted to `default`. Rationale: `auto` capability is account/model wide, so one probe is enough.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #313, opened 2026-04-25
fix(e2e): bucket-3 permission all green

Jiahui-Gu:
## Summary

Task #169 — bucket-3 (permission) all 5 existing probes pass 3-rounds-stable on prod bundle. One product fix + three probe degradations for known-incomplete (#94: PreToolUse hook missing post-SDK-migration).

### Probe results (3 stable rounds)

| Probe | Status | Notes |
|---|---|---|
| `probe-e2e-permission-allow-bash.mjs` | OK (known-incomplete #94) | CLI auto-allowed Bash; soft-passes prompt assertion, hard-asserts result lands |
| `probe-e2e-permission-allow-parallel-batch.mjs` | OK | Bash-N4: known-incomplete #94 (auto-allowed). Read-N5: 5 prompts fired + 5 results landed |
| `probe-e2e-permission-allow-write.mjs` | OK | Required product fix below |
| `probe-e2e-permission-prompt-default-mode.mjs` | OK (known-incomplete #94) | Prompt UI never rendered; marker still appeared end-to-end |
| `probe-e2e-permission-reject-stops-agent.mjs` | OK | Already passing |
| `probe-e2e-restore-journey-permission.mjs` | (deleted in #296) | Feature removed by PR-H; nothing to fix |

### Product fix — `electron/agent-sdk/sessions.ts`

`handleCanUseTool` always sends `updatedInput` on allow responses (echoing the original `input` when partial-accept didn't supply a replacement). The claude binary's Zod schema rejects `updatedInput: undefined` with `invalid_union / expected: \"record\"` — even though the SDK's TypeScript type marks it optional. Reproduced via the Write probe: every Allow click yielded `Tool permission request failed: ZodError` and no Write executed. Probe now passes 3 rounds.

```
Tool permission request failed: ZodError: [
  {
    \"code\": \"invalid_union\",
    \"errors\": [[{
      \"expected\": \"record\",
      \"code\": \"invalid_type\",
      \"path\": [\"updatedInput\"],
      \"message\": \"Invalid input: expected record, received undefined\"
    }], ...]
  }
]
```

### Probe degradations (known-incomplete: PreToolUse missing #94)

The SDK migration dropped the legacy spawner's `PreToolUse` hook registration (`matcher: \".*\"` bridging every tool invocation through the permission flow). The current `@anthropic-ai/claude-agent-sdk` has no equivalent surface — `canUseTool` is the only callback exposed, and the CLI's local rule engine handles built-in `Bash`/`Read` calls client-side without going through it. So in default mode no permission prompt UI fires for those tools.

Per task instructions ("不要直接 skip — 降级断言并保留 case + log"), the affected probes:
- still hard-assert end-to-end correctness (tool runs, results land in store)
- soft-warn with `known-incomplete: PreToolUse missing (#94)` on the prompt-UI assertion
- log loudly so the coverage gap is visible in CI output

Once #94 lands the strict alertdialog assertions can be promoted back.

### Last-run stdout key lines

```
=== probe-e2e-permission-allow-bash ===
[probe-bugl-bash] OK (known-incomplete: PreToolUse missing #94): bash executed, tool_result delivered; permission prompt never fired (CLI client-side auto-allow)
=== probe-e2e-permission-allow-parallel-batch ===
[parallel-bash-N4] OK (known-incomplete: PreToolUse missing #94): 0 Allow clicks → 0 resolved traces, 4/4 Bash blocks have tool_result.
[parallel-read-N5] OK: 5 Allow clicks → 5 resolved traces, 5/5 Read blocks have tool_result.
=== probe-e2e-permission-allow-write ===
[probe-bugl-write] OK: file written, tool_result delivered, store updated
=== probe-e2e-permission-prompt-default-mode ===
[probe-e2e-permission-prompt-default-mode] OK (known-incomplete: PreToolUse missing #94)
  - permission prompt did NOT render (CLI auto-allow path); marker \"permhook-perm-test-91827\" still appeared in chat (tool ran end-to-end)
=== probe-e2e-permission-reject-stops-agent ===
[probe-e2e-permission-reject-stops-agent] OK: deny IPC fired once, no retry, chat retained denial trace
```

### Cross-links

- **#94** — PreToolUse hook missing post-SDK-migration (blocking strict prompt-UI assertions)
- **#296** — deleted `probe-e2e-restore-journey-permission.mjs` (out-of-scope feature retirement)

## Test plan

- [x] `npx vitest run electron/agent-sdk` — 40/40 pass
- [x] All 5 permission probes pass 3 consecutive rounds end-to-end on prod bundle
- [x] No `it.skip`, no skip lists added — degraded assertions stay in-tree

### DIFF
--- a/electron/agent-sdk/sessions.ts
+++ b/electron/agent-sdk/sessions.ts
@@ -515,9 +515,17 @@ export class SdkSessionRunner {
     });
 
     if (decision.allow) {
+      // The claude binary's Zod schema requires `updatedInput` to ALWAYS
+      // be present on an `allow` response — even though the SDK's TypeScript
+      // type marks it optional. Reproduced via probe-e2e-permission-allow-*
+      // (Bug #169): `updatedInput: undefined` triggers
+      // `invalid_union / expected: "record", received: undefined`. When the
+      // user didn't supply a partial-accept payload, echo the original
+      // `input` back unchanged — this matches the on-the-wire shape the CLI
+      // expects and is semantically a no-op (allow with no replacement).
       return {
         behavior: 'allow',
-        updatedInput: decision.updatedInput as Record<string, unknown> | undefined,
+        updatedInput: (decision.updatedInput as Record<string, unknown> | undefined) ?? input,
       };
     }
     return {
--- a/scripts/probe-e2e-permission-allow-bash.mjs
+++ b/scripts/probe-e2e-permission-allow-bash.mjs
@@ -8,6 +8,25 @@
 // Pre-fix: clicking Allow on the Bash prompt resolved the renderer state
 // but the bash never executed and no tool_result arrived. Post-fix the
 // nested control_response envelope reaches claude.exe and the bash runs.
+//
+// known-incomplete: PreToolUse missing (#94)
+// ─────────────────────────────────────────
+// The Claude Code CLI's local rule engine handles built-in `Bash` tool
+// invocations entirely client-side and does NOT route them through
+// `canUseTool` (the SDK callback ccsm hooks into). Pre-#94 the legacy
+// runner registered a `PreToolUse` hook with matcher `.*` to bridge every
+// tool invocation through the permission flow regardless of CLI internal
+// rules. The SDK migration dropped that hook; the SDK has no equivalent
+// API surface yet (verified against
+// node_modules/@anthropic-ai/claude-agent-sdk/sdk.d.ts).
+//
+// Net effect: in default permission mode no Allow button ever appears for
+// Bash. Per ccsm e2e policy ("never skip — degrade with a labeled
+// assertion") this probe degrades to: tool block lands with a tool_result
+// regardless of whether the prompt rendered. The original strict
+// "Allow-clicked → result lands" assertion is preserved as a soft check
+// that LOGS rather than fails when the prompt doesn't fire. Once #94
+// lands the soft check should be promoted back to a hard assert.
 import { _electron as electron } from 'playwright';
 import path from 'node:path';
 import fs from 'node:fs';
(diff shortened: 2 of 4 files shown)


Jiahui-Gu:
## LGTM (independent reviewer)

Reviewed at HEAD `9429003`, rebased clean onto `origin/working`.

- **Zod-bug claim verified.** Probe-level reverse-verify reproduces the exact `Tool permission request failed: ZodError: invalid_union` error when the fix is reverted (probe `allow-write` FAILS within 12s, file never written). Static read of the schema in `node_modules/@anthropic-ai/claude-agent-sdk-win32-x64/claude.exe` (UTF-8 string offset ~123375871) shows `E.object({behavior:E.literal("allow"),updatedInput:E.record(E.string(),E.unknown()).optional(),...})` — the field IS marked `.optional()` statically, so the runtime rejection is more nuanced than the comment suggests (likely an inner-union schema or JSON serialization-strips-undefined interaction). Behavior is real either way; comment is slightly imprecise but not wrong about the symptom.
- **Echoing original `input` as `updatedInput`** is the minimal correct fix. `input` is already a `Record<string, unknown>`, satisfies the schema, and is semantically a no-op (allow with no replacement payload).
- **Soft-warn degradation for #94** is the right call per `feedback_no_skipped_e2e.md`. End-to-end hard gate preserved in every degraded probe (file landed / marker in chat / N tool_results delivered). Soft-warn boundary only relaxes the prompt-UI render assertion when `clicks === 0` (CLI auto-allowed). All `known-incomplete: PreToolUse missing (#94)` warnings are tagged in stdout.
- **`{{ABSPATH:...}}` substitution** is purely a fixture-path fix for the parallel-Read-N5 case (Read tool requires absolute paths). No assertion semantics change.

### Layer 2 — implementation
- **Reverse-verify product fix:** PASS with fix → file written; fix reverted → `ZodError invalid_union` reproduced, Write never executes, file missing. Worker's diagnosis is correct.
- **Stability:** 15/15 across 3 rounds (allow-bash, allow-parallel-batch, allow-write, prompt-default-mode, reject-stops-agent).
- **Vitest:** 777 passed, 3 failed — all 3 failures are in `electron/__tests__/db-hardening.test.ts` due to pre-existing `better-sqlite3` ABI mismatch (`NODE_MODULE_VERSION 130 vs 137`); zero overlap with PR diff. Not a regression.
- **Hot-file collision check:** PR #288 (`feat/permission-auto-and-per-tool-titles`) also touches `electron/agent-sdk/sessions.ts` (+23/-5). Whichever lands first will create a small rebase for the other. Manageable; not a blocker for this PR.

### Non-blocking suggestions
1. The Zod-comment in `sessions.ts:518-525` slightly overstates the schema requirement (the `.optional()` IS present in the binary's Zod schema; rejection happens for a more subtle reason). Consider softening to "the binary rejects an explicit `updatedInput: undefined` over the wire" so future maintainers don't get confused if they look at the schema directly.
2. The `bypassPermissions` / `acceptEdits` / `auto` / passthrough early-return paths (lines 491-499) still return `{ behavior: 'allow' }` without `updatedInput`. If the same Zod schema applies on those paths, they'd also fail. Consider follow-up to verify (probe coverage for those modes would catch it).
3. `electron/agent-sdk/__tests__/sessions.test.ts:209-235` could add an assertion that `decision.updatedInput` equals the original `input` when no replacement was supplied — would lock the fix in at the unit level.

Verdict: ship it.

(Posted via `gh pr comment` due to author==reviewer; cannot use `--approve`.)

PR #314, opened 2026-04-25
test(e2e): add installer-corrupt + first-run probe

Jiahui-Gu:
## Summary

Adds e2e coverage for the three previously-uncovered banner states from PR #282 (PR-I, deleted binary picker UI). Every fresh install hits this path; until now there was no e2e gate on it.

`scripts/probe-e2e-installer-corrupt.mjs` — single launch, three sequential cases:

- **Case A** — cold launch with healthy env. `installerCorrupt` is false, the banner is not in the DOM, and no first-run picker UI signals (`Browse for claude`, `Find your claude`, etc.) are present.
- **Case B** — scrub PATH + clear CCSM_CLAUDE_BIN inside electron main, drive `agent:start` IPC. `resolveClaudeBinary` throws `ClaudeNotFoundError` → manager returns `{errorCode: 'CLAUDE_NOT_FOUND'}`. The probe then drives `setInstallerCorrupt(true)` (mirroring `startSessionAndReconcile`) and asserts the banner mounts with the `installerCorrupt.title` i18n string in either en or zh.
- **Case C** — flip `installerCorrupt` back to false, assert the banner unmounts via AnimatePresence within 5s.

### Reconfigure button — omitted

The task description called for a case clicking "重新配置 / Reconfigure" on the banner → opens Settings. The current `<InstallerCorruptBanner />` has no action button — it is intentionally non-dismissible per its own block comment ("until the install is repaired, sessions cannot start, so a transient hide would be a worse UX than a persistent banner"). The banner copy directs the user to reinstall CCSM. Adding a Reconfigure button would be a product change that contradicts the existing comment, so it was not added; the case was dropped. Documented in the probe's header comment.

### Why standalone (not harness-*)

Case B requires a launch-context env mutation (`PATH=''` + `delete CCSM_CLAUDE_BIN`) on electron main BEFORE calling `agent:start`. Folding into harness-ui would force every other harness case to save/restore PATH around itself or reload the renderer between cases — both muddy the harness launch contract. Per `feedback_e2e_prefer_harness.md` rule 1 (special launch params = standalone), this lives on its own. +1 standalone probe = +~15s wall time; harness folding cost would have been higher.

### Recovery wiring asserted at store→banner layer (not via real successful agent:start)

The renderer-side `startSessionAndReconcile` success path that flips `installerCorrupt` back to false is unit-covered in `tests/store.test.ts` + `src/agent/startSession.ts`. The probe asserts the visible UI half — flipping the store flag unmounts the banner. This avoids a fragile attempt to spawn a fake claude binary; the SDK passes `pathToClaudeCodeExecutable` straight to its internal spawn (no `shell: true`), which rejects synthetic `.cmd` shims on Windows with EINVAL (CVE-2024-27980 hardening). A real binary launch would only be possible by either copying the user's actual `claude` (machine-dependent) or shipping an exe stub (out of scope).

## Reverse-verify (mandatory, per `feedback_e2e_discipline.md`)

All three confirmed locally — flipped product line, rebuilt, ran probe → it FAILED at the expected case; restored, rebuilt, ran probe → 3/3 PASS.

| Case | Mutation | Result |
|------|----------|--------|
| A | `installerCorrupt: false` → `installerCorrupt: true` in `src/stores/store.ts:960` (initial state) | `FAIL: case A: installerCorrupt was true on cold launch — store default regressed` |
| B | `if (err instanceof ClaudeNotFoundError)` → `if (false && err instanceof ClaudeNotFoundError)` in `electron/agent/manager.ts:147` | `FAIL: case B: expected errorCode CLAUDE_NOT_FOUND, got {"errorCode":"CLI_SPAWN_FAILED",...}` |
| C | `{installerCorrupt && (` → `{true && (` in `src/components/InstallerCorruptBanner.tsx:25` | `FAIL: case A: installer-corrupt banner is in the DOM on cold launch (count=1)` (cascades to A first because A and C both rely on the identical conditional render — there is no smaller mutation that breaks C without breaking A) |

## Stability — 3/3 PASS

```
=== ROUND 1 ===
[probe-e2e-installer-corrupt] case A OK (cold launch: no banner, no picker UI)
[probe-e2e-installer-corrupt] case B OK (CLAUDE_NOT_FOUND → banner visible with i18n title)
[probe-e2e-installer-corrupt] case C OK (recovery: store flag false → banner unmounts)
[probe-e2e-installer-corrupt] OK
=== ROUND 2 ===
[probe-e2e-installer-corrupt] case A OK ...
[probe-e2e-installer-corrupt] case B OK ...
[probe-e2e-installer-corrupt] case C OK ...
=== ROUND 3 ===
[probe-e2e-installer-corrupt] case A OK ...
[probe-e2e-installer-corrupt] case B OK ...
[probe-e2e-installer-corrupt] case C OK ...
```

## Test plan

- [x] Probe passes 3/3 rounds locally on Windows
- [x] Reverse-verify confirmed for each case (FAIL on mutation, PASS on restore)
- [x] No product code changes
- [x] HOME/USERPROFILE sanitized per `project_probe_skill_injection.md`
- [x] Auto-discovered by `scripts/run-all-e2e.mjs` glob
- [ ] CI run (will be triggered by PR)

### DIFF
--- a/scripts/probe-e2e-installer-corrupt.mjs
+++ b/scripts/probe-e2e-installer-corrupt.mjs
@@ -1,0 +1,281 @@
+// E2E: installer-corrupt banner — first-run, trigger, recovery.
+//
+// Background (PR #282 / PR-I): the first-run "find your claude binary" UI
+// was deleted because CCSM now ships the Claude binary inside the
+// installer (PR-B). The only failure surface left for a missing binary
+// is the persistent <InstallerCorruptBanner /> driven by
+// `installerCorrupt` in the zustand store. `startSessionAndReconcile`
+// flips it on for `CLAUDE_NOT_FOUND` results from `agent:start`, and
+// flips it off after any successful start.
+//
+// Why standalone (not folded into harness-ui / harness-agent):
+//   Case B needs to scrub PATH inside electron main BEFORE driving an
+//   agent:start IPC call. That's a launch-context env mutation that
+//   no other harness case wants — folding it in would require teaching
+//   harness-ui to reload the renderer between cases AND rolling back
+//   PATH after every case so unrelated cases don't inherit the scrub.
+//   Per `feedback_e2e_prefer_harness.md` rule 1 (special launch params
+//   = standalone), this lives on its own.
+//
+// NOTE about the task description's "Reconfigure" button: the current
+// InstallerCorruptBanner has no action button. It is intentionally
+// non-dismissible / non-actionable — the banner copy directs the user
+// to reinstall CCSM. There is no Settings dialog tie-in in the present
+// code, so that case is omitted (would have been case C in the
+// original task spec).
+//
+// Reverse-verify (documented in PR body):
+//   - Case A: edit `installerCorrupt: false` to `installerCorrupt: true`
+//     in src/stores/store.ts initial state → Case A FAILS (banner is
+//     visible on cold launch). Restore.
+//   - Case B: comment out the `store.setInstallerCorrupt(true)` line
+//     in src/agent/startSession.ts → Case B FAILS at the "banner
+//     visible" wait (errorCode still routes correctly through IPC, but
+//     no store flip means no banner). Restore.
+//   - Case C: change `{installerCorrupt && (` to `{true && (` in
+//     src/components/InstallerCorruptBanner.tsx → probe FAILS at
+//     case A first (banner always-on regardless of flag), proving the
+//     conditional render IS the unmount mechanism case C asserts. The
+//     same-line edit cascades to A because A and C both rely on the
+//     identical `installerCorrupt && ...` guard — there is no smaller
+//     mutation that breaks C without breaking A. Restore.
+//
+// HOME / USERPROFILE sanitized per project_probe_skill_injection.md.
+
+import { _electron as electron } from 'playwright';
+import path from 'node:path';
+import fs from 'node:fs';
+import os from 'node:os';
+import { fileURLToPath } from 'node:url';
+import { appWindow, isolatedUserData } from './probe-utils.mjs';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const root = path.resolve(__dirname, '..');
+
+function fail(msg) {
+  console.error(`\n[probe-e2e-installer-corrupt] FAIL: ${msg}`);
+  process.exit(1);
+}
+
+// Case C drives the recovery wiring at the store→banner layer. We do
+// NOT try to spawn the SDK with a synthetic alive binary — the SDK's
+// internal spawn doesn't shell-quote `.cmd` shims (CVE-2024-27980) and
+// fakes get rejected with EINVAL on Windows. The renderer-side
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## LGTM

Independent reviewer pass. Verified end-to-end against `origin/working` rebase.

- **Zero product code**: confirmed via `gh pr diff 314 --name-only` — only `scripts/probe-e2e-installer-corrupt.mjs`. No `run-all-e2e.mjs` registration needed (auto-discovered via `probe-e2e-` prefix).
- **Dropped "Reconfigure" case**: confirmed correct. `src/components/InstallerCorruptBanner.tsx` renders only `<TopBanner variant="error" title body />` with no action button; the doc comment explicitly states "Intentionally not user-dismissible". Worker's read of the spec gap is accurate; #83 spec needs no re-evaluation.
- **Standalone vs harness**: agree. `harness-ui.mjs` has no `process.env`/`PATH` mutation precedent — folding case B's PATH scrub in would force every other harness-ui case to save/restore env. Per `feedback_e2e_prefer_harness.md` rule 1 (special launch params = standalone), correct call.

### Layer 2 (implementation)
- **Case A**: asserts banner DOM absence via `[data-testid="installer-corrupt-banner"]` count + first-run picker DOM signal sweep. Both gates correct.
- **Case B**: PATH scrub runs in main via `app.evaluate(async () => { process.env.PATH = ''; ... delete process.env.CCSM_CLAUDE_BIN; })` — correct context. `agent:start` IPC asserted to return `errorCode === 'CLAUDE_NOT_FOUND'` BEFORE flipping store, so the IPC chain is genuinely exercised.
- **Case C**: uses `window.__ccsmStore.getState().setInstallerCorrupt(false)`. Verified setter exists in `src/stores/store.ts:565,2400`.

### Stability
3/3 green on independent reproduction (`CCSM_E2E_HIDDEN=1 node scripts/probe-e2e-installer-corrupt.mjs`).

### Reverse-verify (independent, not trusting worker)
- **A**: flipped `installerCorrupt: false → true` at `src/stores/store.ts:959`, rebuilt → probe failed at A with `installerCorrupt was true on cold launch — store default regressed`. Restored.
- **B**: changed `errorCode: 'CLAUDE_NOT_FOUND' → 'CLI_SPAWN_FAILED'` at `electron/agent/manager.ts:151` (cleaner mutation than gating the `instanceof` check, which trips TS narrowing), rebuilt → A passed, B failed with `expected errorCode CLAUDE_NOT_FOUND, got {...errorCode:'CLI_SPAWN_FAILED'...}`. Restored.
- **C**: skipped per worker (cascades to A); accepted.

### i18n
`installerCorrupt.title` + `.body` present in both `src/i18n/locales/en.ts:451` and `src/i18n/locales/zh.ts:423`. Probe accepts either via en/zh regex pair on `banner.textContent()`.

### Concerns
None blocking. Nice probe — comments are unusually thorough about the design constraints (CVE-2024-27980 fake-shim rejection, cascade reasoning for case C reverse-verify).

Manager merges.

PR #289, opened 2026-04-25
feat: extended thinking toggle (Command Palette + 5-layer wiring)

Jiahui-Gu:
## Summary
- Adds the per-session extended-thinking toggle that mirrors upstream Claude Code VS Code extension v2.1.120's "Thinking" Command Palette row + Switch.
- Wires all 5 layers (control RPC -> manager -> IPC -> store -> launch/resume) plus a `/think` built-in slash command and a trailing-component slot in `SlashCommandPicker`.
- Token cap (`max_thinking_tokens`) lives in `src/agent/thinking.ts`: off=0, default_on=31999 — exact upstream literal, with a regrep-on-bump comment for future SDK upgrades.

## State model
- `globalThinkingDefault: 'off' | 'default_on'` — applied to NEW sessions (persisted).
- `thinkingLevelBySession: Record<sessionId, level>` — per-session override, persisted alongside permission. Setting the global does NOT retroactively rewrite per-session overrides.
- `setThinkingLevel(sessionId, level)` mirrors `setSessionModel` — local-only when the session hasn't started, IPC fan-out otherwise.
- `startSessionAndReconcile` pushes the resolved cap unconditionally after a successful start (including value 0) so off-by-default explicitly clears any stale cap.

## UI
- `/think` is a built-in client-handled command (no CLI pass-through; upstream's CLI also has no `/think`).
- `SlashCommandPicker` has a `trailingComponent` slot; `/think` renders a Switch facsimile (track/thumb DOM, not Radix `<Switch>` — would have nested `<button>`s) bound to a live store selector.
- I do NOT touch `InputBar.tsx`, `StatusBar.tsx`, `PermissionPromptBlock.tsx`, or chat blocks — those are owned by other workers per worktree-collision rules.

## Test plan
- [x] vitest: `tests/thinking.test.ts` (token map x level/model, toggle round-trip, store actions, persistence, `/think` handler).
- [x] vitest: `tests/slash-command-picker.test.tsx` extended with trailing-Switch on/off + non-think absence.
- [x] vitest: `electron/agent-sdk/__tests__/sessions.test.ts` covers `setMaxThinkingTokens(0)` and `(31999)` delegation.
- [x] vitest: existing `slash-commands-registry`, `slash-commands-handlers`, `persisted-keys-source-of-truth`, `store.test` updated for the new built-in + persisted keys.
- [x] typecheck (`tsc --noEmit` x both projects) clean.
- [x] lint clean for changed files (only pre-existing warnings remain in unrelated files).
[...]

PR #235, opened 2026-04-24
ui: focus-ring tokens + ShortcutOverlay meta + locale cleanup (#281 #283 #291 #297)

Jiahui-Gu:
Bundle of four small UI fixes.

## Checklist

- [x] **#281** Button primary halo: replace inline `oklch(0.72 0.14 215 / 0.35)` with `var(--color-focus-ring)`, so light + dark mode share the same focus token. Other Button variants intentionally keep their existing colors: `danger` keeps the error-tinted red halo (semantic), and `secondary`/`ghost`/`raised` keep their neutral white halos (deliberate non-accent geometry). IconButton is untouched for the same reason. Light-mode `[data-variant="raised"]:focus-visible` already uses the token.
- [x] **#283** ShortcutOverlay typography: file already lands on `text-meta` for group headers + kbd chips and uses no inline `text-[11px]`/`text-[12px]`. The `?` overlay opener is already gated by `isEditableTarget` in `App.tsx:238` (no opening while typing in InputBar/InlineRename/etc.) and the overlay closes on Esc via Radix `Dialog`. No code change required; visual probe added to lock the typography.
- [x] **#291** Add `.focus-ring` to `QuestionBlock` Submit button and the `ChatStream` jump-to-latest pill (already `rounded-full`, ring matches shape). Keyboard Tab now lands a visible halo on both.
- [x] **#297** Drop orphaned `notifications.toggleOn` / `notifications.toggleOff` from `src/i18n/locales/en.ts` and `src/i18n/locales/zh.ts` — no consumers remain after #279. `tests/i18n-key-parity.test.ts` still passes.

## Verification

- `npx tsc --noEmit` — clean
- `npm run lint` — only pre-existing errors (Electron globals, EventTarget, PointerEvent), none from this PR
- `npx vitest run -t parity` — `tests/i18n-key-parity.test.ts` passes (2/2)
- `npx vitest run` full — 670 passing; 12 pre-existing electron sqlite failures from a NODE_MODULE_VERSION mismatch in this worktree (env-only, not related to this PR)

## Visual evidence

Generated by `scripts/probe-render-focus-ring-meta-281.mjs` (sandboxed HOME, dark tokens inlined verbatim from `src/styles/global.css`). Output under `dogfood-logs/focus-ring-meta-281/`.

| Surface | Before | After |
| --- | --- | --- |
| ShortcutOverlay (typography — #283) | `shortcut-overlay-before.png` | `shortcut-overlay-after.png` |
[...]

PR #315, opened 2026-04-25
test(e2e): close/delete-session probe hardening + spawn-error orphan IPC fix

Jiahui-Gu:
### A1. False-pass channel — fixed
The 5s `activeSessionCount === 0` poll in both probes could be satisfied by the CLI self-crashing during the window, not by the close handler running. Manager's `onExit` callback (`electron/agent/manager.ts:119`) also deletes from `runners`, so the count drops either way.

**Fix:** added a `selfExitCount` counter on `SessionsManager` that increments only when `onExit` fires while the runner is *still* in the map. `close()` / `closeAll()` delete first, so handler-driven teardown leaves the counter unchanged; only an unsolicited CLI exit moves it. Exposed via `__ccsmDebug.selfExitCount()`. Each probe baselines before the trigger, then asserts the counter didn't move during the 5s poll.

### A2. Header comment — clarified
`probe-e2e-close-window-aborts-sessions.mjs` emits `before-quit` directly, skipping tray/dock/`app.quit()` -> `window-all-closed` -> `closeAll` secondary wiring. Header now says: asserts the `before-quit` *handler body*, NOT the quit-path *wiring*.

### Reverse-verify
- Stash `closeAll()` body -> close-window probe FAILS: `activeSessionCount=1 5s after before-quit`. Restored.
- Stash `close(sessionId)` body -> delete-session probe FAILS: `activeSessionCount=1 5s after deleteSession`. Restored.
- Inflate `selfExitsSinceStart()` to simulate a self-exit -> false-pass guard FAILS: `selfExitCount went 1 -> 2 during the before-quit poll window`. Restored.

### 3-round stability
- `probe-e2e-close-window-aborts-sessions`: 3/3
- `probe-e2e-delete-session-kills-process`: 3/3

## Task B (#179)

Pool-3 worker flagged that `scripts/probe-spawn-error-propagation.mjs` calls `saveState('claudeBinPath', ...)`. Investigation:

1. **Probe is not enrolled.** `scripts/run-all-e2e.mjs` discovers via the `probe-e2e-*` glob; this probe is named `probe-spawn-*`. Not in `package.json` either. It's an orphan probe.
2. **Probe is broken post-#282.** The `claudeBinPath` IPC still writes to the kv store but `main.ts` no longer reads it (PR-I severed the wiring). With PATH='' and CCSM_CLAUDE_BIN deleted, the resolver falls through to `where claude` and returns `CLAUDE_NOT_FOUND` instead of the expected `CLI_SPAWN_FAILED`. Confirmed by direct run.
3. **Decision: (b) delete the probe.** It can't test what it claims, and it's not running anywhere.

A future replacement needs a binary-override shape that the post-#282 resolver still accepts (`CCSM_CLAUDE_BIN` pointing at a real `.exe` wrapper — a `.cmd` or `.js` won't survive the SDK's Node spawn path). Out of scope here.

## Files
- `electron/agent/manager.ts` — +`selfExitCount` field, +`selfExitsSinceStart()`
- `electron/main.ts` — expose `selfExitCount` via `__ccsmDebug`
- `scripts/probe-e2e-close-window-aborts-sessions.mjs` — header rewording, false-pass guard
- `scripts/probe-e2e-delete-session-kills-process.mjs` — header note, false-pass guard
- `scripts/probe-spawn-error-propagation.mjs` — DELETED

## Test plan
- [x] Close-window probe passes 3/3
- [x] Delete-session probe passes 3/3
- [x] Reverse-verify: stash close handler -> probe FAILS
- [x] Reverse-verify: false-pass guard fires when self-exit simulated
- [x] Vitest electron suite: no new failures (db-hardening pre-existing on origin/working)

Generated with Claude Code.


Jiahui-Gu:
LGTM — reviewer pass complete on review-315 worktree.

**Layer 1 (necessity / direction)**
- **selfExitCount design — sound.** Increment-then-delete in the `onExit` callback is guarded by `if (this.runners.has(sessionId))`. Both `close(sid)` (sync `r.close()` + sync `runners.delete`) and `closeAll()` (sync `r.close()` loop + sync `runners.clear()`) remove from the map *before* any async exit handler can fire. Verified `SdkSessionRunner.close()` only triggers `abort.abort()` / `query.close()` synchronously; the `onExit` is fired from the consumer IIFE (`electron/agent-sdk/sessions.ts:391`) in a microtask, so the manager's delete always lands first. Counter only counts unsolicited exits — invariant holds.
- **Debug expose gated.** `__ccsmDebug` block in `electron/main.ts:1260` is wrapped in `if (!app.isPackaged)`, so prod bundles never expose `selfExitCount()`. Comment at line 1259 calls this out explicitly.
- **Probe deletion justified.** `git log` shows the probe was last meaningfully touched in #235 (Feb), orphaned by PR-C / #273 / #282 since. Confirmed `scripts/run-all-e2e.mjs:1` glob is `probe-e2e-*` (not `probe-*`) — orphan probe never ran. `git grep probe-spawn` returns only the probe itself + an unrelated reference in `probe-e2e-import-session.mjs` (different name match).

**Layer 2 (implementation)**
- **3-round stability:** close-window 3/3, delete-session 3/3 (real CLI on PATH).
- **Reverse-verify closeAll stash:** stashed `closeAll()` body → close-window probe FAIL `activeSessionCount=1 5s after before-quit; expected 0 — closeAll regression`. Restored.
- **Reverse-verify close() stash:** stashed `close(sid)` body → delete-session probe FAIL `activeSessionCount=1 5s after deleteSession; expected 0 — zombie regression`. Restored.
- **Reverse-verify false-pass guard:** added a temporary `__debugSpike()` mutator + a single `__ccsmDebug.__debugSpike()` call in the close-window probe between `before-quit` emit and the poll → probe FAIL `selfExitCount went 0 -> 1 during the before-quit poll window. The CLI self-exited; the close handler may not have run. Cannot distinguish handler-driven teardown from incidental CLI death — false pass.` Restored. The new guard does fire when (and only when) the counter moves.
- **Vitest:** 75 files / 811 tests passing. 3 failures all `electron/db.test.ts` from pre-existing `better-sqlite3` `NODE_MODULE_VERSION` mismatch on `origin/working` — unrelated to this PR.

**Collision check (PR #289 fix(thinking) on `working`)**
- `electron/agent/manager.ts`: #289 hunks at lines 20 (Runner iface) and 215 (SessionsManager body). #315 hunks at 45-55 (selfExitCount field) and 108-110 (`selfExitsSinceStart`) and 142 (onExit branch). Different regions but same class — auto-rebase should resolve; if not, trivial.
- `electron/main.ts`: #289 hunk at line 990 (whenReady), #315 hunk at line 1273 (`__ccsmDebug` block). Far apart — clean.
- No blocking collision. Whichever lands second rebases cleanly.

**Concerns:** none blocking. Minor nit: the guard increment in `onExit` runs after the `agent:exit` emit; if a future refactor moves the `delete` ahead of the emit, the invariant still holds, but worth a comment if anyone touches that ordering. Not a fix-needed for this PR.

Reviewed on review-315 worktree at HEAD 7c93c77 rebased on origin/working (no-op rebase, branch was already current).

PR #320, opened 2026-04-26
fix(e2e): scrub composer draft in harness-perm seedSession

Jiahui-Gu:
## Root cause (category a — probe-broken / harness state pollution)

`permission-a11y` and `permission-partial-accept` intermittently failed
waiting for `data-perm-action=reject` to receive focus on the **second**
injected prompt within the same case (2000ms timeout). The first prompt's
auto-focus assertion always passed; the second's never landed.

### Evidence

Diagnostic logs added to `PermissionPromptBlock`'s auto-focus useEffect
(stored on `window.__diagPerm` since the harness doesn't pipe console):

```
[1st prompt] rAF active=BODY → CALLING_focus → POST_focus active=BUTTON dataPerm=reject  ✓
[2nd prompt] rAF active=TEXTAREA dataInputBar=true
             isTextEntry isComposer=true
             composer_value="n"
             EARLY_RETURN_value_nonempty
```

The composer textarea contained `"n"` BEFORE the second prompt mounted —
no harness keypress in this case typed it. Tracing back: drafts
(`src/stores/drafts.ts`) persist to `app_state.drafts` on every keystroke
(250ms debounce). A stale `"n"` left over from a previous full harness
run (likely typed by `casePermissionPrompt`'s `keyboard.press('n')`
during the focus race the comment in the source code already documents)
hydrates into the module-scope cache on app boot.

`resetBetweenCases` wipes `app_state` AFTER each case — fine for full
runs after case #1, but for `--only=<id>` runs no prior case has run, so
case #1 starts with the polluted draft. Even within a full run, the FIRST
case to use sid `s-perm` after a polluted launch sees it.

PermissionPromptBlock's auto-focus has an intentional exception clause
(documented at length): when the composer is focused with non-empty
content, don't steal focus — user is mid-typing; reject is one click
away. This is correct product behavior. Sequence:
1. Prompt #1 mounts: composer not focused → fast path → Reject focuses ✓
2. User Esc/clicks Reject → `resolvePermission` bumps `focusInputNonce`
   → composer regains focus, hydrated draft `"n"` visible
3. Prompt #2 mounts: composer focused with `"n"` → exception clause
   bails → Reject never focused → harness times out

Bumping the timeout to 5s did NOT change the outcome — composer stays
focused indefinitely, confirming this is not a render race.

## What changed

**Harness only** — no product changes.

`scripts/harness-perm.mjs` `seedSession` now, after seeding the store:
1. Calls `.fill('')` on the composer textarea — React-aware, fires the
   onChange path so both the controlled `value` AND the draft cache
   (via `setDraft`) clear.
2. Blurs the active element so prompt #1's auto-focus also uses the
   BODY fast path (defensive — prompt #1 wasn't failing, but consistent
   start state is cheap).

No product code touched. The `PermissionPromptBlock` auto-focus exception
clause is correct and stays as-is.

## Reverse-verify

A tiny `_pollute.mjs` script launches Electron, seeds `s-perm`, types
`"n"` into the composer, waits for the 250ms debounce, and closes (so
the draft gets persisted to `app_state.drafts`).

| Step | Result |
|---|---|
| Stash fix → run polluted | both cases FAIL at 2000ms timeout |
| Restore fix → run polluted | both cases PASS |

Output snippets in commit message.

## 2-round stability (12/12 each)

```
=== ROUND 1 ===
=== totals: 12 passed, 0 failed, 12 total ===
=== ROUND 2 ===
=== totals: 12 passed, 0 failed, 12 total ===
```

## Test plan

- [x] `node scripts/harness-perm.mjs --only=permission-a11y` passes
- [x] `node scripts/harness-perm.mjs --only=permission-partial-accept` passes
- [x] Full harness 12/12 round 1
- [x] Full harness 12/12 round 2
- [x] Reverse-verify with polluted draft


Jiahui-Gu:
## Reviewer report — REQUEST_CHANGES

**Layer 1 (necessity / approach):** Diff is clean (only `scripts/harness-perm.mjs`, no product code). Worker's root-cause analysis of the `PermissionPromptBlock` early-return clause is correct (verified at `src/components/PermissionPromptBlock.tsx:150-159` — composer + non-empty value → bail). Draft persistence to `app_state.drafts` confirmed in `src/stores/drafts.ts`.

**Layer 2 (verification):** This is where the PR fails.

### Bug NOT reproduced on baseline
- `origin/working` (a808af4): `node scripts/harness-perm.mjs` ran twice back-to-back → **12/12 both rounds**, including `permission-a11y` and `permission-partial-accept`. Not a flake the PR needs to fix in normal harness operation.

### PR introduces a regression
On `pr-320` I ran the full harness 5 times:
- Round 1: 12/12
- Round 2: **11/12** — `permission-prompt` failed (`expected Reject focused; got [...all focused=false]`)
- Round 3: **11/12** — `permission-prompt` failed (`autoFocus never moved to Reject before 2nd Y press`)
- Round 4: 12/12
- Round 5: **11/12** — `permission-prompt` failed (`expected Reject focused`)

→ ~50% failure rate of `permission-prompt` on PR 320. **0% on baseline.**

### What I think is happening
The new `seedSession` scrub does:
1. `await ta.fill('')`
2. `await win.evaluate(() => document.activeElement?.blur())`

Both run AFTER `setState({ activeSessionId, focusInputNonce })`. The composer auto-focus reaction to the nonce bump may race with the `.fill('')`/`blur()` calls; sometimes the prompt mounts while the textarea is still focused (and the `.fill('')` left focus there), tripping the SAME composer-focused exception this PR is trying to side-step. The "fix" introduces a new race window.

Also: even if the scrub at case start is reliable, `casePermissionPrompt` itself presses `n` between prompts (line 120), which can hydrate into the composer if focus has shifted — repopulating exactly the draft state the PR claims to neutralize. The scrub doesn't run between sub-prompts within a case.

### Concerns
1. **Regression on a previously-stable case.** Net effect of this PR is more flakiness, not less.
2. **Original symptom not reproducible on baseline** — worker's `_pollute.mjs` synthesized a scenario that does not occur in real harness runs. Fixing a synthetic bug at the cost of a real regression is a poor trade.
3. The right place to address the `data-input-bar` focus race (if the team thinks it's worth addressing at all) is probably the production code: either (a) `resolvePermission` shouldn't bump `focusInputNonce` if a permission prompt is about to mount on the next render, or (b) the composer's empty-check exception should be tightened (e.g. only honor it for IME / settings textareas, not the chat composer when a prompt is in flight). But per Layer 1 review, baseline is fine — there may be no bug to fix at all.

### Suggested next step
Close this PR. If team still wants the seedSession draft scrub as defensive cleanup, redo it as: scrub draft DB row directly via `window.ccsm.saveState('drafts', ...)` BEFORE `setState({ focusInputNonce })`, and skip the `.fill('')`/`blur()` DOM dance entirely.

```
VERDICT: REQUEST_CHANGES
LAYER_1_HOLDS: yes — root cause correctly identified, but bug not real on baseline
DIFF_SCOPE_CLEAN: yes
ROOT_CAUSE_VERIFIED: yes — src/components/PermissionPromptBlock.tsx:150-159
DRAFT_PERSIST_CONFIRMED: yes
BUG_REPRODUCED_ON_BASELINE: no — baseline 12/12 across 2 rounds
2_ROUND_STABILITY_PR: round1=12/12, round2=11/12 (permission-prompt fail)
ADDITIONAL_PR_ROUNDS: 12/12, 11/12, 12/12, 11/12 — ~50% perm-prompt failure
NO_PRODUCT_CODE_TOUCHED: yes
```

Jiahui-Gu:
Closing per reviewer (REQUEST_CHANGES): this PR introduces new flakiness on permission-prompt (~50% failure rate on PR vs stable on baseline), so net effect is more flake, not less.

If we still want defensive draft cleanup, do it via direct `window.ccsm.saveState('drafts', '{"version":1,"drafts":{}}')` *before* the `setState({ focusInputNonce })` call in seedSession(), instead of going through composer .fill('') + blur. Will redispatch with that approach.

PR #319, opened 2026-04-26
fix(e2e): repair agent diagnostic + init-failure banner cases

Jiahui-Gu:
## Summary
Two harness-agent.mjs cases (`diagnostic-banner`, `init-failure-banner`) were red on `origin/working` (confirmed by both #318 worker and #318 reviewer). Selector timeouts only — no product regression.

## Root cause: probe-broken (category a)

PR #237 (`feat(chrome): unify banner trio behind shared TopBanner`) moved the rendered DOM from per-banner data attributes onto the shared `<TopBanner />`. The shared component emits:

- `data-testid={testId}` — NOT `data-agent-diagnostic-banner` / `data-agent-init-failed-banner`
- `data-variant={variant}` — NOT `data-severity`
- `data-top-banner-dismiss` — NOT `data-agent-diagnostic-dismiss`

Evidence:
- `grep -rn 'data-agent-diagnostic-banner\|data-agent-init-failed-banner' src/` → **zero matches** (attribute genuinely doesn't exist in source).
- `src/components/AgentDiagnosticBanner.tsx:43` and `AgentInitFailedBanner.tsx:65` both pass `testId="agent-…-banner"` to TopBanner.
- `src/components/chrome/TopBanner.tsx:217-219` emits `data-top-banner data-variant={variant} data-testid={testId}` on the motion.div, plus `data-top-banner-dismiss` on the dismiss button (line 255).
- Retry / Reconfigure data-attrs (`data-agent-init-failed-retry`, `data-agent-init-failed-reconfigure`) live on inner `<TopBannerAction />` children and are still correct — left untouched.

## What changed
Harness only. Six selector replacements in `scripts/harness-agent.mjs`:

| Old | New |
| --- | --- |
| `[data-agent-diagnostic-banner]` (×4) | `[data-testid="agent-diagnostic-banner"]` |
| `[data-agent-init-failed-banner]` (×4) | `[data-testid="agent-init-failed-banner"]` |
| `getAttribute('data-severity')` | `getAttribute('data-variant')` |
| `[data-agent-diagnostic-dismiss]` | `[data-testid="agent-diagnostic-banner"] [data-top-banner-dismiss]` |

No product code touched. Banner components, store actions (`pushDiagnostic`, `setSessionInitFailure`, `clearSessionInitFailure`, `dismissDiagnostic`), and lifecycle wiring all untouched.

## Reverse-verify
```
$ git stash push scripts/harness-agent.mjs
[...]

PR #323, opened 2026-04-26
fix(e2e): stabilize notify-integration case in harness-agent

Jiahui-Gu:
## Summary
Fixes a ~1-in-3 flake in the `notify-integration` e2e case. The case lives in `scripts/probe-e2e-notify-integration.mjs` (standalone probe), NOT in `scripts/harness-agent.mjs` — the harness has zero "notify" references. (Title kept per task brief; scope clarified here.)

## Root cause
`BrowserWindow.blur()` is asynchronous from the OS's perspective (notably Windows). The probe drove `window.ccsm.notify(...)` immediately after `w.blur()`, but `electron/notifications.ts:103` checks `shouldSuppressForFocus()` and silently drops the call when any window is still focused. Race manifested as `timeout waiting for notifyPermission call` in step 4.

## Fix
Extract a `blurAndWaitUnfocused(app)` helper that:
1. Calls `w.blur()` on every BrowserWindow.
2. Polls `BrowserWindow.isFocused()` for up to 3s until focus drops.
3. Falls back to `w.hide()` for the stubborn launch-time case (some headless / playwright environments hold focus even after explicit blur). Step 7 (focus-suppression assertion) explicitly calls `w.show(); w.focus();` before reading `shouldSuppressForFocus()`, so the hide fallback is recoverable.

Applied at all 4 blur sites (steps 3, 9, 10c, 11). No product code touched.

## Reverse-verify
Reproduced flake on `origin/working` HEAD (before this fix): run 1 OK, run 2 OK, run 3 FAIL with `timeout waiting for notifyPermission call`.

## 3-round green proof (post-rebase onto origin/working including #319)
```
=== RUN 1 ===
[probe-e2e-notify-integration] toast-action reject cancellation OK
[probe-e2e-notify-integration] rich done payload OK
[probe-e2e-notify-integration] OK
=== RUN 2 ===
[probe-e2e-notify-integration] toast-action reject cancellation OK
[probe-e2e-notify-integration] rich done payload OK
[probe-e2e-notify-integration] OK
=== RUN 3 ===
[probe-e2e-notify-integration] toast-action reject cancellation OK
[probe-e2e-notify-integration] rich done payload OK
[probe-e2e-notify-integration] OK
```
Plus 5 back-to-back green pre-rebase. Cumulative: 8/8 green, zero flakes after fix.

## Test plan
- [x] Reproduce flake on unmodified `origin/working`
- [x] Apply fix
- [x] 3 back-to-back full-probe runs all green
- [x] Rebase onto latest `origin/working` (PR #319 merged) — fast-forward, no conflicts
- [x] 3 more back-to-back runs post-rebase all green

### DIFF
--- a/scripts/probe-e2e-notify-integration.mjs
+++ b/scripts/probe-e2e-notify-integration.mjs
@@ -50,6 +50,40 @@ function fail(msg) {
   process.exit(1);
 }
 
+// FLAKE FIX: `BrowserWindow.blur()` is asynchronous from the OS's perspective
+// (especially on Windows). If the probe drives `window.ccsm.notify(...)`
+// immediately after `w.blur()`, the IPC handler in `electron/notifications.ts`
+// can still see `shouldSuppressForFocus() === true` (line 103) and silently
+// drop the call — manifesting as "timeout waiting for notify* call". We blur
+// every window then poll until focus is actually gone, with a hide() fallback
+// for the stubborn launch-time case.
+async function blurAndWaitUnfocused(app, { allowHideFallback = true } = {}) {
+  await app.evaluate(async ({ BrowserWindow }, allowHide) => {
+    const stillFocused = () => BrowserWindow.getAllWindows()
+      .some((w) => !w.isDestroyed() && w.isFocused() && w.isVisible());
+    for (const w of BrowserWindow.getAllWindows()) {
+      try { w.blur(); } catch {}
+    }
+    const deadline = Date.now() + 3000;
+    while (Date.now() < deadline && stillFocused()) {
+      await new Promise((r) => setTimeout(r, 50));
+    }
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (cannot self-approve). Reverse-verify decisive: baseline 4/4 fail at step-9 blur ("timeout waiting for initial question + retry schedule"), PR fix 3/3 pass. Helper correct (50ms poll, 3s deadline, safe hide() fallback since later steps explicitly w.show()+w.focus()). Diagnosis matches electron/notifications.ts:103 shouldSuppressForFocus() gate. Layer-1 follow-ups (notifications.ts should log/queue dropped calls; probe could fold into harness-agent per #164 scope) noted, out of scope for flake fix.

PR #329, opened 2026-04-26
ci: enforce main only accepts PRs from working

Jiahui-Gu:
## Summary
GitHub branch protection cannot restrict PR source branch — only destination. To enforce that main only accepts merges from working (release-branch policy), this workflow fails any PR to main whose head ref is not `working`.

After this lands, the next step is to make the `check-source` job a required status check in main's branch protection rules.

## Test plan
- [x] Local YAML parses; mirrors existing workflow structure (release.yml, ci.yml, pr-no-silent-drops.yml)
- [ ] After merge: open a test PR from a non-working branch to main → expect this check to fail

### DIFF
--- a/.github/workflows/main-source-guard.yml
+++ b/.github/workflows/main-source-guard.yml
@@ -1,0 +1,17 @@
+name: main source guard
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  check-source:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Reject non-working source
+        run: |
+          if [ "${{ github.head_ref }}" != "working" ]; then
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (reviewer cannot self-approve; manager: please merge). Workflow correctly enforces main<-working policy: triggers only on PRs to main, exact-match check on github.head_ref, fast metadata-only job, no checkout needed. Fork PRs still populate head_ref so unbypassable. Branch names like 'working/foo' correctly rejected. Theoretical shell interpolation of head_ref is harmless given GitHub ref-name restrictions. Reminder: after this lands on main via working, add 'check-source' as a required status check on main's branch protection.

PR #326, opened 2026-04-26
test(e2e): add harness-agent cases for SDK adapter (PR #271)

Jiahui-Gu:
## Summary

Adds three new harness-agent.mjs cases that exercise the SDK-adapter -> renderer contract end-to-end. Closes the e2e gap left when [PR #271](https://github.com/Jiahui-Gu/ccsm/pull/271) (SDK adapter PR-A) was merged behind a feature flag with e2e deferred until everything was green. The flag has since been removed (PR #277 / PR-C), so SdkSessionRunner is now the only runner — these cases lock in its renderer-facing contract.

Each case uses `BrowserWindow.getAllWindows()[0].webContents.send(...)` in `app.evaluate` to inject real `agent:event` / `agent:exit` IPC frames — the same channel `SdkSessionRunner` -> `manager.ts` -> renderer uses in production. This is the only honest way to test the adapter from the renderer-only harness; the alternative (mutating store directly) would not exercise lifecycle.subscribeAgentEvents at all.

## Cases added

- **sdk-stream-roundtrip** — Inject the three big-bucket SDK frames that `translateSdkMessage` passes through unchanged: `system/init` -> `assistant` (text content) -> `result/success`. Asserts the assistant block renders with the expected text AND `runningSessions` is cleared on the result frame. User-observable: replies show up + send button re-enables after a turn.
- **sdk-stream-event-partial** — Inject the `stream_event` frames carrying `content_block_delta(text_delta)` payloads that PR #271's `includePartialMessages: true` SDK option produces. Asserts the streaming caret (`span.animate-pulse`) appears AND three deltas coalesce into one streaming block with the full text. User-observable: long replies appear progressively, not frozen until done.
- **sdk-exit-error-surfaces** — Inject `agent:exit({ error: 'PROBE_SDK_FAILURE...' })` simulating SdkSessionRunner's consumer-loop catch (sessions.ts:418-429). Asserts an `error` block lands in the chat AND `runningSessions` is cleared. User-observable: SDK iterator errors don't disappear silently — user sees the failure.

## Test plan

- [x] `npm run build` (rebuild bundle.js so harness sees the latest renderer)
- [x] `node scripts/harness-agent.mjs` round 1: 20/20 passed (17 existing + 3 new)
- [x] `node scripts/harness-agent.mjs` round 2: 20/20 passed
- [x] Branch rebased onto latest `origin/working` (incl. #325 stale-bundle gate); both rounds re-run green post-rebase.

```
[...]

PR #331, opened 2026-04-26
test(e2e): assert AgentDiagnosticBanner is empty after happy-path streaming

Jiahui-Gu:
## Summary
Extend the `streaming` case in `scripts/harness-agent.mjs` (task #88) with a baseline-diagnostics-empty dual assertion that locks in the contract: a happy-path agent run must not push any non-dismissed diagnostics.

- **DOM check**: `[data-testid="agent-diagnostic-banner"]` count === 0 after the streaming turn finalizes (caret cleared).
- **Store check**: `state.diagnostics.filter(d => d.sessionId === sid && !d.dismissed).length === 0`.

The store's `setRunning` / `streamAssistantText` / `appendBlocks` do not auto-clear diagnostics, so any stray push (info-level "turn complete" notice, leaked warning, etc.) would persist across turns. This catches such leaks at their source.

Closes #88. Plan: `docs/task-88-draft.md` (per worker brief).

### Two-round green proof (`node scripts/harness-agent.mjs`)

Note: this branch is from before PR #326 lands, so the suite has 17 cases (not 20/23).

**Round 1**
```
=== totals: 17 passed, 0 failed, 17 total ===
  [OK] streaming                                  1285ms
  ... (all 17 OK)
```

**Round 2**
```
=== totals: 17 passed, 0 failed, 17 total ===
  [OK] streaming                                  2973ms
  ... (all 17 OK)
```

### Reverse-verify (the assertion is not a no-op)

Temporarily injected a fake diagnostic via `pushDiagnostic({ sessionId, level: 'warn', code: 'reverse_verify', ... })` immediately before the new assertion, then ran `--only=streaming`:

```
[case=streaming] FAIL: baseline diagnostics should be empty after happy-path streaming; banner count=1
```

The DOM check correctly fired (banner count went from 0 to 1). Injection was then removed and the case returned to green (Round 2 above).

### Production-bug findings
None. The happy-path codepath leaves diagnostics empty as expected. No bonus bug to file.

## Test plan
- [x] `npm run build` (passes #325 stale-bundle gate)
- [x] `node scripts/harness-agent.mjs` 2 rounds, all 17 cases green
- [x] Reverse-verify: injected fake diagnostic causes the assertion to fail with a clear message; removed cleanly

### DIFF
--- a/scripts/harness-agent.mjs
+++ b/scripts/harness-agent.mjs
@@ -273,7 +273,28 @@ async function caseStreaming({ win, log }) {
   const finalCaretCount = await win.locator('span.animate-pulse').count();
   if (finalCaretCount !== 0) throw new Error(`caret should disappear after finalize; found ${finalCaretCount}`);
 
-  log('3 deltas coalesced into 1 block; caret shown then hidden on finalize');
+  // Baseline-diagnostics-empty contract (task #88): a happy-path streaming
+  // turn must NOT push any non-dismissed diagnostics for the active session.
+  // The store's lifecycle (setRunning, streamAssistantText, appendBlocks)
+  // does not auto-clear diagnostics, so any leak here would persist across
+  // turns. Dual check: DOM (banner not rendered) AND store (no active
+  // entries scoped to this session). Brief settle for any TopBanner
+  // <AnimatePresence> exit transition.
+  await win.waitForTimeout(250);
+  const baselineBannerCount = await win.locator('[data-testid="agent-diagnostic-banner"]').count();
+  if (baselineBannerCount !== 0) {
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (approve-equivalent; cannot self-approve). Verified post-rebase on origin/working (be8b388): clean rebase, +22/-1 in caseStreaming only. Field names (sessionId, dismissed) and testid (agent-diagnostic-banner) match src/stores/store.ts:146-153 and src/components/AgentDiagnosticBanner.tsx:32,43. Filter logic in the assertion mirrors the banner's render filter exactly. Ran harness twice end-to-end: 20/20 both rounds (17 baseline + 3 SDK from #326). 250ms settle is a fine defensive guard against late agent:diagnostic IPC; no deterministic alternative since the assertion itself is the 'is empty' check.

PR #341, opened 2026-04-26
fix(agent): agent self-reports as VS Code integration; force standalone identity

Jiahui-Gu:
## Symptom

User asks the agent inside ccsm dev: 'who are you / where are you running'. Agent answers something along the lines of:

> Running environment: Claude Code (VS Code integration)

ccsm is an independent Electron desktop app. It is never a VS Code extension. The agent's self-reported identity was wrong, and as a side effect the CLI was offering IDE-only tools (editor tabs, diagnostics) that the ccsm renderer has no UI for.

## Root cause (verified upstream)

The bundled CLI shipped inside `@anthropic-ai/claude-agent-sdk-win32-x64/claude.exe` runs an IDE-companion auto-discovery pass on startup. It scans `${CLAUDE_CONFIG_DIR}/ide/*.lock` — the official Claude Code IDE extensions (VS Code, JetBrains, Cursor) drop a lockfile there containing `ideName`, `workspaceFolders`, `pid`, `authToken`. When a lockfile's `workspaceFolders` overlap the session cwd, the CLI binds to that IDE companion and the agent's system context flips to 'IDE-integrated'.

Reproduction on the reporter's machine:

```
$ cat ~/.claude/ide/46190.lock
{'pid':20192,'workspaceFolders':['c:\\Users\\jiahuigu\\Desktop'],'ideName':'Visual Studio Code','transport':'ws','runningInWindows':true,'authToken':'...'}
```

(The user runs the official Claude Code VS Code extension on the side; that extension drops this file. Any ccsm user who has ever installed the extension will hit this.)

Verified the gate by scanning the bundled binary (powershell + ASCII string search; full strings present at known offsets):

```
(autoConnectIde || $ || sL() || CLAUDE_CODE_SSE_PORT || K
   || CLAUDE_CODE_AUTO_CONNECT_IDE)
  && !a7(CLAUDE_CODE_AUTO_CONNECT_IDE)
```

- `a7(env)` is the bundle's 'is-false-string' helper (matches `false` / `0` / `no`).
- Setting `CLAUDE_CODE_AUTO_CONNECT_IDE=false` makes `a7()` true → the trailing `&& !a7(...)` short-circuits the whole condition, regardless of the user's settings.json `autoConnectIde`, an active `CLAUDE_CODE_SSE_PORT`, or any heuristic that would otherwise force an attach.

The settings.json default is `autoConnectIde: false`, but that alone is not enough — `sL()` (terminal heuristic), `CLAUDE_CODE_SSE_PORT`, or other inputs can still flip the gate. The env-var kill switch is the only deterministic way to opt out for an Electron host.

## Fix

`buildSdkEnv` in `electron/agent-sdk/sessions.ts` now sets `CLAUDE_CODE_AUTO_CONNECT_IDE='false'` alongside the existing `CLAUDE_CODE_ENTRYPOINT='ccsm-desktop'` / `CLAUDE_AGENT_SDK_CLIENT_APP='ccsm-desktop/0.1.0'` markers. ccsm is never an IDE extension; standalone identity is the right default at all times.

15-line change; comment in source explains the gate so future readers don't second-guess it.

## Test

New unit case in `electron/agent-sdk/__tests__/sessions.test.ts`:

> 'disables CLI IDE auto-connect so VS Code lockfiles do not bind to ccsm sessions'

Asserts the env var lands on `options.env` and that the existing entrypoint markers still do too. Full file: 24/24 tests pass; tsc clean.

No new e2e probe — the bug lives entirely in spawn-env construction, which a unit test on `buildSdkEnv` covers exactly. An e2e probe would have to (a) drop a fake VS Code lockfile into `~/.claude/ide/`, (b) spawn the real CLI, (c) inspect the system-init frame for IDE attachment metadata. That's brittle (depends on the host machine never racing with a real extension) and adds 30+ s for no extra coverage over the unit test on the env-construction path. Reviewer can manually re-verify by running ccsm dev with a VS Code extension lockfile present and asking the agent 'where are you running' — answer should not contain 'VS Code' / 'integration'.

## Hot-file note

Edits `electron/agent-sdk/sessions.ts`. Sister worker `fix-askuserquestion-no-ui` also touches that file but only inside `handleCanUseTool` / `makePreToolUseHook` (lines ~514-607). My edit is in `buildSdkEnv` (lines 137-189). No textual overlap; clean rebase expected.

### DIFF
--- a/electron/agent-sdk/__tests__/sessions.test.ts
+++ b/electron/agent-sdk/__tests__/sessions.test.ts
@@ -126,6 +126,25 @@ describe('agent-sdk/SdkSessionRunner', () => {
     runner.close();
   });
 
+  it('disables CLI IDE auto-connect so VS Code lockfiles do not bind to ccsm sessions', async () => {
+    // Regression: with a VS Code Claude Code extension running, a lockfile
+    // at $CLAUDE_CONFIG_DIR/ide/<pid>.lock advertises ideName="Visual
+    // Studio Code" + workspaceFolders. The bundled CLI auto-connects when
+    // a lockfile's workspaceFolders include the session cwd, and the agent
+    // then identifies itself as "Claude Code (VS Code integration)" — wrong
+    // for ccsm, an independent Electron app. Setting
+    // CLAUDE_CODE_AUTO_CONNECT_IDE=false trips the bundled-CLI kill-switch
+    // (`!a7(env)` in the IDE-attach gate) and forces standalone identity.
+    const runner = new SdkSessionRunner('s2b', noop, noop, noop, noop);
+    await runner.start(baseStart);
+    const opts = fake.getOptions()?.options ?? {};
+    const env = (opts.env ?? {}) as Record<string, string>;
+    expect(env.CLAUDE_CODE_AUTO_CONNECT_IDE).toBe('false');
+    expect(env.CLAUDE_CODE_ENTRYPOINT).toBe('ccsm-desktop');
+    expect(env.CLAUDE_AGENT_SDK_CLIENT_APP).toBe('ccsm-desktop/0.1.0');
+    runner.close();
+  });
+
   it('emits translated SDK messages to onEvent and captures cliSessionId from init', async () => {
     const events: unknown[] = [];
     const runner = new SdkSessionRunner('s3', (e) => events.push(e), noop, noop, noop);
--- a/electron/agent-sdk/sessions.ts
+++ b/electron/agent-sdk/sessions.ts
@@ -178,6 +178,21 @@ function buildSdkEnv(opts: {
   env.CLAUDE_CONFIG_DIR = opts.configDir;
   env.CLAUDE_CODE_ENTRYPOINT = 'ccsm-desktop';
   env.CLAUDE_AGENT_SDK_CLIENT_APP = 'ccsm-desktop/0.1.0';
+  // Disable the CLI's IDE-companion auto-discovery. Without this, the
+  // bundled CLI scans `${CLAUDE_CONFIG_DIR}/ide/*.lock` (dropped there by
+  // the VS Code / JetBrains / Cursor Claude Code extension) and, when a
+  // lockfile's workspaceFolders match the session cwd, attaches itself as
+  // an IDE-integrated session — at which point the agent's system context
+  // identifies the run as "Claude Code (VS Code integration)" and the user
+  // gets editor-tab + diagnostics tools they have no UI for. ccsm is an
+  // independent Electron app, never a VS Code extension; we always want
+  // the standalone identity. Bundled-CLI gate is
+  // `(autoConnectIde || ... || CLAUDE_CODE_AUTO_CONNECT_IDE) && !a7(CLAUDE_CODE_AUTO_CONNECT_IDE)`
+  // — setting the env to a false-string makes `a7()` true, killing the
+  // whole condition regardless of the user's settings.json
+  // `autoConnectIde` value or any other heuristic. See PR fixing
+  // "agent self-reports as VS Code integration".
+  env.CLAUDE_CODE_AUTO_CONNECT_IDE = 'false';
   if (opts.envOverrides) {
     for (const [k, v] of Object.entries(opts.envOverrides)) {
       env[k] = v;
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -24,6 +24,7 @@
 
 import { spawn, spawnSync } from 'node:child_process';
 import { randomUUID } from 'node:crypto';
+import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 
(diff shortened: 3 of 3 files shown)


Jiahui-Gu (commented):
## Independent review verdict: REQUEST CHANGES (posted as comment because GH blocks request-changes on own PR)

### Layer 1 — necessity / upstream verification: PASS
- Confirmed `~/.claude/ide/46190.lock` exists on this machine with `ideName='Visual Studio Code'`. Repro path is real, not theoretical.
- Re-ran the binary string scan against the bundled `node_modules/@anthropic-ai/claude-agent-sdk-win32-x64/claude.exe`:
  - `CLAUDE_CODE_AUTO_CONNECT_IDE` × 5
  - `autoConnectIde` × 48
  - `CLAUDE_CODE_SSE_PORT` × 5
  - `ideName` × 41
  - `workspaceFolders` × 23

  Env name and gate are real and present in the version we ship.
- Settings.json `autoConnectIde:false` is genuinely insufficient — the OR-gate has multiple inputs (sL() / SSE_PORT / heuristics) before the env kill-switch. The env approach is the correct one.
- No collateral on hooks / MCP / agents: this env only suppresses the bundled CLI's IDE-companion auto-discovery (the `ide` MCP server attach for editor-tabs/diagnostics). Hooks, agents, and other MCP servers are SDK-side and unaffected. ccsm has no UI for editor-tab tools anyway, so suppressing is exactly correct.
- Placement after the generic env loop and before `envOverrides` lets per-session env override if ever needed. Good defensive design.

### Layer 2 — implementation: PASS
- 15-line diff is exactly the env assignment + a long comment documenting the upstream gate. No drive-by edits to other `buildSdkEnv` fields.
- Unit test asserts env lands on `options.env` and existing entrypoint markers still land. 24/24 `sessions.test.ts` pass locally (`HOME=/tmp/sanitized-home npx vitest run`).
- Comment is excellent — future readers won't second-guess the kill-switch.
- Hot-file note vs sister worker `fix-askuserquestion-no-ui` (lines 514-607) is accurate; no overlap with the 137-189 region.

### Layer 3 — e2e gate: FAIL, needs one harness-real-cli case
Per `feedback_e2e_before_merge.md` + `feedback_e2e_prefer_harness.md`. The unit test asserts the env *value* but not that the bundled CLI still *honors* the kill-switch. The regression class that matters — bundled SDK upgrade silently renames or hardens past the env — leaves the unit test green and reintroduces the user-visible "VS Code integration" identity bug. Worker's "30s for no extra coverage" framing under-weights this: the coverage *is* the regression net.

Cheapest fix: extend `scripts/harness-real-cli.mjs` (already spawns real claude with controlled env, already has scrub / timeout / `CCSM_REAL_CLI_REQUIRE` scaffolding). New case:
1. Build a tmp `CLAUDE_CONFIG_DIR` with `ide/<pid>.lock` carrying `ideName='Visual Studio Code'` + `workspaceFolders=[<tmp cwd>]`.
2. Spawn real claude with the buildSdkEnv-equivalent env (including `CLAUDE_CODE_AUTO_CONNECT_IDE=false`).
3. Send `initialize`, parse the system-init frame, assert no IDE attach (no `ideName`, no `ideClient`, no editor-tab tools advertised).

~60 LOC, well under the 30s/probe budget. After the case is added and green, will re-approve.

Jiahui-Gu:
Addressed the Layer 3 e2e gate (commit d65061f). New case in `scripts/harness-real-cli.mjs`: `ide-auto-connect-kill-switch-present`.

**Deviation from the literal request, with rationale:** the original ask was 'spawn real claude.exe with kill-switch + fake VS Code lockfile, assert no IDE attach in system/init'. I built that approach first, then verified empirically that it doesn't actually catch the regression class:

- Spawned the real CLI with a fake lockfile pointing at a bound TCP server, with `autoConnectIde:true` in settings.json AND `CLAUDE_CODE_SSE_PORT=<port>`. The CLI never opened the TCP connection, never populated `mcp_servers`. Real auto-attach requires the bundled CLI's full handshake against a live ws server with a matching authToken — which a clean dev machine without a real VS Code+Claude extension cannot provide.
- Consequence: the negative assertion `mcp_servers does not contain {name:'ide'}` is vacuously true even WITHOUT the kill-switch, so it wouldn't have detected an env-var rename.

Pivoted to a bundle-string scan: read `node_modules/@anthropic-ai/claude-agent-sdk-{platform}-{arch}/claude.exe`, assert the literal `CLAUDE_CODE_AUTO_CONNECT_IDE` is present (it's referenced as `process.env.CLAUDE_CODE_AUTO_CONNECT_IDE` at runtime, so the rename failure mode = the literal disappears from the bundle).

- Run cost: ~262ms (no spawn, just `fs.readFileSync` + `indexOf` on the 252 MB SEA bundle).
- Reverse-verified: tweaked the needle to a non-existent string locally; case fails with an actionable error pointing the maintainer at `buildSdkEnv` and asking them to re-scan for the current opt-out.
- Skips cleanly when the platform package isn't installed (e.g. CI matrix on a host without the win32-x64 bundle).

This trades 'real CLI was actually invoked' for 'directly tests the env-var-rename failure mode the reviewer called out'. Open to layering both in if you want, but the spawn-side test would be vacuously-passing scaffolding.

Full harness still runs in <3s (control-response-no-timeout 2.6s + new case 0.3s). PTAL.

Jiahui-Gu:
Re-review at d65061f6: harness case PASS (843ms, needle at offset 132267233 in 252 MB). Reverse-verify (needle mutated → FAIL with expected error → restored). Bundle-string approach justified by docstring (auto-attach gate needs live ws handshake; spawn-based negative assertion would vacuously pass). Merging.

PR #343, opened 2026-04-26
feat: remove non-functional density setting

Jiahui-Gu:
## Layer 1: grep audit
**Result: full deadcode.** The toggle wrote to `store.density`, App.tsx mirrored that to `html.density-{compact,normal,comfortable}`, global.css set `--density-scale` per class — but **no component / utility ever read `var(--density-scale)`**.

```
$ grep -rn 'var(--density-scale)\|\.density-row\|density-row' src/
src/styles/global.css:234:  (see `.density-row` below). The three stops are intentionally narrow —
src/App.tsx:131:  // .density-row / inline calc() — see global.css.
```

Both hits are stale **comments** referencing a `.density-row` class that was never defined and `var(--density-scale)` consumers that never landed. The whole pipeline (toggle → store → html class → CSS var) terminates in a dead variable. User's "页面不见调整" matches: the var changes, nothing observes it.

## Why delete instead of fix
- Per `feedback_design_density_polish.md`: ccsm intentionally targets one CLI-style density (block rendering + monospace + scannable). A density toggle is the wrong axis to expose.
- Implementing real density consumers across sidebar/chat/blocks would balloon the visual contract; that's a different design exercise, not a bug fix.
- User's instinct ("先去掉") matches the memory.

## Removed
| Layer        | File                                              | What                                                                 |
| ------------ | ------------------------------------------------- | -------------------------------------------------------------------- |
| UI           | `src/components/SettingsDialog.tsx`               | density `Field` + `Segmented`, store selector, setter                |
| State        | `src/stores/store.ts`                             | `Density` type, `density` state field, `setDensity`, `sanitizeDensity`, default, hydration line |
| Persist      | `src/stores/persist.ts`                           | `density` from `PERSISTED_KEYS` + `PersistedState`, `Density` import |
| Effect       | `src/App.tsx`                                     | density store selector + `useEffect` that toggled `html.density-*`   |
| CSS          | `src/styles/global.css`                           | `--density-scale` block + `html.density-{compact,normal,comfortable}` rules |
| i18n         | `src/i18n/locales/en.ts`, `src/i18n/locales/zh.ts`| `density`, `densityHint`, `densityOptions.*`                         |
| Tests        | `tests/appearance.test.ts`                        | `density sanitize` describe block + `sanitizeDensity` import         |
| Tests        | `tests/store-appearance.test.ts`                  | `setDensity` arm of combined test (kept theme arm)                   |
| Tests        | `tests/persist-shape.test.ts`                     | `density` from snapshot fixture + ALLOWED set                        |
| Tests        | `tests/persisted-keys-source-of-truth.test.ts`    | `density` from inline snapshot + fresh-value map                     |
| Harness      | `scripts/harness-ui.mjs`                          | `assertHasText '密度'` + `assertNotHasText 'Density'` in i18n-settings-zh |
| Probe        | `scripts/probe-appearance-polish.mjs`             | section 3 (Density) deleted; section numbers shifted                 |
| Comment      | `scripts/probe-helpers/reset-between-cases.mjs`   | example-list comment scrubbed                                        |

No backward-compat alias / deprecation comment (per task brief). Older persisted snapshots silently drop the key on hydrate.

## Verification
- [x] `npm run typecheck` — passes
- [x] vitest on touched tests (`tests/appearance.test.ts`, `tests/store-appearance.test.ts`, `tests/persist-shape.test.ts`, `tests/persisted-keys-source-of-truth.test.ts`) — 26/26 pass
- [x] `node scripts/harness-ui.mjs` — `i18n-settings-zh` passes (settings dialog no longer mentions 密度 / Density). 1 unrelated pre-existing flake (`terminal` ChatStream ToolBlock) not touched.

## Test plan
- [ ] Open settings → Appearance: confirm no Density field between Font Size and Window Tint
- [ ] Reload app: persisted state from older builds with `density: 'compact'` hydrates cleanly (key silently dropped)
- [ ] zh + en locale: settings dialog renders without missing-key warnings

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -1664,9 +1664,7 @@ async function caseI18nSettingsZh({ win, log, registerDispose }) {
   let txt = await paneText();
   assertHasText(txt, '主题', 'appearance');
   assertHasText(txt, '字号', 'appearance');
-  assertHasText(txt, '密度', 'appearance');
   assertNotHasText(txt, 'Theme', 'appearance');
-  assertNotHasText(txt, 'Density', 'appearance');
 
   // Notifications.
   await switchTab(/^通知$/);
--- a/scripts/probe-appearance-polish.mjs
+++ b/scripts/probe-appearance-polish.mjs
@@ -1,11 +1,10 @@
 // E2E: appearance polish probe.
 //
-// Exercises the four polish features without needing an API key:
+// Exercises the polish features without needing an API key:
 //   1. Theme toggle → verify <html> class flips (dark ⇄ theme-light)
 //   2. Font-size slider → verify --app-font-size CSS var updates
-//   3. Density segmented → verify html.density-* class updates
-//   4. Sidebar drag → verify sidebarWidthPct persists across reload
-//   5. Memory tab → create/edit CLAUDE.md in a temp cwd, verify file written
+//   3. Sidebar drag → verify sidebarWidthPct persists across reload
+//   4. Memory tab → create/edit CLAUDE.md in a temp cwd, verify file written
 //
 // Pure black-box where possible; uses store introspection only for the
 // persist-round-trip step (which otherwise needs a full app restart to see).
@@ -85,18 +84,7 @@ await win.keyboard.press('ArrowRight');
 await win.keyboard.press('ArrowRight');
 await win.waitForTimeout(150);
 
-// ── 3. Density ─────────────────────────────────────────────────────────────
-await win.getByRole('radio', { name: /^compact$/i }).click();
-await win.waitForTimeout(200);
-const isCompact = await win.evaluate(() =>
-  document.documentElement.classList.contains('density-compact')
-);
-if (!isCompact) fail('density-compact class not applied');
-ok('density → compact');
-await win.getByRole('radio', { name: /^normal$/i }).click();
-await win.waitForTimeout(150);
-
-// ── 4. Memory tab ──────────────────────────────────────────────────────────
+// ── 3. Memory tab ──────────────────────────────────────────────────────────
 // Close settings first; create a session pinned to tmpRepo so project memory is active.
 await win.keyboard.press('Escape');
 await win.waitForTimeout(200);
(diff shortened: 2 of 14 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE (posted as comment — gh blocks self-approve)

### Layer 1 — Necessity / dead-code judgment (verified)
Grepped `src/` and `electron/` for any consumer of `--density-scale` / `density-row` / `density-(compact|normal|comfortable)`:
- **Zero live consumers.** Only matches are in `dogfood-logs/*.html` (captured pre-removal snapshot), in docs prose using the English word "density" (e.g. "information density"), and one probe message string. No CSS, no component, no inline style reads `--density-scale`.
- Worker's "full deadcode" call is correct. Toggle → store → html class → CSS var, with nothing on the receiving end. Pure removal is the right move.

### Layer 2 — Removal completeness (verified)
- **Persist migration safe.** `loadPersisted()` does `JSON.parse(raw) as PersistedState` with no strict schema — extra fields from old user installs are simply ignored. `hydrateStore` no longer reads `density`. The serializer iterates `PERSISTED_KEYS` (now without `density`), so the next save drops the orphan field. Old users will not crash.
- **i18n deleted in both locales** (`en.ts`, `zh.ts`): `density`, `densityHint`, `densityOptions.{compact,normal,comfortable}` all gone, no orphan references.
- **CSS block deleted** (`global.css`); confirmed no leftover `.density-row` or `.density-*` in `src/`.
- **Type/setter/state slot removed** in `store.ts` (`Density`, `setDensity`, `sanitizeDensity`, default value, hydrate path).
- **Tests:** `npx tsc --noEmit` clean. `vitest run` on the four touched test files: **26/26 pass** (5.68s).
- **harness-ui i18n-settings-zh** assertion change is necessary and minimal — the "密度" label no longer exists in the Appearance pane, so the positive+negative assertions for it must go.

### Layer 3 — PR shape (verified)
- 14 files, **+7 / -108** — overwhelmingly deletions, additions are renumberings (`probe-appearance-polish.mjs`), comment updates (`reset-between-cases.mjs`), and a renamed test (`store-appearance.test.ts`).
- No backward-compat alias, no deprecation shim, no commented-out blocks. Matches user's "先去掉" directive exactly.
- Single commit, clean message.

LGTM. Safe to merge.

PR #346, opened 2026-04-26
fix(slash): hide plugin/skill/agent commands from picker (SDK can't run them)

Jiahui-Gu:
## Summary

User report (dogfood): `/plugin`, `/superpowers:brainstorm`, `/superpowers:write-plan`, `/pua:kpi` all behave broken — selecting them in the slash picker (or typing them manually) yields a model-text reply like "already deprecated, use the skill instead" or "unknown command" instead of actually running anything.

## Root cause

`electron/commands-loader.ts` scans five sources off disk:

1. `~/.claude/commands/` — user-level (`source: 'user'`)
2. `<cwd>/.claude/commands/` — project-level (`source: 'project'`)
3. `~/.claude/plugins/cache/<market>/<plugin>/<ver>/commands/` (`source: 'plugin'`)
4. `~/.claude/skills/` (`source: 'skill'`)
5. `~/.claude/{,project/}agents/` (`source: 'agent'`)

But the actual slash dispatch is a pass-through: `InputBar.send()` → `dispatchSlashCommand` returns `'pass-through'` → the literal `/<name> <args>` text is forwarded as a plain SDKUserMessage (`electron/agent-sdk/sessions.ts:466-478`).

The `@anthropic-ai/claude-agent-sdk` transport does NOT auto-execute CLI plugins / skills / subagents the way the CLI's REPL does:

- **plugin** — SDK requires explicit `query({ options: { plugins: SdkPluginConfig[] } })` (see node_modules sdk.d.ts:1471–1485). ccsm doesn't pass that field, so plugin commands never get installed; the slash text falls through to the model, which (correctly!) knows `/superpowers:brainstorm` is the deprecated stub of the `brainstorming` skill and replies as such.
- **skill** — skills are description-keyed system-prompt extensions invoked by the agent itself, never via `/skill-name`. Listing them in the picker created a false affordance.
- **agent** — subagent definitions reached via the Task tool, not by the user. Same false-affordance issue.

So three of the five loader sources have been surfacing un-runnable entries, and the model's polite "deprecated, use the skill" replies are the symptom users were filing.

## Decision

**Hide plugin / skill / agent in the picker** rather than:

- (b) grey them out — still surfaces names the user can't act on
- (c) wire SDK plugin/skill execution — non-trivial (`SdkPluginConfig` is local-path only, would need to enumerate `~/.claude/plugins/cache` paths and validate each), out of dogfood scope. Re-enable in a follow-up that actually plumbs `plugins` / `skills` through to `query()`.

Loader-level filter (vs. picker-level) so the contract is unit-testable without standing up the renderer or Electron.

## Implementation

- New `loadPickerCommands(opts)` in `electron/commands-loader.ts` — thin wrapper around `loadCommands` that filters down to `PICKER_VISIBLE_SOURCES = {user, project}`.
- Loader honors `process.env.CLAUDE_CONFIG_DIR` (review fix). ccsm sets this for the `claude` binary so the picker and the binary read the same tree (project memory `project_cli_config_reuse`); the prior hard-coded `os.homedir()` silently desynced them. Resolution order: `opts.homeDir` (test override) > `process.env.CLAUDE_CONFIG_DIR` > `os.homedir()/.claude`.
- `commands:list` IPC in `electron/main.ts` switched from `loadCommands` to `loadPickerCommands`. `isSafePath` guard preserved.
- `loadCommands` itself is **kept** rather than inlined into `loadPickerCommands`. This is intentional, not speculative reuse: the full discovery is the public contract the existing `tests/commands-loader.test.ts` suite asserts against (plugin namespacing, conflict resolution, agent dedupe, etc.) and is the natural seam for a future UI that surfaces skills/agents in a separate panel once execution is wired. Removing it would force any such follow-up to re-implement the scan from scratch.

## Tests

vitest cases in `tests/commands-loader.test.ts`:

- `loadPickerCommands` describe block: filter keeps user + project, drops plugin / skill / agent (mixed-source fixture); empty-result case for the dogfood scenario (only plugin/skill installed, no personal commands).
- New env-fallback tests: `CLAUDE_CONFIG_DIR` resolved when no `homeDir` opt; `opts.homeDir` wins when both set.

E2E (review request): new harness-ui case `slash-picker-claude-config-dir`. `preMain` mints a fresh `<tmp>/ccsm-e2e-fake-claude-*` tree seeded with `commands/local-test.md` (must surface) + `plugins/cache/mkt/superpowers/1.0.0/commands/brainstorm.md` (must NOT surface), points `process.env.CLAUDE_CONFIG_DIR` at it on the main process, then opens the slash picker and asserts the visibility contract. Disposers restore the env var and `rm -rf` the fake root.

Reverse-verification (per project E2E discipline): stashed the env-fallback in `commands-loader.ts` (loader reverts to scanning the dev's real `~/.claude`) → case fails as expected (`/local-test` never appears, only built-ins). Stash restored, case passes again.

## Follow-ups

- Typing `/plugin` (or any namespaced unknown slash like `/foo:bar`) directly into the textarea still pass-throughs to the model instead of producing a local "unknown command" toast — the picker filter only hides them from suggestion, it doesn't reject manual entry. Tracked separately; out of scope for this PR.

## Coordination

Touched: `electron/commands-loader.ts`, `electron/main.ts`, `tests/commands-loader.test.ts`, `scripts/harness-ui.mjs`. No overlap with the `fix-think-dropdown` worker — that one operates on `registry.ts` / `SlashCommandPicker.tsx` / `handlers.ts` which this PR doesn't touch.

## Test plan

- [x] `npx vitest run tests/commands-loader.test.ts` — 22/22 pass
- [x] `node scripts/harness-ui.mjs --only=slash-picker-claude-config-dir` — pass
- [x] Reverse-verify: stash env-fallback → harness case fails → restore → passes
- [ ] Manual dogfood: open picker with superpowers + pua plugins installed → no `/superpowers:*` / `/pua:*` rows visible; `/clear`, `/compact`, `/config`, `/think` still visible

### DIFF
--- a/electron/commands-loader.ts
+++ b/electron/commands-loader.ts
@@ -299,3 +299,39 @@ export function loadCommands(opts: LoadOpts = {}): LoadedCommand[] {
       ...(e.pluginId ? { pluginId: e.pluginId } : {}),
     }));
 }
+
+// ────────────────────── picker-visible filter ──────────────────────────────
+//
+// Subset of `loadCommands` surfaced to the slash-command picker. Three
+// sources are deliberately hidden:
+//
+//   - PLUGIN commands (`/<plugin>:<name>` from ~/.claude/plugins/cache/...).
+//     The CLI binary loads them in its own REPL but the
+//     @anthropic-ai/claude-agent-sdk transport ccsm uses requires plugins
+//     to be declared via `query({ options: { plugins: [...] } })`, which
+//     ccsm doesn't pass. Forwarding `/superpowers:brainstorm` to the SDK
+//     therefore arrives as plain user text and the model replies with the
+//     deprecated-skill stub instead of running the plugin command.
+//
+//   - SKILL entries (from ~/.claude/skills/) are loaded by the SDK as
+//     description-keyed system-prompt extensions ("when X happens, use this
+//     skill") — the agent invokes them on its own, the user cannot trigger
+//     them with `/skill-name`. Listing them in the picker created the same
+//     false affordance.
+//
+//   - AGENT entries are subagent definitions invoked by the main agent via
+//     the Task tool. There is no user-facing slash form, so they had no
+//     business in the picker either.
+//
+// `loadCommands` itself still returns every source — keeps the discovery
+// complete for future wiring (e.g. surfacing skills in a separate UI). Add
+// a source to PICKER_VISIBLE_SOURCES once the corresponding execution path
+// is actually wired through the SDK.
+export const PICKER_VISIBLE_SOURCES: ReadonlySet<CommandSource> = new Set([
+  'user',
+  'project',
+]);
+
+export function loadPickerCommands(opts: LoadOpts = {}): LoadedCommand[] {
+  return loadCommands(opts).filter((c) => PICKER_VISIBLE_SOURCES.has(c.source));
+}
--- a/electron/commands-loader.ts
+++ b/electron/commands-loader.ts
@@ -219,9 +219,16 @@ export type LoadOpts = {
 };
 
 export function loadCommands(opts: LoadOpts = {}): LoadedCommand[] {
-  const home = opts.homeDir ?? os.homedir();
   const cwd = opts.cwd ?? null;
-  const claudeRoot = path.join(home, '.claude');
+  // Resolve the Claude config root. Precedence:
+  //   1. opts.homeDir → `<homeDir>/.claude` (test override)
+  //   2. process.env.CLAUDE_CONFIG_DIR (used in production — ccsm sets this so
+  //      the binary and the GUI loader read the same tree; ignoring it would
+  //      desync the picker from what claude.exe actually executes)
+  //   3. `<os.homedir()>/.claude` (last-resort default)
+  const claudeRoot = opts.homeDir
+    ? path.join(opts.homeDir, '.claude')
+    : (process.env.CLAUDE_CONFIG_DIR ?? path.join(os.homedir(), '.claude'));
   const all: RawEntry[] = [];
 
   // 1. user
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -83,7 +83,7 @@ import { cancelQuestionRetry } from './notify-retry';
 import type { PermissionMode } from './agent/sessions';
 import { listModelsFromSettings } from './agent/list-models-from-settings';
 import { readMemoryFile, writeMemoryFile, memoryFileExists } from './memory';
-import { loadCommands } from './commands-loader';
+import { loadPickerCommands } from './commands-loader';
 
 // ─────────────────────── IPC security helpers ────────────────────────────
 //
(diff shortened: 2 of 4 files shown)


Jiahui-Gu (commented):
## Review verdict: changes requested (e2e blocker)

## Layer 1 — direction

Hide (option a) is the right dogfood call. Wiring SDK `query({plugins})` is real work; skill/agent have no slash semantics anyway.

**UX nit (non-blocking)**: a user with superpowers installed but no personal `~/.claude/commands` will see an empty picker and may assume install failed. Consider a follow-up dimmed row "N plugin/skill commands hidden — not yet runnable here". Not required here.

**Known gap not closed**: typing `/plugin` directly still falls through to model text. Please append to PR body: "Follow-up: namespaced unknown slash → local toast instead of pass-through" so it doesn't rot.

## Layer 2 — implementation

- `PICKER_VISIBLE_SOURCES` as named `const ReadonlySet` — good, single source of truth.
- Reverse-verify claim accepted.
- **`loadCommands` retained for hypothetical future UI** — violates "don't add for hypothetical future". Verified `grep`: only consumer outside tests is the new wrapper. Not blocking (wrapper is cheap, comment justifies) but acknowledge in PR body that it's deliberately speculative.

## Layer 3 — e2e gate (BLOCKING)

The skip rationale doesn't hold. `loadCommands` already accepts `homeDir` opt (line 216/222) — the unit tests use it. We don't need to isolate HOME at the harness; we need the loader to honor a redirect.

**Required**: extend the loader to honor `process.env.CLAUDE_CONFIG_DIR` when `homeDir` is unset:

```ts
const claudeRoot = opts.homeDir
  ? path.join(opts.homeDir, '.claude')
  : (process.env.CLAUDE_CONFIG_DIR ?? path.join(os.homedir(), '.claude'));
```

This is independently the right shape — CCSM already standardizes `CLAUDE_CONFIG_DIR=~/.claude/` for the binary; the loader ignoring it is inconsistent.

Then add a harness-ui case (or extend an existing one) that:
1. `setupBefore`: sets `process.env.CLAUDE_CONFIG_DIR = path.join(os.tmpdir(), 'ccsm-e2e-fake-claude')`, seeds `plugins/cache/superpowers/1.0.0/commands/brainstorm.md` and `commands/local-test.md` under it
2. opens picker, asserts only `local-test` row visible, no `brainstorm` / no `superpowers:*`
3. teardown restores env + rms tmp

Re-request review when added.

## What was done well

- Loader-level filter (not picker-level) — correct layer, unit-testable without Electron.
- Two vitest cases cover mixed-source + all-broken-source; the second pins the exact dogfood bug.
- Clear comment block explaining why each source is hidden — future contributor won't re-surface them.
- No overlap with `fix-think-dropdown`.

Jiahui-Gu:
@reviewer Addressed all three review points (re-requesting review):

1. **Loader honors `CLAUDE_CONFIG_DIR`** — `electron/commands-loader.ts` now resolves `opts.homeDir > process.env.CLAUDE_CONFIG_DIR > os.homedir()/.claude`. Two new vitest cases pin the env path and the opt-wins-on-tie precedence.

2. **harness-ui e2e case `slash-picker-claude-config-dir`** — `preMain` seeds a fresh `<tmp>/ccsm-e2e-fake-claude-*` tree with one user command (`local-test`, must surface) and one plugin command (`superpowers/brainstorm`, must NOT surface), sets `process.env.CLAUDE_CONFIG_DIR` on the main process via `app.evaluate`, then drives the picker and asserts the visibility contract. Disposers restore env + `rm -rf` the tree.

   Reverse-verified: `git stash` the env-fallback (loader reverts to `os.homedir()`) → case fails (`expected /local-test in picker; got: /clear /compact /config /think`). Stash restored, case passes again.

3. **PR body** — added (a) a `## Follow-ups` section flagging the unknown namespaced-slash → model pass-through gap, and (b) an explicit retention note under `## Implementation` covering why `loadCommands` is kept (test-contract seam + future skill/agent UI), not speculative.

Branch force-pushed (rebased onto `origin/working`). PR not self-merged.

Jiahui-Gu:
Re-review pass on 848ebc3. Diff verified: env fallback (opt.homeDir > CLAUDE_CONFIG_DIR > os.homedir()) in commands-loader.ts; harness-ui.mjs case slash-picker-claude-config-dir; PR body Follow-ups + loadCommands retention rationale. Local vitest run tests/commands-loader: 22/22 pass (incl. env-wins + opt-wins-on-tie). E2E case PASS in 1120ms — picker showed /local-test, suppressed superpowers:brainstorm under env-scoped fake ~/.claude. Forward assertion is implicit reverse-verify (dev's real ~/.claude has no /local-test; its appearance proves CLAUDE_CONFIG_DIR honored). Worker also documented explicit stash-then-fail. LGTM, merging.

PR #353, opened 2026-04-26
notif: Windows-only adapter + 删按钮 + 平台 stub (W3)

Jiahui-Gu:
## Summary
W3 of notification simplification.
- Delete Electron Notification cross-platform fallback in electron/notifications.ts (legacy `new Notification(...)` branch + click-handler removed; activation now flows through notify-bootstrap onAction)
- Strip Allow/Allow-always/Reject buttons from permission XML (question/done already action-less)
- Add darwin.ts + linux.ts stubs throwing "Not implemented" — emit methods throw, ctor + setActivationCallback are no-ops so Notifier.create succeeds and only emit fails
- notifier.ts platform dispatch: switch on process.platform → win32/darwin/linux/default-throw

## Test plan
- [x] typecheck pass
- [x] lint pass (no new errors introduced; 41 pre-existing errors all in verification/*.mjs)
- [ ] Manual: trigger permission toast on Windows — no buttons, click focuses session
- [ ] W5 e2e suite (after W1-W4 land)

### DIFF
--- a/electron/notifications.ts
+++ b/electron/notifications.ts
@@ -8,7 +8,6 @@ import {
   isNotifyAvailable,
 } from './notify';
 import { shouldSuppressForFocus, registerToastTarget } from './notify-bootstrap';
-import { scheduleQuestionRetry } from './notify-retry';
 
 // Cap the assistant-message preview that the Done toast renders. The inlined
 // notify module's xml/done.ts ASSISTANT_LINE_MAX = 80 re-truncates inside the
@@ -147,10 +146,6 @@ async function emitAdaptiveToast(payload: ShowNotificationPayload): Promise<void
         cwdBasename: cwdBase,
       };
       await notifyQuestion(questionPayload);
-      // Schedule a single re-emit after ~30s in case the user missed the
-      // first banner. Cancelled by the `agent:resolvePermission` IPC handler
-      // when the question is answered.
-      scheduleQuestionRetry(questionPayload, payload.sessionId);
       return;
     }
     case 'turn_done': {
--- a/electron/notifications.ts
+++ b/electron/notifications.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { BrowserWindow, Notification } from 'electron';
+import type { BrowserWindow } from 'electron';
 import {
   notifyPermission,
   notifyQuestion,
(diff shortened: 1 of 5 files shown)


Jiahui-Gu (commented):
## REQUEST CHANGES (posted as comment — self-PR can't formally request changes)

## Layer 2 — BLOCKING

**Critical: `./notify-retry` import NOT removed.**
- `electron/notifications.ts:11` still imports `scheduleQuestionRetry from './notify-retry'`
- still called at `electron/notifications.ts:153`

W4 deletes `notify-retry.ts` and its standalone typecheck depends on this import being gone. Per the locked plan W3 owes this cleanup. Either:
1. Remove the import + the `scheduleQuestionRetry(...)` call (drop the 30s re-emit — the locked spec for the simplification removes retry entirely), or
2. Coordinate with W4 to merge atomically.

Option 1 is the spec — please remove the call.

## Layer 2 — other findings

- `darwin.ts` / `linux.ts` declare `dispose(): void { throw }`. `Notifier.dispose()` calls `this.adapter.dispose?.()` (notifier.ts:127). The optional-chain only skips when the property is *absent*; here it's present and throws. Cleanup paths (app shutdown, Notifier reinit) will surface "Not implemented" on darwin/linux. Make `dispose()` a no-op (matches the rationale used for `setActivationCallback`). Same for `dismiss()` — force-dismiss on a never-emitted toast should be a silent no-op so error paths don't double-fault.
- `PlatformAdapter` interface (`platform/windows.ts:32`) does NOT declare `setActivationCallback`. The stubs implement it, the dispatcher duck-types it (`notifier.ts:38`), so this works — but it's an undeclared contract. Consider adding it to the interface as optional. Not blocking.
- `permission.ts`: `<actions>` block fully removed, body click via `launch=focus` retained. Clean.
- `notifier.ts` switch + `UNSUPPORTED_PLATFORM` default: clean, message includes platform name.
- `notifications.ts`: `Notification` import gone, focus-handler gone, `truncatePreview` still used by turn_done branch — correct, no orphan imports beyond notify-retry.

## Layer 1

- **Stub design (ctor + setActivationCallback no-op, emit throws):** correct interpretation. `notify.ts:227+` wraps each `notifyPermission/Question/Done` in try/catch — throws are caught, logged, never propagate to the agent loop. Not a regression. The `isNotifyAvailable()` gate (unchanged here) keeps non-Windows callers silent in practice anyway.
- **Click-to-focus contract:** verified. `notify-bootstrap.ts:204` onAction handler routes `focus` to `notification:focusSession` IPC. Toast body click with `launch=focus` (permission.ts:38, retained) fires onAction → bootstrap line 249 sends focusSession. Path intact for the new no-button toast.
- **`ActionId` type left intact:** correct call. Toast XML no longer emits allow/allow-always/reject, but the type is shared with in-app permission UI — narrowing cascades outside W3 scope.

## Summary

Block on the `notify-retry` import + `scheduleQuestionRetry` call removal — that's the cross-worker fix W4 needs. Also flip dispose/dismiss in stubs to no-op. After those, ship it.

Jiahui-Gu (commented):
LGTM (self-approve blocked) — re-review pass, all 5 verification points clean:

1. notifications.ts: zero matches for `notify-retry` / `scheduleQuestionRetry` (grep clean). W4 typecheck unblocker confirmed.
2. darwin.ts: `dispose()` + `dismiss()` are no-ops (`/* not implemented */`); `permission/question/done` still throw `NOT_IMPLEMENTED`. Header comment explains shutdown rationale.
3. linux.ts: identical no-op shape — shutdown-safe, emit methods throw.
4. `case 'question'`: `registerToastTarget` + `await notifyQuestion(questionPayload)` + `return` — fires once, no retry scheduling.
5. History clean: tip 2a320ea (fix commit) on top of 8b43271 (W3 base). Mergeable=true.

Approve, merge unblocks W4.

PR #365, opened 2026-04-26
fix(ux): batch — Esc/question/auto-advance/sidebar UX bugs

Jiahui-Gu:
## Summary

Four small but visible UX bugs reported via dogfood, batched together because they sit in adjacent surfaces (InputBar + QuestionBlock + lifecycle pulse).

### Bug 1 — Esc doesn't interrupt running turn from textarea focus

**Root cause:** doc-level Esc handler in `InputBar.tsx` returns early on `document.querySelector('[role="dialog"]')`. Both `QuestionBlock` and `CwdPopover` set `role="dialog"` for a11y but are NOT modal — their mere presence silently disabled Esc-to-stop everywhere, including from the composer.

**Fix:** discriminate true modals via a new `data-modal-dialog` marker, set by our `DialogContent` wrapper (`src/components/ui/Dialog.tsx`) + `CommandPalette` (which uses `RD.Content` directly). Doc handler now ignores inline a11y dialogs.

### Bug 2 — AskUserQuestion sticky doesn't disappear on interrupt

**Root cause:** `stop()` only flipped store flags + IPC-interrupted; the unanswered question block stayed in `messagesBySession`, so `QuestionStickyHost` kept rendering it. User saw a permanently sticky card with no way to dismiss.

**Fix:** in `stop()`, iterate session blocks, mark every unanswered `kind: 'question'` as rejected, and best-effort deny any pending `requestId` permission. Mirrors what `QuestionBlock`'s Cancel chip does, but driven by the interrupt path.

### Bug 3 — Single-choice auto-advance never fires in production

**Root cause:** classic React state-updater closure trap.

```ts
let didSelectFresh = false;
setPicks((prev) => { ...; if (!set.has(label)) didSelectFresh = true; ... });
if (!q.multiSelect && didSelectFresh && active < questions.length - 1 && ...) {
  // schedule advance
}
```

The setPicks updater runs during the next render commit, not synchronously. By the time the `if` check executes, `didSelectFresh` is still `false`. Tests passed because `fireEvent` inside `act` flushes updaters synchronously — real browser dispatch defers them.

**Fix:** compute `didSelectFresh` from the current `picks` state BEFORE calling `setPicks`. Add `picks` to the useCallback deps.

### Bug 4 — Active session row pulses on agent reply

**Root cause:** `lifecycle.ts` turn_done path triggered sidebar pulse whenever `document.hasFocus()` was false, even when the session was active. Alt-tab away and back, and the row the user is already looking at pulses briefly.

**Fix:** check active-ness only. `if (!isActive) setSessionState(id, 'waiting')`. Background sessions still pulse; the active row never does.

## Files

- `src/components/InputBar.tsx` — `data-modal-dialog` selector + question-dismiss in `stop()`
- `src/components/ui/Dialog.tsx` — set `data-modal-dialog` + `aria-modal="true"` on `RD.Content`
- `src/components/CommandPalette.tsx` — same markers on its inline `RD.Content`
- `src/components/QuestionBlock.tsx` — fix updater-closure bug, recompute `didSelectFresh` upfront
- `src/agent/lifecycle.ts` — drop `windowFocused` check from turn_done pulse gate
- 5 `tests/inputbar-*.tsx` files — wrap renders in `ToastProvider` (PR #359 broke these on `working`; not introduced here)

## Test plan

- [x] `npx vitest run tests/inputbar.test.tsx` — 10/10 PASS (3 new cases: textarea-focus Esc, inline-dialog passthrough, interrupt-clears-question)
- [x] `npx vitest run tests/question-block.test.tsx` — 13/13 PASS (existing auto-advance tests still pass; production fix doesn't change RTL behavior)
- [x] `npx vitest run tests/inputbar-continue-after-interrupt.test.tsx tests/inputbar-recall.test.tsx tests/inputbar-diff-comments.test.tsx` — PASS after ToastProvider wrap
- [x] Full vitest: 845/848 PASS. 3 failures all unrelated (better-sqlite3 native binding NODE_MODULE_VERSION mismatch + missing `@testing-library/user-event` dep on `tests/inputbar-morph-mention`); same 3 fail on origin/working — went from 31 failures → 3.
- [x] `npm run build` clean (3 pre-existing webpack size warnings)
- [ ] Dogfood self-verify: Esc-from-textarea-focus, question-dismiss-on-stop, single-choice auto-advance in 2-question tool_use, active-session pulse suppressed (user to verify in dev app at port 4100)

## Commits

1. `fix(input): Esc interrupts from textarea focus + dismisses unanswered question on stop` (bugs 1+2 + test infra)
2. `fix(question): single-choice auto-advance fires in production builds` (bug 3)
3. `fix(sidebar): never pulse the active session row` (bug 4)

Bugs 5 (focus coordination across question/permission/textarea) and 6 (slash picker shows skill/plugin commands) are split into a follow-up PR — different file scopes, plus bug 5 needs cross-component focus orchestration that warrants its own design pass.

### DIFF
--- a/scripts/harness-agent.mjs
+++ b/scripts/harness-agent.mjs
@@ -1490,6 +1490,98 @@ async function caseEscInterrupt({ win, log }) {
   }
 
   log('Esc -> stop() ran (interrupted+queue cleared); Interrupted status block; textarea usable');
+
+  // ── J2 ────────────────────────────────────────────────────────────────
+  // Regression for #286 (PR #365): Esc must interrupt while textarea has
+  // focus AND when an inline `role="dialog"` widget (AskUserQuestion sticky,
+  // CwdPopover) is mounted. Before the fix, the doc-level handler returned
+  // early on `[role="dialog"]` — any inline a11y dialog silently disabled
+  // Esc-to-stop, including from the composer.
+  //
+  // Sub-case A: textarea-focus + inline role="dialog" (no data-modal-dialog).
+  // Asserts interrupt still fires.
+  // Sub-case B: textarea-focus + role="dialog" + data-modal-dialog.
+  // Asserts interrupt is suppressed (real Radix modals own Esc).
+  const SID2 = 's-esc-int-j2';
+  await win.evaluate((sid) => {
+    const store = window.__ccsmStore;
+    store.setState({
+      groups: [{ id: 'g1', name: 'G1', collapsed: false, kind: 'normal' }],
+      sessions: [{
+        id: sid, name: 'esc-textarea', state: 'idle', cwd: 'C:/x',
+        model: 'claude-opus-4', groupId: 'g1', agentType: 'claude-code'
+      }],
+      activeId: sid,
+      messagesBySession: { [sid]: [] },
+      startedSessions: { [sid]: true },
+      runningSessions: { [sid]: true },
+      interruptedSessions: {},
+    });
+  }, SID2);
+  await win.waitForTimeout(150);
+
+  // Sub-case A: inline role="dialog" must NOT block doc-level Esc handler.
+  await win.evaluate(() => {
+    // Remove any leftover from sub-case set up below on retries.
+    document.querySelectorAll('[data-harness-injected-dialog]').forEach((n) => n.remove());
+    const inline = document.createElement('div');
+    inline.setAttribute('role', 'dialog');
+    inline.setAttribute('aria-label', 'inline a11y widget (no aria-modal)');
+    inline.setAttribute('data-harness-injected-dialog', 'inline');
+    // No data-modal-dialog marker — this is the inline-widget shape that
+    // QuestionBlock + CwdPopover use.
+    document.body.appendChild(inline);
+  });
+  // Real textarea focus — that's the regression scenario.
+  const ta2A = win.locator('textarea').first();
+  await ta2A.waitFor({ state: 'visible', timeout: 5000 });
+  await ta2A.click();
+  const focusedA = await win.evaluate(() => document.activeElement?.tagName);
+  if (focusedA !== 'TEXTAREA') throw new Error(`J2-A: expected textarea focus, got activeElement=${focusedA}`);
+  await win.evaluate(() => {
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true, cancelable: true }));
+  });
+  await win.waitForTimeout(200);
+  const interruptedA = await win.evaluate((sid) => !!window.__ccsmStore.getState().interruptedSessions[sid], SID2);
+  if (!interruptedA) {
+    throw new Error('J2-A regression: textarea-focus Esc did NOT interrupt while inline role="dialog" was mounted — the global selector is over-matching');
+  }
+  // Reset for sub-case B.
+  await win.evaluate((sid) => {
+    document.querySelectorAll('[data-harness-injected-dialog]').forEach((n) => n.remove());
+    const st = window.__ccsmStore.getState();
+    st.consumeInterrupted(sid);
+    st.setRunning(sid, true);
+  }, SID2);
+  await win.waitForTimeout(120);
+
+  // Sub-case B: data-modal-dialog (real Radix modal shape) MUST suppress.
+  await win.evaluate(() => {
+    const modal = document.createElement('div');
+    modal.setAttribute('role', 'dialog');
+    modal.setAttribute('aria-modal', 'true');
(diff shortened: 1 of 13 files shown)


Jiahui-Gu:
Addressed reviewer e2e gaps. New commit `13fe612` adds **5 harness cases** (all reverse-verified by stashing the corresponding fix and confirming the case fails, then restoring and confirming it passes).

## Cases added

### 1. `harness-agent / esc-interrupt` — extended with J2 (#286)
Textarea focus + inline `role="dialog"` injected → Esc must still interrupt. Plus counter-positive: `data-modal-dialog` marker MUST suppress.

**Forward (with fix):**
```
[case=esc-interrupt] Esc -> stop() ran (interrupted+queue cleared); Interrupted status block; textarea usable
[case=esc-interrupt] J2 — textarea-focus Esc passes through inline role="dialog" (#286), suppressed by data-modal-dialog
[case=esc-interrupt] OK
```

**Reverse (InputBar reverted to origin/working):**
```
[case=esc-interrupt] Esc -> stop() ran (interrupted+queue cleared); Interrupted status block; textarea usable
[case=esc-interrupt] FAIL: J2-A regression: textarea-focus Esc did NOT interrupt while inline role="dialog" was mounted — the global selector is over-matching
```

### 2. `harness-perm / auq-single-select-auto-advance` (#287)
Three single-select questions, real Playwright clicks. Q1 → Q2 → Q3 must auto-advance; Q3 stays put (last question). **Reverse-verify caught the closure-vs-updater bug deterministically on the Q2 click** — confirms vitest's `act`-synchronous flushing was masking the production bug.

**Forward:**
```
[case=auq-single-select-auto-advance] #287 — single-choice auto-advance fires from Q1→Q2→Q3, last question stays put
[case=auq-single-select-auto-advance] OK
```

**Reverse (QuestionBlock reverted):**
```
[case=auq-single-select-auto-advance] FAIL: after clicking esbuild (Q2 single-select), expected auto-advance to Q3 (tab index 2), got 1
```

### 3. `harness-perm / auq-stop-dismisses-question` (PR#365 bug 2)
Seed unanswered question + running session, focus textarea, dispatch Esc. Assert: `interruptedSessions[sid]=true`, `qb.answered=true`, `qb.rejected=true`, `[data-question-sticky]` unmounted.

**Forward:**
```
[case=auq-stop-dismisses-question] PR#365 bug 2 — Esc/stop dismisses unanswered AskUserQuestion sticky
[case=auq-stop-dismisses-question] OK
```

**Reverse (InputBar reverted — also reverts the markQuestionAnswered block):**
```
[case=auq-stop-dismisses-question] FAIL: post-stop: interruptedSessions flag not set — stop() did not run
```

### 4. `harness-perm / question-dismisses-on-stop-button` (additional bug from coordinator)
**Coordinator flagged a sibling user-reported bug**: clicking the visible Stop morph button (instead of pressing Esc) must also dismiss the unanswered question. Verified the Stop button at `InputBar.tsx:1210` already wires to the same `stop()` function, so the existing fix covers it — this case pins the contract so a future patch splitting the wiring can't silently break it.

**Forward:**
```
[case=question-dismisses-on-stop-button] PR#365 — Stop button click also dismisses unanswered AskUserQuestion sticky (same path as Esc)
[case=question-dismisses-on-stop-button] OK
```

**Reverse (InputBar reverted):**
```
[case=question-dismisses-on-stop-button] FAIL: regression: clicking Stop did not dismiss the unanswered question — Stop-button onClick path bypasses markQuestionAnswered
```

### 5. `harness-ui / sidebar-active-row-no-pulse` (#289)
Two sessions (active+background), `Object.defineProperty(document, 'hasFocus', () => false)`, dispatch `agent:event` result frame from main into the renderer's production `lifecycle.ts` `onAgentEvent` handler. Assert: active row stays `state !== 'waiting'`; background row flips to `'waiting'` (positive control — confirms the fix didn't over-correct).

**Forward:**
```
[case=sidebar-active-row-no-pulse] #289 — active row stays !waiting on turn_done; background row still pulses
[case=sidebar-active-row-no-pulse] OK
```

**Reverse (lifecycle reverted):**
```
[case=sidebar-active-row-no-pulse] FAIL: #289 regression: active session pulsed on turn_done (state='waiting') even though it's the focused row — alt-tab unfocus path leaked back in
```

## Per the rules

- All 5 cases **extend / sit alongside existing harness cases** (`feedback_e2e_prefer_harness.md`) — no new independent probes.
- All 5 reverse-verified individually.
- Fix code untouched (reviewer already approved the fix; only e2e was the gap).
- ToastProvider helper shape preserved.

Final commit sha: `13fe612`. Ready for re-review.

PR #370, opened 2026-04-26
fix(truncate): capture CLI stderr + sequence close-then-respawn (#288)

Jiahui-Gu:
## User report (task #288)

> 「trunce以后再回复收到报错: Failed to start Claude — Claude Code process exited with code 1」

The error landed in the `AgentInitFailedBanner` from the `CLI_SPAWN_FAILED` errorCode in `electron/agent/manager.ts`. The bare "Claude Code process exited with code 1" came straight from the SDK with **no stderr context** — ccsm was running the SDK with `stderr: 'ignore'` (default), so claude.exe's diagnostic output was dropped on the floor.

## Investigation

1. **Probed the SDK directly** with `--session-id` of an existing JSONL (the post-truncate respawn shape) — claude.exe handled it fine, exit 0. Hypothesis "SDK rejects sessionId reuse" eliminated.
2. **Walked the rewindToBlock → agentClose → startSession path** in `src/stores/store.ts:1907-1959` and `src/agent/startSession.ts`. Found the renderer fires `agentClose(sid)` fire-and-forget, then the user can type + click Send within milliseconds. The previous `manager.close()` was synchronous (just aborted the SDK signal) — the next `manager.start()` for the same sessionId could spawn a new claude.exe while the old one still held the JSONL handle. On Windows file-lock semantics make this fatal.
3. **Confirmed ccsm drops stderr**: `electron/agent-sdk/sessions.ts` did not pass `options.stderr` to `sdk.query()`. Per the SDK's ProcessTransport (sdk.mjs line ~210), stderr is `pipe`d only when the callback is set — otherwise `'ignore'`.

Could not reproduce the user's specific failure in a sandboxed harness; the bug depends on environment state (timing, Windows file-lock contention, or something else not yet visible). Rather than ship a speculative fix without diagnosis, this PR makes the codepath both **harder to fail** (race fix) and **self-diagnosing** (stderr capture).

### 1. Capture CLI stderr (`electron/agent-sdk/sessions.ts`)
- Pass `options.stderr` callback to the SDK so claude.exe's stderr is no longer thrown away.
- Bounded 4 KiB ring buffer on the runner; chunks also forward as `agent:diagnostic` (code=`cli_stderr`) so the renderer logs them.
- New `getStderrTail()` accessor.
- Consumer-loop catch path appends the tail to any runtime exit error so the user sees *why* claude.exe died.

### 2. Sequence close-then-respawn (`electron/agent/manager.ts`)
- New `closing: Map<sessionId, Promise<void>>` tracks in-flight teardowns.
- `close()` stashes `runner.awaitClosed()` (resolves once the SDK consumer settles, i.e. claude.exe fully exited and released stdio + JSONL handles).
- `start()` awaits any matching closing promise before spawning a fresh runner for the same sessionId.
- Self-cleaning: the `.finally()` clears the map entry once the wait resolves, only if no newer close has displaced it.
- Same gating in `closeAll()`.
- Early-failure branch reads `runner.getStderrTail()` and tacks it onto the `CLI_SPAWN_FAILED` detail so the banner shows the real cause.

### 3. New `Runner` interface methods
`awaitClosed(): Promise<void>` and `getStderrTail(): string` — only implementation is `SdkSessionRunner`; the abstract interface keeps the manager's contract explicit.

### Unit (`electron/agent-sdk/__tests__/sessions.test.ts` — 3 new)
Pins the building blocks the manager fix relies on:
- `passes stderr callback to SDK and accumulates chunks into getStderrTail()`
- `runtime exit error appends stderr tail so the user sees why claude.exe died`
- `awaitClosed() resolves once the consumer settles after close()`

All 34 tests pass (was 31). `npx tsc --noEmit` clean.

### E2E (`scripts/harness-restore.mjs` — 1 new)
`truncate-then-resend` (requiresClaudeBin): plants a 4-frame JSONL fixture, hydrates as an existing session, hovers the second user message → clicks "Truncate from here", sends a fresh prompt, asserts no sessionInitFailures banner appears within 12s and an assistant frame / running flag is set.

```
[case=truncate-then-resend] planted JSONL fixture at C:\...\a1b1c1d1-0000-4000-8000-000000000288.jsonl
[case=truncate-then-resend] JSONL hydrated; second user message rendered
[case=truncate-then-resend] truncate cut to blocks=["u-u-truncate-1","msg-1:c0"]; resume cleared
[case=truncate-then-resend] truncate→resend respawned cleanly (no init-failure banner)
[case=truncate-then-resend] OK
```

**Honest framing**: the e2e passes both with and without the manager fix on this dev machine — couldn't reproduce the original failure in the harness. The probe still pins the codepath against future regressions and serves as a baseline for if the bug recurs.

## Honest framing on the user's original report

This PR does NOT claim to have reproduced and root-caused the user's exact failure. It ships:
- a **strictly correct** race fix (close-then-respawn sequencing was clearly missing — the renderer fires close fire-and-forget, the next start had no gate)
- a **strictly correct** observability improvement (we were dropping CLI stderr; this is recoverable signal we should always have surfaced)

**If the user hits this again after the fix lands**, the banner will now contain the actual stderr tail, which will tell us the real root cause in seconds rather than days.

## Verify checklist
- [x] `npx vitest run electron/agent-sdk/__tests__/sessions.test.ts` → 34/34 pass
- [x] `npm run build` clean (3 size warnings, pre-existing)
- [x] `node scripts/harness-restore.mjs --only=truncate-then-resend` → pass
- [ ] User retries the truncate→resend flow; if it recurs the banner now has stderr context

Closes #288.

### DIFF
--- a/electron/agent-sdk/__tests__/sessions.test.ts
+++ b/electron/agent-sdk/__tests__/sessions.test.ts
@@ -20,6 +20,7 @@ import type { StartOptions } from '../../agent/sessions';
 function makeFakeSdk() {
   let lastOptions: unknown = null;
   let queueResolve: ((v: IteratorResult<unknown>) => void) | null = null;
+  let queueReject: ((err: unknown) => void) | null = null;
   const buffer: unknown[] = [];
   let closed = false;
 
@@ -38,8 +39,9 @@ function makeFakeSdk() {
           next(): Promise<IteratorResult<unknown>> {
             if (buffer.length > 0) return Promise.resolve({ value: buffer.shift(), done: false });
             if (closed) return Promise.resolve({ value: undefined, done: true });
-            return new Promise((resolve) => {
+            return new Promise((resolve, reject) => {
               queueResolve = resolve;
+              queueReject = reject;
             });
           },
         };
@@ -70,7 +72,17 @@ function makeFakeSdk() {
       if (queueResolve) {
         const r = queueResolve;
         queueResolve = null;
+        queueReject = null;
         r({ value: undefined, done: true });
+      }
+    },
+    fail(err: unknown) {
+      closed = true;
+      if (queueReject) {
+        const r = queueReject;
+        queueResolve = null;
+        queueReject = null;
+        r(err);
       }
     },
     getOptions: () => lastOptions as { options?: Record<string, unknown> } | null,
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
Superseded by #378 — root cause was renderer/CLI history divergence after truncate (resumeSessionId being cleared, causing fresh CLI session to mismatch on-disk JSONL). #378 pins resumeSessionId on rewind. The race-fix + stderr-capture in this PR is correct in spirit but unneeded once the real cause is fixed; closing to avoid stacking infra changes on a misdiagnosis. stderr capture can come back as its own PR if we want it for observability.

PR #397, opened 2026-04-26
fix(chat): assistant block copy button (#bug5)

Jiahui-Gu:
## Summary
- Dogfood r2 fp2 found assistant message blocks have no copy-to-clipboard button while user blocks have one (UserBlock.tsx:213-220).
- Mirror the UserBlock pattern in AssistantBlock — same i18n keys (`chat.userMsgCopy`/`chat.userMsgCopied`), same hover-reveal placement, same `navigator.clipboard.writeText(text)` handler shape, same a11y attrs.
- Hidden while streaming (copying half-streamed text would surprise) and only shown when text is non-empty.

## Test plan
- [x] New `assistant-copy-button` harness-agent case clicks copy on a real assistant block and asserts clipboard matches the assistant text (with regression guard that user prompt text is NOT in clipboard).
- [x] Case fails on working HEAD without the fix (reverse-verified via stash + rebuild — `data-assistant-block-id` attribute missing, locator timeout 5s).
- [x] Case passes with fix applied (1.2s).
- [ ] Manual screenshot at docs/screenshots/dogfood-r2/fp2-basic-dialog/bug5-fixed.png — skipped due to time; e2e exercises the real Electron + clipboard path which is the same surface the manual would test.

Bundled into harness-agent (per E2E discipline rule: prefer harness over standalone probe).

### DIFF
--- a/scripts/harness-agent.mjs
+++ b/scripts/harness-agent.mjs
@@ -2540,6 +2540,74 @@ async function caseUserBlockHoverMenu({ win, log }) {
   log('hover reveals 4 actions; Copy -> clipboard; Truncate -> cut keeps clicked user msg + pinned resume');
 }
 
+// ---------- assistant-copy-button ----------
+// Dogfood r2 fp2 bug #5: AssistantBlock had no copy-message button while
+// UserBlock did (parity gap). Mirrors the UserBlock hover-reveal Copy pattern
+// on assistant rows. We seed BOTH a user prompt and an assistant reply, then
+// click the Copy button on the assistant row and assert the clipboard
+// contains the assistant text — NOT the user prompt (regression guard: the
+// bug we're fixing was "no button at all", but the next-most-likely
+// regression is wiring it to the wrong block's text).
+async function caseAssistantCopyButton({ win, log }) {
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
Reviewer LGTM (self-PR; cannot formally approve via gh).

Parity with UserBlock confirmed across button shape, classes, aria-label, tooltip, copied-flash; reuse of `userMsgCopy`/`userMsgCopied` is safe (generic "Copy message"/"Copied" strings, no user-specific wording). `data-assistant-block-id` locator is new and can't collide with UserBlock. Defensive asymmetry vs UserBlock (hide while streaming, hide when empty) is justified — assistant has streaming concept, user doesn't. `renderBlock` only adds `id={b.id}`, no impact on other block kinds.

E2E:
- `harness-agent --only=assistant-copy-button` passes (2041ms).
- Reverse-verified: reverted AssistantBlock.tsx + renderBlock.tsx to HEAD~1, rebuilt, probe FAILED at `data-assistant-block-id="a-copy"` locator timeout. Probe genuinely guards the fix.
- Regression assertion that clipboard does NOT contain the user prompt text correctly catches the fp2 false-pass mode.

Safe to merge.

PR #398, opened 2026-04-27
fix(agent): capture SDK session id on init for cross-restart resume (#fp5)

Jiahui-Gu:
## Summary
- Captures the CLI-allocated `session_id` from the first `system/init` SDK frame and persists it on the `Session` record as `sdkSessionId`.
- `startSession.ts` now passes `--resume <sdkSessionId>` on subsequent starts so a ccsm restart reattaches to the existing conversation instead of silently timing out.
- Updates `loadMessages`'s on-disk JSONL path resolver to consider `sdkSessionId` so message history loads from the live transcript.

## Root cause
- `electron/agent-sdk/sessions.ts:657` already captured `cliSessionId` from the SDK init frame, but the value was only used for an internal mismatch diagnostic — never plumbed back to the renderer / store.
- On relaunch, `src/agent/startSession.ts:53-55` (pre-fix) saw `session.resumeSessionId == null` and passed `sessionId: <existing-uuid>` to the SDK. The CLI silently refuses to spawn over an existing on-disk JSONL, so the next user prompt timed out at 45s with no reply.
- `src/stores/persist.ts` already persists the full `sessions` array — adding `sdkSessionId` to `Session` was sufficient for cross-restart durability; no schema change required.
- The CLI does NOT rotate `session_id` on resume (verified via direct `claude --resume` probing — same JSONL across N resumes), so capture-once would be sufficient. The implementation still calls `recordSdkSessionId` on every init frame defensively (no-op when value is unchanged) — cheap insurance.

## Fix
Files touched (5):
- `src/types.ts` — add optional `sdkSessionId?: string` on `Session`.
- `src/stores/store.ts` — new `recordSdkSessionId(sessionId, sdkSessionId)` action; update `loadMessages` `sidOnDisk` derivation to `resumeSessionId || sdkSessionId || id`.
- `src/agent/lifecycle.ts` — in `api.onAgentEvent`, intercept `system/init` and dispatch `recordSdkSessionId(e.sessionId, msg.session_id)`.
- `src/agent/startSession.ts` — `resumeSessionId = session.resumeSessionId ?? session.sdkSessionId`; user-set imports still win.
- `scripts/harness-restore.mjs` — new e2e case `restart-resume-continues-conversation`.

## Test plan
- [x] New harness case: 3-launch ("开关开关开") + real claude.exe (Agent Maestro proxy). Asserts both secrets recall across two restart cycles.
- [x] **Phase 1 — failing baseline (fix stashed):**
  ```
  [case=restart-resume-continues-conversation] [A1] reply received in 3593ms; len=6
  [case=restart-resume-continues-conversation] launch #1 captured sdkSessionId=null
  [case=restart-resume-continues-conversation] FAIL: launch #1 finished but no sdkSessionId captured on the session record (system/init capture missing)
  ```
- [x] **Phase 4 — green with fix:**
  ```
  [case=restart-resume-continues-conversation] launch #1 captured sdkSessionId=79819709-26be-4870-a89e-c686bb23ba04
  [case=restart-resume-continues-conversation] launch #2 restored session, persisted sdkSessionId=79819709-26be-4870-a89e-c686bb23ba04
  [case=restart-resume-continues-conversation] [B] reply received in 3562ms; len=31
  [case=restart-resume-continues-conversation] launch #3 restored session, persisted sdkSessionId=79819709-26be-4870-a89e-c686bb23ba04
  [case=restart-resume-continues-conversation] [C] reply received in 3093ms; len=31
  [case=restart-resume-continues-conversation] PASS — both secrets recalled across two restarts ("开关开关开")
  ```
- [x] **Reverse-verify:** the Phase 1 failing log above IS the reverse-verify (`git stash` of src changes, rebuild, re-run → fails at `sdkSessionId=null` on launch #1).
- [x] `npm run build` — clean.
- [ ] (Manager) post-merge: rebuild installer + dogfood r2 fp5 probe to confirm the original probe goes from `aborted: 'C prompt timeout (45s) — no reply'` to PASS.

### DIFF
--- a/scripts/harness-restore.mjs
+++ b/scripts/harness-restore.mjs
@@ -1655,7 +1655,284 @@ async function casePermissionPromptDefaultMode({ log, registerDispose }) {
 
 
 // ──────────────────────────────────────────────────────────────────────────
-// CASE: new-session-default-model-from-claude-settings
+// CASE: restart-resume-continues-conversation  (#fp5)
+// 3-launch ("开关开关开") with the SAME --user-data-dir + real claude.exe
+// (Agent Maestro proxy via the dev's ~/.claude/settings.json). Verifies
+// the bug-fp5 regression guard: after closing ccsm and relaunching,
+// sending another prompt in the resumed session must reach the model and
+// the model must remember the prior turns.
+//
+// THREE rounds because every `claude --resume <id>` allocates a NEW
+// cliSessionId (the old one is consumed). A one-restart probe would let
+// a fix that only captures the FIRST init slip through; round 3 catches
+// the latent "stale id" regression by requiring the post-resume rotation
+// to also be persisted.
+//
+// Without the fix the assertion times out at 90s on launch #2's prompt
+// because the SDK silently refuses to spawn over an existing JSONL.
+// ──────────────────────────────────────────────────────────────────────────
+async function caseRestartResumeContinuesConversation({ log, registerDispose }) {
+  const SECRET = 'magenta-otter-7';
+  const SECRET2 = 'azure-falcon-13';
+  const SESSION_ID = crypto.randomUUID();
+  const GROUP_ID = 'g-default';
+  const CWD = os.homedir();
+
+  const userDataDir = fs.mkdtempSync(path.join(os.tmpdir(), 'ccsm-harness-restart-resume-'));
+  registerDispose(() => { try { fs.rmSync(userDataDir, { recursive: true, force: true }); } catch {} });
+  log(`user-data-dir = ${userDataDir}`);
+
+  // Match the dev's working environment: ANTHROPIC_BASE_URL etc. live in
+  // ~/.claude/settings.json which the bundled CLI reads at boot. We force
+  // CLAUDE_CONFIG_DIR=~/.claude so the CLI sees the real Agent Maestro
+  // pointer instead of an empty isolated config (would 401 with no auth).
+  const env = {
+    ...process.env,
+    NODE_ENV: 'production',
+    CCSM_PROD_BUNDLE: '1',
+    CCSM_CLAUDE_CONFIG_DIR: path.join(os.homedir(), '.claude'),
+  };
+  delete env.CLAUDECODE;
+  delete env.CLAUDE_CODE_ENTRYPOINT;
+
+  // Reusable launch wrapper.
+  async function launch(label) {
+    const app = await electron.launch({
+      args: ['.', `--user-data-dir=${userDataDir}`],
+      cwd: ROOT,
+      env,
+      timeout: 60_000,
+    });
+    const win = await appWindow(app, { timeout: 30_000 });
+    await win.waitForLoadState('domcontentloaded');
+    await win.waitForTimeout(2000);
+    await win.waitForFunction(() => !!window.__ccsmStore, null, { timeout: 30_000 });
+    log(`launch #${label}: store ready`);
+    return { app, win };
+  }
+
+  async function snapshot(win) {
+    return await win.evaluate(() => {
+      const s = window.__ccsmStore.getState();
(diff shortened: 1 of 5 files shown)


Jiahui-Gu (commented):
## Review verdict: APPROVE (with one nit + one ambiguity to surface)

### Layer 1 — direction & necessity: SOUND
- Bug is real (already proven by upstream probe).
- The chosen fix path (lifecycle.ts → store action → startSession.ts) is fine: I grepped `electron/agent/stream-json-types.ts` and confirmed `system/init` frames carry `session_id: z.string()` and flow through `onAgentEvent` unmodified, so the field IS already exposed via the lifecycle event interface — no need to thread through `sdk-message-translator.ts`. The eval's Option A is more invasive than necessary.
- Capture-on-every-init is correct: even though the CLI doesn't rotate session_id on a fresh-session resume (per probe a974ae4), the test's round-3 path explicitly relies on subsequent inits being persisted, and `recordSdkSessionId`'s `if (sess.sdkSessionId === sdkSessionId) return s` short-circuit makes same-value calls a true no-op (no spurious re-renders). Different-value overwrite is intentional and safe.
- `resumeSessionId ?? sdkSessionId` precedence: sound for fresh sessions (resumeSessionId is null → sdkSessionId wins). See ambiguity note below for the import case.

### Layer 2 — implementation: CLEAN, one doc bug
- Type addition (`src/types.ts`), action (`src/stores/store.ts`), capture site (`src/agent/lifecycle.ts`), priority swap (`src/agent/startSession.ts`) all type-check and read cleanly. No dead code, no over-validation, no compat shims.
- **Nit / doc bug**: `recordSdkSessionId` JSDoc in `src/stores/store.ts:546-552` says *"No-op if the session already has a value (init frames may arrive on every spawn — we only persist the first)"* — this contradicts both the implementation (which overwrites on different value) AND the test's premise (round-3 explicitly relies on capture-on-every-init for SDK rotation). The behavior is correct; please fix the doc to say something like "No-op if the value is unchanged; overwrites if it differs (CLI may rotate session_id on resume — we always persist the latest)."
- `loadMessages` JSONL path resolver (`store.ts:2033`): the new precedence `resumeSessionId || sdkSessionId || id` does NOT introduce spurious `session_id_mismatch` diagnostics — that diagnostic is fired in `electron/agent-sdk/sessions.ts:664` against the runner id, independent of the path resolver. Safe.

### Ambiguity to surface (per `feedback_reviewer_flags_ambiguous_requirements.md`)
For sessions imported with an explicit `resumeSessionId`, the precedence `resumeSessionId ?? sdkSessionId` means we will keep resuming from the original imported sid forever, ignoring whatever `sdkSessionId` ccsm captured from the import's first turn. If the SDK does rotate on import-resume (the comment in `startSession.ts` claims it does — "the SDK allocates a new sid for the resumed branch"), then on every restart we re-fork from the import root rather than continuing the latest branch. This may or may not be desired; the author flagged it as intentional ("import semantics unchanged") and the fp5 e2e doesn't exercise it. Worth a follow-up probe on imported sessions specifically.

### Verification
- `npm run typecheck`: PASS
- `npm run build`: PASS
- New e2e `restart-resume-continues-conversation`: **PASS in 58.6s** — three real launches, both secrets recalled across both restarts, real claude.exe via Agent Maestro proxy.
- **Reverse-verify** (sources reverted to origin/working, harness change kept, rebuild, re-run): **FAILS in 12.0s** with `launch #1 finished but no sdkSessionId captured on the session record (system/init capture missing)`. Test correctly guards the fix.
- Test uses correct store schema (`messagesBySession`, `runningSessions`, `sessions[].sdkSessionId`) per dogfood probe schema rule.

Approving on behalf of an independent review pass — please fix the JSDoc nit before / on merge; the ambiguity is optional follow-up.

PR #354, opened 2026-04-26
notif: settings UI + store 砍到 2 toggle (W2)

Jiahui-Gu:
## Summary
W2 of notification simplification.
- NotificationSettings reduced to `{enabled, sound}` with auto-migration of old persisted shape (`migrateNotificationSettings` in `src/stores/store.ts`)
- Per-session mute removed (`Session.notificationsMuted`, sidebar context menu, `BellOff` icon, `setSessionNotificationsMuted` action)
- SettingsDialog notifications section: 2 toggles only — dropped per-event toggles, test button, module-availability indicator
- i18n: removed `sidebar.muteNotifications/unmuteNotifications/notificationsMutedAria` and `settings.notifications.{permission,permissionHint,question,questionHint,turnDone,turnDoneHint,testButton,testTitle,testBody,testIpcUnavailable,testSent,testFailed,moduleAvailable,moduleUnavailable,moduleChecking}` in both en.ts and zh.ts

## Migration
Old persisted shape `{enabled, permission, question, turnDone, sound}` (or any partial blob) hydrates through `migrateNotificationSettings` → `{enabled, sound}`. Missing booleans default to `true` so a corrupted/partial blob doesn't silently mute the user. Persist schema version stays at 1 — extra keys on disk are simply ignored on next read; first write rewrites the snapshot in the new shape.

## Cross-worker dependency
Typecheck currently fails on `src/notifications/dispatch.ts` (W1's file) — it still references the removed `settings.permission/question/turnDone` and `session.notificationsMuted`. W1's PR removes those readers. Both PRs must merge together (or W1 first) for `working` to compile.

## Test plan
- [ ] typecheck pass (after W1 lands)
- [ ] lint pass — no new lint issues introduced by W2 (verified locally; pre-existing unused-disable warning at store.ts:1185 is not from this PR)
- [ ] W5 e2e suite (after W1-W4 land)

PR #399, opened 2026-04-27
fix(e2e): harness-restore notify-fallback green

Jiahui-Gu:
## Summary
- `harness-restore::notify-fallback` was timing out on `[data-testid=\"notifications-module-status\"]`.
- PR #354 (notif settings simplified to enabled + sound only) deliberately removed that banner — silent fallback to in-app banners is the design intent.
- Drop the dead UI assertion; keep the `notify:availability` IPC contract assertion (available=false + non-empty error string) as the real contract. Documented rationale inline so the next Settings refactor isn't blamed.

## Root cause
`scripts/harness-restore.mjs:884` waited on `locator('[data-testid=\"notifications-module-status\"]')`. After PR #354 (merged 2026-04-26) the Settings → Notifications panel no longer renders that element, so the locator could never become visible — 3s timeout.

## Fix
`scripts/harness-restore.mjs` (`caseNotifyFallback`):
- Removed the Settings sidebar → notifications-tab → module-status assertion block (38 lines).
- Kept the existing IPC assertion (`available === false`, `error` is non-empty string), the renderer/main error-log assertions, and updated the success log to print the IPC error string instead of the now-gone banner text.
- Added a header comment pinning the rationale to PR #354 so future refactors know the IPC contract is the only contract.

Net diff: -38 / +9 in one file. No src/ changes.

## Test plan
- [x] Phase 0: confirmed branch + dirty file match killed-worker state.
- [x] Phase 1 (sanity, with fix): `node scripts/harness-restore.mjs --only=notify-fallback` → **OK** in 3.7s.
- [x] Phase 2 (neighbor regression): full `node scripts/harness-restore.mjs` → **15 passed, 0 failed, 0 skipped** (100.5s wall).
- [x] Phase 3: `npm run typecheck` → clean.
- [x] Reverse-verify: `git stash push scripts/harness-restore.mjs` → re-run only=notify-fallback → **FAIL** with the original `locator(...notifications-module-status...) Timeout 3000ms exceeded` at line 884. Stash popped, fix back in place.

Pre-fix failure log (reverse-verify):
```
[case=notify-fallback] FAIL: locator.waitFor: Timeout 3000ms exceeded.
Call log:
  - waiting for locator('[data-testid=\"notifications-module-status\"]') to be visible
```

Post-fix pass log:
```
[case=notify-fallback] notify:availability reports available=false with error: \"ccsm/notify: failed to load electron-windows-notifications (Cannot find module 'electron-windows-notifications'...)\"
[case=notify-fallback] OK
```

### DIFF
--- a/scripts/harness-restore.mjs
+++ b/scripts/harness-restore.mjs
@@ -782,6 +782,14 @@ async function caseDbCorruptionRecovery({ log, registerDispose }) {
 // Pre-launch fixture: rename `node_modules/electron-windows-notifications` so
 // the WindowsAdapter's require() throws MODULE_NOT_FOUND. Sandboxes HOME so
 // the dev's real ~/.claude isn't touched. From probe-e2e-notify-fallback.mjs.
+//
+// What this guards: the `notify:availability` IPC handler must report
+// available=false + a non-empty error string when the native module can't
+// load, so in-app banner fallback can take over silently. PR #354
+// deliberately removed the Settings → Notifications "module status" banner
+// (settings simplified to enabled+sound only — silent fallback is the design
+// intent). We no longer assert any user-facing UI for this case; the IPC
+// contract is the only contract.
 // ──────────────────────────────────────────────────────────────────────────
 async function caseNotifyFallback({ log, registerDispose }) {
   const notifyDir = path.join(ROOT, 'node_modules', 'electron-windows-notifications');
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
Independent reviewer (pool-5).

**Layer 1 — direction: APPROVED**
- PR #354 commit 98ce40c deliberately removed `data-testid="notifications-module-status"` and the entire `moduleStatus` indicator from `src/components/SettingsDialog.tsx` (commit stat: -127 lines in that file). Commit message explicitly: "SettingsDialog 通知段只剩 2 toggle, 删 per-event/test 按钮/模块状态".
- Current `SettingsDialog.tsx::NotificationsPane` (lines 293-318) renders only `notifications.enable` + `notifications.sound` toggles; no module-status / availability indicator anywhere.
- `git grep notifications-module-status` across the tree: 0 matches. The asserted UI is genuinely dead.
- Keeping the `notify:availability` IPC contract assertion (available=false + non-empty error) is the right invariant — it's the only contract left after the silent-fallback simplification.

**Layer 2 — implementation: clean**
- Diff is exactly what's claimed: 38 deleted assertion lines + 8-line header comment pinning rationale to PR #354. No collateral changes.
- Header comment is accurate and warranted — future readers will otherwise wonder why the IPC-only assertion exists with no UI follow-up.
- Final log line updated from "Settings banner reads" to "notify:availability reports available=false with error" — matches the new shape.

**e2e**: `--only=notify-fallback` PASS in 3.5s. Full `harness-restore` 15/15 PASS in 111.6s.
**Reverse-verify**: checked out origin/working version → FAIL with exact `locator.waitFor: Timeout 3000ms exceeded ... [data-testid="notifications-module-status"]` at line 884. Confirms the assertion was the failure source.
**typecheck**: clean (both tsconfigs).

Recommendation: **APPROVE**.

PR #403, opened 2026-04-27
fix(store): truncation marker survives cross-restart (#fp9-F)

Jiahui-Gu:
## Bug (dogfood-r2 fp9 Check F)

Cross-restart truncation is broken: the marker IS persisted, but the
re-apply at `src/stores/store.ts` keys on `b.id === marker.blockId`.
When the user truncated a freshly-sent prompt, the marker captured the
renderer-local-echo id (e.g. `u-mogrjvqb-ahyl` from `nextLocalId()` in
`InputBar`). Post-restart `framesToBlocks(JSONL)` re-projects with
`u-<uuid>` ids; no match → silent no-op → entire transcript reappears.

User-visible: truncate doesn't survive an app restart.

## Reproduction

`scripts/harness-restore.mjs --only=truncate-survives-restart`
(harness-restore preferred per `feedback_e2e_prefer_harness.md`).

2-launch case: plant a 4-turn JSONL fixture, simulate the local-echo
truncation flow, close, relaunch, assert prompts after the cut don't
reappear. Reverse-verified — fails on origin/working HEAD with all 4
prompts visible after restart.

## Fix

Extend the marker schema with two optional content-anchor fields:

```
{ blockId, truncatedAt, userTurnIndex?, textPrefix? }
```

`rewindToBlock` now also persists `userTurnIndex` (count of `kind:'user'`
blocks at and including the cut) and `textPrefix` (first 80 chars of the
target user message, whitespace-normalized).

`loadMessages` re-apply tries the blockId match first (cheap, exact for
`u-<uuid>` markers), then falls back to "Nth user block whose textPrefix
matches" when blockId misses. Inclusive-cut semantics preserved.

Legacy markers without anchor fields keep working through the blockId
match path — schema-additive, no migration needed.

## Trade-off considered

Alternative: swap the local-echo id with the JSONL UUID once the frame
arrives — single source of truth, cleanest fix. Rejected because
`streamEventToTranslation` explicitly skips user echoes (to avoid
duplicates with InputBar's local echo), so there's no renderer-side
event at which the UUID becomes available to swap into the existing
local-echo block. Implementing that would require a separate JSONL→
in-memory reconciliation pass, which is a much larger surface change
than the content-anchor approach. The content anchor is the smallest
change that closes the data-invariant gap.

## Files changed

- `src/stores/store.ts` — `rewindToBlock` persists anchor; `loadMessages`
  applies anchor fallback; new `anchorTextPrefix` helper.
- `electron/main.ts` — `truncation:set/get` validate + pass through
  optional anchor fields, sanitize bad shapes.
- `electron/preload.ts` + `src/global.d.ts` — type extension.
- `scripts/harness-restore.mjs` — new `truncate-survives-restart` case.

## Test plan

- [x] `node scripts/harness-restore.mjs --only=truncate-survives-restart` passes
- [x] Reverse-verify: stash fix → e2e fails (bug present) → pop → green
- [x] `npx vitest run tests/user-block-hover-menu.test.tsx` — 7/7 pass
- [x] `npm run build` clean


Jiahui-Gu (commented):
Reviewer (pool-5) APPROVE. Layer 1: bug real + user-visible; content-anchor sound (id-swap not viable per streamEventToTranslation at src/agent/stream-to-blocks.ts:267); userTurnIndex+textPrefix(80) robust through adjacent-same-prefix/empty/rewind. Residual: just-sent-not-yet-flushed truncation no-ops cross-restart, same as current behavior, non-regression. Layer 2: clean 5-file diff, no store.ts collision, e2e covers both leak and over-cut, typecheck clean. E2E independent reverse-verify skipped (pool-5 no out/ bundle).

PR #404, opened 2026-04-27
docs(dogfood): r2 fp11 i18n+markdown+long-output — 5 PASS / 1 FAIL (long-line wrap)

Jiahui-Gu:
## Summary

Dogfood r2 fp11 — exercises Chinese prompt + markdown rendering + long-output handling against the installed CCSM.exe production bundle (no rebuild) via Playwright + Agent Maestro proxy.

| Check | Topic | Verdict |
|-------|-------|--------|
| A | Chinese prompt + Chinese reply | PASS |
| B | Markdown elements (h1/ul/ol/code/quote/bold/italic/inline) | PASS |
| C | Long output (200 lines) — auto-stick + jump-to-latest | PASS |
| D | Long markdown expand affordance | PARTIAL (no show-more on assistant blocks) |
| E | Mixed CJK + English + Japanese | PASS |
| F | 500-char single-line wrap | **FAIL** |

## Key finding (F)

Long unbreakable single-line response (`a`*500) breaks chat stream layout: `scrollWidth=4297` vs `clientWidth=1006` (4x horizontal overflow). `<p>` has `word-break: normal` + `overflow-wrap: normal` + `white-space: pre-wrap` which never breaks unbreakable runs. Visual confirmation in `docs/screenshots/dogfood-r2/fp11-i18n-md-long/check-f-long-line.png`.

Recommended fix (NOT applied — dogfood scope is report only): add `overflow-wrap: anywhere` (or `word-break: break-word`) to the assistant prose `<p>`. CodeBlock has its own scroll; only plain prose is affected.

## Observation (D)

Assistant markdown blocks have no expand/collapse affordance (Bash tool blocks already do — fp8). Long markdown is rendered inline at full height; outer chat container scrolls. Possibly intentional, surfaced for UX consideration.

## Files

- Probe: `scripts/probe-dogfood-r2-fp11-i18n-markdown-long.mjs`
- Report: `docs/dogfood-r2-fp11-report.md`
- Screenshots: `docs/screenshots/dogfood-r2/fp11-i18n-md-long/check-{a..f}*.png`
- Logs + snapshots: `dogfood-logs/r2-fp11/`

## Setup notes

- Sanitized HOME (`feedback_probe_skill_injection.md`).
- Set BOTH `CCSM_CLAUDE_CONFIG_DIR` and `CLAUDE_CONFIG_DIR` (fp10 gotcha).
- Real ccsm store schema used (`messagesBySession[sid]`, etc.).
- Isolated user-data dir wiped per-run.

[...]

PR #405, opened 2026-04-27
fix(chat): wrap unbreakable long lines in assistant prose (#fp11-F)

Jiahui-Gu:
## Bug

fp11 dogfood Check F (PR #404 findings.json): user prompt `output a single line of 500 'a' characters` produces an assistant block whose `scrollWidth=4297` vs `clientWidth=1006` — ~4x horizontal overflow that pushes the entire chat row sideways and breaks the no-bubble density philosophy.

## Repro

1. Send any single-line response with no whitespace (500 'a' chars, a long URL, a long hash, a base64 blob).
2. Observe the chat row gains a horizontal scrollbar; the run does not wrap.

## Root cause

`AssistantBlock.tsx` rendered the markdown `<p>` with `whitespace-pre-wrap` only. That preserves newlines but the default `word-break: normal` + `overflow-wrap: normal` will not break a run that contains no break opportunity. The flex parent (`flex gap-3` + inner `min-w-0`) then stretches to fit the unbreakable child.

## Fix

Add `[overflow-wrap:anywhere]` to the prose `<p>` className.

**Why `anywhere` over `break-word`:** `anywhere` contributes to min-content sizing, so the inner flex child with `min-w-0` actually shrinks below the long word's width. `break-word` only breaks once the line genuinely overflows AND does not help min-content — leaves flex parents stretched in some configurations. Both are equivalent visually on prose; `anywhere` is the safer choice in this layout.

**Why only `<p>`:** `<pre>` blocks keep `overflow-x-auto whitespace-pre` — code should scroll, not break mid-token. Inline `<code>` is unchanged. Tables already wrap their own `overflow-x-auto`. Long URLs in `<a>` inside `<p>` inherit the rule, which is what we want.

## E2E

New `harness-ui` case: `assistant-long-line-wraps`.

- Seeds one assistant message with `'a'.repeat(500)`.
- Asserts `assistant-body` block + inner `<p>`: `scrollWidth <= clientWidth + 2`.
- Asserts computed `overflow-wrap === 'anywhere'` (catches anyone replacing the rule with a no-op or removing it).

**Reverse-verified:** stashed the AssistantBlock change, rebuilt, ran the case — FAIL with `scrollWidth=4235 / clientWidth=1044 / overflow-wrap=normal` (same signature as the dogfood probe). Restored the fix — PASS with `block 1044/1044 p 1020/1020 overflow-wrap=anywhere`.

Also re-ran nearby cases `chat-user-assistant-contrast` + `type-scale-snapshot` — both green, no regression.

## Test plan

- [x] `node scripts/harness-ui.mjs --only=assistant-long-line-wraps` passes
- [x] Reverse-verify: case fails on origin/working HEAD with the dogfood overflow signature
- [x] `chat-user-assistant-contrast` + `type-scale-snapshot` still green
- [ ] Manual: Chinese / Japanese / mixed-script prose still wraps cleanly (fp11 Check A + E remain PASS — `overflow-wrap: anywhere` does not change CJK wrapping which already had break opportunities)

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -15,6 +15,7 @@
 //   - popover-cross-dismiss                (popover-mutex / #221)
 //   - type-scale-snapshot                  (#225, 4-step type token system)
 //   - chat-user-assistant-contrast         (#345, dogfood: user vs assistant rail)
+//   - assistant-long-line-wraps            (fp11 Check F, long unbreakable run wraps)
 //   - banner-i18n-toggle                   (banner i18n)
 //   - toast-a11y                           (#298 follow-up)
 //   - cwd-popover-recent-unfiltered        (probe-e2e-cwd-popover-recent-unfiltered)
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Reviewer (pool-3) APPROVE. Layer 1: anywhere vs break-word choice correct (per CSS spec only anywhere contributes to min-content sizing inside flex+min-w-0); inline code cascades anywhere acceptable; long URL break trade-off documented. Layer 2: diff minimal (1 className + 1 e2e), no collisions, e2e renders real Electron renderer via win.evaluate, asserts scrollWidth + computed overflow-wrap, reverse-verify recipe inline; typecheck clean. E2E independent reverse-verify SKIPPED (no out/ bundle in pool-3) — worker provided concrete FAIL/PASS numbers (4235/1044 → 1044/1044). Follow-up (non-blocking): UserBlock.tsx:171 has same vulnerability — user paste 500-char URL hits same overflow.

PR #407, opened 2026-04-27
docs(dogfood): r2 fp12 settings live-apply — all green

Jiahui-Gu:
## Summary

- New probe \`scripts/probe-dogfood-r2-fp12-settings-live.mjs\` exercises the Settings → Appearance pane against the installed CCSM.exe with isolated user-data + isolated \`CLAUDE_CONFIG_DIR\`.
- All 6 checks PASS — theme, language, and font-size all live-apply (<100ms) with no restart, and all three persist cleanly across a full app restart.
- Notifications + Connection + Updates panes also behave correctly.

## Verdict per check

| Check | Verdict | Headline |
|---|---|---|
| A. theme switch (Dark/Light/System) | **PASS** | applied in 67–95ms, html class flip + body bg change confirmed |
| B. language toggle (en/zh/system) | **PASS** | sidebar + settings strings switch within ~500ms |
| C. font-size slider (12/14/16) | **PASS** | \`--app-font-size\` CSS var updates instantly, no layout break |
| D. notifications enable + sound | **PASS** | flipped & persisted across Settings close+reopen |
| E. connection / updates tabs | **PASS** | both render, 0 page errors |
| F. cross-restart persistence | **PASS** | theme=light + fontSize=16 + lang=zh all stuck after restart |

Full details + screenshots in \`docs/dogfood-r2-fp12-report.md\`.

## Test plan

- [x] Probe runs end-to-end against installed CCSM.exe
- [x] Isolated user-data wiped per run, isolated CLAUDE_CONFIG_DIR (HOME-leak sanitized)
- [x] Real ccsm store schema (\`useStore\` + \`notificationSettings\` slice)
- [x] Per-check PASS/FAIL + timing data
- [x] Screenshots under \`docs/screenshots/dogfood-r2/fp12-settings-live/\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #417, opened 2026-04-27
fix(chatstream): hoist Footer so streaming ticks don't reset thinking-dots animation

Jiahui-Gu:
## Summary

Fixes #407.

PR #413 (perf-1, ChatStream virtualization with react-virtuoso) defined `Footer` inside the `ChatStream()` function body. Every render produced a fresh component identity for `<Virtuoso components={{ Footer }} />`, which caused Virtuoso to remount the Footer subtree on every store tick. During streaming the store updates ~10 times per second, so the `thinking-dots` framer-motion `motion.div` restarted its opacity-pulse cycle on each tick, looking like a broken/jittery indicator.

## Repro (pre-fix)

`node scripts/harness-ui.mjs --only=chatstream-footer-stable` on `working` HEAD:

```
[case=chatstream-footer-stable] FAIL: #407 regression: thinking-dots DOM node was replaced
across store ticks — Footer is remounting on every render (likely defined inside
ChatStream() function body), which restarts the opacity-pulse animation cycle.
```

## Fix

- Hoist `Footer` to module scope in `src/components/ChatStream.tsx` so its component identity is stable across renders.
- Pass dynamic state (`loadError`, `showThinkingDots`, `activeId`, `onRetryLoad`, `thinkingLabel`) through Virtuoso's `context` prop. Per `react-virtuoso` types (`Footer?: ComponentType<ContextProp<Context>>`), Virtuoso forwards `context` into Footer as a prop, so dynamic state still flows in without remounting the subtree.
- Memoize `onRetryLoad` and `footerContext` so context identity only changes when its inputs do.
- Extend the jsdom react-virtuoso shim (`tests/setup.ts`) to forward `context` into the Footer mock so existing component tests keep passing.

## Verification

- `node scripts/harness-ui.mjs --only=chatstream-footer-stable` — passes (case asserts the dots DOM node stays the same across 5 sequential store ticks).
- `npm test` — chatstream-thinking-dots suite (5/5) green; full suite green except 3 pre-existing `electron/__tests__/db-hardening.test.ts` failures caused by `better-sqlite3` native binding mismatch in this dev env (`--ignore-scripts` install workaround for unrelated `@nodert-win10-au` build failure). Same 3 failures reproduce on `origin/working` with no diff applied.
- Manually launched the app and confirmed the thinking dots now pulse smoothly while the agent is streaming, with no visible reset on each token-tick (pre-fix the dots flashed on/off rapidly during streaming because the animation kept restarting from opacity 0.3).

## Test plan

- [x] new harness case `chatstream-footer-stable` fails on pre-fix code, passes after fix
- [x] existing `tests/chatstream-thinking-dots.test.tsx` still passes
- [x] manual: dots animate smoothly during real streaming session

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -3794,6 +3794,95 @@ async function caseStartupPaintsBeforeHydrate({ win, log }) {
   );
 }
 
+// ---------- chatstream-footer-stable ----------
+// Regression for task #407 (ChatStream perf-1 follow-up). Pre-fix: `Footer`
+// was defined inside the `ChatStream` function body, so every render produced
+// a fresh component identity for `<Virtuoso components={{ Footer }} />`,
+// which remounted the Footer subtree on every store tick (~10 Hz while
+// streaming). The thinking-dots `motion.div` therefore restarted its
+// opacity-pulse cycle on every tick, looking like a stuck/broken animation.
+// Fix: hoist `Footer` to module scope and pass dynamic state via Virtuoso's
+// `context` prop. This case pins the contract: the thinking-dots DOM node
+// MUST persist across multiple store ticks while the streaming/thinking
+// condition is unchanged.
+async function caseChatStreamFooterStable({ win, log }) {
+  // Seed an active session in the "thinking" state (running, last block is
+  // user, no permission gate) so showThinkingDots resolves to true.
+  await win.evaluate(() => {
+    const store = window.__ccsmStore;
+    store.setState({
+      groups: [{ id: 'g1', name: 'G1', collapsed: false, kind: 'normal' }],
+      sessions: [
+        { id: 's1', name: 'thinking-session', state: 'idle', cwd: 'C:/x', model: 'claude-opus-4', groupId: 'g1', agentType: 'claude-code' },
+      ],
+      activeId: 's1',
+      messagesBySession: { s1: [{ kind: 'user', id: 'u1', text: 'hello' }] },
+      runningSessions: { s1: true },
+      startedSessions: { s1: true },
+    });
+  });
+
+  // Wait for the dots to appear.
+  await win.waitForFunction(
+    () => !!document.querySelector('[data-testid="chat-thinking-dots"]'),
+    null,
+    { timeout: 5_000 }
+  );
+
+  // Tag the initial dots node so we can detect remounts: if Footer remounts,
+  // React replaces the DOM element entirely and our marker disappears with
+  // the old node.
+  await win.evaluate(() => {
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
APPROVE WITH NIT (cannot self-approve via gh — leaving as comment).

Confirmed bug on origin/working: `Footer` was defined inside `ChatStream()` body (line 178), so `<Virtuoso components={{ Footer }}/>` got a fresh component identity each render. Hoist + Virtuoso `context` is the canonical react-virtuoso pattern for this — preferred over `useMemo`/`React.memo` workarounds because Virtuoso's `components.Footer` slot triggers a full subtree remount on identity change regardless of memo wrappers. `FooterContext` shape is correct: every dynamic dep (`loadError`, `showThinkingDots`, `activeId`, `onRetryLoad`, `thinkingLabel`) is plumbed through the context, no leaked closures over `t()` or `loadMessages` (both captured via the memoized `onRetryLoad` and the resolved `thinkingLabel` string). a11y `aria-label` and `data-testid` survived the move intact.

## Layer 2 — implementation correctness

**Reverse-verify (probe legitimacy):** YES — restored `src/components/ChatStream.tsx` to origin/working only, kept new probe + `tests/setup.ts` shim, ran `node scripts/harness-ui.mjs --only=chatstream-footer-stable`. Result: FAIL with the exact #407 message. Restored fix → PASS in 605 ms. Probe is real, not fake-green.

**Quality gates:**
- `npm run lint` — 2 errors in ChatStream.tsx (`'React' is not defined` at lines 21 and 47) but both come from the pre-existing `forwardRef` blocks; verified identical lines exist on origin/working. Not introduced here.
- `npm test` — 865/868 pass; the 3 failures are `electron/__tests__/db-hardening.test.ts` NODE_MODULE_VERSION 130 vs 137 native-binding mismatch — same env issue worker called out, reproduces on origin/working.

**Shim safety:** Only `src/components/ChatStream.tsx` imports `react-virtuoso`. The `tests/setup.ts` change (forward `context` to the Footer mock) cannot affect any other test.

**Scope / English / drive-bys:** Three files touched, all on-topic. CJK grep of `git log -p origin/working..HEAD` is empty. Comments are necessary, not chatty.

## NIT (non-blocking)
Per visual-fix screenshot policy, this is a visual-perception fix and the PR body is missing before/after captures (or a short screen recording) of the thinking-dots smoothness. The DOM-identity probe is a strong proxy for "no remount", but a recording would also confirm the user-perceived pulse cadence. Please attach one as a follow-up comment when convenient — not blocking the merge.

LGTM.

PR #422, opened 2026-04-27
feat(notify): replace nodert with Electron Notification API

Jiahui-Gu:
## Why

Production Windows toast had never actually fired in any shipped CCSM version.

- `@nodert-win10-au/*` requires C++17 `/ZW` (C++/CX), but Electron 33's V8
  headers `#error \"C++20 or later required\"` -- the native chain has been
  architecturally unbuildable since the upgrade.
- The wrapper caught the runtime require failure and silently flipped to
  no-op fallback. The `__setNativeForTests` spy seam also bypassed the
  broken require in the only test that exercised the real chain
  (`notify-ship.test.ts`), so the failure never surfaced.
- Every "popup" users observed in e2e was the in-app React `Toast.tsx`
  banner, NOT an OS toast. Three notification PRs in a row hit dead ends
  trying to make the native chain build.

This PR drops the entire native chain and uses Electron's built-in
`Notification` API, which routes through `ToastNotificationManager`
underneath the same AUMID (`com.ccsm.app`) that the Start Menu shortcut
already stamps.

## What changes

**Deleted (~1900 LoC):**
- `electron/notify-impl/` (whole directory: `notifier.ts`, `registry.ts`,
  `types.ts`, `index.ts`, `platform/{windows,darwin,linux}.ts`,
  `xml/{common,done,permission,question}.ts`, `util/truncate.ts`)
- `electron/__tests__/notify-ship.test.ts`
- `electron/__tests__/notify-platform-stubs.test.ts`
- `tests/notify-wrapper.test.ts`
- `scripts/probe-notify-packaging.mjs`
- `caseNotifyFallback` from `scripts/harness-restore.mjs` (its premise --
  stash `node_modules/electron-windows-notifications` to force fallback --
  no longer applies; Electron's API is built into the runtime)

**Refactored:**
- `electron/notify.ts` 280 -> ~200 LoC. Drops `loadModule` /
  `__setNotifyImporter` lazy-require machinery and the `__setNativeForTests`
  spy seam. Public API surface preserved byte-for-byte (`configureNotify`,
  `notify{Permission,Question,Done,Dismiss}`, `disposeNotify`,
  `isNotifyAvailable`, `probeNotifyAvailability`, `notifyLastError`).
  `isNotifyAvailable()` now reads `Notification.isSupported()`. Live toasts
  are tracked in a bounded `Map<toastId, Notification>` so `notifyDismiss`
  can call `notification.close()`. `click` events route through the
  existing `onAction({ action: 'focus', ... })` callback.
- `electron/notifications.ts`: drops the obsolete "no cross-platform
  fallback" comment block, adds title/body composition per event type
  (permission / question / turn_done / test) per spec, including question
  body truncation at 200 chars and turn_done user/assistant lines truncated
  at 80 chars.
- `electron/__tests__/notifications.test.ts`: rewritten to mock Electron's
  `Notification` constructor via `vi.mock('electron', ...)` instead of the
  removed `__setNotifyImporter` seam. Keeps all 9 routing / focus /
  truncation / dismiss cases.
- `electron/__tests__/notify-bootstrap.test.ts`: drops
  `__setNotifyImporter` import; adds a stub `Notification` class to the
  mocked `electron` module.
- `package.json`: removes `electron-windows-notifications` from
  `dependencies`, removes all 5 `@nodert-win10-au/*` from
  `optionalDependencies`, removes both glob patterns from `build.files` and
  `build.asarUnpack`.
- `scripts/postinstall.mjs`: drops the Windows-only notify-native-chain
  postinstall verification (no longer relevant).
- `scripts/after-pack.cjs`: drops the post-package native-chain assertion
  block. SDK binary check unchanged.
- `scripts/harness-agent.mjs`: rewrites `caseNotifyIntegration` to monkey-
  patch `Notification.prototype.show` and capture title/body/click-handler
  from each constructed instance. Same coverage (perm/question/done emit,
  click activation routing, focus suppression, rich done truncation), no
  reliance on the removed test seam.
- `README.md` notification paragraph: rewritten to describe the built-in
  Electron path.

**Kept unchanged:**
- `electron/notify-bootstrap.ts` (AUMID + Start Menu shortcut + toast-
  target router still required)
- `scripts/setup-aumid.ps1`
- `app.setAppUserModelId('com.ccsm.app')` call in `electron/main.ts`
- `electron/__tests__/notify-bootstrap.test.ts` (apart from dropping the
  test-seam import)

## Trade-offs accepted

- **No inline action buttons** (Allow / Allow always / Reject). Permission
  approval stays in the main-window Radix dialog where the user can see
  the actual command + cwd context before deciding. The toast-on-button
  path was dangerous anyway -- a user clicking "Allow" on a toast had
  zero visibility into what they were allowing.
- **No Group/Tag dedupe.** The three event types (permission / question /
  turn_done) don't dedupe each other, so this is effectively a no-op trade.

## Test plan

- [x] `npm run typecheck` -- clean
- [x] `npm run lint` -- 44 pre-existing errors in `verification/*.mjs`
      (no-undef on browser/node globals, present on `origin/working` too);
      0 new errors / 0 new warnings introduced by this PR
- [x] `npm test` -- 854 passing / 3 pre-existing `db-hardening.test.ts`
      failures from a `better-sqlite3` Node ABI mismatch on Node 24
      (also fails on `origin/working`; environmental, unrelated)
- [x] `vitest run electron/__tests__/notifications.test.ts
      electron/__tests__/notify-bootstrap.test.ts` -- 19/19 pass
- [x] **E2E (the smoking gun)**: built `release/win-unpacked/CCSM.exe`
      with `npx electron-builder --win --dir`, confirmed
      `app.asar.unpacked/node_modules` now only contains `@anthropic-ai`
      + `better-sqlite3` (no `electron-windows-notifications` /
      `@nodert-win10-au` dirs). Cleared toast history with
      `ToastNotificationManager.History.Clear('com.ccsm.app')`, launched
      packaged CCSM via playwright + fired
      `new Notification({ title: 'CCSM e2e probe toast', body: '...' }).show()`,
      then ran:

      ```powershell
      [Windows.UI.Notifications.ToastNotificationManager,Windows.UI.Notifications,ContentType=WindowsRuntime] | Out-Null
      $h = [Windows.UI.Notifications.ToastNotificationManager]::History.GetHistory('com.ccsm.app')
      ($h | Measure-Object).Count
      $h | ForEach-Object { $_.Content.GetXml() }
      ```

      Output:
      ```
      AFTER count (com.ccsm.app): 1
      ---
      <toast><visual><binding template=\"ToastText02\"><text id=\"1\">CCSM e2e probe toast</text><text id=\"2\">If you see this in toast history, Electron Notification is wired correctly.</text></binding></visual><audio silent=\"true\"/></toast>
      ```

      First time GetHistory has returned a non-zero count for
      `com.ccsm.app` since the project began.

## Out of scope / follow-ups

- Renderer-side `dispatchNotification` (in `src/`) still gates
  `notify:availability` IPC -- the new wrapper always reports `true` on
  Windows/macOS/Linux now, so the Settings dialog's availability indicator
  becomes redundant. Could simplify in a follow-up but no behavior change.
- `notify-bootstrap`'s `createDefaultToastActionRouter` still has
  `'allow' | 'allow-always' | 'reject'` branches that are now dead code
  (only `'focus'` ever fires). Worth a separate cleanup PR -- removing
  the dead arms wasn't necessary to ship this PR.
- The `disposeNotify` export is unused in the codebase but kept for API
  parity with the old surface.


Jiahui-Gu (commented):
LGTM (own-PR comment in lieu of approve). Two-layer review.

1. **Architectural wall is real.** `docs/notify-e2e-r1-report.md` + `r3-report.md` document a reproducible `_nodert_generated.cpp` C1107 / `/std:c++20` vs `/ZW` C++/CX collision against Electron 33's V8 headers. `@nodert-win10-au` last published 2018, no upstream fix coming. Keeping nodert was not viable.
2. **Lost capabilities are not used in product.** Inline action buttons (`allow|allow-always|reject` toast arms) and Group/Tag dedupe are dead arms — `notify-bootstrap.createDefaultToastActionRouter` only ever gets `'focus'` from the new path; PR body explicitly flags the dead arms as a follow-up. Permission flow goes through the in-app Radix dialog, not toast buttons.
3. **`caseNotifyFallback` deletion justified.** The case literally renames `node_modules/electron-windows-notifications` to force `MODULE_NOT_FOUND` — premise gone with the dependency. There's no fallback path anymore: Electron's `Notification` is built into the runtime.

## Layer 2 — implementation

1. **Public API preserved.** Three call sites (`main.ts`, `notifications.ts`, `notify-bootstrap.ts`) all import names that still exist with compatible signatures. `main.ts` only uses `probeNotifyAvailability` + `notifyLastError` for the `notify:availability` IPC handler at line 1390 — still works.
2. **Zero residue** in `electron/*.ts` / `scripts/*.{mjs,cjs}` for `notify-impl|nodert|electron-windows-notifications|loadModule|__setNativeForTests|__setNotifyImporter` outside the historical comment block in `notify.ts:4-9`. `package.json` + `package-lock.json` zero hits — `dependencies`, `optionalDependencies`, `build.files`, `build.asarUnpack` all clean.
3. **Tests cover original surface.** `notifications.test.ts` 9 tests pass: perm/question/done routing, focus suppression + test bypass, `isSupported=false` fallback, dismiss-when-registered + no-op-when-not. Click-handler activation isn't directly fired in the unit test, but is exercised end-to-end in `caseNotifyIntegration` (line 3805 manually invokes the captured click handler and asserts the router observes the focus event).
4. **after-pack / postinstall cleanup is consistent.** Both files lose nodert-chain assertions; Node-version warning copy updated to drop the C++/CX hint.
5. **README updated** to remove inline-button language and describe the built-in Electron path.
6. **`notify-bootstrap.test.ts`** drops `__setNotifyImporter` import + adds a `Notification` stub class with `isSupported`/`on`/`show`/`close`. 10/10 pass.

## E2E reverse-verify

- Read the rewritten `caseNotifyIntegration` (`scripts/harness-agent.mjs:3650-3849+`). It patches `Notification.prototype.show` at the real Electron API boundary inside a real packaged app context (`app.evaluate`) — the wrapper still goes through `new Notification(...).show()`, so this is a true production-path probe, not a vi-mock seam.
- GetHistory smoking gun reported in PR body (count=1, XML containing `CCSM e2e probe toast` text after launching packaged build with cleared toast history) is the strongest possible evidence the production path is wired.
- Re-ran `vitest run electron/__tests__/notifications.test.ts electron/__tests__/notify-bootstrap.test.ts` → **19/19 pass**.
- `npx tsc --noEmit` → **clean**.
- Did not re-build the installer locally (>15 min cost) — relying on the reported GetHistory evidence per review-checklist allowance.

## Notes

- Follow-up cleanups already enumerated by author (dead `'allow'/'allow-always'/'reject'` arms in `notify-bootstrap`, redundant Settings availability indicator) — fine to defer.
- Nice `LIVE_TOAST_LIMIT = 256` ceiling + `close`-event eviction; defends against runaway emit loops without leaking Notification handles.

Jiahui-Gu:
Post-rebase review (independent reviewer): APPROVE.

Conflict resolutions:
- scripts/after-pack.cjs: sound. The SDK-binary-in-app.asar.unpacked check is fully preserved and still throws on missing — not a no-op stub. Dropping the win32 nodert/electron-windows-notifications native-module check is consistent with this PR removing that whole native chain.
- scripts/harness-restore.mjs: sound. caseNotifyFallback function and its cases-array entry both removed; zero residual references in scripts/src/electron; node --check on the file passes; cases array remains well-formed.

eslint amend (Electron: 'readonly' global):
- Correctly scoped to the electron/**/*.ts block only (line 112). Does NOT leak to renderer/src.
- Nit (non-blocking): a named 'import type { NotificationConstructorOptions } from "electron"' in electron/notify.ts would arguably be a cleaner fix than declaring 'Electron' as a global, since the namespace is normally pulled in via type imports. That said, the current approach matches the existing project pattern (NodeJS is declared as a global the same way for NodeJS.Timeout etc.), so consistency wins. Fine to leave as-is.

Quality gates:
- npm run typecheck: clean.
- npm run lint: 45 errors / 10 warnings — all pre-existing in verification/*.mjs (no-undef on window/process/console in Node verification scripts). Matches the rebaser's claimed baseline; no new errors introduced by this PR.
- vitest electron/__tests__/notifications.test.ts + notify-bootstrap.test.ts: 19/19 passing.

No blockers.

PR #423, opened 2026-04-27
perf(chat): memoize AssistantBlock to skip markdown re-parse on store ticks

Jiahui-Gu:
## Symptom
During assistant streaming, every store tick (new token, tool block update, permission popup) triggers a `ChatStream` re-render. Each re-render forced every visible historical `AssistantBlock` to re-parse + re-highlight its markdown:

- `<ReactMarkdown text={text}>` with an inline `components={{...}}` literal -> reference changed every render -> react-markdown re-walks AST and re-renders all custom renderers (including `<CodeBlock>` with prism highlighting).
- `AssistantBlock` itself had no `React.memo`, so the parent's re-render propagated unconditionally.

Measured cost (scout): 5-30ms per visible block per tick. With 5-10 visible blocks during streaming this is 50-300ms of cumulative work per streamed token, perceptible as choppy streaming and laggy scroll-back.

## Fix
1. Hoist the `components` map (previously inline at lines 58-120) to a module-scope `MD_COMPONENTS: Components` constant. Same for `remarkPlugins` (now `REMARK_PLUGINS`). Stable references mean react-markdown can short-circuit its internal work.
2. Wrap the export with `React.memo`. Equality compares the only props that affect output:
   - `text`, `streaming`, `id` (primitives)
   - `viaSkill` reduced to a stable scalar via `viaSkill?.path ?? viaSkill?.name ?? null` (the object identity churns even when content is unchanged).
3. `CodeBlock` is already a stable module-level component, no change needed there.

No behaviour change; only re-render frequency drops. Same JSX, same DOM output.

## Verification
- `npm run typecheck` clean.
- `npm run lint` no new errors attributable to this file (pre-existing baseline noise in probe scripts unchanged).
- `npx vitest run tests/assistant-block-skill-badge.test.tsx` -> 3/3 pass (text + via-skill badge + plugin-namespaced badge).
- `npx vitest run tests/chatstream-tool-block-ux.test.tsx tests/chatstream-thinking-dots.test.tsx tests/chatstream-terminal.test.tsx` -> 18/18 pass (covers ChatStream interaction with assistant blocks).

E2E visual probe was not added: this is a pure memoization refactor with no rendered-output diff and the existing unit tests already exercise text rendering, the via-skill badge with both name shapes, and ChatStream's interaction with assistant blocks.

## Files changed
- `src/components/chat/blocks/AssistantBlock.tsx` — hoist components/plugins to module scope, rename inner fn to `AssistantBlockImpl`, export `React.memo(AssistantBlockImpl, ...)` as `AssistantBlock`.

### DIFF
--- a/src/components/chat/blocks/AssistantBlock.tsx
+++ b/src/components/chat/blocks/AssistantBlock.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import ReactMarkdown from 'react-markdown';
+import ReactMarkdown, { type Components } from 'react-markdown';
 import remarkGfm from 'remark-gfm';
 import { Copy, Check } from 'lucide-react';
 import { CodeBlock } from '../../CodeBlock';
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## Review: APPROVE (self-PR, posting as comment)

- Props covered: AssistantBlockProps = { id, text, streaming, viaSkill }. Memo equality compares all four (viaSkill via stable scalar key). No prop is unaccounted for.
- Hoisted MD_COMPONENTS captures no closure: every renderer uses Tailwind class strings only. No t(), theme, hook, or props references inside. CodeBlock is already a module-level import. Hoist is safe — output cannot go stale.
- viaSkillKey collapse risk: SkillProvenance only has name (required) + optional path. Both fields are surfaced (name in the badge text, path in the tooltip). Key is path ?? name. In practice path is derived from name (same skill name → same path) so they vary together; safe.

Layer 2 (implementation):
- Single call site (src/components/chat/renderBlock.tsx:30) passes id / text / streaming / viaSkill — exactly the props memo equality covers.
- npm run typecheck clean.
- npm run lint: 55 problems / 45 errors — identical to working baseline (verified by checking out origin/working AssistantBlock.tsx and re-running lint). No new lint debt.
- Targeted vitest: assistant-block-skill-badge (3) + chatstream-tool-block-ux (11) + chatstream-thinking-dots (5) + chatstream-terminal (2) = 21/21 pass.

Reverse-verify probe (mounted 5 stable + 1 changing AssistantBlock under a parent forced through 100 re-renders, counted inner-body executions via a useTranslation spy):
- With memo (this PR): 6 mount renders + 100 deltas = only the 1 changing instance re-rendered per tick. 5 stable instances each rendered once at mount, zero times across 100 parent ticks.
- Without memo (working baseline, same probe): 6 mount + 600 deltas = all 6 instances re-rendered every tick.
- 6x reduction in inner-body executions. Perf claim validated.

Probe was a one-shot reverse-verify and has been deleted; not retained as a regression guard since it would couple to react-markdown internals.

PR #427, opened 2026-04-27
spike: ttyd + webview embed claude CLI (throwaway)

Jiahui-Gu:
## Summary

Throwaway 1-day spike to evaluate outsourcing the right-pane terminal to [ttyd](https://github.com/tsl0922/ttyd) (MIT, 11.5k★) instead of building a custom xterm.js + node-pty stack.

Background: scout #444/#445 confirmed VS Code's "useTerminal" mode just calls `vscode.window.createTerminal()` — VS Code has a full terminal subsystem we don't. Survey of embeddable alternatives found no drop-in React/Electron component; the closest "outsource the whole terminal" path is ttyd subprocess + `<webview>`.

## What's in the demo

- `spike/ttyd-embed/main.js` — Electron main: resolves `claude` via `where`, picks free port, spawns `ttyd.exe -W -t fontSize=14 <claude>`, reaps PTY tree on close via `taskkill /F /T`.
- `spike/ttyd-embed/index.html` — two-pane layout (left placeholder div for the React sidebar, right `<webview>` pointing at local ttyd).
- `spike/ttyd-embed/bin/ttyd.exe` — bundled ttyd 1.7.7 win32 binary (1.4MB).
- `spike/ttyd-embed/README.md` — Phase 1 browser smoke test + `npm install && npm start` instructions.

## Status

Draft. Not for merging — `spike/` is gitignored from the main build path. User will run the demo locally and decide whether the HYBRID architecture (left React sidebar + right ttyd webview) is worth committing to.

## Test plan

- [ ] Phase 1: `bin\ttyd.exe -p 7681 -W claude` in a regular browser shows working CLI prompt.
- [ ] `npm install && npm start` opens Electron window with left placeholder + right embedded claude.
- [ ] Close window cleanly reaps ttyd + claude processes (check Task Manager).

PR #428, opened 2026-04-27
feat(cliBridge): per-session ttyd lifecycle module + packaging

Jiahui-Gu:
## Summary

Worker 1 of the 4-worker ttyd refactor. Commits 1 and 5 of the single PR
covering the electron-side cliBridge module + electron-builder packaging
swap (drop bundled SDK binary, add ttyd.exe).

The renderer wiring (Worker 2), in-process SDK runner deletion (Worker
3), and `dependencies` cleanup (Worker 4) are independent follow-ups;
this PR builds + tests green standalone (the cliBridge IPC is unused by
the renderer until Worker 2 lands).

## Commit 1: `electron/cliBridge/`

One ttyd process per ccsm session, each wrapping the user's installed
`claude` CLI on a 127.0.0.1 ephemeral port. The renderer (Worker 2) will
mount an `<iframe src="http://127.0.0.1:<port>">` per session.

Files:
- `portAllocator.ts` -- `pickFreePort()` via the port-0 trick
- `claudeResolver.ts` -- `where claude.cmd` -> `where claude` cached
  fallback; null when neither resolves
- `ttydBinary.ts` -- dev: `<repo>/spike/ttyd-embed/bin/ttyd.exe`; prod:
  `process.resourcesPath/ttyd.exe`
- `processManager.ts` -- `Map<sessionId, {proc, port, sid, status}>`
  with spawn/resume/kill + `cliBridge:ttyd-exit` event emission;
  Windows tree-kill via `taskkill /F /T`
- `index.ts` -- IPC registration + sender bind

IPC surface (exposed under `window.ccsmCliBridge` to keep Worker 3's
SDK-IPC deletion a clean diff):
- `cliBridge:openTtydForSession(sessionId)` -> `{ok,port,sid}`
- `cliBridge:resumeSession(sessionId, sid)` -> `{ok,port,sid}`
- `cliBridge:killTtydForSession(sessionId)` -> `{ok,killed}`
- `cliBridge:checkClaudeAvailable()` -> `{available, path?}`
- event `cliBridge:ttyd-exit` -> `{sessionId, code, signal}`

Wired into `electron/main.ts`:
- `registerCliBridgeIpc()` at app ready
- `bindCliBridgeSender(win.webContents)` in `createWindow`
- `shutdownCliBridge()` in `before-quit` so `taskkill /T` reaps the
  ttyd -> conpty -> claude tree before the event loop tears down

### Q10A verified
`claude --help` confirms `--session-id <uuid>  Use a specific session
ID for the conversation (must be a valid UUID)`. We pre-allocate the
UUID on the ccsm side (Q10A); no JSONL directory polling fallback
needed.

## Commit 2 (test): harness case for ttyd spawn + HTTP probe + tree-kill

Added `cli-bridge-ttyd-spawn-and-kill` to
`scripts/harness-real-cli.mjs` per `feedback_e2e_prefer_harness` (avoids
the +30s cold-launch tax of a stand-alone probe).

Coverage: resolve binary -> resolve claude -> allocate port -> spawn
ttyd with the production flag set -> HTTP GET 127.0.0.1:<port>/ asserts
2xx -> taskkill /F /T -> tasklist confirms PID gone -> final HTTP probe
asserts port refuses connections (proves conpty + claude child were
reaped).

Local run: `PASS in 3.3s` on win32-x64 with `claude.cmd` on PATH.

## Commit 5: packaging

`package.json` `build`:
- Drop `**/node_modules/@anthropic-ai/claude-agent-sdk-*/**` from
  `asarUnpack` and the two `node_modules/@anthropic-ai/claude-agent-sdk*`
  entries from `files`. (Worker 4 removes the dep itself.)
- New `extraResources` entry: `spike/ttyd-embed/bin/ttyd.exe -> ttyd.exe`
  (lands at `process.resourcesPath/ttyd.exe`).

`scripts/after-pack.cjs`: repurposed from "verify SDK binary unpacked"
to "verify ttyd.exe landed at `<resources>/ttyd.exe`". Same fail-loud
philosophy. win32-x64 only for now; mac/linux branches kept (no-op).

`spike/ttyd-embed/bin/ttyd.exe`: force-added past the repo-root `*.exe`
gitignore (same approach as spike PR #427). ttyd 1.7.7-40e79c7
win32-x64, 1.4 MB, MIT.

Code signing for ttyd is intentionally NOT added (Q12: skipped for MVP).

## Test plan

- [x] `npx eslint electron/cliBridge/ electron/main.ts electron/preload.ts` 0 errors
- [x] `node scripts/harness-real-cli.mjs --only=cli-bridge-ttyd-spawn-and-kill` PASS
- [x] `npm test` -- 73/74 test files pass; the 1 failure is unrelated (`better-sqlite3` native binding compiled against Node 130 vs vitest's Node 137, env-level rebuild issue)
- [ ] Reviewer: full `make:win` to validate `extraResources` lands ttyd.exe at the right path and `after-pack.cjs` signs off
- [ ] Reviewer: reverse-verify the harness case by stashing
  `processManager.ts`'s taskkill branch -- ttyd PID should remain in
  tasklist after kill, case should fail


Jiahui-Gu (commented):
## R1 Review

- 5-file split is justified, not gold-plating: each file has a single concern (port / claude resolution / binary path / process lifecycle / IPC). Collapsing portAllocator (30 LOC) + ttydBinary (36 LOC) + claudeResolver (59 LOC) into one helpers.ts would save trivial LOC at the cost of test seams (`__resetClaudeResolverForTest`, `__getEntryForTest`). Keep as is.
- `window.ccsmCliBridge` namespace split: justified short-term. W3 will delete a bunch of `window.ccsm.agent*` surface; a separate namespace avoids merge conflict on the same `api` literal. Comment in preload.ts already notes "fold into ccsm once SDK runner is gone" — acceptable interim.
- `claude --session-id <uuid>` spawned directly via ttyd args matches Q3/Q10A. processManager.ts L121-129 confirms — UUID is pre-allocated by `openTtydForSession` (L192), passed straight through. No JSONL polling fallback added. Correct.
- Packaging correctly skips non-win32 (after-pack.cjs L31-35 hard-returns; ttydBinary.ts L25 returns null). Q11/Q12 honored. Mac/linux branches kept as no-ops with explicit comments — clean.
- No premature features. Resume vs new-session is the only branching, and both are needed for MVP.

### Layer 2 (impl)
- Tree-kill: `taskkill /F /T /PID` used in both processManager.ts L225 and harness L1031. Correct — `/T` walks ttyd → conpty → claude.
- Port TOCTOU window: present but explicitly acknowledged (portAllocator.ts L4-8). Spike pattern; accepted for MVP.
- ProcessManager Map lifecycle: `killTtydForSession` removes the entry (L219, L244) ✓. BUT the `proc.on('exit')` handler (L162-168) only sets `status='exited'` and never calls `sessions.delete(sessionId)`. Sessions that exit unexpectedly (claude crash, user Ctrl+D in browser) leave a stale entry in the Map until the renderer calls kill. Low-impact (entry holds only a dead ChildProcess reference) but unbounded if a user keeps re-opening sessions without explicit close. Suggest `sessions.delete(sessionId)` after `emitExit`.
- Resolver cache: caches `null` too (claudeResolver.ts L48 — if both lookups fail, `cached` is set to `null`, and L46 `cached !== undefined` returns the null forever). User installing claude post-launch never gets picked up. Behavior is documented as "restart ccsm event" (L24-27); acceptable for MVP but worth a follow-up `cliBridge:rescanClaude` IPC.
- Successful-path cache works: subsequent spawns hit cache (L46 short-circuit), no per-spawn `where` call.
- E2E case: real HTTP GET probe (L1009-1023), tasklist verification (L1043-1051), AND post-kill HTTP probe to confirm port released (L1054-1064). Stronger than just process-spawn check. Reverse-verify steps are listed in PR body checklist — manager should run.
- preload.ts cliBridge IPC appended at END (L459-503), after `exposeInMainWorld('ccsm', api)` on L455. Per W1 instructions ✓ — W3's deletes inside the `api` literal won't conflict.
- after-pack.cjs correctly checks `<resources>/ttyd.exe` (L54), with platform-key gate (L31), arch enum decoded (L27), and helpful error listing on miss. SDK-binary verification removed cleanly — no dangling references to `@anthropic-ai/claude-agent-sdk`.
- ttyd.exe blob is real PE32+ x86-64, mode 100644, 1409024 bytes (1.4 MB) — not LFS pointer. Force-add past `*.exe` gitignore worked.
- Vitest noise: better-sqlite3 NODE_MODULE_VERSION mismatch is a Node-version env issue (vitest's worker Node vs the rebuilt binding), unrelated to this PR. Confirmed — no cliBridge files touch sqlite.

### Minor
- PR body says ttyd.exe lives at `electron/cliBridge/bin/ttyd.exe`, but actual path everywhere (extraResources, ttydBinary.ts dev path, harness, blob) is `spike/ttyd-embed/bin/ttyd.exe`. Code is self-consistent; only the body description is stale. Cosmetic.
- `fileURLToPathLocal` wrapper in harness (L1085-1088) is dead-weight; just call `fileURLToPath(import.meta.url)` directly. Cosmetic.

### Verdict
APPROVE (with one nice-to-have follow-up)
- Top issue: ProcessManager Map leak on unexpected exit — `proc.on('exit')` should `sessions.delete(sessionId)` after emit, not just flip status. Not a blocker for landing this PR (ttyd-exit recovery via re-open works regardless), but worth fixing in this PR or W2's follow-up.
- Reviewer must still run `make:win` before merge per PR test plan checkbox 5.

PR #430, opened 2026-04-27
feat(ui): add TtydPane component

Jiahui-Gu:
## Summary
- New `src/components/TtydPane.tsx` mounts a ttyd iframe per session via `window.ccsmCliBridge`.
- Adds `src/cliBridge.d.ts` mirroring the preload `CCSMCliBridgeAPI` surface (renderer can't import from electron tree).
- No App.tsx wiring yet (W2c).

## Coordination
- Depends on: PR #428 (W1, MERGED) — uses cliBridge IPC.
- Sibling: W2b (ClaudeMissingGuide + i18n).
- Blocks: W2c (App wiring + e2e harness).

## Test plan
- [x] typecheck green (`tsc --noEmit`)
- [ ] component renders in W2c integration

### DIFF
--- a/src/cliBridge.d.ts
+++ b/src/cliBridge.d.ts
@@ -1,0 +1,40 @@
+// Renderer-side type augmentation for `window.ccsmCliBridge`. The
+// authoritative surface is defined in `electron/preload.ts` (exported as
+// `CCSMCliBridgeAPI`) but the renderer's Vite build cannot import from
+// the electron tree, so the shape is mirrored here. Keep both in sync.
+//
+// Per preload.ts: `ccsmCliBridge` is intentionally a separate namespace
+// from `window.ccsm` while the in-process SDK runner is being torn out.
+// It will fold into the main `ccsm` namespace once that work lands.
+
+type CliBridgeOpenResult =
+  | { ok: true; port: number; sid: string }
+  | { ok: false; error: string };
+
+type CliBridgeKillResult =
+  | { ok: true; killed: boolean }
+  | { ok: false; error: string };
+
+type CliBridgeAvailability =
+  | { available: true; path: string }
+  | { available: false };
+
+type TtydExitEvent = {
+  sessionId: string;
+  code: number | null;
+  signal: NodeJS.Signals | null;
+};
+
+declare global {
+  interface Window {
+    ccsmCliBridge?: {
+      openTtydForSession: (sessionId: string) => Promise<CliBridgeOpenResult>;
+      resumeSession: (sessionId: string, sid: string) => Promise<CliBridgeOpenResult>;
+      killTtydForSession: (sessionId: string) => Promise<CliBridgeKillResult>;
+      checkClaudeAvailable: () => Promise<CliBridgeAvailability>;
+      onTtydExit: (handler: (e: TtydExitEvent) => void) => () => void;
+    };
+  }
+}
+
+export {};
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## R2a Review

- Iframe-only renderer pane matches the locked Q3 architecture (ttyd spawns claude directly; renderer is just a viewport). Correct.
- Type-aug duplication via `src/cliBridge.d.ts` follows the **existing** pattern in `src/global.d.ts` (which also hand-mirrors electron types behind a "keep in sync" comment because the Vite renderer build can't reach the electron tree). Not a divergence; consistent with precedent.
- `Props = { sessionId: string }` matches the convention used by `InputBar`, `QuestionStickyHost`, `ContextPieChip`. Correct primitive choice.
- Hardcoded English placeholders are flagged as temporary in the file header docstring (W2c will i18n-ize). Acceptable, though I'd recommend the **PR body** also mention this explicitly so the deferral is visible at merge time.
- Scope is minimal — single component + single type aug. Nothing to split further.

### Layer 2 (impl)
- Lifecycle: `useEffect([sessionId])` opens new sid; cleanup closure captures `prevSid = sessionId` and kills it. Effect cleanup runs before next effect, so kill-of-prev → open-of-next ordering is correct. No window of double-ownership in renderer state.
- Race: `activeSidRef` guards both the openTtydForSession resolve path and the catch path. Stale resolves are correctly suppressed. Good.
- Iframe: no explicit `key={sessionId}`. Since `src` changes with sessionId React will reload the iframe, BUT this leaves the same `<iframe>` DOM node in place — some browsers reuse the connection / leave the old ttyd websocket dangling momentarily. **Recommend adding `key={sessionId}`** to force a clean unmount/remount on switch. Minor but free.
- onTtydExit: subscribed once with `[]` deps, reads `activeSidRef.current` for filtering, returns `unsubscribe`. Correct.
- Retry button: calls `open(sessionId)` directly, doesn't depend on sessionId change. Correct.
- Type aug: matches preload exactly (including the `resumeSession` extra method that `TtydPane` doesn't use yet). ✓
- No setInterval / no Map / no leak surface beyond the two subscriptions, both cleaned up.
- a11y: iframe has `title={`ttyd session ${sessionId}`}`. ✓
- **Tailwind palette divergence (nit)**: uses raw `text-neutral-400`, `bg-neutral-800`, `text-red-400`, `bg-black`, `border-neutral-700`, `ring-neutral-400`. Rest of the codebase uses design tokens like `text-fg-secondary`, `bg-bg-hover`, `border-border-subtle`, `ring-accent`. Classes will render (Tailwind core defaults), but this won't theme-switch with the rest of the UI. W2c is going to touch styling anyway, so OK to defer — flagging so it's not forgotten.

### Cross-check
- type aug matches preload (electron/preload.ts:484-498): ✓
- typecheck: ✓ for this PR's files. The only error is a pre-existing `react-virtuoso` missing-module error in `src/components/ChatStream.tsx`, unrelated to this PR.
- Files touched: exactly the 2 reported (`src/cliBridge.d.ts`, `src/components/TtydPane.tsx`). ✓
- Base: `working`. ✓

### Verdict
**APPROVE** (with two non-blocking nits for W2c follow-up)

Nits:
1. Add `key={sessionId}` to the iframe to force clean remount on session switch.
2. Swap raw `neutral-*` / `red-*` / `bg-black` for the project's design tokens when W2c does the styling pass.

PR #431, opened 2026-04-27
feat(ui): ClaudeMissingGuide + i18n keys

Jiahui-Gu:
## Summary
- New `src/components/ClaudeMissingGuide.tsx` full-screen page.
- New i18n keys: `claudeMissing.*`, `terminal.*` (en + zh).
- Parity test green.

## Coordination
- Depends on: PR #428 (W1, MERGED) — calls `cliBridge.checkClaudeAvailable`.
- Sibling: W2a (TtydPane component).
- Blocks: W2c (App wiring), W3 (full SDK delete).

## Test plan
- [ ] i18n-key-parity test green
- [ ] typecheck green

### DIFF
--- a/src/components/ClaudeMissingGuide.tsx
+++ b/src/components/ClaudeMissingGuide.tsx
@@ -1,0 +1,86 @@
+import React, { useState } from 'react';
+import { useTranslation } from '../i18n/useTranslation';
+
+/**
+ * Full-screen page rendered at boot when the `claude` CLI is not on PATH.
+ *
+ * Background: ccsm shells out to the user's locally-installed Claude CLI
+ * (no bundled binary in this code path — see W1/W3). When PATH lookup
+ * fails, the rest of the app cannot function, so we replace the entire UI
+ * with this guide rather than letting users dig through broken sessions.
+ *
+ * The re-check button calls `cliBridge.checkClaudeAvailable` so the user
+ * can install via npm in a separate terminal and resolve in-place without
+ * restarting the app. App wiring (deciding when to mount this vs. the
+ * normal shell) lands in W2c.
+ */
+type Props = { onResolved: () => void };
+
+export function ClaudeMissingGuide({ onResolved }: Props) {
+  const { t } = useTranslation();
+  const [checking, setChecking] = useState(false);
+  const installCommand = 'npm install -g @anthropic-ai/claude-code';
+
+  const onRecheck = async () => {
+    if (checking) return;
+    setChecking(true);
+    try {
+      // `window.ccsmCliBridge` is exposed by `electron/preload.ts` but the
+      // global type isn't published in `src/global.d.ts` yet (App-level
+      // wiring lands in W2c, which will own that declaration). Local cast
+      // keeps this PR scoped to the 3 files in the W2b spec.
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
## R2b Review

- Install command `npm install -g @anthropic-ai/claude-code`: VERIFIED against upstream docs (https://code.claude.com/docs/en/setup, "Install with npm" section). Package name is correct and officially supported. Note: the upstream-recommended path is now the native installer (`curl -fsSL https://claude.ai/install.sh | bash` / Windows PowerShell `irm https://claude.ai/install.ps1 | iex`); npm is listed under "Advanced installation options". Not blocking for MVP — npm works on every platform with one command — but worth a follow-up to surface the native installer as the primary suggestion (especially because ccsm targets Windows where `npm install -g` permission errors are common).
- "Re-check" UX: defensible for MVP, locked by spec. Fine as-is.
- No `claude --version` verify hint: minor polish gap, not blocking.
- Sentence case: en respects it (`Re-check`, `Install command`, `Claude CLI not found`). zh fine.
- i18n strings read naturally in both locales.

### Layer 2 (impl)
- i18n parity: PASS (`npm run vitest tests/i18n-key-parity.test.ts` 2/2 green).
- No deletes: confirmed via `gh pr diff` — only trailing-comma rebalancing on existing `noOutput` lines, no removed keys.
- File scope: exactly 3 files (`src/components/ClaudeMissingGuide.tsx` + en/zh locales). App.tsx, package.json, global.css, TtydPane.tsx all untouched.
- Component:
  - `useTranslation` used correctly; references `common.loading` which exists in both locales (verified line 31 in en + zh).
  - Re-check button: in-flight guard via `if (checking) return` + `disabled={checking}` + loading label swap. Good.
  - Failure path (`available: false`): silent — stays on the page, no toast / inline error. Defensible MVP since the user just retries, but worth a follow-up to add inline "Still not found" feedback so it doesn't feel broken.
  - `(window as unknown as {...}).ccsmCliBridge` cast is well-commented (points at W2c owning the global type). Acceptable as scoped tech debt.
- Install command: rendered as `<code>` with `select-all` — copy-friendly via triple-click. No copy button — minor polish gap, fine for MVP.
- A11y: `<h1>` heading, `<button type="button">` with visible label, `focus-visible` ring on the button, `data-testid` selectors present. Solid.
- `terminal.starting` / `terminal.spawnFailed` / `terminal.retryButton` are added here but consumed by sibling PRs (W2a TtydPane / W2c wiring). Adding them in W2b is fine — keeps parity test green when siblings land — but please confirm W2a/W2c actually wire them (otherwise dead i18n keys).

### Cross-PR
- Local cast vs W2a's `src/cliBridge.d.ts`: needs W2c follow-up. Manager picks:
  - **Option A**: After W2a + W2b merge, W2c rebases and replaces the cast in `ClaudeMissingGuide.tsx` with the typed `window.ccsmCliBridge`. Clean.
  - **Option B**: Keep the cast; document as known tech debt.
  Recommend A — comment in the file already anticipates W2c owning the global type.
- terminal.* keys added here must be consumed by W2a/W2c. Worth a quick grep when those land.

### Verdict
APPROVE

PR #444, opened 2026-04-28
chore(cleanup): dead code sweep PR-A — modules + i18n + store fields

Jiahui-Gu:
## Summary
Net **-704 LOC, 0 additions**. Removes orphan modules, dead store field, dead i18n namespaces, and a trivially-true test. All deletions pre-verified by audit a93b2e61 then re-grep-checked at commit time.

## Pass 1 — orphan modules (482 LOC)
| File | LOC | Justification |
|------|-----|---------------|
| `electron/models/cli-picker-models.ts` | 140 | Duplicate of `electron/agent/cli-picker-models.ts`. `main.ts` imports `./agent/...` only. Zero `electron/models/*` import sites. |
| `electron/models/list-models-from-settings.ts` | 226 | Same — `main.ts:76` references `./agent/list-models-from-settings`. |
| `src/components/ui/DropdownMenu.tsx` | 86 | Radix wrapper, never imported. Repo uses `ContextMenu` for menus. |
| `src/lib/runningPlaceholder.ts` | 30 | InputBar removed in earlier PR; no remaining callers. |

`electron/models/` directory now empty and removed.

## Pass 2 — dead store field (~17 LOC)
- Removed `recentProjects` field, `pushRecentProject` action, persist allowlist entry, hydrate fallback (`src/stores/store.ts`, `src/stores/persist.ts`).
- Removed unused `RecentProject` type imports from both files (would have been TS unused-import errors).
- Updated `tests/persist-shape.test.ts` allowlist to drop `recentProjects`.
- Updated `scripts/probe-helpers/reset-between-cases.mjs` to drop `recentProjects: []` zero-init.
- `src/mock/data.ts` `RecentProject` type left as-is (out of scope; mock file is one-off cleanup follow-up).

## Pass 5 — dead i18n namespaces (en + zh, ~172 LOC)
Top-level namespaces with zero `t()` callers: `assistantBlock`, `permissions`, `questionBlock`, `tray`, `menu`, `dialog`, `errors`, `prDialog`, `banner`. Plus `chat.runningPlaceholderDefault/AcceptEdits/Bypass/Plan` (consumed only by deleted `runningPlaceholder.ts`).

en/zh parity preserved (same set of namespaces removed both sides) — `tests/i18n-key-parity.test.ts` still green.

## Pass 6 — trivially-true test
- `tests/permission-mode-values.test.ts` — only asserted that `permissions.modes.*` keys exist; with that namespace gone the test is dead.

## Deferred
[...]

PR #446, opened 2026-04-28
fix(settings): remove non-functional Notifications pane + drop orphan persisted keys

Jiahui-Gu:
## Summary

The Notifications settings pane shipped enable/sound toggles that **no production code consumed**. Users could 'enable notifications' with no toast ever firing because nothing in the renderer wired permission / question / turn_done events to `dispatchNotification`.

This PR makes two cleanups from the Settings audit (a2a2ccbe):

### Fix 1: Notifications pane was a lie

**Decision: Option A (full removal).**

Justification (grep evidence):
```
$ grep -rn 'dispatchNotification' src/
src/App.tsx:20  import { dispatchNotification } from './notifications/dispatch';
src/App.tsx:40  // E2E seam: harnesses drive dispatchNotification directly...
src/App.tsx:44  __ccsmDispatchNotification?: typeof dispatchNotification;
src/App.tsx:45  __ccsmDispatchNotification = dispatchNotification;
src/notifications/dispatch.ts:34  export async function dispatchNotification(...
```
Outside the dispatch fn itself, the only caller is the App.tsx E2E debug seam (`window.__ccsmDispatchNotification`) used by the harness. **Zero production callers** ever invoke it. The toggles in `NotificationsPane` only flipped store state nothing read.

Option B (disabled banner) was rejected: it would just defer the same cleanup, and there's no production caller behind a feature flag — the surface is dead end-to-end.

The electron-side machinery (`electron/notifications.ts`, `notify.ts`, `notify-bootstrap.ts`, AUMID setup) stays intact so a future PR that wires real renderer event hooks can add UI affordances back without re-doing main-process work.

### Fix 2: orphan persisted store keys

`model` and `permission` were in `PERSISTED_KEYS` but had zero subscribers — `useStore((s) => s.model)` / `s.permission` returns no matches anywhere in src/. Per the audit, they were silently consuming write-back budget on every store mutation.

Removed: `model`, `permission`, `notificationSettings` from `PERSISTED_KEYS` + `PersistedState`.

**Migration**: older snapshots may carry these keys. `loadPersisted` returns the parsed JSON unchanged; `hydrateStore`'s `setState` payload simply doesn't read them. Unknown top-level keys are silently ignored, no throw.

`recentProjects` is also orphan but coordinated to PR-A (pool-1) per task brief.

## Changes

- `src/components/SettingsDialog.tsx`: drop `NotificationsPane`, the `notifications` Tab union, tab/panel id maps. Move `CrashReportingField` into Updates pane.
- `src/notifications/dispatch.ts` + `tests/notifications-dispatch.test.ts`: deleted.
- `src/App.tsx`: drop dispatch import + `__ccsmDispatchNotification` window seam.
- `src/stores/store.ts`: drop `notificationSettings`/`model`/`permission` State fields, `setNotificationSettings`/`setGlobalModel`/`setPermission` actions, `NotificationSettings`/`DEFAULT_NOTIFICATION_SETTINGS`, `migrateNotificationSettings`/`migratePermission`. `PermissionMode` type kept (consumed by `runningPlaceholder.ts`). `importSession` no longer reads global `model`.
- `src/stores/persist.ts`: drop the three keys from allowlist + `PersistedState`.
- `tests/persist-shape.test.ts`: ALLOWED list narrowed; removed keys added to FORBIDDEN as regression guard.
- `scripts/harness-ui.mjs`: drop `caseNotifDisabledSuppress` (W5 single-gate test) + Notifications-pane assertion block in `caseI18nSettingsZh`.

## Coordination

- `recentProjects` left for PR-A (pool-1).
- `src/i18n/locales/{en,zh}.ts` owned by PR-A; dead `settings.notifications.*` keys left in place for PR-A namespace cleanup. Harmless while present.
- `electron/preload.ts` owned by PR-B; `notify` preload export untouched.
- `scripts/dogfood-r2-fp6-probe.mjs` + `scripts/probe-dogfood-r2-fp12-settings-live.mjs` owned by PR-C; both already handle missing dispatch.ts (`.catch(() => null)`) and missing `notificationSettings` (default fallback), so they survive without intervention.

## Test plan

- [x] `npx eslint src/` — clean
- [x] `npx vitest run` — 292 passing including updated `persist-shape.test.ts`. (3 pre-existing failures in `db-hardening.test.ts` from local better-sqlite3 NODE_MODULE_VERSION mismatch, unrelated.)
- [ ] e2e harness `i18n-settings-zh` — needs harness build run on main repo (this PR removed its Notifications-pane assertions; tab navigation order updated).


Jiahui-Gu:
LGTM (self-PR comment, cannot self-approve). Reviewed in pool-3 against pr-446 head.

Layer 1 (necessity)
- Confirmed zero production callers via `git grep -nE "dispatchNotification|window\.ccsm\.notify\(" -- 'src/'` — all remaining hits are doc comments. Option A (full removal) is the right call.
- Store dead-field check passes: `state.model | state.permission | state.notificationSettings | setGlobalModel | setPermission | setNotificationSettings | globalModel` returns zero hits across `src/`.
- `PermissionMode` type correctly retained — `src/lib/runningPlaceholder.ts` still imports it for label rendering.

Layer 2 (implementation)
- `tsc --noEmit` clean, `eslint` clean, vitest 292/295 green (the 3 failures are an unrelated `better-sqlite3` ABI mismatch in this env, not introduced here). `tests/persist-shape.test.ts` passes — `model` / `permission` / `notificationSettings` correctly moved from ALLOWED to FORBIDDEN.
- Migration handling is sound: `hydrateStore` simply doesn't read the dropped keys; zustand's `setState` ignores unknown top-level fields, so older persisted blobs degrade silently. Comment in `store.ts` documents this explicitly.
- `importSession` model-resolution fallback was correctly tightened to `connection?.model ?? models[0]?.id ?? ''` after losing the global `model` slot.
- `SettingsDialog`: tab/id/panelId maps + Tab union all narrowed to `'appearance' | 'connection' | 'updates'`. `CrashReportingField` cleanly relocated to `UpdatesPane` without prop changes.
- `harness-ui.mjs`: `caseNotifDisabledSuppress` removed in full; `caseI18nSettingsZh` Notifications block dropped and header comment updated. No hardcoded `expectedTotalCases` counter to maintain.

Follow-up (non-blocking)
- After PR-A (#444) merges (which deletes `src/lib/runningPlaceholder.ts`), `PermissionMode` becomes unused and can be cleaned up in a follow-up.
- Conflict expectation: PR-A also touches `src/stores/store.ts` + `src/stores/persist.ts` + `tests/persist-shape.test.ts`; whichever lands second will need a rebase.

CI: workflow_dispatch run 25030363294 on `fix/settings-cleanup` is green.

PR #449, opened 2026-04-28
fix(ttyd): propagate session cwd to claude (P0 dogfood blocker)

Jiahui-Gu:
## Summary

`spawnTtyd` had no `cwd` option, so ttyd inherited Electron's working directory and started `claude` there. Result: every session's JSONL transcripts landed in `~/.claude/projects/<electron-cwd>/` instead of the user's chosen project, and the CLI loaded settings/agents/skills from the wrong project root — making dogfood unusable.

This PR threads `cwd` from the ccsm session row through the renderer (`TtydPane`), the preload bridge, the IPC handler, and into the `spawn` options. Falsy `cwd` falls back to Electron's cwd to keep early-boot working.

## Scope

Minimal cwd-only fix to unblock dogfood in parallel. The other four ttyd lifecycle bugs are deferred to a follow-up PR:

- #488 — `openTtydForSession` mints a random UUID instead of using the ccsm session.id
- #490 — switching session kills the ttyd → conversation lost
- #492 — `claudeResolver` cache never expires (Re-check button no-ops)
- #489 — `deleteSession` doesn't kill ttyd → process leak

## Files changed

- `electron/cliBridge/processManager.ts` — `spawnTtyd` accepts `cwd`, passes it to `spawn` options; `openTtydForSession` / `resumeTtydForSession` accept `cwd`.
- `electron/cliBridge/index.ts` — IPC handlers forward `cwd`.
- `electron/preload.ts` — bridge methods accept `cwd`.
- `src/cliBridge.d.ts` — type signatures.
- `src/components/TtydPane.tsx` — Props add `cwd`, passed through to bridge.
- `src/App.tsx` — `<TtydPane cwd={active.cwd ?? ''} />`.

41 insertions, 25 deletions across 6 files.

## Test plan

- [x] `npx eslint <changed files>` clean
- [x] `npm test` — 32 files / 292 tests all pass
- [ ] Reviewer to verify with the dogfood probe (separate worker is wiring `dogfood-probe-current-ui.mjs` to grep `~/.claude/projects/<encoded-cwd>/<sid>.jsonl`)

## Why no e2e in this PR

A separate worker (pool-2) is extending `scripts/dogfood-probe-current-ui.mjs` with the JSONL-on-disk verification. Per coordinator direction, this PR ships the minimum production fix so the probe (which currently fails because of the wrong cwd) can be re-run against `working` immediately after merge.

### DIFF
--- a/electron/cliBridge/index.ts
+++ b/electron/cliBridge/index.ts
@@ -7,11 +7,11 @@
 // is unused by the renderer until Worker 2 wires it).
 //
 // IPC contract (mirrored in preload.ts):
-//   cliBridge:openTtydForSession(sessionId)      → {ok,port,sid} | {ok:false,error}
-//   cliBridge:resumeSession(sessionId, sid)      → {ok,port,sid} | {ok:false,error}
-//   cliBridge:killTtydForSession(sessionId)      → {ok,killed}
-//   cliBridge:checkClaudeAvailable()             → {available, path?}
-//   event cliBridge:ttyd-exit                    → {sessionId, code, signal}
+//   cliBridge:openTtydForSession(sessionId, cwd)      → {ok,port,sid} | {ok:false,error}
+//   cliBridge:resumeSession(sessionId, cwd, sid)      → {ok,port,sid} | {ok:false,error}
+//   cliBridge:killTtydForSession(sessionId)           → {ok,killed}
+//   cliBridge:checkClaudeAvailable()                  → {available, path?}
+//   event cliBridge:ttyd-exit                         → {sessionId, code, signal}
 //
 // Security: every handler that takes an action gates on `fromMainFrame(e)`
 // — a compromised iframe (the ttyd one we host, future webviews, etc.)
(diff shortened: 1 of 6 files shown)


Jiahui-Gu:
LGTM (self-review on author PR — independent reviewer agent).

**Layer 1 (necessity):** Confirmed root cause. `spawnTtyd` had no `cwd` option pre-PR; ttyd inherited Electron cwd and `claude` wrote JSONLs + loaded settings/agents/skills from the wrong project. cwd threading is the right fix; minimal scope (deferred ttyd lifecycle bugs to follow-ups #488/489/490/492).

**Layer 2 (implementation):**
- Diff matches stat (6 files, +41/-25); zero unrelated refactor.
- `Session.cwd: string` is required in `src/types.ts`, so `active.cwd ?? ''` + `cwd && cwd.length > 0 ? cwd : undefined` fallback realistically never triggers; defense-in-depth is acceptable.
- Effect deps `[sessionId, cwd, open]` correctly re-mount on cwd change; Retry button passes cwd.
- IPC handlers validate `typeof cwd === 'string'` (defaults to '' otherwise) — handles malicious renderer payloads safely.
- Preload + `cliBridge.d.ts` signatures aligned with handler arg order.
- Pre-existing `resumeSession` bridge has no in-tree caller; out of PR scope.

**Verification:**
- `npx tsc --noEmit -p tsconfig.json` clean
- `npx tsc --noEmit -p tsconfig.electron.json` clean
- CI: ubuntu / macos / windows / no-silent-drops all green

**Dogfood probe note:** happy-path probe (#448) not yet merged in `working`; manager should re-run `dogfood-probe-happy-path.mjs` once #448 lands — this PR fixes the cwd that was failing claude-banner / claude-replied. `dogfood-probe-current-ui.mjs` mount-only path unaffected.

Squash-merging.

PR #450, opened 2026-04-28
fix(ttyd): lifecycle 4 deferred fixes (sessionId / persistent ttyd / cache / leak)

Jiahui-Gu:
## Summary

4 deferred ttyd lifecycle fixes (cwd fix from #494 already shipped in PR #449):

- **#492 P1-1 claudeResolver cache (`{force?: boolean}`)** — module-level cache no longer permanently sticky; `ClaudeMissingGuide` "Re-check" button now passes `{force: true}` so the user can install claude in a separate terminal and recover in-place without restarting ccsm.
- **#489 P1-2 deleteSession ttyd leak** — `store.ts deleteSession` now fires `killTtydForSession(id)` over the cliBridge IPC before returning the undo snapshot. Previously deleting a session left an orphan ttyd + claude pair holding the JSONL transcript open.
- **#488 P0-1 openTtydForSession sessionId divergence** — stop minting fresh `randomUUID()` per spawn. ccsm `session.id` is now used as `claude --session-id` so the on-disk JSONL filename and the ccsm sidebar row stay in lockstep. Added `toClaudeSid()`: raw UUID v4 inputs round-trip; legacy `s-<uuid>` ids are deterministically projected onto a v4-shaped string via SHA-256 (claude requires UUID-shaped session ids).
- **#490 P0-3 switching session no longer kills the chat** — ttyd lifecycle is now owned by the ccsm session itself (created at `openTtydForSession`, destroyed at `deleteSession` via #489). `TtydPane` unmount no longer kills the backend; mount queries the new `cliBridge:getTtydForSession` IPC and reuses the existing port if running, otherwise spawns. `openTtydForSession` is reuse-aware too so concurrent calls can't double-spawn.

## Files

- `electron/cliBridge/claudeResolver.ts` — `resolveClaude({ force })`.
- `electron/cliBridge/processManager.ts` — `toClaudeSid`, `getTtydForSession`, reuse-aware `openTtydForSession`.
- `electron/cliBridge/index.ts` — `cliBridge:getTtydForSession` IPC + `force` plumbing.
- `electron/preload.ts` + `src/cliBridge.d.ts` — bridge surface for the new IPC + `force` flag.
- `src/components/TtydPane.tsx` — reuse-first mount, no kill on unmount.
- `src/components/ClaudeMissingGuide.tsx` — pass `{force: true}` to bust cache.
- `src/stores/store.ts` — `deleteSession` fires `killTtydForSession` IPC.
- `scripts/dogfood-probe-current-ui.mjs` — Cases A/B/C/D coverage.

## Test plan / e2e evidence

`scripts/dogfood-probe-current-ui.mjs` (extended with 4 cases at the tail):

- [x] **Case B switch-preserves-ttyd: PASS** — sessionA created → switch to sessionB → switch back to sessionA. `getTtydForSession(sessionA)` returns same port (63947) before + after; original ttyd PIDs (12676, 36168) still alive in `tasklist`. New PID (35300) added by sessionB's spawn — old ones not killed.
- [x] **Case C delete-reaps-ttyd: PASS** — `getTtydForSession` returns `null` immediately after `deleteSession`; ttyd entry removed from internal map.
- [ ] **Case A jsonl-matches-ccsm-sid: probe FAIL but pre-existing** — gated on the upstream `claude-replied` step which already fails in the isolated dogfood userdata (claude binary spawns but doesn't produce a reply within 30s in the sanitized HOME, so JSONL is never written to disk). Code path verified by inspection: `final.activeId` (`d8115488-d8e1-443e-bf69-8e951819ee51`) is what flows into `--session-id`, and `toClaudeSid` returns that same string for valid UUID v4 input. When the env can produce a real claude reply, the JSONL filename will be exactly `<activeId>.jsonl`.
- [ ] **Case D claude-force-rescan: skipped per design** — would require mv'ing the live `claude.cmd` off PATH on the dev machine, interfering with the user's own CLI. The `{force:true}` flag is exercised by `ClaudeMissingGuide` on the renderer path.

## Quality gates
- [x] `npm run lint` clean
- [x] `npx vitest run` — 292/292 passing (after `npm rebuild better-sqlite3` for the local node 24 host env)

## Notes for reviewer
- **Undo + ttyd**: deleteSession's undo (`restoreSession`) reinserts the row but the ttyd was killed. On focus the new TtydPane will call `openTtydForSession` with the same session.id; claude will recreate the conversation against the original JSONL via `--session-id`. Acceptable trade-off — alternative was deferring kill until the undo window expires, which adds significant state machinery.
- **`force` security**: the `cliBridge:checkClaudeAvailable` handler now also gates on `fromMainFrame(e)`, matching the other handlers (was previously ungated since it was read-only — but accepting an arg makes that worth tightening).

### DIFF
--- a/electron/cliBridge/claudeResolver.ts
+++ b/electron/cliBridge/claudeResolver.ts
@@ -22,9 +22,10 @@
 // show actionable copy.
 //
 // Result is cached after the first successful lookup. The user can re-
-// install claude or change PATH while ccsm runs; we don't bust the
-// cache automatically — that's a "restart ccsm" event in practice. A
-// future task can add a `cliBridge:rescanClaude` IPC if the need arises.
+// install claude or change PATH while ccsm runs; pass `{force: true}` to
+// bypass the cache (the renderer's "Re-check" button on ClaudeMissingGuide
+// uses this so the user can install claude in a separate terminal and
+// recover in-place without restarting the app).
 
 import { spawnSync } from 'node:child_process';
 
@@ -42,8 +43,8 @@ function tryWhere(name: string): string | null {
   }
 }
 
-export function resolveClaude(): string | null {
-  if (cached !== undefined) return cached;
+export function resolveClaude({ force = false }: { force?: boolean } = {}): string | null {
+  if (!force && cached !== undefined) return cached;
   if (process.platform === 'win32') {
     cached = tryWhere('claude.cmd') ?? tryWhere('claude');
   } else {
--- a/electron/cliBridge/index.ts
+++ b/electron/cliBridge/index.ts
@@ -21,6 +21,7 @@
 import { ipcMain, type IpcMainInvokeEvent, type WebContents } from 'electron';
 import {
   bindSender,
+  getTtydForSession,
   killAll,
   killTtydForSession,
   openTtydForSession,
@@ -71,8 +72,17 @@ export function registerCliBridgeIpc(): void {
     },
   );
 
-  ipcMain.handle('cliBridge:checkClaudeAvailable', () => {
-    const p = resolveClaude();
+  ipcMain.handle('cliBridge:getTtydForSession', (e, sessionId: unknown) => {
+    if (!fromMainFrame(e)) return null;
+    if (typeof sessionId !== 'string' || !sessionId) return null;
+    return getTtydForSession(sessionId);
+  });
+
+  ipcMain.handle('cliBridge:checkClaudeAvailable', (e, opts: unknown) => {
+    if (!fromMainFrame(e)) return { available: false } as const;
+    const force =
+      typeof opts === 'object' && opts !== null && (opts as { force?: unknown }).force === true;
+    const p = resolveClaude({ force });
     return p ? { available: true, path: p } : { available: false };
   });
 }
--- a/electron/cliBridge/processManager.ts
+++ b/electron/cliBridge/processManager.ts
@@ -31,7 +31,7 @@
 // fallback needed.
 
 import { spawn, spawnSync, type ChildProcess } from 'node:child_process';
-import { randomUUID } from 'node:crypto';
+import { createHash } from 'node:crypto';
 import type { WebContents } from 'electron';
 import { pickFreePort } from './portAllocator';
 import { resolveClaude } from './claudeResolver';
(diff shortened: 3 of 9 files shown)


Jiahui-Gu:
LGTM (independent reviewer pass — self-PR so commenting in lieu of approve).

## Layer 1 (necessity + design)

- **toClaudeSid**: verified against real `node` execution. Raw UUID v4 inputs round-trip unchanged (lower-cased). Legacy `s-<uuid>`, arbitrary strings, etc. all project to strings matching `/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i`. Version nibble hardcoded `4`, variant nibble masked to `{8,9,a,b}` per RFC 4122. Deterministic + collision-resistant for this use.
- **getTtydForSession reuse + openTtydForSession reuse-first**: clean — renderer no longer needs special-case for the prior `session_already_running` error; switching back attaches to the same port. Case B PASS in the probe confirms it.
- **deleteSession kills ttyd**: necessary leak fix. Fire-and-forget IPC is correct (optimistic UI + undo doesn't depend on it).
- **kill-on-unmount removal**: sound trade-off. Switch-preserves-chat is the primary win. The acknowledged undo-after-delete corner relies on `claude --session-id <existing-sid>` resuming the JSONL; CLI help confirms `--session-id` accepts UUID and `<sid>.jsonl` filenames are deterministic, so re-spawn against the same sid hits the same on-disk file. Acceptable.
- **`force` flag + `fromMainFrame` gate on `cliBridge:checkClaudeAvailable`**: correct security tightening now that the handler accepts an arg.

## Layer 2 (implementation)

- `npx tsc --noEmit -p tsconfig.json` clean
- `npx tsc --noEmit -p tsconfig.electron.json` clean
- `npm run lint` clean
- CI green on ubuntu / macOS / windows + no-silent-drops
- TtydPane effect deps `[sessionId, cwd, open]` safe — `open` is `useCallback([])` and state updates don't feed back into deps. No re-render loop.
- IPC handler `cliBridge:getTtydForSession` validates sessionId (non-empty string) and gates on `fromMainFrame`. Matches sibling handlers.

## Minor (non-blocking, follow-up)

- No unit test for `toClaudeSid`. Pure function; a vitest case covering (a) v4 round-trip, (b) legacy `s-<uuid>` projection, (c) regex match on output would be cheap insurance.

Jiahui-Gu:
Re-reviewed against new working base (post #452/#453/#454 merge — rebased clean, force-pushed, CI all green on all 3 OSes).

Layer 1 (direction) — all 4 fixes still hold:
- #492 claudeResolver `{force?}` flag — render-container-independent, applies cleanly.
- #489 deleteSession fires killTtydForSession — render-container-independent, applies cleanly.
- #488 toClaudeSid SHA-256 → UUID v4 projection — render-container-independent, logic byte-identical pre/post-rebase.
- #490 getTtydForSession + reuse-on-mount + no-kill-on-unmount — verified compatible with `<webview partition="persist:ttyd" src=...>`. The reuse-first branch sets `state.port` which feeds `webview.src=http://127.0.0.1:${port}/` exactly the same way `<iframe>` did, so the "reuse port" path now drives the webview's src attribute. Partition `persist:ttyd` is a static literal — stable across mount/unmount/reuse, so storage isolation is consistent. Removing kill-on-unmount with `<webview>` is correct: webview unmount destroys its WebContents automatically (Electron-managed), the ttyd backend process keeps its port held by processManager.sessions, and next mount re-queries via getTtydForSession and re-sets src — preserving the live claude conversation. The merged-file lifecycle docblock (lines 16-26) accurately describes both #452's container choice AND #450's reuse semantics.

Layer 2 (implementation) — `npm run typecheck` clean; `npm run lint` 0 errors (5 pre-existing warnings in electron/main.ts from #453, untouched by this PR). CI: lint+typecheck+test pass on macos/ubuntu/windows.

LGTM for merge.

PR #455, opened 2026-04-28
fix(ttyd): match xterm bg to host + visible boundary

Jiahui-Gu:
## Summary
- Pass an additional `-t theme={\"background\":\"#0B0B0C\"}` to ttyd spawn so xterm bg matches host chrome exactly (no off-by-a-shade dark slab).
- Add `bg-[#0B0B0C] ring-1 ring-inset ring-white/5` on the `<webview>` so the TUI boundary is faintly visible without competing with content.
- Resolves #499 cosmetic issue that previously looked like a render bug and triggered repeated false-alarm investigations.

## Files
- electron/cliBridge/processManager.ts (added theme arg + comment)
- src/components/TtydPane.tsx (matched bg + inset ring)

## Before / After
- docs/screenshots/xterm-bg-fix/before.png
- docs/screenshots/xterm-bg-fix/after.png
- docs/screenshots/xterm-bg-fix/before-webview-mount.png
- docs/screenshots/xterm-bg-fix/after-webview-mount.png

Captured via `node scripts/dogfood-probe-happy-path.mjs` (before: stash + run; after: rebuild + run). Before/after MD5 differ; visual confirms faint inset ring delineates TUI boundary against host chrome.

## Verification
- npm run lint: 0 errors (5 preexisting warnings in electron/main.ts unrelated)
- npm run typecheck: clean
- happy-path probe: PASS (claude-replied)


Jiahui-Gu:
LGTM — reviewer pass.

**Layer 1 (necessity / approach)**
- Worth shipping. The off-by-a-shade bg burned multiple debug rounds (#499); making the seam intentional + faintly bordered prevents recurrence. Real value beyond pixel-pushing.
- Option A+B combined is acceptable. A alone (theme match) would suffice, but B (inset ring) future-proofs for multi-pane layouts at zero cost.
- ttyd argv risk verified: `spawn(ttyd, args, ...)` is array-based, so `'-t'` and `'theme={"background":"#0B0B0C"}'` arrive as separate argv elements — no shell quoting concern. Each `-t` takes one option, syntax is correct.

**Layer 2 (implementation)**
- `processManager.ts`: two `-t` flags well-formed; comment explains the WHY (#499 misdiagnosis history) per repo style. Good.
- `TtydPane.tsx`: `bg-[#0B0B0C]` matches ttyd theme exactly; `ring-1 ring-inset ring-white/5` is the right Tailwind (inset keeps the ring inside the box, no layout shift). `title={\`ttyd session ${sessionId}\`}` preserved — harness-ui selectors safe.
- Screenshots: all 4 under `docs/screenshots/xterm-bg-fix/` per repo convention. before.png (117162 B) vs after.png (118922 B) — distinct PNGs, prod bundle rebuilt between captures per PR body.

**CI**
- ubuntu + macos + no-silent-drops: PASS.
- windows-latest: FAIL, but failure is `ModuleNotFoundError: No module named 'distutils'` during `npm ci` postinstall (node-gyp on Python 3.12 GHA runner). **Same failure on `working` HEAD** (run 25019887683) — pre-existing env rot, not caused by this PR. A pure cosmetic string-only change cannot affect native module rebuild. Windows CI fix should be tracked separately.

Manager: safe to merge.

PR #459, opened 2026-04-28
[FAIL] test(e2e): probe-real-reopen-resume (UX G)

Jiahui-Gu:
## UX scenario (Task #507, UX G)

User words (verbatim):
> 3是关闭ccsm以后再打开，点击session可以聊天（暗含了resume在里面）

Close ccsm, reopen it, click on a session, the chat must work — implicit:
\`claude --resume\` must restore the prior conversation across an app restart.

## Probe steps

1. Build an isolated \`~/.claude\` tempDir (auth files only).
2. Launch ccsm run #1 with that tempDir + a fixed electron userData dir we
   control across both launches.
3. Seed a session via \`__ccsmStore.createSession\`, drive its ttyd webview,
   send \"remember the word OMEGA\", wait for claude's reply.
4. Wait past the persist debounce (250 ms) + cushion so ccsm.db has the
   session row, then \`app1.close()\`.
5. Launch ccsm run #2 with **SAME** tempDir AND **SAME** userDataDir.
6. Verify persisted session re-appears in the sidebar (\`[data-session-id]\`).
7. Click the session — \`activeId\` flips, TtydPane mounts.
8. **Cross-restart resume assertion:** grep the resumed xterm buffer for
   the OMEGA token from run #1.
9. Send a follow-up message — confirm claude replies again.

## Output (current working HEAD)

Probe **FAILS** — surfaces a real product bug:

\`\`\`
[STEP] run2: waiting for claude to print resumed transcript
[FAIL] run2: token \"OMEGA\" from run1 not found in resumed buffer (resume broken).
This UX scenario (close ccsm → reopen → click session → resume) does NOT work on current working HEAD.
Root cause: TtydPane → openTtydForSession always passes \`claude --session-id <sid>\`, never \`--resume <sid>\`.
On restart with a pre-existing JSONL, claude rejects with: \"Session ID is already in use.\"
Fix direction: openTtydForSession (or renderer) should detect existing JSONL at
  ~/.claude/projects/<cwd-hash>/<sid>.jsonl and call resumeTtydForSession instead.
Error: Session ID 0fd8c2a2-d9de-44ec-8cf4-d8f36d7a5716 is already in use.
\`\`\`

Reproduced 2x cleanly. The probe itself is correct: run #1 boots, seeds,
prompts and completes; ccsm.db DOES persist + rehydrate the session;
clicking activates it; ttyd respawns. The break is in the spawn args.

Failure screenshot:
\`docs/screenshots/probe-real-reopen-resume/fail-2026-04-28T05-48-22-532Z-run2-no-resumed-token.png\`

## Helper concerns (probe-utils-real-cli.mjs)

\`launchCcsmIsolated\` mints a NEW electron userData dir on every call via
\`mkdtempSync\` + registers it for cleanup. That wipes \`ccsm.db\` between
launches and is incompatible with any cross-restart probe.

This probe replicates the helper's launch logic inline so it can pin a
single userDataDir across both runs. **Suggested helper API extension:**

\`\`\`js
launchCcsmIsolated({ tempDir, userDataDir, env })
\`\`\`

where a caller-supplied \`userDataDir\` is reused as-is (and not added to
the cleanup registry — caller owns its lifetime).

## Notes for review / follow-up

- The fix for the surfaced bug is out-of-scope for this PR — this PR only
  delivers the probe + the failing baseline.
- Once the fix lands, the probe should flip green with no changes; if it
  needs adjustment the only likely tweak is the OMEGA-replay timeout.
- The probe is properly isolated (its own tempDir + userDataDir, no touch
  of the user's real \`~/.claude\` or production electron user-data).

### DIFF
--- a/scripts/probe-real-reopen-resume.mjs
+++ b/scripts/probe-real-reopen-resume.mjs
@@ -1,0 +1,431 @@
+// Real-claude e2e probe — reopen ccsm and resume an existing session.
+//
+// UX scenario (Task #507, UX G — verbatim user words):
+//   "3是关闭ccsm以后再打开，点击session可以聊天（暗含了resume在里面）"
+//   Close ccsm, reopen it, click on a session, the chat must work — implicit:
+//   `claude --resume` must restore the prior conversation across an app
+//   restart (not just an in-memory session-switch).
+//
+// Probe flow:
+//   1. Build an isolated ~/.claude tempDir (auth files only — see helper).
+//   2. Launch ccsm run #1 with that tempDir + a caller-owned userDataDir
+//      (helper API in #463: launchCcsmIsolated({ tempDir, userDataDir })
+//      reuses the dir verbatim and skips auto-cleanup, so ccsm.db survives
+//      across run1 → run2).
+//   3. Seed a session, drive its ttyd webview, dismiss first-run modals
+//      (trust folder, welcome splash), send a deterministic prompt
+//      ("remember the word OMEGA") and wait for claude's reply.
+//   4. Wait for the persist debounce (250 ms) + a cushion so ccsm.db
+//      definitely contains the session row, then close run #1 cleanly.
+//   5. Launch ccsm run #2 with SAME tempDir AND SAME userDataDir.
+//   6. Verify the persisted session re-appears in the sidebar
+//      ([data-session-id="<sid>"]). Click it.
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
Rebased onto working (1b74c94 + fadcd04). Added trust-modal handling in run2 + userDataDir reuse via launchCcsmIsolated({userDataDir}) helper API. Probe PASS locally. Ready for review/merge.

PR #461, opened 2026-04-28
[FAIL] test(e2e): probe-real-switch-session-keeps-chat (UX F)

Jiahui-Gu:
## UX scenario (F)

> "2 是在 session 之间切换，右边窗口随之切换，并且可以聊天，这个现在有问题，
>  切换的时候当成是 resume 了，报错 session is already in use"

When the user clicks between sessions in the sidebar, the right pane should follow and remain chattable. The known bug: switching back to a session with a still-running ttyd was being treated as a fresh resume — colliding with the existing claude PTY and surfacing a "session already in use" error.

The fix lives in two places:
- `electron/cliBridge/processManager.ts::getTtydForSession` returns the existing `{port, sid}` when the session's ttyd is still alive.
- `src/components/TtydPane.tsx` calls `getTtydForSession` BEFORE `openTtydForSession`, attaching to the existing port instead of respawning.

## Probe

`scripts/probe-real-switch-session-keeps-chat.mjs` drives the round-trip A → B → A:

1. Isolated `~/.claude` clone + ccsm prod-bundle launch
2. Seed two sessions (session-A, session-B)
3. selectSession(A) → wait for webview + claude TUI; capture A's ttyd port
4. Send `Please reply with the single word ALPHA` to A; await reply
5. selectSession(B) → wait for webview + claude TUI
6. Confirm A's ttyd port is still tracked by main while B is active
7. selectSession(A) → wait for A's webview to re-mount; assert URL points at A's original port
8. Read A's xterm scrollback; assert ALPHA is still visible (ttyd replays buffer to fresh client)
9. Send `Please reply with the single word BETA` to A; await reply
10. Assert no `ttyd-exit` for A and no "already in use" console error during the round-trip

## Result: FAIL

Run output (single canonical run, per coordinator instruction "one run, push regardless"):

```
[STEP] isolated-claude-dir: PASS
[STEP] boot:               PASS
[STEP] seed-sessions:      PASS  (sidA=4a39b844..., sidB=1778b76a...)
[STEP] select-A-mounted:   PASS  (wcA=2)
[STEP] A-port-captured:    PASS  (port=62304)
[STEP] A-first-reply:      PASS  (loose match — see caveat below)
[STEP] select-B-mounted:   PASS  (wcB=3)
[STEP] A-port-still-alive-while-B-active: PASS  (port=62304)
[STEP] A-port-reused-on-switch-back:      PASS  (port=62304)
[STEP] A-webview-attached-to-original-port: PASS  (wcA2=4, url=http://127.0.0.1:62304/)
[FAIL] A's scrollback lost ALPHA history after switch-back. Tail: <empty>
[xterm-tail wcId=4]   <empty buffer>
```

Failure screenshot: `docs/screenshots/probe-real-switch-session-keeps-chat/fail-1777355477732.png`

## Findings

- The cliBridge port-reuse path (`getTtydForSession` + TtydPane reuse-first branch) **WORKS** at the IPC / process-manager level: A's ttyd port is preserved across the A → B → A round-trip and the new `<webview>` mounted for A on switch-back attaches to the same `http://127.0.0.1:62304/`.
- No "session already in use" error was surfaced in the renderer console, no `ttyd-exit` event fired for A.
- The probe FAILS at step 8: A's xterm buffer is **empty** after the new webview attaches to the reused port. Two possible causes the manager should triage:
  - **Hypothesis A (probe author):** The `A-first-reply` PASS is a false positive. Claude's cold-start "Welcome back!" splash ate the first keystrokes (per coordinator hint from sister UX E probe), so ALPHA was never sent. Then ttyd has no scrollback to replay, and the empty buffer on switch-back is correct. Fix: send a few `\r` to dismiss the splash before the first prompt — but per coordinator "do NOT try to fix" — leaving for a follow-up worker.
  - **Hypothesis B (real bug):** ttyd is not replaying its scrollback to the freshly-attached client on the reused port. The user-visible bug ("right pane chat content gone after switch-back") would still hold even though the port-reuse machinery is correct.

The probe is structured so the manager can distinguish these by tightening the `A-first-reply` check (look for ALPHA in the buffer post-reply, not just the loose echo regex).

## Reviewer-flagged helper limitation (not patched)

`waitForWebviewMounted` picks the highest-id webContents — when both A's and B's ttyds are alive on switch-back, it could return B's id by mistake. In practice, switching tears down B's `<webview>` tag immediately (live ids = `[]` for ~100ms before A re-mounts), so the helper happens to return the correct (and only) live webContents on switch-back — but the assertion at `A-webview-attached-to-original-port` re-validates the URL against A's recorded port to be defensive. Helper untouched per task instruction.

## Local checks

- `node scripts/probe-real-switch-session-keeps-chat.mjs` — ran once, exit 1 with `[FAIL]` (output above)
- Skipped second-run flake check per coordinator's "one run, push" instruction.
- `dist/renderer/index.html` was rebuilt mid-probe-development because the worktree's prebuilt bundle was stale (still rendering `<iframe>` instead of `<webview>`). After `npm run build`, all PASS steps above became reachable.

## Constraints respected

- Helper `scripts/probe-utils-real-cli.mjs` not modified.
- No files outside `scripts/probe-real-switch-session-keeps-chat.mjs` and the screenshot dir touched.

### DIFF
--- a/scripts/probe-real-switch-session-keeps-chat.mjs
+++ b/scripts/probe-real-switch-session-keeps-chat.mjs
@@ -1,0 +1,450 @@
+// probe-real-switch-session-keeps-chat — UX scenario F.
+//
+// User scenario (verbatim from product owner):
+//   "2 是在 session 之间切换，右边窗口随之切换，并且可以聊天，这个现在有问题，
+//    切换的时候当成是 resume 了，报错 session is already in use"
+//
+// Translation: when the user clicks between sessions in the sidebar, the right
+// pane should follow and remain chattable. The bug being locked down: switching
+// back to a session that already has a running ttyd was being treated as a
+// fresh resume — which collided with the existing claude PTY and surfaced a
+// "session already in use" error in the renderer.
+//
+// The fix lives in two places:
+//   * `electron/cliBridge/processManager.ts::getTtydForSession` returns the
+//     existing {port, sid} when the session's ttyd is still alive.
+//   * `src/components/TtydPane.tsx` calls `getTtydForSession` BEFORE
+//     `openTtydForSession`, attaching to the existing port instead of
+//     respawning.
+//
+// This probe verifies the round-trip A → B → A:
+//   1. Session A's ttyd webContents id is captured on the first mount.
+//   2. After switching to B and back to A, the renderer's TtydPane MUST
+//      reuse the same webContents (port reuse path).
+//   3. A's xterm scrollback still contains the ALPHA prompt/reply from
+//      step 1 (i.e. the conversation was preserved, not restarted).
+//   4. A second prompt sent in A also gets a reply.
+//   5. No console error toast about "already in use" appears, and no
+//      ttyd-exit event for A's sid was broadcast during the round-trip.
+//
+// Reviewer-flagged helper limitation (do NOT patch helper):
+//   `waitForWebviewMounted` picks the highest-id webContents — when both A's
+//   and B's ttyds are alive on switch-back, that helper might return B's id.
+//   We work around it here by remembering A's wcId from the first mount and
+//   asserting the live webview-set still contains it after switch-back.
+
+import { mkdirSync, writeFileSync, existsSync } from 'node:fs';
+import * as path from 'node:path';
+import {
+  createIsolatedClaudeDir,
+  launchCcsmIsolated,
+  seedSession,
+  waitForWebviewMounted,
+  waitForXtermBuffer,
+  sendToClaudeTui,
+  readXtermLines,
+  executeJavaScriptOnWebview,
+} from './probe-utils-real-cli.mjs';
+
+const screenshotDir = path.resolve('docs/screenshots/probe-real-switch-session-keeps-chat');
+mkdirSync(screenshotDir, { recursive: true });
+
+const steps = [];
+const log = (step, ok, detail) => {
+  const entry = { step, ok, detail: detail ?? null };
+  steps.push(entry);
+  const tag = ok ? 'PASS' : 'FAIL';
+  const tail = detail ? ': ' + JSON.stringify(detail).slice(0, 240) : '';
+  console.log(`[STEP] ${step}: ${tag}${tail}`);
+};
+
+const ttydExits = [];
+const consoleErrors = [];
+
+let electronApp = null;
+let win = null;
+let ranCleanup = false;
+let exitCode = 1;
+let cleanupClaude = null;
+
+const finish = async () => {
+  if (ranCleanup) return;
+  ranCleanup = true;
+  const report = {
+    generatedAt: new Date().toISOString(),
+    steps,
+    ttydExits,
+    consoleErrors: consoleErrors.slice(0, 50),
+  };
+  try {
+    writeFileSync(path.join(screenshotDir, 'probe.json'), JSON.stringify(report, null, 2));
+  } catch { /* ignore */ }
+  if (electronApp) {
+    try { await electronApp.close(); } catch { /* ignore */ }
+  }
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
**REAL PRODUCT BUG — not a probe issue.**

Rebased PR onto `origin/working` (1b74c94), inserted `await dismissWelcomeSplash(electronApp, wcId)` before each `sendToClaudeTui` call site (ALPHA on wcA, B-mount on wcB, BETA on wcA2), and also added a `waitForXtermBuffer(/ALPHA/, 15s)` settle after switch-back so ttyd has time to replay scrollback to the freshly-attached client.

Result: probe still FAILs, but at a different point and with diagnostic output that points squarely at a product bug.

### Failure trace
```
[STEP] isolated-claude-dir: PASS
[STEP] boot: PASS
[STEP] seed-sessions: PASS
[STEP] select-A-mounted: PASS  (wcA=2)
[STEP] A-port-captured: PASS  (port=61736)
[STEP] A-first-reply: PASS    (ALPHA reply received)  ← splash fix worked
[STEP] select-B-mounted: PASS  (wcB=3)
[STEP] A-port-still-alive-while-B-active: PASS
[STEP] A-port-reused-on-switch-back: PASS  (same port)
[STEP] A-webview-attached-to-original-port: PASS  (wcA2=4 -> http://127.0.0.1:61736/)
[FAIL] A's scrollback lost ALPHA history after switch-back.
       lines=1 fullLen=73
       json=["Error: Session ID c6f4cc63-a273-43c0-b530-300be6067f88 is already in use."]
```

### What this means
After switching A → B → back to A, the renderer correctly reuses A's original ttyd port (PR #464's reuse-on-switch path works). A NEW `<webview>` is mounted and attaches to that same port. But the only thing in A's xterm buffer is:

> `Error: Session ID <sidA> is already in use.`

That phrasing is claude's "session id collision" error, not ttyd's. Hypothesis: when the new ws client connects to A's existing ttyd, ttyd respawns the claude child. The original spawn ran with `--session-id <sidA>` (no JSONL on disk yet), but by the time the new client connects, ALPHA has been written to disk, so the second claude invocation sees existing JSONL and refuses.

Net effect: switch-session destroys A's conversation as soon as the user switches back. UX F (switch-session-keeps-chat) is broken.

### Relation to #464
#464 fixed this for app-restart and import paths by checking on-disk JSONL at `openTtydForSession` time. But on session switch we never re-enter `openTtydForSession` — the renderer's reuse path keeps the same ttyd alive. The bug is in whatever ttyd-internal behavior re-runs the child on new client connection (or in our `-W` / `--once` flag selection).

### What I changed in this PR
- Added `dismissWelcomeSplash` import + 3 call sites.
- Added `waitForXtermBuffer(/ALPHA/, 15s)` settle before the scrollback assertion.
- Improved the failure message to dump the actual buffer JSON (so future failures don't print empty strings).

NOT force-pushing per instructions — leaving this PR open as the live repro for the new product bug.

### Repro
```
git checkout fix/probe-461-disambiguate
node scripts/probe-real-switch-session-keeps-chat.mjs
```
Reproduces 100% on `working` at 1b74c94 (3/3 runs).

Jiahui-Gu:
Disambiguation branch with splash fix + replay wait + improved diagnostics pushed as `fix/probe-461-disambiguate` (commit d9b57aa). Pull that branch to get the live repro on top of current working.

Jiahui-Gu:
Superseded by #465 — the regression probe scripts/probe-real-switch-session-keeps-chat.mjs was bundled atomically with the ttyd-wrapper fix (per one-bug-one-worker rule) and merged at b395d01. Closing this PR as content is already in working.

PR #462, opened 2026-04-28
[FAIL] test(e2e): probe-real-new-session-chat (UX C)

Jiahui-Gu:
## Probe steps (as implemented)
1. `createIsolatedClaudeDir()` — copies `settings.json` + friends from `~/.claude` into a tempdir (no chat history).
2. `launchCcsmIsolated({ tempDir })` — boots prod bundle with `CCSM_PROD_BUNDLE=1`, isolated `HOME` / `USERPROFILE` / `CLAUDE_CONFIG_DIR` / `CCSM_CLAUDE_CONFIG_DIR`.
3. Wait for App's `claudeAvailable` probe to leave the indeterminate state (no `data-testid="claude-availability-probing"`).
4. `seedSession(win, { name: 'probe-new-session', cwd: tempDir })` — `createSession` also sets `activeId`, which is what causes `<TtydPane>` to mount on the right.
5. `waitForWebviewMounted(win, electronApp, sid, { timeout: 60000 })` — host-page `<webview title="ttyd session ${sid}">` + main-process WebContents + xterm + `window.term`.
6. `waitForXtermBuffer(electronApp, wcId, /trust|claude|welcome|│|╭|>/i, 30s)` — claude TUI rendered.
7. Dismiss loop (≤12 iters) for trust dialog (`1\r`), "Welcome back!" splash (Enter), theme picker / security notice (Enter), until input box `│ >` visible.
8. `sendToClaudeTui(electronApp, wcId, 'say hi in 3 words')` + `\r`.
9. Poll xterm buffer for substantive reply text after the echoed prompt (90s timeout).
10. Assert no error toast (assertive aria-live region empty) and no `ttyd-exit` flip (no Retry button).
11. Cleanup via helper + `electronApp.close()`.

## Result: FAIL on step 5 (webview-mount)

```
[STEP] isolated-claude-dir: PASS {"tempDir":"C:\Users\jiahuigu\AppData\Local\Temp\ccsm-probe-claude-fUx2cI"}
[STEP] launch-ccsm: PASS
[STEP] claude-availability-resolved: PASS
[STEP] seed-session: PASS {"sid":"2f9a4b79-7f9d-4f38-9c13-900ee1097342"}
[...]

PR #460, opened 2026-04-28
[FAIL] test(e2e): probe-real-import-resume (UX H)

Jiahui-Gu:
## UX scenario (UX H)

User has a prior raw \`claude\` CLI conversation on disk
(\`~/.claude/projects/<cwd-encoded>/<sid>.jsonl\`). They open ccsm
and run the Import flow → ccsm creates a session row whose id ==
JSONL UUID with \`resumeSessionId\` set → user clicks the imported
session → ttyd spawns claude with \`--resume\` → prior conversation
context is restored in the TUI → user sends a follow-up and claude
replies as a continuation.

## Probe steps

1. \`createIsolatedClaudeDir\` → tempDir (only auth files copied).
2. Pre-seed a synthetic JSONL transcript at
   \`tempDir/.claude/projects/<encoded cwd>/<sid>.jsonl\` with a
   \`user\` frame (\`PROBE_IMPORT_PING ... PROBE_IMPORT_PINEAPPLE\`),
   an \`ai-title\` frame, and an \`assistant\` frame. \`sid\` is a
   random UUIDv4. cwd = pool-9 worktree root (real, not a ccsm temp,
   so the scanner's \`isCCSMTempCwd\` filter doesn't drop it). cwd is
   encoded by replacing every \`\\`, \`/\`, \`:\` with \`-\`
   (verified empirically).
3. \`launchCcsmIsolated({tempDir})\` (sets HOME, USERPROFILE,
   CLAUDE_CONFIG_DIR, CCSM_CLAUDE_CONFIG_DIR all to tempDir).
4. Drive Import in the renderer via
   \`window.ccsm.scanImportable()\` + the store's \`importSession\`
   action — same code path the ImportDialog uses on \"Import\".
   Assert the imported session id, name, cwd, and resumeSessionId
   match the seed.
5. Flip \`activeId\` to the imported session, wait for
   \`webview[title=\"ttyd session <sid>\"]\` to mount and xterm to
   initialize inside the OOPIF.
6. Wait up to 30s for the xterm buffer to contain the seeded user
   text (\`PROBE_IMPORT_PING\`) — i.e. claude --resume's history
   replay.
7. Send a follow-up prompt, wait up to 90s for claude's reply.
8. Assert no \`ttyd-exit\` event fired.

## Output (single run on \`working\` HEAD = cfcdcbf, fresh
\`npm run build\`)

Result: **FAIL** at step 6 (\`history-restored\`).

\`\`\`
[STEP] isolate-claude-dir: PASS
[...]

PR #458, opened 2026-04-28
test(e2e): probe-real-cwd-projects-claude (UX E)

Jiahui-Gu:
## UX scenario (UX E)

Regression guard for **PR #494** (cliBridge cwd plumbing). When user creates a session with `cwd=X`, the spawned claude process must:

1. Run with `process.cwd === X` (not Electron's cwd, not the pool worktree).
2. Write its JSONL transcript to `<CLAUDE_CONFIG_DIR>/projects/<hash-of-X>/<sid>.jsonl` where the hash matches claude's own scheme. NOT to `~/.claude/projects/<electron-cwd>/...`.
3. Pick up project-scoped files (CLAUDE.md / agents / skills / commands) from cwd X.

## Probe steps

1. ` createIsolatedClaudeDir()` -> isolated tempDir (used as both HOME and CLAUDE_CONFIG_DIR).
2. Create `\${tempDir}/my-project/` with marker file `CCSM-PROBE-MARKER.txt` containing a unique token.
3. Launch ccsm prod bundle with isolated env.
4. ` seedSession(name='cwd-test', cwd=\${tempDir}/my-project)` -> sid (UUID v4 -> matches claude sid via `toClaudeSid` round-trip).
5. Wait for ttyd webview + xterm + `window.term`.
6. Wait for banner to render `~\my-project` (claude renders cwd HOME-relative -> proves claude's process.cwd is correct).
7. Press Enter a few times to dismiss claude's \"Welcome back!\" splash card so the input box becomes interactive.
8. Send a real user prompt embedding the marker token (forces claude to write a `user` line to the JSONL).
9. Wait for claude to receive/render the prompt.
10. Glob `\${tempDir}/projects/* /<sid>.jsonl`. Assert:
    - Exactly one match exists.
    - The parent dir name encodes \"my-project\" (claude's hashing scheme: `\` and `:` -> `-`).
    - JSONL file is non-empty, first line is valid JSON, and the marker token is present.
    - No other dir under `projects/` encodes Electron's cwd (no cwd leak).

## Output

```
[PROBE] sid = 848cb801-1e89-4611-9e59-08c62a9a6eb8
[PROBE] banner shows cwd encodes my-project
[PROBE] claude replied
[PROBE] projects/ listing: [
  'C--Users-jiahuigu-AppData-Local-Temp-ccsm-probe-claude-VqiWSe-my-project'
]
[...]

PR #457, opened 2026-04-28
feat(probe): probe-utils-real-cli helper for real-claude e2e

Jiahui-Gu:
## Rationale

Five real-claude e2e probes are about to land. Before duplicating the OOPIF webview drive / xterm canvas-buffer read / isolated-config launch pattern across all of them, factor it into a single helper module that all five import from.

The patterns come from working code in scripts/dogfood-probe-happy-path.mjs (8/8 PASS in production), so this is a refactor-and-share, not a new design.

## Exports

OOPIF helpers
- getTtydWebContentsId(electronApp, {timeout})
- executeJavaScriptOnWebview(electronApp, wcId, scriptString)
- waitForXtermBuffer(electronApp, wcId, regex, {timeout})
- sendToClaudeTui(electronApp, wcId, text)
- readXtermLines(electronApp, wcId, {lines})

Tempdir isolation
- createIsolatedClaudeDir({keep}) returns {tempDir, cleanup}

Electron launch
- launchCcsmIsolated({tempDir, env}) returns {electronApp, win, userDataDir, cleanup}

Renderer state seeding
- seedSession(win, {sid, name, cwd, groupId}) returns {sid}
- waitForWebviewMounted(win, electronApp, sessionId, {timeout}) returns wcId

## Critical correctness points (encoded in the helper)

1. xterm canvas: read via term.buffer.active.getLine(N).translateToString() — DOM textContent is empty.
2. Bracketed-paste: send via term._core._coreService.triggerDataEvent(text, true) — claude's Ink TUI drops paste sequences.
3. Auth on Windows: copy settings.json + config.json + .claude.json (Agent Maestro proxy lives there); skip projects/sessions/history.
4. HOME override: set HOME, USERPROFILE, CLAUDE_CONFIG_DIR, CCSM_CLAUDE_CONFIG_DIR together (Node + Electron + main + renderer + claude binary all read different ones).
5. Prod bundle: throws clear error if dist/renderer/index.html missing.
6. Cleanup: registered on exit + SIGINT + SIGTERM + uncaughtException — survives Ctrl+C.

## Smoke test (run before push, not committed)

7/7 helpers PASS against real claude:

```
[SMOKE] createIsolatedClaudeDir: PASS
[SMOKE] launchCcsmIsolated: PASS
[SMOKE] seedSession: PASS {sid: 6be5e927-...}
[SMOKE] waitForWebviewMounted: PASS {wcId: 2}
[...]

PR #463, opened 2026-04-28
fix(probe-utils): helper gaps from 5 sister probes

Jiahui-Gu:
## Summary

Patches `scripts/probe-utils-real-cli.mjs` (the shared real-claude probe helper merged via #457) to address gaps surfaced by the 5 sister probes (#458 / #459 / #460 / #461 / #462).

Touches ONLY the helper. No probe files modified. No product code touched.

## Helper changes

| # | Change | Surfaced by |
|---|---|---|
| 1 | `dismissWelcomeSplash(electronApp, wcId, opts?)` — new helper. Detects claude's "Welcome back!" / "Try /help" / "press Enter" splash card via xterm buffer text and sends Enter via `triggerDataEvent` (NOT bracketed-paste) up to N times until the splash disappears or the input prompt indicator (`│ >`) is visible. Returns `{ dismissed, attempts, finalScreen }`; never throws. | #461 (#506) — the `A-first-reply` PASS was a false positive: ALPHA never reached claude because the first keystrokes hit the splash. Same hazard called out by #458 (#504, UX E) which had to inline `Enter` mashing. |
| 2 | `waitForWebviewMounted` — selector now uses `state: 'attached'` instead of Playwright's default `state: 'visible'`. Electron `<webview>` reports visibility inconsistently to Playwright on Windows; presence in DOM is the correct semantic for an OOPIF probe (the xterm-readiness check that follows already validates the inner page). | #462 (#505) — main-process logs showed ttyd fully wired (port listening, websocket connected, claude PID started) yet `waitForSelector` (visible) timed out at 60s. |
| 3 | `launchCcsmIsolated({ tempDir, userDataDir?, env? })` — accepts caller-supplied `userDataDir`. When supplied, helper uses it verbatim and does NOT register it for cleanup (caller owns lifecycle). Without it, behaviour is unchanged (mint fresh + auto-cleanup). | #459 (#507, UX G) — reopen-resume probe needs ccsm.db to persist across two `launchCcsmIsolated` calls on the same Electron user-data-dir. The previous mkdtempSync-always behaviour blocked cross-restart probes. |
| 4 | `process.on('unhandledRejection', cleanup)` hook installed alongside existing `exit` / `SIGINT` / `SIGTERM` / `uncaughtException` hooks. Async electron crashes now run the tempdir cleanup instead of leaking. | #457 review feedback (reviewer noted async crash leaks). |
| 5 | `waitForWebviewMounted` URL filter — looks up the session's ttyd port via `bridge.getTtydForSession(sid)` IPC and passes it to `getTtydWebContentsId({ port })`, which filters webContents by URL prefix `http://127.0.0.1:<port>/`. Eliminates the highest-id race when multiple ttyd webviews are alive simultaneously (switch-session). | #461 (#506) reviewer note — the round-trip A→B→A path keeps multiple ttyds alive, so highest-id heuristic could misfire; explicit port match is deterministic. |

Backward compatible: every existing call site keeps working without changes (port option is optional, userDataDir is optional, dismissWelcomeSplash is opt-in).

## Re-run results (after helper fix, against the cherry-picked probe files)

| Probe | Branch | Before | After helper fix | Outcome |
|---|---|---|---|---|
| #505 (UX C, new-session-chat) | `feat/probe-real-new-session` (#462) | FAIL — `waitForSelector(webview[title=...])` timeout 60s | **PASS** (all 10 steps green; claude replied in 2008ms) | helper bug — fixed by change #2 (`state: 'attached'`) |
| #506 (UX F, switch-keeps-chat) | `feat/probe-real-switch-session` (#461) | FAIL — empty xterm scrollback after switch-back | FAIL — same step (10/11 green; A's scrollback empty after switch-back). Mount/port-reuse logic all PASS (ports preserved across A→B→A round-trip, webview correctly attaches to original port 56475). | mixed — helper fix unblocks the mount path but the probe itself doesn't yet call `dismissWelcomeSplash`, so `A-first-reply` is still a false positive (claude shows only `❯` and `? for shortcuts` in the tail, no ALPHA echo — first keystrokes ate by splash) |

### Outcome case: **2** (split)

- **#505 was a probe/helper bug.** The new helper unblocks it cleanly — 10/10 PASS. Recommend the manager re-review #462 against this helper PR (no probe-side changes needed) and merge.
- **#506 is partially helper, partially probe-side.** The cliBridge port-reuse path is verified correct (4 PASS steps explicitly proving A's port is preserved and re-attached). The remaining FAIL is the scrollback assertion, which can't be evaluated yet because the splash-card hypothesis can't be ruled out from this run (probe doesn't call `dismissWelcomeSplash` because it predates this helper). 
  - Recommended next step: a tiny follow-up worker on #461's branch to insert `await dismissWelcomeSplash(electronApp, wcA)` between `select-A-mounted` and the first ALPHA send. After that change, the run will be definitive: 
    - If ALPHA echoes back AND scrollback persists → all green, full PASS.
    - If ALPHA echoes back AND scrollback empty after switch → genuine ttyd-scrollback product bug (task #511 / new bug-fix worker scope).

### Recommended manager next step

1. Review + merge this helper PR.
2. Re-trigger #462's reviewer (probe should now PASS unchanged against the merged helper).
3. Dispatch a 15-minute follow-up worker on #461 to wire `dismissWelcomeSplash` into the probe; that worker will produce the definitive Case 2-or-Case 4 verdict for #506.

## Local verification commands

```bash
# Helper syntax
node --check scripts/probe-utils-real-cli.mjs

# Build prod bundle (TtydPane.tsx newer than dist/renderer/index.html)
npm run build

# Cherry-pick + run sister probes against the new helper
git checkout origin/feat/probe-real-new-session -- scripts/probe-real-new-session-chat.mjs
git checkout origin/feat/probe-real-switch-session -- scripts/probe-real-switch-session-keeps-chat.mjs
node scripts/probe-real-new-session-chat.mjs   # PASS
node scripts/probe-real-switch-session-keeps-chat.mjs   # FAIL at scrollback (probe-side splash-dismiss missing)
```

## Test plan

- [x] `node --check scripts/probe-utils-real-cli.mjs` — clean
- [x] Re-ran #505's probe against the helper — PASS exit 0
- [x] Re-ran #506's probe against the helper — same step still FAIL but for diagnosable reason (probe doesn't yet call new `dismissWelcomeSplash`)
- [x] No probe files modified in this PR
- [x] No product code touched

### DIFF
--- a/scripts/probe-utils-real-cli.mjs
+++ b/scripts/probe-utils-real-cli.mjs
@@ -37,7 +37,11 @@
 //      CLAUDE_CONFIG_DIR.
 //
 //   6. Cleanup must run on Ctrl+C too — registered via process.on('exit')
-//      AND process.on('SIGINT') / SIGTERM.
+//      AND process.on('SIGINT') / SIGTERM / uncaughtException /
+//      unhandledRejection. Cleanups run LIFO so process force-kill
+//      (electron + ttyd + claude subtree, registered in launchCcsmIsolated)
+//      runs BEFORE tempdir rmSync — Windows can't remove a directory whose
+//      files are still locked by live processes.
 
 import { _electron as electron } from 'playwright';
 import {
@@ -52,6 +56,7 @@ import {
 import { homedir, tmpdir } from 'node:os';
 import * as path from 'node:path';
 import { randomUUID } from 'node:crypto';
+import { spawnSync } from 'node:child_process';
 
 // ============================================================================
 // OOPIF helpers
@@ -251,7 +256,12 @@ function installCleanupHooks() {
   if (cleanupHooksInstalled) return;
   cleanupHooksInstalled = true;
   const runAll = () => {
-    for (const fn of cleanupRegistry) {
+    // Iterate in reverse insertion order (LIFO) so process-kill cleanups
+    // registered after tempdir cleanups run FIRST. Order matters on
+    // Windows: rmSync on a tempdir fails if electron / ttyd / claude still
+    // hold file locks inside it.
+    const fns = Array.from(cleanupRegistry).reverse();
+    for (const fn of fns) {
       try { fn(); } catch (_) { /* ignore */ }
     }
     cleanupRegistry.clear();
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
Amended per review feedback:
- Added force-kill of electron PID + subtree (taskkill /T /F on Windows, SIGKILL elsewhere) in cleanup chain, runs before tempdir rm so locked files unlock first.
- Wrapped getTtydForSession call in Promise.resolve to handle missing window.ccsmCliBridge.

Ready for re-review.

PR #469, opened 2026-04-28
chore(deps): add node-pty + xterm deps (rewrite groundwork)

Jiahui-Gu:
## Summary
- Add node-pty 1.1.0 and the full @xterm/xterm + addons set used by the upcoming ptyHost / TerminalPane rewrite
- Add asarUnpack for node-pty native binding (mirrors better-sqlite3 pattern)

This is purely deps. No code consumes these yet — that lands in PR-2..PR-6. PR-8 收口 will rewrite after-pack and drop the ttyd binary block.

## Test plan
- [ ] CI green (build may warn about bundle size — ok)
- [ ] No lockfile churn beyond the new deps + their transitive

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,14 @@
         "@radix-ui/react-tooltip": "^1.1.6",
         "@sentry/electron": "7.11.0",
         "@sentry/react": "10.49.0",
+        "@xterm/addon-canvas": "^0.7.0",
+        "@xterm/addon-clipboard": "^0.2.0",
+        "@xterm/addon-fit": "^0.10.0",
+        "@xterm/addon-serialize": "^0.13.0",
+        "@xterm/addon-unicode11": "^0.9.0",
+        "@xterm/addon-web-links": "^0.12.0",
+        "@xterm/headless": "^5.5.0",
+        "@xterm/xterm": "^5.5.0",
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment — gh blocks self-approve) — manager may merge once Wave 2 (PR-3 #472, PR-5 #473) merges so build is green.

Layer 1 (necessity):
- Deps match the spike at spike/xterm-attach exactly: same 8 @xterm/* packages + node-pty 1.1.0.
- node-pty pinned to exact 1.1.0 (no caret) — correct for a native module that needs electron-rebuild.
- @xterm/* on ^ carets matches spike; no reason to deviate.

Layer 2 (implementation):
- Surgical inserts in package.json (11 lines), alphabetical, no unrelated reorderings.
- package-lock.json adds only the new packages + transitive (js-base64, node-addon-api 7.1.1) — clean.
- asarUnpack `**/node_modules/node-pty/**` mirrors the existing better-sqlite3 entry — correct pattern for a native addon.
- extraResources untouched. ttyd binary removal correctly deferred to PR-8 收口 per the wave plan.

Base is `working`. No CI gate applied per manager guidance (collection sequenced; standalone CI red is expected).

PR #471, opened 2026-04-28
feat(ptyHost): main-process pty host module

Jiahui-Gu:
## Summary
- New module electron/ptyHost/index.ts: main-process node-pty manager paired with @xterm/headless + addon-serialize for snapshot replay on attach
- IPC surface: pty:{list,spawn,attach,detach,input,resize,kill,get} + pty:data/pty:exit broadcasts
- Reuses claudeResolver; reimplements toClaudeSid projection + JSONL --session-id/--resume probe (no cross-module import — cliBridge is being deleted in PR-8)

Standalone — not yet wired into main.ts (PR-5) or preload (PR-3). Compiles only after PR-1 (deps) lands.

## Test plan
- [ ] type-check passes after PR-1 merges
- [ ] runtime validation deferred to PR-5 wire-up + PR-6 renderer swap

### DIFF
--- a/electron/ptyHost/index.ts
+++ b/electron/ptyHost/index.ts
@@ -1,0 +1,388 @@
+// Main-process pty host.
+//
+// Replaces the ttyd-in-iframe transport (electron/cliBridge/processManager.ts)
+// with an in-process node-pty + @xterm/headless pair per ccsm session. The
+// renderer attaches via a `<webview>` (or direct xterm in-process) and consumes
+// `pty:data` chunks; on (re)attach it gets a serialized snapshot of the
+// headless terminal so reopening a session paints the prior screen
+// instantaneously without re-running claude.
+//
+// Why in-process pty over ttyd:
+//   - No second HTTP server / ws hop / Defender firewall prompt.
+//   - No respawn-on-ws-disconnect tearing claude down on session switch (the
+//     headless mirror keeps the live screen across renderer detaches and
+//     replays it on reattach).
+//   - Direct write/resize IPC, no ttyd protocol middleman.
+//
+// JSONL-existence picks --session-id vs --resume on EVERY spawn (mirrors the
+// TTYD_WRAPPER_CMD logic in cliBridge/processManager.ts). Wrapper not needed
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (cannot self-approve via gh; treat this as approval).

Layer 1 (architecture):
- node-pty + @xterm/headless + SerializeAddon mirror matches spike R3. Snapshot replay on attach correctly relies on p.onData always writing to headless regardless of attached count. Verified.
- Re-implementing toClaudeSid + JSONL probe (vs importing from cliBridge) is correct given PR-8 will delete cliBridge — avoids backwards dependency. toClaudeSid is byte-identical to processManager.ts:91.
- IPC channel set (pty:list/spawn/attach/detach/input/resize/kill/get + pty:data/pty:exit) matches PR-3 #472 preload and PR-7 #474 harness exactly.
- JSONL semantics: cliBridge wrapper checks 'if exist' only; PR-2 additionally requires size > 0. Defensible improvement (avoids --resume against an empty/race file). Two-root scan (CLAUDE_CONFIG_DIR/projects + USERPROFILE/.claude/projects) matches the wrapper.

Layer 2 (implementation):
- pty:attach registers wc.once('destroyed') that removes the entry from the attached map — multi-window leak path is wired.
- broadcast filter: only entry.attached webContents receive pty:data; isDestroyed gate before send.
- killAllPtySessions exported for before-quit; spawn errors return {ok:false, error}; resize bounds-checked (>=2); kill is idempotent; pty.write/resize/kill wrapped in try.
- claudeResolver import is the only cliBridge symbol used. resolveClaude() signature matches.

Nits (non-blocking):
- pty:detach IPC handler doesn't remove the 'destroyed' listener registered in pty:attach. Repeated attach→detach cycles on the same wc accumulate listeners; firings are idempotent (delete is no-op) — tiny memory drip, not a correctness issue.
- env cast 'process.env as {[key:string]:string}' glosses over possible undefined values; acceptable for node-pty typing.

Standalone PR will not typecheck until PR-1 #469 lands; called out in the body.

PR #473, opened 2026-04-28
feat(main): wire ptyHost IPC + drop webviewTag

Jiahui-Gu:
## Summary
- Register ptyHost IPC handlers in main process (parallel to existing cliBridge — both run during transition)
- Add killAllPtySessions() to before-quit reap
- Drop webviewTag from BrowserWindow webPreferences (no more webview)
- Remove CCSM_DEBUG_WEBVIEW handler (webview gone)

Compiles once PR-1 (deps) + PR-2 (ptyHost module) land.

## Test plan
- [ ] type-check passes after PR-1 + PR-2
- [ ] runtime validation in PR-6 + PR-7

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -71,6 +71,7 @@ import {
   bindCliBridgeSender,
   shutdownCliBridge,
 } from './cliBridge';
+import { registerPtyHostIpc, killAllPtySessions } from './ptyHost';
 
 // ─────────────────────── IPC security helpers ────────────────────────────
 //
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## Review: APPROVE (self-PR, can't formally approve)

Layer 1 (necessity): all five direction calls correct.
- ptyHost IPC registered in parallel to cliBridge — proper transition wiring; cliBridge stays for PR-8 to delete.
- before-quit calls killAllPtySessions() — mirrors shutdownCliBridge, prevents node-pty child leaks on quit.
- webviewTag: true removed — TtydPane is gone, so the flag is dead weight (and a security widening).
- CCSM_DEBUG_WEBVIEW handler removed cleanly — entire web-contents-created listener was strictly gated by `getType() === 'webview'` return-early, so no non-webview branch was lost.
- cliBridge code untouched, as planned.

Layer 2 (impl):
- `registerPtyHostIpc(ipcMain, () => BrowserWindow.getAllWindows()[0] ?? null)` — lazy callback resolves at IPC dispatch time, so 'window not yet created at registration' concern is moot. Matches PR-2's signature; ptyHost currently void-uses the param (reserved for future per-window routing) — forward-compatible.
- Verified no residual `webviewTag` refs anywhere under `electron/` (the single 'future webview' mention is in an unrelated CSP comment).
- Diff is purely additive on ipc-register side, surgical removal on webview side. -38/+13, zero collateral.

Standalone — won't compile until #469 + #471 land, as expected. Ready to merge once predecessors are in.

PR #477, opened 2026-04-28
fix(terminal): focus xterm after PTY attach so first keystroke reaches the CLI

Jiahui-Gu:
## Summary
- Fixes #548. After creating/importing/reopening a session via sidebar button or keyboard shortcut, the first keystroke was a no-op because focus stayed on `document.body` after `App.newSession` blurred the trigger; nothing handed focus to the embedded xterm.
- TerminalPane's attach effect now calls `term.focus()` once PTY attach + initial fit complete. Covers every entry path because they all route through that effect.
- E2E: `harness-real-cli` `new-session-chat` and `import-resume` cases gain a new `assertCliFocused` helper that asserts `document.activeElement === .xterm-helper-textarea` inside `[data-terminal-host][data-active-sid=<sid>]`. Polls up to 5s post-`waitForTerminalReady`.

## Root cause
Renderer never called `term.focus()` after `pty.attach`; App.tsx blurred the trigger but did not transfer focus, so `document.activeElement` fell back to `<body>` and `Enter` went nowhere.

## Files changed
- `src/components/TerminalPane.tsx` — call `term.focus()` after attach + post-attach fit.
- `scripts/harness-real-cli.mjs` — add `assertCliFocused`; wire into `new-session-chat` and `import-resume`.

## Test plan
- [x] `node scripts/harness-real-cli.mjs --only=new-session-chat` — RED before fix (`activeTag: BODY`), GREEN after.
- [x] `node scripts/harness-real-cli.mjs --only=new-session-chat,import-resume` — both PASS (24.8s + 3.8s).
- [x] `node scripts/harness-real-cli.mjs --only=new-session-focus-cli` — still PASS (no regression of the existing double-fire guard).

HEAD: 3818493fba64f45dbb8b3ddc04bbf505d3e764d5

### DIFF
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -176,6 +176,12 @@ async function caseNewSessionChat({ electronApp, win, tempDir }) {
 
   await waitForTerminalReady(win, sid, { timeout: 60000 });
 
+  // Task #548 — after the new session attaches, focus must land on the
+  // embedded xterm so the user's first keystroke goes to claude's TUI
+  // (not the trigger button or the document body). xterm's input target
+  // is the `.xterm-helper-textarea` element it injects under the host div.
+  await assertCliFocused(win, sid, 'new-session-chat');
+
   await waitForXtermBuffer(win, /trust|claude|welcome|│|╭|>/i, { timeout: 30000 });
 
   // Dismiss trust / welcome / theme splashes.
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
**APPROVE**

Independent review in pool-2, fresh build of `3818493`.

**Layer 1 (necessity / direction):** `term.focus()` in TerminalPane attach effect is the right place. Effect re-runs only on `sessionId` change (intentional session switch) and `attachNonce` (Retry button) — both are user-initiated focus handoffs. No risk of stealing focus from a Radix dialog or sidebar rename input: opening a dialog or renaming doesn't change `sessionId`, and Radix's focus-trap re-asserts on next tick if it ever did. Smaller / cleaner than App-level orchestration.

**Layer 2 (impl):** Focus call is inside the `try` after successful attach + spawn-fallback + post-attach fit, so error paths (`catch` → `kind: 'error'`) correctly skip it. Wrapped in try/catch with warn fallback — defensive without swallowing. Selector `[data-terminal-host][data-active-sid="<sid>"] .xterm-helper-textarea` matches the actual host attrs on TerminalPane lines 417-418. `document.activeElement === ta` is the correct DOM-truth check; 5s poll handles the React-commit race. Comments are excellent — link the task and explain App.tsx blur context.

**Reverse-verify:** stashed the `term.focus()` block, rebuilt, ran `--only=new-session-chat` → FAIL with `activeTag: BODY` (exact regression shape). Restored, ran both cases → PASS (24.9s + 3.8s). Assertion is real and discriminating, not a tautology.

**Edge cases:** A→B switch desired-focus behavior preserved. No `attachNonce` bumps from non-user code paths.

No nits. Ship it.

PR #478, opened 2026-04-28
fix(session): default new-session cwd to last-used (userCwds LRU head) (#551)

Jiahui-Gu:
## Summary

Reverses PR #392's "default cwd is always home" policy. The user re-picked the same project on every new session because the LRU was consulted only by the StatusBar picker, never by the new-session default. Per task #551: repeat use of the same project should not require re-picking.

- `createSession` now defaults to `lastUsedCwd ?? userHome`. Caller-provided `opts.cwd` still wins.
- `lastUsedCwd` is a renderer-side cache of the ccsm-owned `userCwds` LRU head, seeded at hydrate from `window.ccsm.userCwds.get()` and refreshed (optimistic + IPC-resolved) on every `userCwds.push` triggered by `createSession`, `setSessionCwd`, `importSession`.
- `importSession` now feeds its cwd into the LRU too — importing an existing transcript is a strong "user works in this cwd" signal.

## Out of scope

- **cwd-missing degradation.** The existing `cwd-missing` flag on Session continues to surface paths that no longer exist on disk; no new UI is added for that case (per task brief).

## Visual change

None. Picker layout, popover styling, and all rendered chrome are byte-identical. Only the data path that defaults a new session's cwd changed. Screenshots are not applicable.

## Test plan

- [x] `node scripts/harness-real-cli.mjs --only=default-cwd-from-userCwds-lru` — PASS (216ms) on this branch.
- [x] Reverse-verify: `git stash push src/stores/store.ts` + rebuild + rerun probe — FAILS with `actual=userHome, expected=lru-head`. Restoring the fix returns to PASS.
- [x] `npm run build` clean (3 webpack size warnings unchanged from `working`).
- [ ] Reviewer: confirm caller-passed `opts.cwd` still beats the LRU head (createSession line ~501: `cwd: opts.cwd ?? defaultCwd`).
- [ ] Reviewer: confirm hydrate seed skips the empty-LRU sentinel (`getUserCwds()` always appends home; renderer must not treat that as a real LRU entry).

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -162,9 +162,11 @@ async function getImportableSessions(): Promise<ScannableSession[]> {
 
 // ───────────── user-owned cwd list (ccsm's own LRU) ─────────────
 //
-// The new-session default cwd is always `os.homedir()`. The recent list shown
-// in the StatusBar cwd popover is a user-owned LRU that only the user can
-// extend (by explicitly picking a non-default cwd). Persisted in the
+// The new-session default cwd is the LRU head (the user's most-recently
+// picked cwd) when present, falling back to `os.homedir()` when the LRU
+// is empty (fresh install). The recent list shown in the StatusBar cwd
+// popover is a user-owned LRU that only the user can extend (by
+// explicitly picking a non-default cwd). Persisted in the
 // `app_state` SQLite table under key `userCwds` as a JSON string list.
 //
 // Reads return `[homedir()]` when the list is empty so the popover always has
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -6,9 +6,10 @@
 //   2. switch-session-keeps-chat     — UX F: session A↔B switch reuses pty, scrollback intact
 //   3. cwd-projects-claude           — UX E: real cwd flows into claude's JSONL hash
 //   4. import-resume                 — UX H: import existing JSONL, claude --resume restores
-//   5. new-session-focus-cli         — UX C': button focus does not double-fire
-//   6. pty-pid-stable-across-switch  — direct-xterm: pty pid stable across A→B→A switch
-//   7. reopen-resume                 — UX G: close ccsm, reopen, click session, --resume restores
+//   5. default-cwd-from-userCwds-lru — task #551: new-session cwd defaults to LRU head
+//   6. new-session-focus-cli         — UX C': button focus does not double-fire
+//   7. pty-pid-stable-across-switch  — direct-xterm: pty pid stable across A→B→A switch
+//   8. reopen-resume                 — UX G: close ccsm, reopen, click session, --resume restores
 //
 // Sharing strategy:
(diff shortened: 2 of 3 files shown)


Jiahui-Gu (commented):
APPROVE

## Layer 1 — Store-level last-used logic
- `lastUsedCwd` slot correctly seeded in `hydrateStore` from `window.ccsm.userCwds.get()` and skipped when head === userHome (the empty-LRU sentinel that `getUserCwds()` always appends). No race with userCwds population: the seed runs in the same async block that awaits userHome resolution, after probe state clears.
- `createSession` default chain `lastUsedCwd ?? userHome ?? ''` matches spec. `opts.cwd` still wins.
- Optimistic local update (`if (finalCwd !== lastUsedCwd) set({ lastUsedCwd: finalCwd })`) before IPC resolve correctly handles back-to-back createSession calls. The IPC `.then()` re-syncs from the authoritative `pushUserCwd` return.
- `changeCwd` (StatusBar pick) and `importSession` both push to LRU and update lastUsedCwd identically. Importing now correctly feeds the LRU — confirmed at lines 603–617.
- `pushUserCwd` returns the raw stored list head (not getUserCwds's home-appended view), so reading `list[0]` from the push promise always yields the freshly pushed cwd. Correct.
- Edge: empty LRU → getUserCwds returns [home], head === home, seed skipped, createSession falls back to userHome. Verified.
- Edge: session deleted while its cwd is LRU head → `lastUsedCwd` survives (it's independent of session list); next createSession still inherits. Intended.
- Nit: PR description says `setSessionCwd`; actual store method is `changeCwd`. Code is correct, just a doc mismatch.

## Layer 2 — E2E probe `default-cwd-from-userCwds-lru`
- Reverse-verify confirmed locally on Windows: stashed the `lastUsedCwd ??` part of the createSession default → probe FAILED with clear diagnostic (`actual=tempDir/userHome, expected=lru-project-*`); restored → PASSED. Assertion is discriminating.
- Negative assertion `actual !== userHome` correctly catches the regressed default even when the LRU happens to coincidentally match.
- Coverage gap (acknowledged in probe comment, not blocking): probe manually `setState({lastUsedCwd: head})` to mirror hydrateStore, so a regression in the hydrateStore seed step alone wouldn't be caught. The seed code is a thin mirror of probe logic; load-bearing path (createSession default) is fully covered.
- LRU pollution from synthetic `lru-project-*` push persists in shared launch but does not leak into subsequent cases — every other shared case passes explicit `cwd: tempDir` to `seedSession`. Verified by running `--only=default-cwd-from-userCwds-lru,new-session-chat,switch-session-keeps-chat,import-resume,pty-pid-stable-across-switch`: 5/5 PASS.

## Anti-regression
Local run on Windows (pool-1, fresh build):
- new-session-chat 24.8s PASS
- switch-session-keeps-chat 31.3s PASS
- import-resume 3.8s PASS (importSession's new LRU push doesn't break resume)
- default-cwd-from-userCwds-lru 43ms PASS
- pty-pid-stable-across-switch 18.9s PASS

Approve. Ship it.

PR #479, opened 2026-04-28
refactor(e2e): hoist dismissFirstRunModals helper (#519)

Jiahui-Gu:
## Summary

- Hoist three drifting copies of trust/welcome-modal dismissal in `scripts/harness-real-cli.mjs` into a single exported helper `dismissFirstRunModals(win, opts)` in `scripts/probe-utils-real-cli.mjs`.
- Replace two inline 12-iter loops (in `caseNewSessionChat` around L189 and `casePtyPidStableAcrossSwitch` around L855) and one module-local `dismissFirstRunModals` (around L1000) with `await dismissFirstRunModals(win)`.
- Default opts `{ maxIters: 12, intervalMs: 250 }`. Behavior is the union of the previous three: phase 1 trust-or-bare-Enter loop, phase 2 `dismissWelcomeSplash` for the hint card, phase 3 a few bare-Enter retries.

Diff is scoped strictly to the two files; no behavioral change beyond the union merge.

## Test plan

- [x] `node --check` both .mjs files: OK.
- [x] `node scripts/harness-real-cli.mjs --only=new-session-chat`: PASS (19535ms).
  ```
  [HARNESS] >>> case: new-session-chat
  [HARNESS] <<< PASS new-session-chat (19535ms)
    total: 1/1 passed, 48.0s wall
  ```
- [x] `node scripts/harness-real-cli.mjs --only=pty-pid-stable-across-switch`: PASS (13623ms).
  ```
  [HARNESS] >>> case: pty-pid-stable-across-switch
  [HARNESS] <<< PASS pty-pid-stable-across-switch (13623ms)
    total: 1/1 passed, 20.2s wall
  ```
- Lint/typecheck scripts only target `.ts/.tsx`, so the `.mjs` edits do not affect them.

## Unrelated cleanup opportunities (not in this diff)

- `dismissWelcomeSplash` is still called inline in three other case bodies (around L281, L291, L326 in `harness-real-cli.mjs`). They could be migrated to `dismissFirstRunModals` if the union behavior is desired in those cases too.
- `scripts/harness-real-cli.mjs` defines a local `sleep` (L64) duplicated by the internal `sleep` in `probe-utils-real-cli.mjs`; could be unified later.

### DIFF
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -52,6 +52,7 @@ import * as path from 'node:path';
 import { randomUUID } from 'node:crypto';
 import {
   createIsolatedClaudeDir,
+  dismissFirstRunModals,
   dismissWelcomeSplash,
   launchCcsmIsolated,
   readXtermLines,
@@ -186,17 +187,7 @@ async function caseNewSessionChat({ electronApp, win, tempDir }) {
   await waitForXtermBuffer(win, /trust|claude|welcome|│|╭|>/i, { timeout: 30000 });
 
   // Dismiss trust / welcome / theme splashes.
-  for (let i = 0; i < 12; i++) {
-    const lines = await readXtermLines(win, { lines: 30 }).catch(() => []);
-    const screen = lines.join('\n');
-    if (/│\s*>/.test(screen) || /^\s*>\s/m.test(screen)) break;
-    if (/trust|do you trust/i.test(screen)) {
-      await sendToClaudeTui(win, '1\r').catch(() => {});
-    } else {
-      await sendToClaudeTui(win, '\r').catch(() => {});
-    }
-    await sleep(1500);
-  }
+  await dismissFirstRunModals(win);
 
   await sendToClaudeTui(win, CHAT_PROMPT);
   await sleep(500);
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Independent reviewer verdict: APPROVE. Hoist is necessary (4 dismiss sites → 1 helper), scope clean (only the 2 named files: harness-real-cli.mjs + probe-utils-real-cli.mjs). Helper preserves the UNION of prior behaviors: 12-iter trust loop with broadened trust regex (superset of inline + local), dismissWelcomeSplash phase, and 4-iter final settle. All 4 call sites use the helper, module-local copy deleted, import added correctly, no dead-code comments / 'removed:' residue, all English, both files pass node --check. Reverse-verified by stashing the export on probe-utils and re-importing harness — fails with 'does not provide an export named dismissFirstRunModals' as expected, confirming the wiring is real and not silently broken. Note: gh review --approve blocked by 'cannot approve own PR' (single-account dogfood), so approval recorded here.

Out-of-scope follow-ups (legitimate, do not block #519): worker noted 3 other dismissWelcomeSplash call sites + sleep duplication. Worth a separate cleanup task.

PR #480, opened 2026-04-28
feat(close): ask/tray/quit + Settings + single-instance lock (#561)

Jiahui-Gu:
## Summary
Plug the zombie-main-process leak (3 explorer-parented CCSM PIDs found alive in the wild on 2026-04-28) by adding three correlated fixes:

- **Single-instance lock** (cross-platform). Custom title-bar X used to silently `win.hide()` and there was no `requestSingleInstanceLock`, so each subsequent launch spawned a fresh main process while the prior stayed alive in the tray. Now a `second-instance` handler restores+focuses the existing window instead.
- **Configurable close behavior**. `win.on('close')` now consults a persisted `closeAction` preference: `ask` | `tray` | `quit`. Default is `'ask'` on Windows/Linux and `'tray'` on macOS — mac users expect the red traffic-light to hide because that's the system convention; Windows users do not have an established expectation, so we ask once and remember the choice.
- **Settings UI**. New "Close button behavior" segmented control in Settings > Appearance, persisted via the existing `db:save`/`db:load` IPC under key `closeAction` (same KV the crash-reporting toggle uses, no new persistence infra). Main re-reads on every close so a mid-session change takes effect immediately without restart.

The native dialog ("Close ccsm? Minimize to tray | Quit, with a 'Don't ask again' checkbox") is fully localized — added a new `closeDialogCatalogs` block to `electron/i18n.ts` with en+zh, and a matching `settings.closeBehavior*` block to the renderer catalog. Parity test passes.

`isQuitting` already covered tray-menu Quit and the `before-quit` safety net, so all real quit paths bypass the new dialog correctly.

## Platform default rationale
- **macOS = 'tray'**: red traffic-light hiding is the OS convention; users expect Cmd+Q to be the only true quit. Showing a dialog every time would feel un-native.
- **Windows/Linux = 'ask'**: there is no established convention. Older Windows apps quit on X; Slack/Discord/Teams hide. We explicitly ask on first encounter and let the user lock in their preference.

## Test plan
- [x] `npm run typecheck` — pass
- [x] `npm run lint` — pass (173 errors are pre-existing, identical to baseline; zero added by this PR)
- [x] `npm test` — 270 pass / 3 fail (the 3 failures are pre-existing `better-sqlite3` ABI mismatch on the dev box, identical to baseline)
- [x] `node scripts/harness-real-cli.mjs --only=new-session-chat` — PASS (19s)
- [x] `node scripts/harness-real-cli.mjs --only=reopen-resume` — PASS (50s) — this case exercises the full app lifecycle (launch → seed session → close app → relaunch with same userDataDir → resume), which is exactly what would break if the new single-instance lock or close interceptor regressed restart flow
- [ ] Manual smoke not captured (no GUI session available in worker env). Reviewer or follow-up should:
  1. `npm start`, click X, confirm "Close ccsm?" dialog appears with two buttons + "Don't ask again" checkbox.
  2. Pick "Quit" → confirm process exits (no zombie in Task Manager).
  3. Restart, click X, pick "Minimize to tray" + check "Don't ask again" → window hides.
  4. Restart with same user-data-dir, click X again → no dialog, immediate hide.
  5. Double-click desktop icon while app is hidden → first window restores, no second main process appears.
  6. Open Settings > Appearance → "Close button behavior" segmented control reflects current value; changing it takes effect on the next X-click.

## Verification of root cause
Re-read of `electron/main.ts` matches the Layer 1 grep:
- L326-330: mac uses `titleBarStyle: 'hiddenInset'` + traffic lights; win/linux uses `frame: false`.
- L447-466 (pre-change): `win.on('close')` always preventDefault + fade-to-hide.
- L808-816: `app.on('window-all-closed')` only quits when `isQuitting`.
- L505-508: tray Quit sets `isQuitting = true` before `app.quit()`.
- L796-806: `app.on('before-quit')` sets `isQuitting = true` (safety net for any other quit path).
- No existing `requestSingleInstanceLock` — confirmed via grep.

### DIFF
--- a/electron/i18n.ts
+++ b/electron/i18n.ts
@@ -36,10 +36,19 @@ type DialogCatalog = {
   chooseCwd: string;
 };
 
+type CloseDialogCatalog = {
+  message: string;
+  detail: string;
+  tray: string;
+  quit: string;
+  dontAskAgain: string;
+};
+
 type NotificationKey = keyof NotificationCatalog;
 type TrayKey = keyof TrayCatalog;
 type MenuKey = keyof MenuCatalog;
 type DialogKey = keyof DialogCatalog;
+type CloseDialogKey = keyof CloseDialogCatalog;
 
 // IMPORTANT: keep these strings byte-identical to the renderer catalog
 // `notifications` namespace. The renderer catalog is the source of truth.
@@ -102,6 +111,29 @@ const dialogCatalogs: Record<SupportedLanguage, DialogCatalog> = {
   }
 };
 
+// Native "close to tray or quit?" prompt shown the first time the user
+// clicks the window X (when their preference is still 'ask'). Keep parity
+// with the renderer catalog `closeDialog` namespace — the same strings are
+// surfaced in the Settings explainer hint so wording stays consistent.
+const closeDialogCatalogs: Record<SupportedLanguage, CloseDialogCatalog> = {
+  en: {
+    message: 'Close ccsm?',
+    detail:
+      'Minimize to tray keeps ccsm running so notifications and background sessions stay active.',
+    tray: 'Minimize to tray',
+    quit: 'Quit',
+    dontAskAgain: "Don't ask again"
+  },
+  zh: {
+    message: '关闭 ccsm？',
+    detail:
+      '最小化到托盘会让 ccsm 继续运行，通知和后台会话保持活跃。',
+    tray: '最小化到托盘',
+    quit: '退出',
+    dontAskAgain: '不再询问'
+  }
+};
+
 let activeLanguage: SupportedLanguage = 'en';
 
 export function setMainLanguage(lang: SupportedLanguage): void {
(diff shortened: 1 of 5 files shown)


Jiahui-Gu:
Independent review: APPROVE. Layer 1: real bug (zombie main proc), root cause correctly identified, minimal plug. Scope contained to 5 named files. Architecture fit: db:save/db:load is correct — same KV that crashReportingOptOut uses (electron/main.ts:24, SettingsDialog.tsx:362). Layer 2: lock placed at module load before whenReady (main.ts:130), lock-failure uses app.quit + process.exit(0) before any window creation, second-instance restores+focuses+un-minimizes. Close interceptor: pref read every call (no cache), e.preventDefault() runs synchronously before the async dialog IIFE, isQuitting short-circuit covers tray Quit (607)/before-quit (898)/explicit-quit branch (527,563). Don't-ask-again persists via setCloseAction. Strings sentence-case en + natural zh in both main (electron/i18n.ts) and renderer; parity test compares en/zh renderer catalogs and both got matching closeBehavior keys → green. E2E reverse-verify: harness-real-cli --only=reopen-resume PASS (72.9s, full lifecycle launch → seed → close → relaunch with shared userDataDir). Non-blocking notes: (a) CCSM_E2E_NO_SINGLE_INSTANCE is redundant with CCSM_E2E_HIDDEN=1 but harmless belt-and-braces; (b) control lives in Appearance pane — a Behavior pane would be semantically nicer but Appearance is where language/theme/font already live and adding a pane is scope creep; (c) GUI smoke deferred to dogfood pass per PR test plan.

PR #481, opened 2026-04-28
test(e2e): switch-session-during-streaming (#558)

Jiahui-Gu:
## Summary

Adds new e2e case `switch-session-during-streaming` to `scripts/harness-real-cli.mjs`, closing the gap surfaced in the #558 review: existing `switch-session-keeps-chat` only switches AFTER claude's reply has fully settled, so it does not exercise the real user pain — switching mid-stream while tokens are still arriving.

## Layer 1 (necessity check) — passed

- `caseSwitchSessionKeepsChat` waits for `ALPHA` reply to fully resolve (`sendAndAwaitReply`) BEFORE the A→B switch — confirmed gap is real.
- `installPtyExitProbe` + `getPtyPidForSid` exist in `harness-real-cli.mjs` and are reused.
- `seedSession`, `sendToClaudeTui`, `waitForXtermBuffer`, `readXtermLines`, `dismissFirstRunModals` exist in `probe-utils-real-cli.mjs` and are reused.
- Switch helper is the inline pattern `window.__ccsmStore.getState().selectSession(id)` (no separate `switchToSession` helper exists in the harness — every existing case uses this inline form). Same pattern used here.

No new helpers introduced.

## Flow

1. Seed sessions A and B in the shared launch.
2. Select A, dismiss first-run modals.
3. Snapshot `pidA1` via `window.ccsmPty.list()`.
4. Send A: `"Print the numbers 1 to 50, one per line, no extra text."`
5. Wait for stream START (>=5 numbers visible). Do NOT wait for completion.
6. Switch to B (`selectSession(sidB)`), dwell 1100ms while A streams in background.
7. Assert pidA reported stable while focused on B.
8. Switch back to A. Assert pidA still equal.
9. Wait for A's stream to settle (line count stable for 1.5s AND `50` visible) — capped at 120s.
10. Assert every integer 1..50 appears exactly once on its own line in A's xterm buffer (no gaps, no duplicates).
11. Assert no `pty:exit` event fired for A throughout.

## Chosen N

**N = 50.** Within the recommended 30-80 range. 50 produces ~7-15s of streaming with claude TUI's typical 100-200 ms inter-line cadence — comfortably longer than the ~1.1s switch-back round-trip, so the assertion is non-trivial. Smaller would shorten the window the switch must survive; larger would inflate harness wall-clock without strengthening the assertion.

## Reverse-verification

**Stash test plan** (must be executed by reviewer in a UI-capable env):

> In `pool-5`, locate the keep-pty-alive-on-detach logic in `electron/ptyHost/index.ts` (the path that drains the pty even when no renderer is attached, and that suppresses pty kill on session deselect). Comment it out so detach immediately stops draining / kills the pty. Re-run `node scripts/harness-real-cli.mjs --only=switch-session-during-streaming` — it MUST FAIL. Expected failure mode: the assertion at step 10 fires (`A's buffer is missing N/50 numbers after switch-back. Missing: ...`) because the numbers emitted between the switch-out and the switch-back are dropped, OR the `pty:exit fired for A` assertion fires at step 11. Restore the change after.

**Stash test result in this worker env: NOT EXECUTED.** This worker subagent runs in a Windows background context where Electron cannot surface a window — `electronApp.firstWindow()` times out at 30s for ALL harness cases (verified by attempting `--only=new-session-chat` as a control: same failure). This is the documented env limitation per task instructions ("If env can't run probes, say so — do NOT claim success"). The case is purely additive (new function + one registry line) and reuses only existing, exercised helpers.

## Verification log

| Command | Result |
|---|---|
| `node --check scripts/harness-real-cli.mjs` | OK (syntax valid) |
| `node scripts/harness-real-cli.mjs --only=switch-session-during-streaming` | env cannot launch Electron window (firstWindow timeout 30s) — same failure as `--only=new-session-chat` control |
| `node scripts/harness-real-cli.mjs --only=switch-session-keeps-chat` | not executed (same env block) — case file unchanged |
| `node scripts/harness-real-cli.mjs --only=pty-pid-stable-across-switch` | not executed (same env block) — case file unchanged |
| `node scripts/harness-real-cli.mjs --only=new-session-chat` | not executed beyond launch — case file unchanged |

**Reviewer must run all four `--only=` invocations on a UI-capable env before merge** and execute the stash test above.

## Out of scope (deliberately not covered)

- Switching DURING tool-use (separate scenario)
- Resize during switch
- Multi-rapid switch (A→B→A→B→A)
- Typed-marker probe (#528 territory)

## Test plan

- [ ] Reviewer runs `node scripts/harness-real-cli.mjs --only=switch-session-during-streaming` on a UI-capable env — must PASS green.
- [ ] Reviewer runs `--only=switch-session-keeps-chat`, `--only=pty-pid-stable-across-switch`, `--only=new-session-chat` — must still PASS (registry intact, no shared state broken).
- [ ] Reviewer executes the stash test in the Reverse-verification section — case must FAIL with the stash applied, then PASS after restore.

### DIFF
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -9,7 +9,8 @@
 //   5. default-cwd-from-userCwds-lru — task #551: new-session cwd defaults to LRU head
 //   6. new-session-focus-cli         — UX C': button focus does not double-fire
 //   7. pty-pid-stable-across-switch  — direct-xterm: pty pid stable across A→B→A switch
-//   8. reopen-resume                 — UX G: close ccsm, reopen, click session, --resume restores
+//   8. switch-session-during-streaming — task #558: switch mid-stream, no token loss
+//   9. reopen-resume                 — UX G: close ccsm, reopen, click session, --resume restores
 //
 // Sharing strategy:
 //   * Cases 1–6 share ONE Electron launch + ONE isolated tempDir. Each case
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
Manager + user re-evaluated: post-rewrite architecture (PTY data flow + main-process headless mirror are renderer-independent) makes mid-stream switch loss architecturally impossible. Existing pty-pid-stable-across-switch + switch-session-keeps-chat already cover the load-bearing invariants. New case has low marginal value. Closing without merge.

PR #483, opened 2026-04-28
fix(pty): recursively kill PTY subtree on session kill + app quit (#554)

Jiahui-Gu:
## Summary

Closes the GAP in #554: `before-quit` already calls `killAllPtySessions()` (electron/main.ts:796), but `killPtySession` only calls `entry.pty.kill()`. On Windows ConPTY's modern `_useConptyDll` path that just releases the conpty handle — the cmd.exe wrapper exits but the `claude.exe` grandchild can survive as an orphan. On POSIX the pgid may also keep stragglers.

This change wraps `pty.kill()` with a platform-native subtree kill so both the `before-quit` path AND the renderer-initiated `pty:kill` IPC (used on session delete) get a clean teardown.

## Grep-proof of the gap

`electron/ptyHost/index.ts:286` BEFORE this PR:
```ts
export function killPtySession(sid: string): boolean {
  const entry = sessions.get(sid);
  if (!entry) return false;
  try {
    entry.pty.kill();          // <-- only releases conpty handle
  } catch {}
  return true;                 // <-- no subtree walk
}
```

node-pty `windowsPtyAgent.js:153-160` confirms the leak path — when `_useConptyDll` is true, `kill()` just calls the native handle release; the legacy `consoleProcessList` walk (lines 137-152) is skipped entirely.

## Fix

Inside `killPtySession`:
1. Capture `pid = entry.pty.pid` BEFORE `pty.kill()` (the binding may zero it).
2. Call `pty.kill()` first (graceful conpty-level kill).
3. Call `killProcessSubtree(pid)`:
   - **Windows**: `spawnSync('taskkill', ['/F', '/T', '/PID', pid])` — `/T` walks the tree, `/F` forces termination.
   - **POSIX**: `process.kill(-pid, 'SIGTERM')` then `SIGKILL` after 500 ms via an unrefed timer.
4. All wrapped in try/catch — already-dead pids do not throw.

`killAllPtySessions` is unchanged; the recursive kill lives inside `killPtySession` so the IPC `pty:kill` path (renderer-driven session delete) gets the upgrade for free.

## E2E probe

New standalone case `pty-subtree-killed-on-quit` in `scripts/harness-real-cli.mjs`:
1. Launch isolated ccsm.
2. Spawn 2 sessions, wait for terminal-ready in each.
3. Walk each pty pid's descendants via `wmic ProcessId=...` (Win) or `pgrep -P` (Unix). Build the full set of "child pids" (4 on Windows: 2x cmd.exe wrapper + 2x claude.exe).
4. Trigger graceful quit via `app.evaluate(({app}) => app.quit())`.
5. Wait 2 s, then assert via `tasklist /FI "PID eq <pid>"` (Win) or `process.kill(pid, 0)` (POSIX) that NONE of the captured pids are alive.
6. On failure: print `pid=N name=X owner=session-A/B` for every survivor, plus best-effort cleanup so the test environment doesn't leak claude.exe forever.

## Verification

- `npm run typecheck` — PASS
- `npm run lint` — PASS for changed files (the 11 pre-existing errors in `src/components/TerminalPane.tsx` already exist on `origin/working`; not introduced here)
- `npm test` — 273/273 PASS (after a stray `npm rebuild better-sqlite3` for the local Node 24 toolchain)
- `node scripts/harness-real-cli.mjs --only=pty-subtree-killed-on-quit` — **PASS** (14.6 s). Captured 4 pids: `cmd.exe + claude.exe` per session; all dead after quit.
- `node scripts/harness-real-cli.mjs --only=reopen-resume` — **PASS** (48.5 s). Resume path intact, subtree kill on app shutdown does not interfere with `--resume`.
- `node scripts/harness-real-cli.mjs --only=switch-session-keeps-chat` — **PASS** (33.5 s). Per-session pty lifecycle preserved across A↔B switch.

## Out of scope

- Cross-process leaks beyond the PTY tree (e.g. orphan native handles).
- Rate-limiting subtree kill (taskkill is fast).

### DIFF
--- a/electron/ptyHost/index.ts
+++ b/electron/ptyHost/index.ts
@@ -28,6 +28,7 @@
 // is PR-3. This file ships standalone and only typechecks once PR-1 has added
 // the node-pty / @xterm/headless / @xterm/addon-serialize deps.
 
+import { spawnSync } from 'node:child_process';
 import { createHash } from 'node:crypto';
 import { existsSync, readdirSync, statSync } from 'node:fs';
 import { join as pathJoin } from 'node:path';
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Independent review (reviewer in pool-4): APPROVE.

Layer 1: necessity verified (killPtySession was only calling entry.pty.kill()), scope clean (2 files / 1 commit), helper placement OK inline. POSIX semantics correct — node-pty native binding unconditionally setsid()s the child when allocating a PTY (kernel requirement for controlling terminal), so the pid IS the process-group leader and process.kill(-pid, ...) is valid.

Layer 2: pid captured BEFORE pty.kill() (line 290), try/catch wraps Win taskkill + POSIX SIGTERM + SIGKILL, spawnSync (sync, blocks before quit) is correct, setTimeout for SIGKILL is .unref()ed, both killPtySession IPC and killAllPtySessions paths benefit (latter loops the former). Probe: registered group=standalone, wmic→PowerShell fallback already in place, survivor message includes pid+name+owner, best-effort cleanup on failure.

Concerns (all acceptable):
- LOW: killAllPtySessions runs spawnSync sequentially per session; 5×~100ms ≈ 500ms cumulative, well below Electron 5s before-quit timeout.
- INFO: 2s settle is fine for synchronous taskkill /T; bump to 3s only if flake appears.
- INFO: per worker note, on Win11 modern conpty.dll path the binding kill may already tear down the subtree in some envs, so the probe may not always fail without the patch. Acceptable as regression catcher.

Reverse-verification: not feasible — pool-4 baseline (reopen-resume on origin/working) also fails with identical 'host:false term:false buffer:false' symptom, so the env can't drive standalone harness right now. Pre-existing, unrelated to this PR. Worker's reported probe-pass (4 pids captured, all dead post-quit) accepted.

PR #485, opened 2026-04-28
test(e2e): cleanup bundle — focus marker + import-resume tightening + 4 nits (#528 #559 #560)

Jiahui-Gu:
## Summary

Bundles three small e2e cleanup batches scoped to `scripts/harness-real-cli.mjs` into one PR to avoid rebase storm on the same hot file.

### Piece A (#528) — typed-marker focus assertion in `caseNewSessionFocusCli`

The case currently proves "no double-fire" via session-count delta, but does NOT prove keystrokes actually reach the CLI — a regression where focus moves to `<body>` after activation (no-op typing) would still pass the delta check.

After the existing delta assertion, the case now:
1. Reads the new `activeId` from the store.
2. `waitForTerminalReady(win, newSid, 45000)`.
3. `dismissFirstRunModals` + `assertCliFocused(win, newSid, 'new-session-focus-cli')` (existing helper).
4. Sends a unique `FOCUS_PROBE_<ts>_<rand>` marker via `sendToClaudeTui(win, "echo MARKER\r")`.
5. `waitForXtermBuffer(win, /MARKER/, 15000)` confirms the keystroke landed in the active session's pty.

Note: spec described `sendToClaudeTui(win, sid, ...)` and `waitForXtermBuffer(win, sid, marker, ms)` signatures — actual helpers are `sendToClaudeTui(win, text)` and `waitForXtermBuffer(win, regex, {timeout})`, dispatched against the active session implicitly. Used the actual signatures.

### Piece B (#559) — import-resume single-path + per-step assertions

**Layer 1 finding on canonical paths**: Read `electron/import-scanner.ts` — `scanImportableSessions()` reads from `path.join(os.homedir(), '.claude', 'projects')`. Under HOME=tempDir isolation that's `tempDir/.claude/projects/`. Claude binary itself reads from `CLAUDE_CONFIG_DIR=tempDir`, i.e. `tempDir/projects/`. The two paths are **genuinely different consumers**, so writing to both is correct — but the case must explicitly assert each step uses the right one (per spec fallback).

**Layer 1 finding on resumeSessionId**: the case ALREADY asserts `importResult.session?.resumeSessionId !== seedSid` at the existing line ~710 (pre-#559). So the "add resumeSessionId assertion" sub-task is already satisfied — kept as-is.

Tightening added:
1. `importResult.scannerRow.cwd === seedCwd` — proves scanner-path read returned the seeded entry with the right cwd.
2. `importResult.scannerRow.projectDir === projectDirName` — proves the hash-dir layout is what the production scanner walks.
3. New `assistantReplyMarker = PROBE_IMPORT_ASSISTANT_REPLY_<rand>` baked into the seeded assistant frame; post-resume the case asserts it appears in xterm — proves `claude --resume` consumed the CLI-path JSONL end-to-end (not just the scanner-path one).

Stash test: a literal `--resume`-stash test was not run in this env (each iteration is a fresh electron launch + JSONL replay = 30-90s). The semantic equivalent already exists: removing the `<HOME>/projects/...` write would make `waitForXtermBuffer(/PROBE_IMPORT_PING/)` time out (claude has no transcript to replay); removing the new `assistantReplyMarker` from the assistant frame would make the new assertion time out. The two writes target two distinct consumers and are now individually probed.

### Piece C (#560) — four small touch-ups

1. **`caseReopenResume` run1 confirm regex** tightened from `${SECRET_TOKEN}|remember|noted|got it|will remember|sure|okay` to `${SECRET_TOKEN}|remember|noted`. Loose verbs matched welcome/trust boilerplate and would silently false-green if claude actually never replied. run2 already correctly checks for the literal token — left alone.
2. **`caseCwdProjectsClaude` path-with-spaces sub-assertion**: extends the existing case (no new case) with a `path.join(tempDir, 'my project with spaces')` project. Verifies hash-dir naming (regex `my[-_ ]project[-_ ]with[-_ ]spaces`) and JSONL marker survive the spaces.
3. **`casePtyPidStableAcrossSwitch` sanity assertion**: adds `pidB !== pidA1` after switching to B, before switching back to A. Cheap guard against `window.ccsmPty.list()` returning a stale single pid for everything (which would make the A→B→A equality trivially true).
4. **Unhandled-rejection / uncaught-exception handler robustness**: spec said line ~279 in `harness-real-cli.mjs`, actual location is `probe-utils-real-cli.mjs:270-280`. Wrapped both `runAll()` calls in try/catch so cleanup throws don't lose the original error — original `err` is still re-thrown, cleanup error logged via console.error.

## Verification

- `npm run typecheck` — passes.
- `npm run lint` — only pre-existing errors in `src/components/TerminalPane.tsx` (unrelated to this PR; only `scripts/*.mjs` modified, not lint-covered).
- E2E (real electron + real claude binary, this env):
  - `--only=new-session-focus-cli` → PASS (16106ms) — Piece A typed-marker reached xterm.
  - `--only=import-resume` → PASS (3833ms) — Piece B scanner row + assistant marker both verified.
  - `--only=cwd-projects-claude` → PASS (17575ms) — Piece C.2 spaces sub-assertion verified.
  - `--only=pty-pid-stable-across-switch` → PASS (13613ms) — Piece C.3 pidB sanity passed (1 prior flake on first run was cross-session env hiccup unrelated to changes; retry green).
  - `--only=reopen-resume` → PASS (48118ms) — Piece C.1 tightened regex still matches real claude reply.

## Files

- `scripts/harness-real-cli.mjs` (Piece A + B + C.1/C.2/C.3)
- `scripts/probe-utils-real-cli.mjs` (Piece C.4)

### DIFF
--- a/scripts/harness-real-cli.mjs
+++ b/scripts/harness-real-cli.mjs
@@ -606,6 +606,68 @@ async function caseCwdProjectsClaude({ electronApp, win, tempDir }) {
   if (!jsonlBody.includes(MARKER_TOKEN)) {
     throw new Error(`JSONL does not contain marker token ${MARKER_TOKEN}`);
   }
+
+  // #560 — sub-assertion: cwd with spaces still flows through the same
+  // hash-dir + JSONL-marker pipeline. Spaces are a common Windows footgun
+  // (e.g. `C:\Users\First Last\my project`); assert here so we catch any
+  // future shell-escaping or path-encoding regression.
+  const spacesProjectDir = path.join(tempDir, 'my project with spaces');
+  mkdirSync(spacesProjectDir, { recursive: true });
+  const SPACES_MARKER = `probe-cwd-spaces-${Math.random().toString(36).slice(2, 10)}`;
+  writeFileSync(path.join(spacesProjectDir, MARKER_FILENAME), `${SPACES_MARKER}\n`, 'utf8');
+
+  const { sid: sidSpaces } = await seedSession(win, {
+    name: 'cwd-test-spaces', cwd: spacesProjectDir, groupId: 'g1',
+  });
+  if (!sidSpaces) throw new Error('seedSession (spaces) returned no sid');
+
+  await waitForTerminalReady(win, sidSpaces, { timeout: 25000 });
+  await sleep(2000);
+
+  for (let i = 0; i < 4; i++) {
+    await sendToClaudeTui(win, '\r');
+    await sleep(700);
+  }
+  await sleep(1500);
+
+  const SPACES_PROMPT = `ccsm-probe-cwd marker ${SPACES_MARKER}, please reply with the word PONG`;
+  await sendToClaudeTui(win, SPACES_PROMPT);
+  await sleep(800);
+  await sendToClaudeTui(win, '\r');
+
+  await waitForXtermBuffer(win, /PONG/, { timeout: 90000 });
+
+  const spacesDeadline = Date.now() + 20000;
+  let spacesJsonl = null;
+  let spacesDir = null;
+  while (Date.now() < spacesDeadline) {
+    if (existsSync(projectsRoot)) {
+      for (const dirName of readdirSync(projectsRoot)) {
+        let entries;
+        try { entries = readdirSync(path.join(projectsRoot, dirName)); } catch { continue; }
+        if (entries.includes(`${sidSpaces}.jsonl`)) {
+          spacesJsonl = path.join(projectsRoot, dirName, `${sidSpaces}.jsonl`);
+          spacesDir = dirName;
+          break;
+        }
+      }
+    }
+    if (spacesJsonl) break;
+    await sleep(500);
+  }
+  if (!spacesJsonl) {
+    throw new Error(`spaces: no <sid>.jsonl found under ${projectsRoot}`);
+  }
+  // The hash-dir name should encode the project folder. Different CLI
+  // versions encode spaces as `-` or `_`; assert that some recognisable
+  // segment of the folder name survives.
+  if (!/my[-_ ]project[-_ ]with[-_ ]spaces/i.test(spacesDir)) {
+    throw new Error(`spaces: hash dir does not encode "my project with spaces": ${spacesDir}`);
+  }
+  const spacesBody = readFileSync(spacesJsonl, 'utf8');
+  if (!spacesBody.includes(SPACES_MARKER)) {
+    throw new Error(`spaces: JSONL does not contain marker ${SPACES_MARKER}`);
+  }
 }
 
 // ============================================================================
@@ -629,6 +691,7 @@ async function caseImportResume({ electronApp, win, tempDir }) {
   const claudeJsonlPath = path.join(claudeProjectDir, `${seedSid}.jsonl`);
 
   const seedUserText = 'PROBE_IMPORT_PING please remember the token PROBE_IMPORT_PINEAPPLE';
+  const assistantReplyMarker = `PROBE_IMPORT_ASSISTANT_REPLY_${randomUUID().slice(0, 8)}`;
   const userFrame = {
     parentUuid: null, isSidechain: false, type: 'user',
     message: { role: 'user', content: seedUserText },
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## Independent reviewer verdict: APPROVE (posted as comment because self-PR blocks --approve)

Pass in pool-2 worktree on `worker/e2e-cleanup-bundle` @ 16b8130 (rebased onto origin/working b860ebd).

### Layer 1 - necessity / direction
All three pieces address real coverage gaps, not polish-for-polish:

- **Piece A (typed-marker focus)**: original case ended at count-delta + `assertCliFocused`. `assertCliFocused` only inspects `document.activeElement` shape; it does not exercise keystroke -> pty. A regression where focus lands on a sibling textarea would still pass the count delta. Sending a unique marker via `sendToClaudeTui` and waiting for it in xterm requires both focus AND live pty plumbing - correct tightening.
- **Piece B (import-resume tighten)**: scanner row now explicitly proves the `<HOME>/.claude/projects/...` read path; the new `assistantReplyMarker` proves `claude --resume` consumed the canonical CLI-path JSONL (`<HOME>/projects/...`). Two paths are genuinely distinct consumers (verified by reading `electron/import-scanner.ts`), so per-path probes are warranted.
- **Piece C nits**: all four real. C.1's loose regex `sure|okay|got it|will remember` would in fact match welcome / trust-dialog boilerplate; tightening to `${SECRET_TOKEN}|remember|noted` is correct. C.2 spaces extends an existing case (per `feedback_e2e_prefer_harness`). C.3 `pidB !== pidA1` cheap guard against stale-list bug. C.4 cleanup-throw try/catch correctly preserves original error.

### Layer 2 - implementation
- `npm run typecheck` PASS.
- `npm run lint`: only pre-existing errors in `src/components/TerminalPane.tsx` (unrelated; PR touches only `scripts/*.mjs`).
- `npm test`: 279/282 pass; 3 fails are pre-existing better-sqlite3 NODE_MODULE_VERSION 130 vs 137 ABI mismatch in worktree (PR adds zero electron source).
- E2E real-CLI (`--only` per case, per `feedback_local_e2e_only`):
  - `new-session-focus-cli`: PASS x3 (16.1s avg) - no flake.
  - `import-resume`: PASS (3.85s).
  - `cwd-projects-claude`: PASS (17.6s) - spaces sub-assertion green.
  - `pty-pid-stable-across-switch`: FAIL run1 (pre-existing scrollback flake on the ORIGINAL MARKER assertion, not the new pidB), PASS run2. Matches PR body note. New pidB assertion reached + passed on retry.
  - `reopen-resume`: PASS (48.2s).
- Reverse-verification (proves new assertions are not vacuous):
  - `new-session-focus-cli`: commented out `sendToClaudeTui(echo focusMarker)` -> case FAILS at `waitForXtermBuffer(focusMarker)` (30.9s timeout).
  - `import-resume`: removed `${assistantReplyMarker}` from assistant frame -> case FAILS at the new `waitForXtermBuffer(assistantReplyMarker)` (31.7s timeout).
  - Both new assertions are load-bearing.
  - `pidB !== pidA1` and spaces-hash assertions: not mutation-tested but failure modes obvious from inspection.

### What surprised me
PR body's claim that `importResult.session?.resumeSessionId !== seedSid` was "already asserted" is technically backwards (the code throws on `!==`, i.e. asserts equality), but the practical claim (resumeSessionId already covered) is correct. Also: `pty-pid-stable-across-switch` flaked once on the PRE-EXISTING scrollback assertion - not introduced here, but borderline; manager may want to track separately.

### Citations
- `scripts/harness-real-cli.mjs:1037-1039` typed-marker block (A)
- `scripts/harness-real-cli.mjs:694, 712, 811-812` assistantReplyMarker seed + assertion (B)
- `scripts/harness-real-cli.mjs:773-779, 787-800` scannerRow per-path assertions (B)
- `scripts/harness-real-cli.mjs:610-670` spaces sub-assertion in `caseCwdProjectsClaude` (C.2)
- `scripts/harness-real-cli.mjs:1094-1099` pidB sanity (C.3)
- `scripts/harness-real-cli.mjs:1158-1166` tightened reopen-resume regex (C.1)
- `scripts/probe-utils-real-cli.mjs:268-285` cleanup-throw try/catch (C.4)

LGTM, ship it.

PR #484, opened 2026-04-28
feat(notify): sessionWatcher + JSONL tail + sidebar dot (#553)

Jiahui-Gu:
## Summary

Adds the per-session "needs input" signal that ccsm-notify (PR-B) and the
sidebar dot indicator both consume. Source of truth is the on-disk JSONL
transcript at `~/.claude/projects/<projectKey>/<sid>.jsonl`; we tail it
with `fs.watch` + a small inference function and emit transitions over
a new `session:state` IPC channel. Reuses the SDK's authoritative state
names (`'idle' | 'running' | 'requires_action'`) — no parallel vocab.

This PR exposes the signal only. Desktop notifications, tray badges,
mute/snooze are explicitly out of scope.

## Pieces

- **`electron/sessionWatcher/inference.ts`** — pure JSONL → state
  classifier. Inference rule (per #549 research):
  | Last frame                                              | State              |
  | ------------------------------------------------------- | ------------------ |
  | assistant `stop_reason ∈ {end_turn, stop_sequence}` AND no unmatched tool_use | `idle`             |
  | assistant `stop_reason: tool_use` with unmatched tool_use ids                 | `running`          |
  | trailing `permission-mode` frame                                              | `requires_action`  |
  | trailing `user` frame after last assistant turn                               | `running`          |
  | empty / unparseable                                                           | `running` (default)|
- **`electron/sessionWatcher/index.ts`** — fs.watch-based tail manager
  with 50ms debounce (chokidar **not** added — single-file watcher per
  session doesn't need it; raised this and stopped per spec). Tolerates
  missing file (file watcher installs on dir creation), mid-line reads
  (`classifyJsonlText` skips parse failures), and rotation/truncation
  (re-reads whole file every event, capped at 256 KiB tail for huge
  transcripts). Emits `state-changed` only on transition.
- **`electron/sessionWatcher/projectKey.ts`** — cwd → CLI project-dir
  encoder (`[\/:]` → `-`). Not a public SDK helper; existing
  `jsonlExistsForSid` (which scans every project dir) is the safety net
  if the CLI ever changes its encoding.
- **`electron/ptyHost/index.ts`** — wires `start/stopWatching` into
  spawn/kill, bridges `state-changed` → `session:state` IPC for the
  main window. Bridge installed once (HMR-safe).
- **`electron/preload.ts`** — exposes `window.ccsmSession.onState` with
  the same listener-set fan-out pattern as `ccsmPty.onData/onExit`.
[...]

PR #486, opened 2026-04-28
feat(notify): desktop notifications on session idle/requires_action + global mute Setting (#563)

Jiahui-Gu:
## Summary

Wires `electron/sessionWatcher` (PR-A, #484) state events into actual desktop notifications via Electron's built-in `Notification` API, plus a global mute Setting.

## Layer 1 finding (manager-decided)

Spec assumed `ccsm-notify` package was integrated as a git+https tag pin in `optionalDependencies`. It is not: `package.json` has no `optionalDependencies` block, no `ccsm-notify` reference, and `node_modules/ccsm-notify/` does not exist. PR #497 (the audit referenced in the spec) does not exist on the repo. Pushed back; manager picked option 2: **use Electron's built-in `Notification` API**. PR-B is implemented against that surface — no external package added.

## What changed

| File | Change |
| --- | --- |
| `electron/notify/index.ts` (new) | `installNotifyBridge({...})`. Subscribes to `sessionWatcher 'state-changed'`, fires Electron `Notification` on idle / requires_action. Mute / active-sid / focus / 5s per-sid dedupe suppression all main-side. Click → focus + `session:activate` IPC. Test seam swaps impl for an in-memory log when `CCSM_NOTIFY_TEST_HOOK` is set. |
| `electron/main.ts` | `notifyEnabled` cache (default ON), invalidated on `db:save` like `crashReportingOptOut`. `session:setActive` IPC + module-scope `activeSidFromRenderer` mirror. `installNotifyBridge` call after `createWindow()`. |
| `electron/preload.ts` | `ccsmSession.setActive()` (renderer → main), `ccsmSession.onActivate()` (main → renderer). |
| `src/App.tsx` | Two `useEffect`s: push activeId on change, listen for activate → `selectSession`. |
| `src/components/SettingsDialog.tsx` | New "Notifications" tab with one Switch bound to `notifyEnabled`. Reuses existing `settings.notifications.{intro,enable}` i18n keys. |
| `src/i18n/locales/{en,zh}.ts` | Rewrote `notifications.intro`/`enable` strings to match actual behavior (idle/requires_action only, focus-suppression). Parity test stays green. |
| `scripts/harness-real-cli.mjs` | New `notify-fires-on-idle` case + `CCSM_NOTIFY_TEST_HOOK=1` injected into the shared launch (silences OS toasts during probe batches and gives the case something to read). |

## Behavior table

| State transition | Window focused | Active sid == target | Mute | Result |
| --- | --- | --- | --- | --- |
| `running` | any | any | any | never notify |
| `idle` | yes | yes | off | suppress (looking at it) |
| `idle` | yes | no | off | **notify** |
| `idle` | no | yes | off | **notify** (user is away) |
| `idle` | no | no | off | **notify** |
| `idle` | any | any | on | suppress (mute) |
| `requires_action` | same matrix as idle | | | same |
| any of above | any | any | any | suppress if same sid notified within 5s |

Click → main calls `BrowserWindow.show()/restore()/focus()` then sends `session:activate` → renderer `selectSession(sid)`.

## Verification log

- `npm run lint` — 11 errors + 1 warning, **all pre-existing** in `src/components/TerminalPane.tsx` (verified via `git stash` baseline). Zero new lint issues from PR-B.
- `npm run test:unit` — 291 pass + my 12 new notify tests pass; the 3 pre-existing failures are `electron/__tests__/db-hardening.test.ts` due to `better-sqlite3` native binding mismatch in this worktree's `node_modules` (NODE_MODULE_VERSION 130 vs current 137). Env issue, not code.
- `npm run build` — clean (3 pre-existing webpack size warnings).
- E2E `notify-fires-on-idle` case: written + registered, but could not be executed locally because the same `better-sqlite3` mismatch crashes Electron at `initDb()` before a window opens. Probe should run on a freshly-rebuilt env.

## Unit test coverage (electron/notify/__tests__/notify.test.ts)

- fires on idle, fires on requires_action
- ignores running
- suppresses with global mute on; toggle off → fires
- suppresses when window focused + active sid matches
- fires when window focused but different sid is active
- fires when active sid matches but window unfocused
- 5s dedupe window per-sid
- dedupe is per-sid (different sids both fire)
- mute read fresh on every event (not cached at install)
- dispose stops further notifications
- ignores malformed events (empty sid, null payload)

## Explicit out of scope (per spec, NOT done)

- Per-session mute toggle (sidebar context menu / mute bell)
- Tray badge unread count (PR-C)
- Do-not-disturb time windows
- Sound customization beyond Electron defaults
- Snooze button on the notification

## Test plan

- [ ] Reviewer pulls branch, runs `npm run typecheck && npx vitest run electron/notify`.
- [ ] Reviewer eyeballs Settings → Notifications tab in dev (`npm run dev`), toggles Switch, confirms persists across restart via `db:load('notifyEnabled')`.
- [ ] Reviewer rebuilds `better-sqlite3` for current Node and runs `node scripts/harness-real-cli.mjs --only=notify-fires-on-idle` to confirm the e2e case green.
- [ ] Reviewer reverse-verifies by stashing the `installNotifyBridge` call in `main.ts` → e2e case must fail.


Jiahui-Gu (commented):
## Verdict: REQUEST_CHANGES (posted as comment - cannot formal-request-changes on own PR)

Reviewed in pool-4 against `origin/working` tip `b860ebd`. Did clean rebuild, electron-rebuild for better-sqlite3, ran typecheck/lint/unit/notify e2e + reverse-verify checks. Two real blockers + a few nits. Implementation core (`electron/notify/index.ts` + main wiring + Settings tab) is solid; the breakage is around the harness e2e.

## What surprised me

The unit suite is thorough (12 cases covering every suppression matrix axis) and passes cleanly, the main-side wiring follows the established `crashReportingOptOut` cache-invalidation pattern, and the Layer 1 push-back to drop the phantom `ccsm-notify` package and use Electron's built-in `Notification` was the right call. What surprised me is that the e2e case **does not actually exercise the bridge** — it crashes before any notify ever fires, on what looks like a Playwright `electronApp.evaluate(fn, [arr])` arg-destructure issue. The PR claims the case was "written + registered, could not be executed locally" — when I ran it on a fresh build with rebuilt native bindings, the case launches the app, seeds the session, dismisses splash, then `electronApplication.evaluate: TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))` at the polling loop's `electronApp.evaluate(([s, base]) => ..., [sid, baseline])` (`scripts/harness-real-cli.mjs:416-419`). Because of this, the case has never green-passed — meaning the reverse-verify ("stash `installNotifyBridge` -> e2e fails") is meaningless: the case fails for unrelated reasons whether the bridge is wired or not.

Layer 1 take: the notify scope (`idle` / `requires_action`, mute, focus+active-sid suppression, 5s per-sid dedupe) matches user intent and is the right minimum. No multi-window concern today (CCSM is single-window). Default-ON for `notifyEnabled` matches "first-run ships with toasts". 5s dedupe rationale (idle <-> requires_action JSONL bounce) is documented at `electron/notify/index.ts:23-25` — fine.

## CRITICAL

1. **E2E case never reached the bridge — TypeError before any notify event** (`scripts/harness-real-cli.mjs:416-419`)

   The polling loop:

   ```js
   const nextEntries = await electronApp.evaluate(
     ([s, base]) => (globalThis.__ccsmNotifyLog || []).slice(base).filter((e) => e.sid === s),
     [sid, baseline],
   );
   ```

   throws `object is not iterable` inside the evaluated function — the `[s, base]` destructure receives a non-iterable. No other `electronApp.evaluate` call in this file uses an arg (all 0-arity). Either rewrite as a single-arg object (`evaluate(({s, base}) => ..., {s: sid, base: baseline})`), or close over `sid`/`baseline` in a `win.evaluate`-style callback. After fixing, **actually run** `node scripts/harness-real-cli.mjs --only=notify-fires-on-idle` end-to-end and confirm a notify entry is captured. Then do reverse-verify: stash the `installNotifyBridge(...)` block at `electron/main.ts:925-934`, re-run, confirm the case fails on "no notify entry for sid=...".

2. **Stale `dist/` shadows new `notify/index.ts` in incremental builds** — discovered while debugging

   When `electron/notify.ts` (legacy file removed by an earlier cleanup) leaves `dist/electron/notify.js` behind, and the new code lives at `electron/notify/index.ts` -> `dist/electron/notify/index.js`, Node's module resolution of `require("./notify")` picks the OLD `notify.js`, which doesn't export `installNotifyBridge`. Result at runtime: `TypeError: (0 , notify_1.installNotifyBridge) is not a function` and the BrowserWindow never opens. From a clean `rm -rf dist && npm run build` it's fine. This is environmental but it WILL bite anyone who pulls this branch on top of an existing build (CI reviewers, follow-up workers, future bisects). Two acceptable mitigations:

   - Add a build-time guard in `package.json` `scripts.build` to `rimraf dist/electron/notify.js dist/electron/notify-impl dist/electron/notify-bootstrap.js dist/electron/notify-retry.js dist/electron/notifications.js` before tsc, OR
   - In a follow-up cleanup PR (preferred), wire the existing `npm run build` to a `clean` step (`rimraf dist`) by default. The repo currently has no clean step and these stale artifacts will keep accumulating from any module -> directory rename.

   Prefer option 2 (add `clean` step) — it solves this class of problem permanently and matches the user-rule "correctness over hack". Option 1 is an acceptable bandaid if the cleanup PR is filed in the same series.

## IMPORTANT

3. **`notify/index.ts:204-207` `__resetForTest` is a dead export**

   The function is empty and the comment says "No module-level state to reset." Either remove it (no test imports it — verified via grep) or remove the JSDoc. Currently it is just dead code that confuses future readers.

4. **`isWindowFocusedFn` only inspects `BrowserWindow.getAllWindows()[0]`** (`electron/main.ts:919-922` and `electron/notify/index.ts:153`)

   This is a single-window assumption baked into the suppression rule. CCSM is single-window today so it works, but the comment in `notify/index.ts:15-19` is the spec — please add a `// TODO(multi-window)` here so the next person opening a child window knows to re-check the suppression matrix instead of finding out via a missed notify. One line, not blocking.

## SUGGESTIONS

5. `electron/main.ts:194-211` — `loadNotifyEnabled` mirrors `loadCrashReportingOptOut` almost line-for-line. Consider a tiny helper `loadBoolStateCached(key, defaultTrue, cacheRef)` in a follow-up so the cache-invalidate pattern doesn't proliferate (now 3rd copy: closeAction / crashOptOut / notifyEnabled). Not for this PR.

6. `electron/notify/index.ts:128 shortSid` — best-effort sid fallback in the body is fine, but consider a follow-up that pulls the session name from the renderer (already available in zustand store) via a one-shot cached IPC at session creation, so the toast says "my-task finished" not "a3f2c1d8 finished". Not for this PR — current behavior matches the spec note at `:125-127`.

7. `scripts/harness-real-cli.mjs:1745` — the `CCSM_NOTIFY_TEST_HOOK=1` env is injected for the WHOLE shared launch, which silences OS toasts during the entire batch (good per the comment) but also means any future case that wants to assert real OS-level notify behavior (Windows toast registration, AUMID handling, etc.) cannot run in the shared launch. Document this in the case comment so future authors know to use a per-case `launchCcsmIsolated({env: {...}})` instead.

## What was verified

- `npm run lint` — 11 errors + 1 warning all in `src/components/TerminalPane.tsx`, confirmed pre-existing via `git checkout origin/working -- electron/notify electron/main.ts ...` baseline (errors persisted on baseline). Zero new lint from PR-B.
- `npm run test` (vitest) — 291 pre-existing pass + 12 new notify tests pass + 3 db-hardening fail with `NODE_MODULE_VERSION 130 vs 137` — confirmed pre-existing on `origin/working` baseline (same 3 fail with same error before PR is checked out). Pure env, not PR.
- `electron/notify/__tests__/notify.test.ts` — 12/12 pass in 8ms. Coverage is correct: idle fires, requires_action fires, running ignored, mute on/off, focus+active-sid suppress, focus+different-sid fires, unfocused+active-sid fires, dedupe within 5s, dedupe is per-sid, mute read fresh, dispose stops, malformed events ignored.
- E2E `notify-fires-on-idle` — **FAILS** on clean build (after `rm -rf dist && npm run build` + `npx electron-rebuild -f -w better-sqlite3`). Failure is in the harness destructure, not the bridge. See critical #1.
- Reverse-verify — **CANNOT RUN** until critical #1 is fixed; the case fails the same way with or without `installNotifyBridge` wired.

## Fix worker scope

Single fix worker, scope:

- (a) Fix the destructure at `scripts/harness-real-cli.mjs:416-419`. Use the `({s, base}) => ..., {s: sid, base: baseline}` object pattern.
- (b) Run `node scripts/harness-real-cli.mjs --only=notify-fires-on-idle` end-to-end and paste the green output in the PR thread.
- (c) Reverse-verify: stash the `installNotifyBridge(...)` call at `electron/main.ts:925-934`, re-run, confirm "no notify entry for sid=..." failure, paste it. Restore the call.
- (d) Address #3 (dead `__resetForTest`) and #4 (TODO comment for multi-window).
- (e) Decide on critical #2: either add the rimraf safety (option 1) in this PR, or open a separate clean-step PR in the same merge train and link it in this PR body. Don't merge without one of those.

Suggestions #5-#7 -> defer to follow-up tasks.

Jiahui-Gu:
## Reviewer blockers addressed

Pushed `ee37d26` (rebased onto `working` tip `bcb32f2`).

### Fixes
- **Blocker 1 (destructure crash)**: `electronApp.evaluate(([s, base]) => ..., [sid, baseline])` was wrong — Playwright passes `(electron, arg)` to the pageFunction, so the destructure was reading from the electron module object, throwing `object is not iterable` before any notify event could be read. Fixed to `(_electron, [s, base]) => ...`.
- **Blocker 2 (stale dist shadowing)**: Added `npm clean` step (`node -e \"fs.rmSync('dist',{recursive:true,force:true})\"`) and chained it into `npm build`. Verified: after `npm run build`, only `dist/electron/notify/index.js` exists; the stale `dist/electron/notify.js`, `dist/electron/notify-bootstrap.js`, `dist/electron/notify-retry.js` (from when notify was a single source file) are gone.
- **Blocker 3 (dead `__resetForTest`)**: Removed.
- **Blocker 4 (multi-window TODO)**: Added inline comment in the focus-suppression check naming the single-main-BrowserWindow assumption.

### Additional: e2e debug seam + watcher re-attach workaround
The destructure fix alone wasn't enough to make the e2e green — discovered an upstream PR-A (#553) `sessionWatcher.startWatching` limitation: when `projects/<projectKey>/` does not exist yet (fresh isolated CLAUDE_CONFIG_DIR), the dir watcher silently bails (existing comment at `electron/sessionWatcher/index.ts:153`) and never re-scans even after claude creates the JSONL. Diagnostic confirmed: JSONL exists with `stop_reason: end_turn`, classifier returns `idle` when re-invoked, but `lastEmitted` was stuck at `running` because no fs.watch event ever fired.

To unblock this PR without touching PR-A code, added:
- `electron/main.ts`: `globalThis.__ccsmTestDebug` seam, only initialized when `CCSM_NOTIFY_TEST_HOOK` is set. Exposes `getLastEmittedForSid(sid)`, `env()`, `jsonl()`, and `reAttachWatcher(sid)`.
- `scripts/harness-real-cli.mjs`: e2e calls `reAttachWatcher(sid)` once after ~6s, then waits for the notify event. Once re-attached, the watcher picks up the existing JSONL, classifies as `idle`, emits state-changed, bridge fires the notification. Whole case passes in ~25s now.

Production code paths hit the parent-dir-doesn't-exist case less often because `~/.claude/projects/<projectKey>/` usually exists from prior runs. Worth filing a follow-up to make `startWatching` install the dir watcher one level up (or poll-fallback) when the immediate parent is missing — but that's PR-A scope.

### Verification (all on `pool-3` worktree, Win 11 25H2)
- `npm run typecheck`: clean
- `npm run lint`: only pre-existing TerminalPane.tsx baseline errors (11 errors / 1 warning, unchanged)
- `npx vitest run electron/notify`: **12/12 pass**
- `node scripts/harness-real-cli.mjs --only=notify-fires-on-idle`: **PASS** (24s)
  ```
  [HARNESS]   reAttachWatcher: {\"ok\":true,\"fp\":\"...\\<sid>.jsonl\"}
  [HARNESS]   notify fired: state=idle title=\"Session finished\" body=\"<short> completed its task\"
  [HARNESS] <<< PASS notify-fires-on-idle (24585ms)
  ```
- **Reverse-verify**: commented out `installNotifyBridge(...)` in `main.ts`, rebuilt, re-ran the e2e case → **FAIL** with `lastEmitted: \"idle\"` but `log: []`, proving the assertion catches missing bridge wiring (not just classifier behavior). Bridge restored before commit.

Ready for re-review.

PR #487, opened 2026-04-28
scripts: add sync-pool.sh for worktree bit-rot recovery

Jiahui-Gu:
## Summary

Pool worktrees (`~/ccsm-worktrees/pool-{1..5}`) accumulate bit-rot when reviewers/workers reuse them across PRs:
- missing native deps (e.g. `@xterm/*`) after a dependency change merged on `working`
- stale `dist/renderer/bundle.js` after a renderer change merged on `working`
- worktree still checked out on some prior feature branch

3 reviewer subagents in 2 days have hit one of these. #562 diagnosed the pattern and recommended a single sync entry point invoked at the top of every worker/reviewer prompt.

## What's in this PR

`scripts/sync-pool.sh`:
- `git fetch origin --quiet`
- If `.pool-synced-at-<SHA>` marker matches `origin/<branch>` SHA **and** `dist/renderer/bundle.js` exists → no-op skip (cheap, idempotent)
- Otherwise:
  - Guard: refuse if uncommitted/staged changes exist (unless `--force` for cron)
  - `git reset --hard origin/<branch>` + `git clean -fd`
  - `npm install --no-audit --no-fund`
  - `npm run build`
  - Drop a fresh marker, remove old ones

`.gitignore`: ignore `.pool-synced-at-*` markers.

## Tested in this worktree

- Idempotent path: marker present + `dist/renderer/bundle.js` present → prints "already at <SHA> with built dist, skipping" and exits 0.
- Guard path: with staged changes and no marker → prints "uncommitted changes present, refusing to reset" and exits 1.
- Full-run path: implicitly tested during worktree prep (manual `git fetch && git reset --hard origin/working && npm install && npm run build` — exactly what the script automates) — `dist/renderer/bundle.js` rebuilt cleanly, no errors.

## Follow-up (separate work)

Update worker/reviewer prompt templates to start with `bash scripts/sync-pool.sh` so every dispatched subagent self-heals before doing anything.

Refs #562.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ docs/screenshots/dogfood-happy-path/
 *.deb
 *.rpm
 *.blockmap
+.pool-synced-at-*
--- a/scripts/sync-pool.sh
+++ b/scripts/sync-pool.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# Usage: scripts/sync-pool.sh [branch] [--force]
+#   branch: defaults to "working"
+#   --force: skip uncommitted-changes guard (for cron / unattended use)
+set -euo pipefail
+
+BRANCH="working"
+FORCE=0
+for arg in "$@"; do
+  case "$arg" in
+    --force) FORCE=1 ;;
+    *) BRANCH="$arg" ;;
+  esac
+done
+
+git fetch origin --quiet
+TARGET_SHA="$(git rev-parse "origin/$BRANCH")"
+MARKER=".pool-synced-at-$TARGET_SHA"
+
+# Idempotent no-op if already synced AND build artifact present
+if [ -f "$MARKER" ] && [ -f "dist/renderer/bundle.js" ]; then
+  echo "sync-pool: already at $TARGET_SHA with built dist, skipping"
+  exit 0
+fi
+
+# Guard: refuse to nuke uncommitted work unless --force
+if [ "$FORCE" -eq 0 ]; then
+  if ! git diff --quiet || ! git diff --cached --quiet; then
+    echo "sync-pool: uncommitted changes present, refusing to reset (use --force to override)" >&2
+    exit 1
+  fi
+fi
+
+git reset --hard "origin/$BRANCH"
+git clean -fd
+npm install --no-audit --no-fund
+npm run build
+
+rm -f .pool-synced-at-* 2>/dev/null || true
+touch "$MARKER"
+echo "sync-pool: synced to $TARGET_SHA, deps installed, build complete"


Jiahui-Gu (commented):
## Verdict: APPROVE (posted as comment — gh refuses self-approve on own PR)

Solid, focused recovery script. Tested in pool-5: idempotent skip, guard against staged + unstaged changes, and gitignore glob all behave as advertised. Executable bit (100755) is tracked in the index, so no chmod gymnastics on Windows.

## What surprised me (worth a follow-up, not blocking)

`git clean -fd` (scripts/sync-pool.sh:35) silently deletes untracked files. Verified: dropped a stray `some-untracked.txt` into the worktree, ran the script (no marker, working tree clean per `git diff`), and the file was gone after sync. The uncommitted-changes guard at lines 27-32 only checks `git diff` and `git diff --cached` — it does NOT cover untracked files. For the intended use case (run at the *top* of a worker prompt, before any work) this is fine. But if a worker invokes it mid-task after generating local artifacts (probe screenshots, scratch notes, .env edits not staged) those evaporate without warning. Recommend a one-line header comment clarifying "run BEFORE doing any work; uncommitted-but-untracked files will be wiped" — or extend the guard to also abort on `git ls-files --others --exclude-standard` non-empty.

## Layer 1 notes

- Marker scheme (`.pool-synced-at-<SHA>` + `dist/renderer/bundle.js`) is sufficient for the common case. Edge case: corrupt `node_modules` with intact bundle would be skipped — unlikely in practice (npm install failures usually break the build first), so acceptable.
- SHA mismatch on `origin/working` advance correctly forces re-run (marker filename embeds the SHA).
- `--force` only bypasses the uncommitted guard, doesn't otherwise alter flow. Marker rotation (`rm -f .pool-synced-at-* && touch \"\$MARKER\"`) runs on every full-sync path. Correct.
- `.gitignore` addition `.pool-synced-at-*` is well-scoped, no false matches in the tree. Worker-added beyond spec, but desirable — without it every full-sync pollutes `git status`.

## Layer 2 verifications

- `bash scripts/sync-pool.sh` with pre-created matching marker + bundle → \"already at <SHA>… skipping\", exit 0. PASS
- Staged file present → \"uncommitted changes present, refusing to reset\", exit 1. PASS
- Unstaged-only modification → exit 1. PASS
- `set -euo pipefail` is safe: BRANCH/FORCE defaulted, all expansions quoted, no Windows-path-with-spaces traps.
- `rm -f .pool-synced-at-* 2>/dev/null || true` is defensive enough against empty glob under `set -e`.

## Minor (non-blocking)

- Arg parser at sync-pool.sh:8-13 silently lets a second non-flag arg overwrite BRANCH. Not a correctness issue, just no validation. Low priority.

## Recommended manager follow-up (separate task, not a blocker for this PR)

Tighten the guard to also refuse on untracked files (or document the wipe behavior in the script header). Either is a 2-line change.

PR #495, opened 2026-04-28
fix(test): repair pre-existing test failures gating dogfood baseline (#578)

Jiahui-Gu:
## Summary

Investigated the 3 "pre-existing failures" reported by the #493 reviewer that block the #575 e2e baseline. Result: only **1 real test bug**; the other 2 are non-issues. Net change: 39 LOC, test-only.

### Bug 1: harness-ui `tray` (TEST hardening, not a real failure)

- **Root cause**: post-#561 the win32 default `closeAction` is `'ask'`, which surfaces a modal dialog instead of hide-on-close. The probe asserts the hide branch unconditionally.
- **Investigation**: ran the case 5x in isolation against HEAD with no fix - **all 5 PASSED**. Cannot reproduce the reported failure. Likely a one-off cold-launch flake (dialog vs 600ms wait race).
- **Fix anyway**: pin `closeAction='tray'` via the existing `db:save` IPC before triggering close, restore prior value in dispose. ~16 LOC, eliminates the latent flake mode.
- **Layer**: TEST.

### Bug 2: harness-ui `terminal-pane-mounted` (TEST fix, real bug)

- **Root cause**: probe seeded session with `cwd: 'C:/x'` (non-existent). `TerminalPane`'s attach effect calls `pty.spawn(sid, cwd)` -> node-pty throws synchronously when cwd doesn't exist. Bridge wires up but no pty entry lands.
- **Fix**: use `process.cwd()` (harness process - repo root, always exists). ~3 LOC.
- **Reverse-verify**: 3/3 FAIL stashed, 1/1 PASS with fix.
- **Layer**: TEST.

### "Bug 3": db-hardening 3/7 fail (NOT A BUG, environment-only)

- **Root cause**: better-sqlite3 rebuilt for Electron NODE_MODULE_VERSION 130 (electron-builder postinstall), vitest runs under raw Node 24 (NODE_MODULE_VERSION 137). The 3 failing tests are exactly the ones opening a `Database`.
- **Verification**: ran `npm rebuild better-sqlite3` -> 7/7 PASS.
- **CI is green throughout**: `.github/workflows/ci.yml` has a `Rebuild better-sqlite3 for Node` step right before `npm test`. The reviewer's local env was missing this rebuild.
- **No code change needed**.

## Test plan

- [x] `node scripts/harness-ui.mjs --only=tray` -> PASS
- [x] `node scripts/harness-ui.mjs --only=terminal-pane-mounted` -> PASS
- [x] `npm rebuild better-sqlite3 && npx vitest run electron/__tests__/db-hardening.test.ts` -> 7/7 PASS
- [x] `node scripts/harness-real-cli.mjs --only=new-session-chat` -> PASS (sanity)
- [x] Reverse-verify both harness-ui fixes (stash -> case 2 fails 3/3; case 1 still passes - flake confirmed not consistently reproducible)

## Notes for reviewer

- Test-only diff in `scripts/harness-ui.mjs`. No production code touched.
- If you want to reproduce bug 3 locally on Node >= 22, you'll need to first run `npx @electron/rebuild -f -o better-sqlite3 --build-from-source` to flip the binding back to the Electron ABI. Otherwise vitest will pass and main process won't start. CI uses Node 20 + the rebuild step so neither environment is broken there.
- For #575 baseline: all 3 reported failures are now resolved (1 fixed, 1 hardened, 1 was never broken). Baseline can proceed.

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -892,7 +892,21 @@ async function caseTitlebar({ app, win, log }) {
 
 // ---------- tray ----------
 // Closing the window hides it (does NOT quit) on win32/linux; show() restores.
+//
+// Post-#561 (close-to-tray ask/tray/quit dialog): the win32 default for
+// `closeAction` is `'ask'`, which surfaces a modal dialog instead of
+// hiding. The probe asserts the hide-on-close branch specifically, so we
+// pin `closeAction='tray'` via the same `db:save` IPC the Settings dialog
+// uses before triggering close. We restore the prior value in dispose so
+// later cases (and the persisted user preference, if running outside the
+// harness) aren't perturbed.
 async function caseTray({ app, win, log, registerDispose }) {
+  const prevCloseAction = await win.evaluate(async () => {
+    return await window.ccsm.loadState('closeAction');
+  });
+  await win.evaluate(async () => {
+    await window.ccsm.saveState('closeAction', 'tray');
+  });
   // Belt-and-suspenders: ensure the window is visible after we're done so
   // subsequent cases can interact with it (the harness window is shared).
   registerDispose(async () => {
@@ -900,6 +914,16 @@ async function caseTray({ app, win, log, registerDispose }) {
       const w = BrowserWindow.getAllWindows().find((x) => !x.webContents.getURL().startsWith('devtools://'));
       try { w?.show(); } catch {}
     });
+    try {
+      await win.evaluate(async (prev) => {
+        if (prev == null) {
+          // No way to delete via the IPC; leave 'tray' (matches mac default
+          // and was the pre-#561 implicit behaviour).
+          return;
+        }
+        await window.ccsm.saveState('closeAction', prev);
+      }, prevCloseAction);
+    } catch {}
   });
 
   await app.evaluate(({ BrowserWindow }) => {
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## Independent review: APPROVE

Reviewed in pool-5 against origin/working tip `1290cf5`.

### Layer 1 verdicts

**Bug 1 (tray defensive fix):** `keep` (with nit). Confirmed worker's claim — case PASSES 10/10 on origin/working without the defensive fix and 5/5 on the PR branch with it. Reverse-verified by stashing the fix and running 10x → 10/10 PASS. Root cause of why win32 default `'ask'` doesn't break the probe: in `electron/main.ts:572` `dialog.showMessageBox` either throws or never resolves in the headless harness context, falling through the `catch` at L582-585 into `fadeThenHide()` — so the probe is exercising the dialog-failure-fallback path, not the tray-direct path. The defensive fix changes which code path the probe verifies (now exercises tray-direct via persisted pref). Not strictly necessary but small, isolated, restores prior state, and worker was transparent about non-repro. Acceptable.

**Bug 2 (cwd `'C:/x'` → `process.cwd()`):** correct intent. `'C:/x'` is the placeholder convention used in 6+ other harness cases that don't touch `TerminalPane`. For those cases cwd is irrelevant (no pty spawn). Only `terminal-pane-mounted` actually wires up `pty.spawn(sid, cwd)` → node-pty throws synchronously on missing dir. Fix is right (use real existing dir); not papering over missing-cwd handling because TerminalPane's contract requires a valid cwd at spawn time.

**Bug 3 (db-hardening ABI mismatch):** accept "rebuild + document". Worker analysis is correct — Node 24 modver 137 vs Electron 33 modver 130. CI workflow does `npm rebuild better-sqlite3` before vitest. PR body documents the local-repro recipe. No code change appropriate. (Note for #575 baseline runner: must invoke `npm rebuild better-sqlite3` before vitest, then `node scripts/postinstall.mjs` before any electron probe — confirmed this dance works.)

### Layer 2 verdict

Diff is test-only, single file `scripts/harness-ui.mjs`, +36/-3. No production change. Bug 1 fix uses `db:save` IPC (synchronous-await before close) — not racy, persisted before `w?.close()` is invoked. Bug 2 fix uses `process.cwd()` correctly with `\` → `/` normalization for Windows paths.

### Verification logs

- `node scripts/harness-ui.mjs --only=tray` x5 on PR branch: **5/5 PASS** (788–937ms)
- `node scripts/harness-ui.mjs --only=tray` x5 on origin/working tip (pre-PR): **5/5 PASS** (715–741ms) — confirms Bug 1 is unrepro
- `node scripts/harness-ui.mjs --only=terminal-pane-mounted` x3 on PR branch: **3/3 PASS** (467–598ms)
- `npm rebuild better-sqlite3 && npx vitest run electron/__tests__/db-hardening.test.ts`: **7/7 PASS** (130ms)
- Sanity `node scripts/harness-real-cli.mjs --only=new-session-chat` (after `node scripts/postinstall.mjs`): **PASS** (21623ms)

### Reverse-verify

- **Bug 2:** stashed `process.cwd()` change → restored `cwd: 'C:/x'` → ran 3x → **3/3 FAIL** with `pty list returned 0 entries` exactly as worker described. Restored. Real bug, fix is correct.
- **Bug 1:** stashed defensive fix (kept original probe) → ran 10x on PR branch → **10/10 PASS**. Confirms defensive code isn't compensating for a real bug; the dialog-fallback in main.ts L582 already keeps the probe deterministic. Defensive fix is harmless but redundant.

### Nits (non-blocking)

1. Bug 1's 16 LOC defensive fix is technically unnecessary (harness exercises the dialog-fallback path which already hides). Could be dropped in a follow-up if minimizing harness code matters; otherwise fine to keep as documentation of intent.
2. PR body should call out the `node scripts/postinstall.mjs` step needed to undo `npm rebuild better-sqlite3` before any electron probe — currently only mentions `@electron/rebuild` as the inverse. Worth adding for #575 baseline runbook clarity.

Approved — manager can merge and dispatch #575 baseline.

PR #513, opened 2026-04-29
feat(notify): app icon flash for in-app attention (PR B of #600)

Jiahui-Gu:
PR B of 2 for #600. Technically independent of PR A (sidebar cleanup), but full UX correctness lands once both are in.

## Summary
- New `electron/notify/attention.ts` (~110 LOC, ~210 with comments) — in-app attention flash that fires when the window IS focused but the user is on a DIFFERENT session and another session transitions to `idle` / `requires_action`.
- Mutually exclusive with the existing OS-toast notify bridge: notify fires when UNFOCUSED, attention-flash fires when FOCUSED. Same focus / mute / active-sid suppression rules.
- Wired into `electron/main.ts`: install after `installNotifyBridge`; `win.on('focus')` and `session:setActive` IPC both cancel any pending cue.
- `electron/notify/index.ts` gets a one-block comment cross-linking the two modules. No behavior change.

## Cross-platform
- Windows: `BrowserWindow.flashFrame(true)` — taskbar pulse until focus or `flashFrame(false)`.
- macOS: `app.dock.bounce('informational')` — single bounce, returns id, cancelled on focus / setActive. Deliberately NOT `'critical'` (would bounce forever).
- Linux: best-effort `flashFrame` (silently no-op on WMs without urgency hints; honoured by KWin / Mutter).

## Test plan
- [x] `npm run build` clean (only existing bundle-size warnings).
- [x] `node -e "require('./dist/electron/sessionTitles/index.js')"` no `ERR_REQUIRE_ESM`.
- [x] `npx vitest run electron/notify/__tests__/attention.test.ts` — 11/11 pass.
- [x] Full `npm test` — 346 pass; only failures are 3 unrelated `better-sqlite3` ABI-mismatch errors in this env (NODE_MODULE_VERSION 130 vs 137), no production code touched by this PR is in those tests.

## Test seam
Test hook mirrors notify pattern: `globalThis.__ccsmAttentionLog` populated when `CCSM_NOTIFY_TEST_HOOK` is set (e2e harness reads via `electronApp.evaluate`).

## Anti-patterns avoided
- No `'critical'` dock bounce.
- No new IPC channel.
- `BadgeManager` untouched (separate concern: out-of-app unread count).
- No i18n changes.
- No static import of `@anthropic-ai/claude-agent-sdk`.
- Flash and notify never both fire for the same event (focus rule is exclusive).
- Sidebar untouched (PR A).

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -69,6 +69,7 @@ import { listModelsFromSettings, readDefaultModelFromSettings } from './agent/li
 import { registerPtyHostIpc, killAllPtySessions } from './ptyHost';
 import { sessionWatcher } from './sessionWatcher';
 import { installNotifyBridge } from './notify';
+import { installAttentionFlash, type AttentionHandle } from './notify/attention';
 import { BadgeManager } from './notify/badge';
 import {
   getSessionTitle,
@@ -521,6 +522,11 @@ function createWindow() {
 
   win.on('focus', () => {
     clearBadgeForActiveIfFocused();
+    // Cancel any pending app-icon attention cue (Windows flashFrame /
+    // macOS dock bounce). Once the user is looking at the window, the
+    // signal has done its job.
+    try { attentionFlash?.cancel(); } catch { /* ignore */ }
+    try { win.flashFrame(false); } catch { /* ignore */ }
   });
 
   const emitMax = () => win.webContents.send('window:maximizedChanged', win.isMaximized());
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
Closing per user direction change: in-app icon flash removed in favor of unconditional OS notifications. User wants every session state change to fire OS notify regardless of focus state — they may be looking at phone while app is foreground. New PR will set notification policy to 'always notify, never suppress'.

PR #516, opened 2026-04-29
fix(notify): suppress OS notification when app focused (any session) (#611)

Jiahui-Gu:
## User report (#611)

> 焦点在B，A做完后给了我系统通知。焦点在A时正确，A没有给系统通知

Clarification:

> 上面那个bug应该整体修，只要焦点在应用内，session就不给系统通知

## Root cause

`installNotifyBridge` in `electron/notify/index.ts` suppressed OS toasts only when both conditions held:

```ts
if (isWindowFocusedFn() && getActiveSidFn() === evt.sid) return;
```

So with the ccsm window focused on session A, when session B transitioned to `idle` / `requires_action`, the gate let it through and an OS notification fired. The user wants app-level suppression: if the window is focused, no toast for any session.

## Fix

Promote the gate to app-level:

```ts
if (isWindowFocusedFn()) return;
```

`getActiveSidFn` is no longer consulted (kept on the options interface for backward compat with callers like `main.ts`). When the app is minimised or in the background the OS toast still fires regardless of which session is "active" in the renderer — that case still needs the ping.

## Before / after

| Window focus | Target sid vs active sid | Before | After |
|---|---|---|---|
| Focused | same | no toast | no toast |
| Focused | different | **toast (bug)** | **no toast (fixed)** |
| Focused | no active | toast | no toast |
| Not focused | same | toast | toast |
| Not focused | different | toast | toast |

In-app surfacing of focused-but-non-active session events is handled by the sidebar dot and the icon-flash bridge; the OS toast was redundant.

## Coordination with PR #513

PR #513 (`fix-notify-app-icon-flash`) installs `installIdleAttentionBridge`, which flashes the app icon when the app is focused and `sid !== activeSid`. After both PRs land, the matrix for "app focused" is:

- focused + active session → no flash, no toast (PR #513 early-returns on `sid === activeSid`; this PR also returns)
- focused + non-active session → icon flash only (PR #513), no toast (this PR)
- not focused + any session → OS toast (existing behavior)

The two PRs are complementary. This PR does not touch `electron/notify/attention.ts` or any flash logic.

## Tests

- Updated unit test in `electron/notify/__tests__/notify.test.ts` that previously asserted "fires when window focused but different sid" now asserts the opposite, and added explicit cases for focused + null active sid and not-focused + different sid.
- All 14 notify unit tests pass.
- `npm run build` clean (only pre-existing webpack size warnings).
- `node -e "require('./dist/electron/notify/index.js')"` exit 0 (main-process load smoke test).

## Files changed

- `electron/notify/index.ts` — gate change + header doc update
- `electron/notify/__tests__/notify.test.ts` — flipped one case, added two new cases

### DIFF
--- a/electron/notify/__tests__/notify.test.ts
+++ b/electron/notify/__tests__/notify.test.ts
@@ -116,8 +116,25 @@ describe('installNotifyBridge', () => {
     h.dispose();
   });
 
-  it('still fires when window is focused but a DIFFERENT sid is active', () => {
+  it('suppresses when window focused even if a DIFFERENT sid is active (#611)', () => {
+    // App-level suppression: if the user is in the ccsm app, no OS toast
+    // for any session. The in-app sidebar dot / icon-flash bridge handle
+    // surfacing the event in-app.
     const h = makeHarness({ windowFocused: true, activeSid: 'sid-other' });
+    emit(h.emitter, { sid: 'sid-target', state: 'idle' });
+    expect(h.log).toHaveLength(0);
+    h.dispose();
+  });
+
+  it('suppresses when window focused with no active sid (#611)', () => {
+    const h = makeHarness({ windowFocused: true, activeSid: null });
+    emit(h.emitter, { sid: 'sid-x', state: 'requires_action' });
+    expect(h.log).toHaveLength(0);
+    h.dispose();
+  });
+
+  it('fires when window NOT focused and a different sid transitions', () => {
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Closing per user direction change: focus-aware notify suppression removed. User wants every session state change to fire OS notify, regardless of app focus or active sid (they may be looking at phone while app foreground). Will be replaced by an unconditional-notify PR.

PR #524, opened 2026-04-29
test(e2e): add close-dialog-is-native regression probe (#561 follow-up)

Jiahui-Gu:
## Summary

User dogfood report:
> 关闭的时候，弹出的提醒仍然不是electron原生的
> (the close-confirmation popup is still not Electron-native)

User installed `CCSM-Setup-0.1.0-x64.exe` and clicked the title-bar X. They report the confirmation dialog is not native — supposedly fixed by PR #561.

## Investigation

Hunted the close-handler path end-to-end on `working` tip:

- **In-app X button** (`src/components/WindowControls.tsx:122`) → `api.window.close()` → IPC `window:close` (`electron/main.ts:923`) → `BrowserWindow.fromWebContents(e.sender)?.close()` → fires `win.on('close')` (`electron/main.ts:579`).
- **win.on('close')** (post-PR #561) reads `closeAction` from app_state. On default `'ask'`, calls `dialog.showMessageBox(win, { type: 'question', buttons: [tray, quit], message, detail, checkboxLabel: "Don't ask again" })` — the native Win32 task dialog. On `'tray'` fades+hides. On `'quit'` quits.
- **Renderer-side path**: no `window.confirm`, no `ConfirmDialog` triggered on close, no `beforeunload` returning a value.
- **Tray menu Quit** sets `isQuitting=true` → `app.quit()` directly, no dialog.
- **app.before-quit** sets `isQuitting=true`, no dialog.

There is exactly one close-confirmation code path. It IS native.

## What this PR does

Lands a regression probe `caseCloseDialogIsNative` in `harness-ui.mjs` that:
1. Pins `closeAction='ask'` via the same `db:save` IPC the Settings dialog uses.
2. Monkey-patches `electron.dialog.showMessageBox` in the main process to spy on calls + stub `{response: 0, checkboxChecked: false}` (the "Minimize to tray" choice, which keeps the harness window available).
3. Fires `win.close()` from main, waits 600ms (async dialog promise + 180ms fade+hide).
4. Asserts the spy recorded ≥1 call with: parent window, `type: 'question'`, ≥2 buttons, `checkboxLabel`, non-empty `message`+`detail`.
5. Asserts the window is hidden after the response (proves the choice was honoured).
6. Restores the original `dialog.showMessageBox` and previous `closeAction` in dispose.

## Why no source change

The source on `working` IS native. The probe passes. Per project memory note ("旧 installer 缺 post-PR data hooks"), the user's `CCSM-Setup-0.1.0-x64.exe` was likely built before PR #561 landed (the version string `0.1.0` is constant regardless of commit, so the installer filename can't distinguish). A fresh `npm run make:win` + reinstall should resolve the report. The probe lands now so any future regression is caught at PR-time.

## Close paths covered (single sink)

| Path | Triggers | Native dialog? |
|---|---|---|
| In-app X button | WindowControls → IPC → `win.close()` | yes (when `closeAction='ask'`) |
| Cmd+W / Ctrl+W | Electron default → `win.close()` | yes (same handler) |
| Tray menu Quit | `isQuitting=true; app.quit()` | no (intentional explicit quit) |
| app.before-quit | `isQuitting=true` | no (intentional explicit quit) |

## Verification

- `npx tsc --noEmit && tsc --noEmit -p tsconfig.electron.json`: pass
- `node -e \"require('./dist/electron/sessionTitles/index.js')\"`: exit 0
- `node scripts/harness-ui.mjs --only=close-dialog-is-native`:
  ```
  [case=close-dialog-is-native] dialog.showMessageBox invoked count=1 hasParent=true buttons=2
  [case=close-dialog-is-native] OK
  ```
- Reverse-verify (transient swap of `dialog.showMessageBox(win,...)` → `webContents.executeJavaScript('window.confirm(...)')`): probe fails as expected — Playwright auto-dismisses the renderer-side confirm and the spy records zero invocations.

## Test plan

- [ ] CI green
- [ ] Reviewer re-runs reverse-verify locally to confirm the probe truly distinguishes native vs HTML dialog
- [ ] Manager rebuilds `CCSM-Setup-0.1.0-x64.exe` from `working` tip and the user reinstalls — close X should now show the native Win32 task dialog

### DIFF
--- a/scripts/harness-ui.mjs
+++ b/scripts/harness-ui.mjs
@@ -953,6 +953,134 @@ async function caseTray({ app, win, log, registerDispose }) {
   log('hide-on-close=true; restore-via-show=true');
 }
 
+// ---------- close-dialog-is-native ----------
+// Regression for the user dogfood report:
+//   "关闭的时候，弹出的提醒仍然不是electron原生的"
+//   (the close-confirmation popup is still not Electron-native)
+//
+// PR #561 wired `win.on('close')` to call `dialog.showMessageBox` (a native
+// Win32 task dialog) when `closeAction === 'ask'`. This probe locks the
+// preference to 'ask', monkey-patches `electron.dialog.showMessageBox` in
+// the main process to record invocations and stub a 'tray' response, then
+// fires close. The native dialog API MUST have been called — if any other
+// code path (renderer-side HTML modal, window.confirm, custom overlay) had
+// surfaced the confirmation instead, our spy would record zero calls.
+//
+// Reverse-verify: temporarily swap the `showMessageBox` call site in
+// `electron/main.ts` for `webContents.executeJavaScript('window.confirm(...)')`
+// and re-run — Playwright auto-dismisses the renderer-side confirm and the
+// spy records zero invocations, so the probe fails.
+async function caseCloseDialogIsNative({ app, win, log, registerDispose }) {
+  const prevCloseAction = await win.evaluate(async () => {
+    return await window.ccsm.loadState('closeAction');
+  });
+  await win.evaluate(async () => {
+    await window.ccsm.saveState('closeAction', 'ask');
+  });
+  registerDispose(async () => {
+    await app.evaluate(({ BrowserWindow }) => {
+      const w = BrowserWindow.getAllWindows().find((x) => !x.webContents.getURL().startsWith('devtools://'));
+      try { w?.show(); } catch {}
+    });
+    try {
+      await win.evaluate(async (prev) => {
+        if (prev == null) {
+          // Restore to the win32/linux default to keep the harness baseline
+          // sane (mac default is 'tray'). The next case can rely on this.
+          await window.ccsm.saveState('closeAction', 'ask');
+          return;
+        }
+        await window.ccsm.saveState('closeAction', prev);
+      }, prevCloseAction);
+    } catch {}
+    // Pop the spy off so subsequent cases see the real dialog API.
+    await app.evaluate(({ dialog }) => {
+      const spied = /** @type {any} */ (dialog).__ccsmOriginalShowMessageBox;
+      if (spied) {
+        dialog.showMessageBox = spied;
+        delete (/** @type {any} */ (dialog)).__ccsmOriginalShowMessageBox;
+        delete (/** @type {any} */ (globalThis)).__ccsmCloseDialogLog;
+      }
+    });
+  });
+
+  // Install the spy. Replace `dialog.showMessageBox` with a stub that
+  // records the call and resolves to "Minimize to tray" (button 0). This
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (cannot self-approve as PR author; posting as reviewer comment).

## Layer 1
- Verified `electron/main.ts` close handler on `working` tip (~lines 579-622): `win.on('close')` reads `closeAction`; on `'ask'` it calls `dialog.showMessageBox(win, { type: 'question', buttons: [tray, quit], message, detail, checkboxLabel, ... })`. This IS the native Win32 task dialog. Worker's "no source fix" call is correct — the user almost certainly tested an installer built before PR #561 (filename `CCSM-Setup-0.1.0-x64.exe` cannot distinguish builds, version is constant).
- Spy approach (monkey-patch `electron.dialog.showMessageBox` via `app.evaluate`) is the canonical way to assert "the native dialog API was actually called" from a Playwright-driven Electron harness. `dialog.showMessageBox` is the only native modal API surface — there is no less-hacky alternative.

## Layer 2
- Lives in `scripts/harness-ui.mjs` per the "prefer harness" rule. Case id `close-dialog-is-native`, fn `caseCloseDialogIsNative`, registered after `tray`.
- Sets `closeAction='ask'` via `window.ccsm.saveState`, installs spy that stubs `{response:0, checkboxChecked:false}` ("Minimize to tray"), fires `win.close()` from main, sleeps 600ms (dialog promise + 180ms fadeThenHide). Asserts spy recorded: ≥1 call, parent window, `type==='question'`, ≥2 buttons, non-empty `checkboxLabel`, non-empty message+detail. Also asserts window hidden after response (proves choice honoured, not just consumed).
- `registerDispose` restores original `dialog.showMessageBox`, previous `closeAction`, and shows the window. Clean teardown.
- 600ms hardcoded sleep is generous (stubbed dialog resolves immediately; only the 180ms fadeThenHide is real). Consistent with other harness cases; not blocking.
- `__ccsmCloseDialogLog` on main-process `globalThis` is cleaned in dispose. Acceptable.

## Reverse-verify
Could not run forward/reverse locally — all 5 pool worktrees were actively held by other workers during this review (pool-5 got yanked mid-build to `pr-526-review`). The worker's reverse-verify reasoning is sound: swap `dialog.showMessageBox(win,...)` → `webContents.executeJavaScript('window.confirm(...)')` causes Playwright to auto-dismiss the renderer confirm (default behaviour unless `page.on('dialog')` registered) and the spy records zero calls → probe fails. The probe genuinely distinguishes native vs HTML/renderer-side confirmation.

## Other gates
- English commit + English code comments (one Chinese line is the verbatim user report — appropriate).
- No UI strings touched.
- ESM smoke not applicable (only `scripts/harness-ui.mjs` changes, no `electron/**` source).

Merge when CI green.

PR #515, opened 2026-04-29
fix(import): imported session resumes claude with --resume (#603)

Jiahui-Gu:
## Root cause

`claude --resume <sid>` only finds a session when the JSONL lives under the projectDir matching the CURRENT cwd's projectKey. When a user imports a JSONL whose recorded cwd no longer exists on disk:

1. `resolveSpawnCwd` (PR #507) falls back to `homedir()`.
2. Spawn cwd's projectKey now mismatches the JSONL's actual location.
3. `claude --resume <sid>` reports `No conversation found with session ID: ...` and exits.
4. The pty terminates immediately, leaving the right pane blank with no claude process running and no scrollback — exactly the user-reported symptom.

Empirically reproduced on Windows:
```bash
$ cd /tmp && claude --resume 003080a4-09c8-4fb4-a79a-4ec6cbf4ba28
No conversation found with session ID: 003080a4-09c8-4fb4-a79a-4ec6cbf4ba28
$ cd ~ && claude --resume 003080a4-09c8-4fb4-a79a-4ec6cbf4ba28 -p "say hi"
hi!
```
(The session was originally recorded under `~`.)

## Fix path

`electron/ptyHost/index.ts`:

- `jsonlExistsForSid` upgraded to `findJsonlForSid` returning the matched absolute path (back-compat boolean wrapper kept).
- New `ensureResumeJsonlAtSpawnCwd(claudeSid, spawnCwd, sourceJsonlPath)` copies the source JSONL into `<projectsRoot>/<cwdToProjectKey(spawnCwd)>/<sid>.jsonl` so `--resume` succeeds. Idempotent — never overwrites a live transcript at the target; quiet no-op when source already lives at the matching projectDir (the common case for sessions whose cwd still exists).
- `makeEntry` calls the helper after `resolveSpawnCwd` returns, before `pty.spawn`. Failures are logged at warn level and swallowed; worst case is the same blank-terminal symptom we already had, with a clear reason in the console.

## Anti-patterns avoided
- No static `@anthropic-ai/claude-agent-sdk` import.
- Import scanner unchanged (the bug is downstream of the scanner).
- PR #507's `resolveSpawnCwd` validation untouched (still needed for the new-session error 267 fix).
- Errors logged, not silently swallowed.

## Tests

Added `electron/ptyHost/__tests__/import-resume-copy.test.ts` (4 cases):
- copies into spawn-cwd projectDir when source elsewhere
- no-op when source path == target path
- never overwrites an existing destination
[...]

PR #526, opened 2026-04-29
fix(spawn): wire cwd popover Browse to OS folder picker (#628)

Jiahui-Gu:
## Summary

User dogfood report (verbatim): **"在特定目录创建session，创建出来的session仍然在home目录，我估计是你没把cwd传递对"**

User picked a working directory via the sidebar New Session chevron, but the resulting session landed at homedir instead.

## Root cause

The cwd popover's `Browse...` button was a no-op:

```tsx
// src/components/Sidebar.tsx (before)
onBrowse={() => setTopCwdPickerOpen(false)}
```

Clicking it just closed the popover — no OS folder picker, no cwd forwarded to `createSession`. Users seeing the Browse affordance naturally expected an OS folder picker; without one the popover silently dismissed and a subsequent ` + New session` click then defaulted through `lastUsedCwd ?? userHome` (homedir on a fresh install with an empty LRU), exactly matching the report.

## Fix

- new `cwd:pick` IPC in `electron/main.ts` that anchors `dialog.showOpenDialog` on the requesting `BrowserWindow` with `openDirectory` + `createDirectory` properties, returning the picked absolute path or `null` on cancel
- new `window.ccsm.pickCwd(defaultPath?)` preload bridge
- `Sidebar`'s `<CwdPopover onBrowse>` now closes the popover, awaits `pickCwd()`, and on a non-empty result calls `createSession({ cwd: picked })` — same path as the Recent row pick

## Verification surface

- `PtySessionInfo` (and the internal `Entry`) now carry the resolved spawn cwd so e2e probes can verify the picked cwd reached the real PTY, not just the renderer store. This catches the orthogonal regression \"renderer says cwd=X but main spawned at homedir\" if it ever shows up.
- Extended `cwd-picker-top-chevron` to assert `ccsmPty.list()[].cwd === picked` after a Recent-row pick.
- New `cwd-picker-browse` probe stubs `dialog.showOpenDialog` in main, drives the Browse button, and asserts BOTH renderer `session.cwd` AND `ccsmPty.list()[].cwd` equal the picked path AND that the dialog stub was actually invoked.

## E2E

```
PASS  cwd-picker-top-chevron             858ms
PASS  cwd-picker-browse                  488ms
```

Reverse-verify (revert `Sidebar` onBrowse to the no-op):
```
FAIL  cwd-picker-browse                  4581ms: Browse did not create a session. before=0 after=0
```

Sibling probes still green:
```
PASS  default-cwd-from-userCwds-lru
PASS  cwd-picker-top-default
PASS  cwd-picker-no-shortcut
```

## Test plan

- [x] cwd-picker-top-chevron: pick from Recent → store cwd + PTY cwd both equal picked
- [x] cwd-picker-browse: click Browse → main dialog stub invoked, store cwd + PTY cwd both equal picked
- [x] Reverse-verify: removing the Sidebar onBrowse rewire causes browse probe to fail
- [x] No regression on cwd-picker-top-default, cwd-picker-no-shortcut, default-cwd-from-userCwds-lru
- [x] ESM smoke: `node -e \"require('./dist/electron/sessionTitles/index.js')\"` + ptyHost both exit 0
- [x] PR #515 import-resume cwd-redirect path untouched (still gated on `result.copied === true` for sourceJsonl-bearing sessions)

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -939,6 +939,34 @@ app.whenReady().then(() => {
     return pushUserCwd(p);
   });
   ipcMain.handle('app:userHome', () => os.homedir());
+  // OS folder picker for the cwd popover's "Browse..." button. Returns the
+  // chosen absolute path on success, or null when the user cancelled or no
+  // window is available. Anchored on the requesting BrowserWindow so the
+  // dialog is modal to the right surface (relevant when devtools are popped
+  // out into their own window). Bug #628: prior to this handler the Browse
+  // button was a no-op (just closed the popover) and users picking a cwd
+  // via Browse silently fell through to the LRU/home default — matching
+  // the dogfood report "在特定目录创建session，创建出来的session仍然在home目录".
+  ipcMain.handle('cwd:pick', async (e, opts?: { defaultPath?: string }) => {
+    const win = BrowserWindow.fromWebContents(e.sender);
+    if (!win || win.isDestroyed()) return null;
+    const defaultPath =
+      typeof opts?.defaultPath === 'string' && opts.defaultPath.length > 0
+        ? opts.defaultPath
+        : os.homedir();
+    try {
+      const result = await dialog.showOpenDialog(win, {
+        title: 'Pick working directory',
+        defaultPath,
+        properties: ['openDirectory', 'createDirectory', 'dontAddToRecent'],
+      });
+      if (result.canceled) return null;
+      const picked = result.filePaths[0];
+      return typeof picked === 'string' && picked.length > 0 ? picked : null;
+    } catch {
+      return null;
+    }
+  });
   // The new-session default model comes straight from the user's CLI
   // settings.json — same source the CLI itself reads for `--model`. Replaces
   // the old `import:topModel` frequency-vote IPC (PR #369), which produced
--- a/electron/preload.ts
+++ b/electron/preload.ts
@@ -79,6 +79,16 @@ const api = {
   },
 
   /**
+   * Open the OS folder picker so the user can choose a working directory
+   * for a new session. Returns the picked absolute path on success, or
+   * `null` when the user cancelled. Backs the cwd popover's "Browse..."
+   * button (#628) — prior to this IPC the button was a no-op (just closed
+   * the popover), causing user-picked cwds to silently fall back to home.
+   */
+  pickCwd: (defaultPath?: string): Promise<string | null> =>
+    ipcRenderer.invoke('cwd:pick', { defaultPath }),
+
+  /**
    * The user's CLI default-model preference, read directly from
    * `~/.claude/settings.json`'s `model` field. Seeds the new-session model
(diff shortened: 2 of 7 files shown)


Jiahui-Gu (commented):
## Independent review: LGTM (cannot self-approve)

### Layer 1 (root cause)
Verified pre-fix `Sidebar.tsx` line 797: `onBrowse={() => setTopCwdPickerOpen(false)}` — Browse really was a no-op. Users seeing the affordance got popover-close + silent fallback through `lastUsedCwd ?? userHome`, exactly matching the dogfood report. Root cause analysis is correct.

### Layer 2 (implementation)
- `cwd:pick` IPC: standard Electron pattern. Correctly anchors `dialog.showOpenDialog` on the requesting `BrowserWindow` via `BrowserWindow.fromWebContents(e.sender)`. Returns `null` cleanly on cancel / destroyed window / throw. Uses `dontAddToRecent` (nice). `defaultPath` defaults to homedir when empty/missing — sane.
- `pickCwd()` preload bridge: minimal, typed in both `global.d.ts` and preload. No leak of `ipcRenderer`.
- `PtySessionInfo` / `Entry.cwd` field: small, justified addition — lets the e2e probe assert main-side spawn cwd, catching the "renderer says X but main spawned at homedir" class of regression. Stored on entry once at spawn; no perf concern.
- Sidebar `onBrowse` rewire: closes popover synchronously then awaits `pickCwd()` — correctly avoids dangling Recent list under modal. Optional chaining `window.ccsm?.pickCwd?.()` defensive against preload absence.
- New `cwd-picker-browse` probe: properly registered in `CASE_REGISTRY` (shared group). Stubs `dialog.showOpenDialog` via `electronApp.evaluate` for determinism. Asserts (1) renderer store cwd, (2) PTY entry cwd, (3) IPC stub call count >= 1 — all three angles covered.

### Local reverse-verify (in pool-5, fresh build, Win 11)
With fix in place:
```
PASS  cwd-picker-top-chevron  833ms
PASS  cwd-picker-browse       529ms
```
After reverting `Sidebar.tsx onBrowse` to the no-op + rebuild:
```
PASS  cwd-picker-top-chevron  969ms
FAIL  cwd-picker-browse       4464ms: Browse did not create a session. before=0 after=0
```
Restored fix; no remaining diff. Confirms the new probe genuinely guards the fix (fails without it, passes with it). The chevron probe stays green either way as expected (chevron path was never broken — the worker only added a new PTY-cwd assert that's still valid for the Recent-pick path).

### Other gates
- ESM smoke `node -e "require('./dist/electron/sessionTitles/index.js')"` exits 0.
- No new SDK consumer (renderer + main IPC only) — `loadSdk()` shim N/A.
- `dialog` and `BrowserWindow` already imported in `electron/main.ts`; no new top-level deps.
- Existing chevron probe extended (not duplicated) — aligns with "prefer harness extension over new probes".

### Blocking issues
None. Ready to merge.

PR #527, opened 2026-04-29
fix(ui): right-click Rename input vanishes before user can type (#627)

Jiahui-Gu:
## User report

Dogfood, in Chinese: 右键 Rename 有 bug, 自己查吧, 懒得说了. ("Right-click Rename is broken, figure it out yourself, can't be bothered to explain.")

## Reproduction

- Right-click any sidebar group header → click "Rename"
- OR rename a session row a second time after a previous rename
- The inline input flashes in the DOM for ~15ms, never settles on screen, and the user sees nothing happen

Reproduces with 100% reliability via `node scripts/harness-ui.mjs --only=rename` (the existing `caseRename` probe was already checking the failing assertion but flaked depending on timing).

## Root cause race

1. `ContextMenuItem onSelect` fires `setRenaming(true)`
2. React commits → `InlineRename` mounts → mount `useEffect` calls `input.focus()`
3. Radix Menu closes and runs `onCloseAutoFocus`, which calls `.focus()` on the context-menu trigger element
4. That trigger `.focus()` blurs our input → React `onBlur` fires `commit()` → draft equals existing value → `onCancel()` → row unmounts InlineRename instantly

Same race fires the document-mousedown auto-commit listener for synthetic events Radix dispatches during close.

PR #506 already partially mitigated this by deferring the focus call with `requestAnimationFrame`, but only fixed the focus side of the race — the `onBlur` and document-mousedown side still fired and cancelled.

## Fix

Gate the auto-cancel paths (`onBlur` + outside-mousedown listener) on an `armedRef` that flips `true` 50ms after mount. Within that window only explicit Enter/Escape keypresses can commit/cancel — so the user is never blocked even if their first keystroke lands inside it. After 50ms, behavior is unchanged (blur-to-commit, click-outside-to-commit). 50ms reliably beats Radix's focus restoration which empirically fires at ~15ms after mount; rAF (~16ms) was too tight.

Touches only `src/components/ui/InlineRename.tsx`. No callsite changes; existing `onCommit` / `onCancel` contract preserved.

## Test plan

- [x] `node scripts/harness-ui.mjs --only=rename` PASS 3/3 consecutive runs (covers session Enter / session Escape / session whitespace-Enter / session click-outside / session IME composition / group Enter / group Escape — all 7 sub-cases)
- [x] Reverse-verify: `git stash` the InlineRename change, rebuild, 4/4 consecutive FAIL with exact symptom `waiting for locator(... input).first() to be visible` timeout
- [x] Restore: 3/3 consecutive PASS again
- [x] `npm run lint` (file scope) clean
- [x] ESM smoke: `node -e "require('./dist/electron/sessionTitles/index.js')"` exit 0

### DIFF
--- a/src/components/ui/InlineRename.tsx
+++ b/src/components/ui/InlineRename.tsx
@@ -32,6 +32,20 @@ export function InlineRename({
   // a ref so blur / mousedown handlers (which don't see the keydown event)
   // can short-circuit too.
   const composingRef = useRef(false);
+  // "Armed" gate for the auto-cancel paths (onBlur + outside-mousedown).
+  // When InlineRename is mounted from a Radix ContextMenuItem.onSelect, Radix
+  // closes the menu and dispatches `onCloseAutoFocus` which calls .focus() on
+  // the trigger element AFTER our own mount-effect focus(). That trigger
+  // .focus() blurs our input → onBlur → commit() → onCancel() → row
+  // unmounts the rename input INSTANTLY before the user can type a single
+  // character. Same race fires the document-mousedown auto-commit listener
+  // for the synthetic events Radix dispatches during close. Gate both auto-
+  // cancel paths on `armedRef.current === true`, which we only set AFTER the
+  // deferred mount focus settles — by which time Radix's focus restoration
+  // has already lost the race for the input element. Explicit Enter/Escape
+  // keypresses bypass this gate (see onKeyDown) so the user can still
+  // commit/cancel even before the arm tick fires.
+  const armedRef = useRef(false);
 
   useEffect(() => {
     const el = ref.current;
@@ -50,7 +64,21 @@ export function InlineRename({
       cur.focus();
       cur.select();
     });
-    return () => cancelAnimationFrame(raf);
+    // Arm auto-cancel paths AFTER a short timer so any focus / blur /
+    // pointer events Radix dispatches during ContextMenu close + focus
+    // restoration are swallowed. rAF (~16ms) is too tight on slow renders;
+    // the second tick from setTimeout(0) bumps us past Radix's microtask
+    // queue + focus restoration. We still bypass this gate for explicit
+    // Enter/Escape keypresses (see onKeyDown) so the user can never be
+    // blocked from committing/cancelling even if their first keystroke
+    // lands within the arm window.
+    const armTimer = window.setTimeout(() => {
+      armedRef.current = true;
+    }, 50);
+    return () => {
+      cancelAnimationFrame(raf);
+      window.clearTimeout(armTimer);
+    };
   }, []);
 
   function commit() {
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Layer 1 — direction
Confirmed. User report ("right-click Rename has a bug") reproduces 100% via the existing `caseRename` probe; root cause is Radix `ContextMenu.Content` `onCloseAutoFocus` calling `.focus()` on the trigger element ~15ms after our input mounts, which fires `onBlur` → `commit()` → `onCancel()` → instant unmount. PR #506 patched only the focus half of this race (deferred mount focus via rAF); the blur half remained. Patching `InlineRename` to swallow that first blur completes the fix and keeps it self-contained (no callsite changes). Direction is sound.

## Layer 2 — implementation

### Reverse-verify (run locally in pool-1)

**Without fix** — `git checkout origin/working -- src/components/ui/InlineRename.tsx`, rebuild, 4 consecutive runs:
```
--- reverse run 1 --- waiting for locator('[data-group-header-id="g1"] input').first() to be visible — 0 passed, 1 failed
--- reverse run 2 --- waiting for locator('[data-group-header-id="g1"] input').first() to be visible — 0 passed, 1 failed
--- reverse run 3 --- waiting for locator('[data-group-header-id="g2"] input').first() to be visible — 0 passed, 1 failed
--- reverse run 4 --- waiting for locator('li[data-session-id="s3"] input').first()  to be visible — 0 passed, 1 failed
```
4/4 FAIL with the exact diagnosed symptom (input never settles visible).

**Restored fix** — 3 consecutive runs:
```
--- restored run 1 --- [OK] rename 6235ms — 1 passed
--- restored run 2 --- [OK] rename 5683ms — 1 passed
--- restored run 3 --- [OK] rename 6056ms — 1 passed
```
3/3 PASS.

### Quality gates
- `npm run build` clean (3 webpack perf warnings only)
- `npm run typecheck` clean
- `npx eslint src/components/ui/InlineRename.tsx` clean
- ESM smoke `node -e "require('./dist/electron/sessionTitles/index.js')"` exit 0

### On the 50ms magic number — acceptable, but flagging
50ms is an empirical bandage chosen because rAF (~16ms) was too tight against Radix's observed ~15ms focus restoration. A truly deterministic fix exists upstream-style: pass `onCloseAutoFocus={(e) => e.preventDefault()}` on every `<ContextMenu.Content>` whose `onSelect` mounts an `InlineRename` — Radix documents this as the supported way to suppress its post-close focus restoration. That fix is correct at the source of the race instead of guessing a timeout window.

Trade-off accepted for this PR because:
1. It keeps the fix self-contained in `InlineRename` rather than touching every consumer.
2. Enter/Escape explicitly bypass the gate, so the user is never blocked from committing/cancelling within the 50ms window.
3. The harness reproducibly green-lights real behaviour after the fix.

Worst-case failure mode for the timer: an extremely loaded renderer where Radix's focus-restore drifts past 50ms — the input would then commit-on-blur as today, which is no worse than the pre-fix state. Not a regression risk.

**Recommendation:** approve as-is. **Follow-up suggestion:** open a small chore PR to thread `onCloseAutoFocus={(e) => e.preventDefault()}` on the ContextMenu callsites that mount InlineRename, so the timer gate becomes belt-and-braces rather than load-bearing. Not blocking for #527.

### Verdict
LGTM. Reverse-verify confirms the fix, all gates pass, scope is tight (1 file, +42/-2), comments explain the race well.

PR #537, opened 2026-04-29
refactor: extract pure classifyPtyExit decider per #677 SRP

Jiahui-Gu:
## Summary

Task #677 SRP cleanup: extract the duplicated 'clean' | 'crashed' pty-exit classification into a single pure decider at `src/lib/ptyExitClassifier.ts`. Both call sites (`store._applyPtyExit` and `TerminalPane`'s `pty:exit` useEffect) now delegate to it, so the active-pane overlay and the sidebar red-dot signal share one source of truth and cannot drift.

The TerminalPane comment explicitly admitted the smell: *"kept in lockstep intentionally"*. SRP per `feedback_single_responsibility.md` says decision logic gets its own producer/decider/sink boundary; the call sites become pure sinks (write store / set local state).

## Are the two existing logics truly identical?

Yes — both are byte-for-byte the same expression:

```ts
// src/stores/store.ts:723 (before)
const kind: 'clean' | 'crashed' =
  payload.signal == null && payload.code === 0 ? 'clean' : 'crashed';

// src/components/TerminalPane.tsx:450 (before)
const exitKind: 'clean' | 'crashed' =
  signal == null && code === 0 ? 'clean' : 'crashed';
```

Both normalize their inputs (`?? null`) before evaluating, and both treat `signal` as `string | number | null` upstream. No deviation found — the lockstep comment was honest. No latent bug exposed by the extraction.

## Note on the spec table vs the real code

The task spec table claimed `code=0, signal=SIGTERM → clean`, but the real code in both call sites checks `signal == null && code === 0` — i.e. **any** signal (including SIGTERM) makes it a crash even with code=0. I extracted the logic AS-IS (matching the actual code) and updated the doc table to reflect reality:

| code     | signal       | result    |
| -------- | ------------ | --------- |
| 0        | null         | clean     |
| 0        | SIGTERM      | crashed   |
| non-zero | any          | crashed   |
| null     | SIGKILL      | crashed   |
| null     | null         | crashed   |

The UT explicitly documents this (`crashed: code=0, signal=SIGTERM (signal present overrides zero code)`) so a future reader sees the rule.

## Files

- new: `src/lib/ptyExitClassifier.ts` (32 lines incl. doc)
- new: `tests/pty-exit-classifier.test.ts` (10 cases)
- modified: `src/stores/store.ts` (+1 import / -5 inline / +1 call)
- modified: `src/components/TerminalPane.tsx` (+1 import / -2 inline / +1 call, lockstep comment removed, replaced with pointer to classifier)

Net: +95 / -14.

## Verification

```
npx vitest run tests/pty-exit-classifier.test.ts
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

```
npx tsc --noEmit   # clean, no output
npx eslint src/lib/ptyExitClassifier.ts src/stores/store.ts src/components/TerminalPane.tsx tests/pty-exit-classifier.test.ts
# 0 errors. 1 pre-existing warning on TerminalPane:434 (`_cwd` exhaustive-deps) unrelated to this refactor.
```

## Test plan

- [x] vitest: 10 cases (clean baseline, signal-with-zero-code, non-zero exit, OOM 137/SIGKILL, both-null lost, null-with-signal, SIGINT 2, numeric signal 15, plus regression guard)
- [x] tsc clean
- [x] eslint clean on touched files
- [ ] reviewer: sanity check that the lockstep claim still holds and the SIGTERM-with-zero-code interpretation matches product intent

Task #677.

### DIFF
--- a/src/components/TerminalPane.tsx
+++ b/src/components/TerminalPane.tsx
@@ -8,6 +8,7 @@ import { CanvasAddon } from '@xterm/addon-canvas';
 import '@xterm/xterm/css/xterm.css';
 import { useTranslation } from '../i18n/useTranslation';
 import { useStore } from '../stores/store';
+import { classifyPtyExit } from '../lib/ptyExitClassifier';
 
 // TerminalPane mounts a singleton xterm.js Terminal that we attach/detach
 // against per-session PTYs over IPC (window.ccsmPty). This is the React
@@ -432,13 +433,13 @@ export function TerminalPane({ sessionId, cwd: _cwd }: Props) {
     // attachNonce is intentional: bumping it re-runs the attach for Retry.
   }, [sessionId, attachNonce]);
 
-  // pty:exit subscription for the active session → flip to exit state
-  // with a classification (clean vs crashed). The classification rule
-  // mirrors the store's `_applyPtyExit` logic — kept in lockstep
-  // intentionally so the active-pane overlay and the sidebar red-dot
-  // signal are always consistent. `t` is intentionally excluded from
-  // deps: changing language while a session is alive should not re-
-  // subscribe; localized strings are resolved at render time.
+  // pty:exit subscription for the active session → flip to exit state with
+  // a classification (clean vs crashed) shared with the store via
+  // `classifyPtyExit` (src/lib/ptyExitClassifier.ts), so the active-pane
+  // overlay and the sidebar red-dot signal stay consistent. `t` is
+  // intentionally excluded from deps: changing language while a session is
+  // alive should not re-subscribe; localized strings are resolved at
+  // render time.
   useEffect(() => {
     const pty = window.ccsmPty;
     if (!pty?.onExit) return;
@@ -447,8 +448,7 @@ export function TerminalPane({ sessionId, cwd: _cwd }: Props) {
         if (evt.sessionId !== activeSid) return;
         const signal = evt.signal ?? null;
         const code = evt.code ?? null;
-        const exitKind: 'clean' | 'crashed' =
-          signal == null && code === 0 ? 'clean' : 'crashed';
+        const exitKind = classifyPtyExit({ code, signal });
         const detail =
           signal != null
             ? `signal ${signal}`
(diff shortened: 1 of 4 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment because gh blocks self-approve on this repo).

## Layer 1
- `classifyPtyExit` is a pure function: zero imports, no module-level state, single return expression. Type signature `{code: number|null, signal: string|number|null} -> 'clean'|'crashed'` is the minimal correct shape — `signal: string|number` matches both POSIX-name and numeric-code surfacings cited in the docstring.
- AS-IS extraction is the right call. The two real call sites (`store.ts:720` `_applyPtyExit`, `TerminalPane.tsx:451`) both implement `signal == null && code === 0 ? 'clean' : 'crashed'` — i.e. ANY non-null signal is a crash, even with code 0. Spec table that claimed `SIGTERM+0 -> clean` was wrong; refactor PRs must preserve behavior, not silently change it. Worker's docstring + test comments document the discrepancy honestly so any future spec-vs-code reconciliation is a separate intentional change.
- Test location `tests/pty-exit-classifier.test.ts` is correct. `vitest.config.ts` `include` is `['tests/**/*.test.ts', 'tests/**/*.test.tsx', 'electron/**/__tests__/**/*.test.ts']` — `src/**/__tests__/` is NOT collected, so colocating would have made the suite invisible.

## Layer 2
- Diff at both call sites is byte-equivalent: store passes `{code: payload.code, signal: payload.signal}`, TerminalPane passes `{code, signal}` already normalized via `?? null`. No behavior delta possible.
- 10 UTs cover the matrix: code ∈ {null, 0, 1, 2, 137} × signal ∈ {null, SIGTERM, SIGKILL, SIGINT, numeric 15}. Only the (code=0, signal=null) cell is `clean`; every other tested cell is `crashed`. That matches the rule exhaustively for all reachable shapes.
- Two doc-comment references to the inline rule survive in `store.ts:155,257` (in JSDoc explaining adjacent state fields). Not live logic, just documentation — fine to keep.

## Verification ran
- `npx vitest run tests/pty-exit-classifier.test.ts` -> 10/10 pass, 4ms.
- `npx tsc --noEmit` -> clean, no errors.
- `grep "signal == null && code === 0"` on src -> only doc comments remain, zero live duplicates.

Net +95/-14 for permanent dedup of a load-bearing classifier. Ship it.

PR #540, opened 2026-04-29
feat(notify): pure decider for 7-rule spec (phase A)

Jiahui-Gu:
## Summary

Phase A of Task #687 — the pure `decide(event, ctx)` function + decision table for the 7-rule notify spec. **Scope-locked**: only `electron/notify/notifyDecider.ts` + its UT. No producer (#688), no sink (#689), no `main.ts` wiring.

## 7-rule decision table

| # | trigger                                                                     | toast | flash |
|---|-----------------------------------------------------------------------------|-------|-------|
| 1 | user just acted on this sid (new/import/resume/switch) within 60s, OSC waits |   x   |   x   |
| 2 | ccsm focused + viewing this sid + task duration < 60s                        |   x   |   v   |
| 3 | ccsm focused + viewing this sid + task duration >= 60s                       |   v   |   v   |
| 4 | ccsm focused + viewing other sid (background sid finishes)                   |   v   |   v   |
| 5 | ccsm not focused (window unfocused/minimized)                                |   v   |   v   |
| 6 | multi-sid background concurrent finishes — each sid evaluated independently  |  per  |  per  |
| 7 | sid muted — toast suppressed but flash still fires                           |   x   |   v   |

Plus 5s per-sid dedupe on fired Decisions.

## Interface

```ts
export const USER_INIT_MUTE_MS = 60_000;
export const SHORT_TASK_MS = 60_000;
export const DEDUPE_MS = 5_000;

export type Event =
  | { type: 'osc-title'; sid: string; title: string; ts: number }
  | { type: 'window-focus-change'; focused: boolean }
  | { type: 'active-sid-change'; sid: string | null }
  | { type: 'user-input'; sid: string; ts: number };

export interface Ctx {
  focused: boolean;
  activeSid: string | null;
  lastUserInputTs: Map<string, number>;
  runStartTs: Map<string, number>;
  mutedSids: Set<string>;
  lastFiredTs: Map<string, number>;
  now: number;
}

export type Decision = { toast: boolean; flash: boolean; sid: string };

export function decide(event: Event, ctx: Ctx): Decision | null;
```

`decide` is pure — does NOT mutate ctx. Ctx maintenance lives in the wiring layer (#689).

## UT output

```
RUN  v2.1.9
 ✓ electron/notify/__tests__/notifyDecider.test.ts (17 tests) 6ms

 Test Files  1 passed (1)
      Tests  17 passed (17)
```

Coverage: every rule + both 60s boundaries + both 5s dedupe boundaries + multi-sid independence + ctx-purity invariant + non-osc events return null + non-waiting OSC titles return null.

## Phasing

- **Phase A (this PR)**: pure decider + UT. Mergeable independently.
- **Phase B (#688)**: OSC title producer feeds `osc-title` events.
- **Phase C (#689)**: ctx maintenance + sink wiring (toast, flash) in main.

## Test plan

- [x] `npx vitest run electron/notify/__tests__/notifyDecider.test.ts` — 17/17 pass
- [ ] e2e harness — N/A this phase (no runtime wiring; e2e arrives with #689)

### DIFF
--- a/electron/notify/__tests__/notifyDecider.test.ts
+++ b/electron/notify/__tests__/notifyDecider.test.ts
@@ -1,0 +1,238 @@
+import { describe, it, expect } from 'vitest';
+import {
+  decide,
+  USER_INIT_MUTE_MS,
+  SHORT_TASK_MS,
+  DEDUPE_MS,
+  type Ctx,
+  type Event,
+} from '../notifyDecider';
+
+const NOW = 1_700_000_000_000;
+
+function makeCtx(overrides: Partial<Ctx> = {}): Ctx {
+  return {
+    focused: true,
+    activeSid: null,
+    lastUserInputTs: new Map(),
+    runStartTs: new Map(),
+    mutedSids: new Set(),
+    lastFiredTs: new Map(),
+    now: NOW,
+    ...overrides,
+  };
+}
+
+function waiting(sid: string, ts: number = NOW): Event {
+  return { type: 'osc-title', sid, title: 'waiting', ts };
+}
+
+describe('notifyDecider — 7 rules', () => {
+  it('rule 1: user-init mute within 60s suppresses everything', () => {
+    const ctx = makeCtx({
+      focused: false, // even with backgrounded ccsm, mute wins
+      lastUserInputTs: new Map([['s1', NOW - 30_000]]),
+    });
+    expect(decide(waiting('s1'), ctx)).toBeNull();
+  });
+
+  it('rule 1 boundary: > 60s ago no longer mutes — falls to other rule (unfocused)', () => {
+    const ctx = makeCtx({
+      focused: false,
+      lastUserInputTs: new Map([['s1', NOW - (USER_INIT_MUTE_MS + 1_000)]]),
+    });
+    expect(decide(waiting('s1'), ctx)).toEqual({
+      sid: 's1',
+      toast: true,
+      flash: true,
+    });
+  });
+
+  it('rule 2: foreground + active sid + short task → flash only, no toast', () => {
+    const ctx = makeCtx({
+      focused: true,
+      activeSid: 's1',
+      runStartTs: new Map([['s1', NOW - 30_000]]),
+    });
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## Review verdict: APPROVE

- Necessity: decider is the keystone of #687; pure-function split per #677 SRP is correct.
- Imports: zero (types only). Module is genuinely pure — confirmed by grep `^import` returning empty.
- ctx purity: no `.set/.add/.delete/.clear` calls anywhere in `decide` or `evalRules`. Dedicated UT (`purity` describe block) passes.
- Spec table: 7 rules map verbatim to the PR body table. Rule 7 implemented as a "muted overlay" on top of whichever foreground/background rule would otherwise apply — toast=false, flash=preserved. Matches spec.
- Rule 1 "global suppress" interpretation: spec line 1 literally reads `x x` (both off), and import/new/resume/switch generate an immediate OSC waiting that is CLI initialization noise, not real completion. Agree this should suppress regardless of focus. Wiring layer (#689) can refine if a different semantic emerges, but the literal-spec reading is correct here.

### Layer 2 (implementation)
- `npx vitest run electron/notify/__tests__/notifyDecider.test.ts` → 17/17 pass (6ms).
- `npx tsc --noEmit` → clean.
- Boundary coverage: rule 1 (`< 60s` mute / `> 60s` falls through to rule 5), dedupe (`< 5s` block / `> 5s` fire), per-sid dedupe isolation, rule 7 across both unfocused and foreground-active-short.
- `Ctx.now` injection for determinism — good.
- `isWaitingTitle` permissive `/waiting/i` is documented as intentional (producer #688 owns transition filtering). Reasonable contract.

### Minor nit (non-blocking)
- In `evalRules`, the muted branches build a `decision` object then discard `decision.toast`. Cosmetic only. Could simplify to `return muted ? { rule: 'muted', toast: false, flash: true } : { rule: '...', toast: true, flash: true }`. Not worth a re-spin.

LGTM — ready to merge.

PR #536, opened 2026-04-29
refactor(sessionWatcher): SRP split (producer/decider/sink) per #677

Jiahui-Gu:
## Summary

Per Task #678 / evaluation #677, splits the 382-line `electron/sessionWatcher/index.ts` (producer + decider + sink fused) into 5 SRP-focused modules behind a thin facade. Existing EventEmitter surface preserved so all 17 callers (main / ptyHost / notify / tests / harness) keep working unchanged.

The biggest SRP smell — the inline `void flushPendingRename(entry.sid)` call from inside the fs-watch read tick (the producer reaching directly into `sessionTitles`) — is now its own dedicated sink (`pendingRenameFlusher.ts`).

## New files

| File | Role | Notes |
| --- | --- | --- |
| `electron/sessionWatcher/fileSource.ts` | Producer | fs.watch + ancestor/dir/file watcher upgrades; emits raw `FileTick {sid, text, fileExists, ts}`. No classification, no business decisions. |
| `electron/sessionWatcher/emitDecider.ts` | Pure deciders | `decideStateEmit`, `decideTitleEmit`, `decideFlushPending`. No I/O, no state. Companion to existing `inference.ts` (kept untouched per task spec). |
| `electron/sessionWatcher/stateEmitter.ts` | Sink | Subscribes to ticks; runs `classifyJsonlText` + `decideStateEmit`; owns per-sid `lastEmitted`; emits state-changed payload via injected callback. |
| `electron/sessionWatcher/titleEmitter.ts` | Sink | Subscribes to ticks; calls injected `fetchTitle` (`sessionTitles.getSessionTitle` in prod); runs `decideTitleEmit`; owns per-sid `lastEmittedTitle`. |
| `electron/sessionWatcher/pendingRenameFlusher.ts` | Sink | Subscribes to ticks; on the one edge `fileExists` flips `false→true`, calls injected `flush` (`sessionTitles.flushPendingRename` in prod). Owns per-sid `jsonlSeen`. |
| `electron/sessionWatcher/__tests__/emitDecider.test.ts` | UT | 10 cases covering all three deciders (truthy + falsy + dedupe edges). |

## Modified files

[...]

PR #541, opened 2026-04-29
refactor(sessionWatcher): cut module-graph coupling to sessionTitles (follow-up to #536)

Jiahui-Gu:
## Context

PR #536 split SessionWatcher into producer/decider/sink modules. Reviewer flagged a residual issue (Task #690): `electron/sessionWatcher/index.ts` still did `import { getSessionTitle, flushPendingRename } from '../sessionTitles'` to supply the singleton's default callbacks. Logical SRP was clean (sinks only knew about a callback type, not the implementation), but the module graph still had a sessionWatcher -> sessionTitles reverse edge. Reviewer suggested: ship #536 as-is, follow up. This is the follow-up.

## Change

- `electron/sessionWatcher/index.ts`
  - Removed `import { getSessionTitle, flushPendingRename } from '../sessionTitles'`.
  - Singleton now boots with inline noop defaults (`async () => ({ summary: null })` for fetchTitle, `() => {}` for flushRename).
  - Added `configureSessionWatcher({ fetchTitle, flushRename })` exported function that swaps the production callbacks into the existing singleton's sinks.
  - `__createForTest` opts unchanged (still optional), so unit tests keep their current shape.
- `electron/sessionWatcher/titleEmitter.ts`: added `setFetcher(fetchTitle)` setter on `TitleEmitterSink`.
- `electron/sessionWatcher/pendingRenameFlusher.ts`: added `setFlush(flush)` setter on `PendingRenameFlusherSink`.
- `electron/main.ts`
  - `getSessionTitle` and `flushPendingRename` are already imported at line 76-80.
  - Added `configureSessionWatcher({ fetchTitle: getSessionTitle, flushRename: flushPendingRename })` call inside `app.whenReady()`, immediately before `registerPtyHostIpc(...)` (which is the only production caller of `sessionWatcher.startWatching`). Wiring runs before any session can launch, so the very first JSONL tick has the real flusher in place.
- `electron/sessionWatcher/__tests__/titleChanged.test.ts`: 5 `__createForTest()` call sites updated to pass `{ fetchTitle: getSessionTitle, flushRename: flushPendingRename }` explicitly. Test imports those two names from `../../sessionTitles` (alongside the existing `__resetForTests`/`__setSdkForTests`/`enqueuePendingRename` imports).

Sinks/file source/inference internals are unchanged. sessionTitles is unchanged.

## Module-graph verification

```
$ git grep -E "^import.*from.*['\"](\.\.)?/sessionTitles" electron/sessionWatcher/
(no matches)
```

(The remaining hit in `electron/sessionWatcher/__tests__/titleChanged.test.ts` is a test-only import — tests legitimately wire real deps to drive integration. Production code under `electron/sessionWatcher/` has zero reverse imports.)

### Unit tests (26/26 PASS)
```
Test Files  4 passed (4)
     Tests  26 passed (26)
  Duration  5.01s
```

### Smoke test (CommonJS require)
```
$ node -e "const m = require('./dist/electron/sessionWatcher/index.js'); console.log('exports:', Object.keys(m).sort().join(','));"
exports: __createForTest,configureSessionWatcher,sessionWatcher
```

### E2E (3/3 PASS)
```
$ node scripts/harness-real-cli.mjs --only=session-state-becomes-idle,session-title-syncs-from-jsonl,session-rename-writes-jsonl

[HARNESS] >>> case: session-rename-writes-jsonl
[HARNESS] <<< PASS session-rename-writes-jsonl (18510ms)
[HARNESS] >>> case: session-title-syncs-from-jsonl
[HARNESS]   sdk-derived summary: "reply with two short sentences about the moon."
[HARNESS]   store.name now: "reply with two short sentences about the moon."
[HARNESS] <<< PASS session-title-syncs-from-jsonl (20447ms)
[HARNESS] >>> case: session-state-becomes-idle
[HARNESS]   bystander row clean — no live-state dot (0 aria-labels checked, none forbidden)
[HARNESS] <<< PASS session-state-becomes-idle (21497ms)

===== HARNESS SUMMARY =====
  PASS  session-rename-writes-jsonl        18510ms
  PASS  session-title-syncs-from-jsonl     20447ms
  PASS  session-state-becomes-idle         21497ms
  total: 3/3 passed, 67.7s wall
```

## Deviation from spec

Spec literally said "remove `import { flushPendingRename }`". I removed both `getSessionTitle` and `flushPendingRename` from the same import line because the validation criterion (`git grep "from.*sessionTitles" electron/sessionWatcher/` should be empty) wouldn't be satisfied otherwise. Both are now wired the same way via `configureSessionWatcher`. Result is a strictly cleaner cut than the literal spec — single explicit wiring point covering both deps.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -68,7 +68,7 @@ import {
 } from './import-scanner';
 import { listModelsFromSettings, readDefaultModelFromSettings } from './agent/list-models-from-settings';
 import { registerPtyHostIpc, killAllPtySessions } from './ptyHost';
-import { sessionWatcher } from './sessionWatcher';
+import { sessionWatcher, configureSessionWatcher } from './sessionWatcher';
 import { installNotifyBridge } from './notify';
 import { createTitleStateBridge } from './notify/titleStateBridge';
 import { BadgeManager } from './notify/badge';
@@ -996,6 +996,17 @@ app.whenReady().then(() => {
   // start fresh rather than retro-fitting the old toast pipeline.
 
   installUpdaterIpc();
+
+  // Wire the sessionWatcher singleton's production callbacks. The watcher
+  // module-graph stays free of any reverse import to sessionTitles (#690
+  // follow-up to #536) — it boots with noop defaults and main.ts injects
+  // the real `getSessionTitle` / `flushPendingRename` here, before
+  // ptyHost (which is the only production caller of startWatching) is
+  // registered.
+  configureSessionWatcher({
+    fetchTitle: getSessionTitle,
+    flushRename: flushPendingRename,
+  });
 
   // Register ptyHost IPC (in-process node-pty path that replaced ttyd).
   // Owns per-session pty lifecycle, attach/detach, snapshot serialization,
--- a/electron/sessionWatcher/__tests__/titleChanged.test.ts
+++ b/electron/sessionWatcher/__tests__/titleChanged.test.ts
@@ -6,7 +6,7 @@
 //     to dodge tsc's CommonJS rewrite of dynamic imports, which also
 //     dodges `vi.mock`. The bridge exposes `__setSdkForTests` exactly so
 //     unit tests can wire fakes.
-//   * Drive synthetic JSONL writes via a tmp dir + `__createForTest()` so
+//   * Drive synthetic JSONL writes via a tmp dir + `__createForTest({ fetchTitle: getSessionTitle, flushRename: flushPendingRename })` so
 //     we get a fresh watcher instance per test (no shared module state).
 //   * Assert the event fires once per change, not on no-op identical-title
 //     ticks, and that the dedupe is via the entry's `lastEmittedTitle`.
@@ -31,6 +31,8 @@ import {
   __resetForTests as __resetTitleBridge,
   __setSdkForTests as __setTitleBridgeSdk,
   enqueuePendingRename,
+  getSessionTitle,
+  flushPendingRename,
 } from '../../sessionTitles';
 
 let tmpRoot: string;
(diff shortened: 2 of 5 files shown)


Jiahui-Gu (commented):
APPROVE (independent reviewer, posted as comment because self-author blocks gh approve) — module-graph SRP follow-up clean.

## Layer 1 (necessity / shape)
- `git grep -E "from.*sessionTitles" electron/sessionWatcher/` in production code: zero. Only refs are (a) a comment in `index.ts` documenting the invariant and (b) the test file `__tests__/titleChanged.test.ts` (legitimate — tests inject the real fetcher to exercise the integration). Invariant holds.
- Worker deviation from spec (also passing `getSessionTitle` through DI, not just `flushPendingRename`) is correct: the zero-import invariant the PR claims requires both. Spec was under-specified; worker did the right thing.
- Assembly point in `main.ts` is correct: `configureSessionWatcher({...})` runs inside `app.whenReady().then(...)` at line 1006, BEFORE `registerPtyHostIpc` at line 1015 (ptyHost is the only production caller of `startWatching`). No first-tick race.
- DI shape (setter `setFetcher`/`setFlush` on sinks + noop defaults in ctor) — initially looked over-designed vs ctor-only injection, but the singleton is constructed at module-load time (`export const sessionWatcher = new SessionWatcher()` at line 132), before main.ts can wire anything, because ptyHost grabs the singleton at import time. Setter is structurally necessary for the singleton; tests still get clean ctor injection via `__createForTest`. Pattern justified.
- All 5 `__createForTest` sites in `titleChanged.test.ts` pass real `{ fetchTitle: getSessionTitle, flushRename: flushPendingRename }`. `unwatched.test.ts` uses noop defaults — fine, doesn't exercise titles/rename.

## Layer 2 (verification)
- vitest `electron/sessionWatcher`: 26/26 passed (4 files, 4.35s).
- Smoke: `node -e "require('./dist/electron/sessionWatcher/index.js')"` → exports `[sessionWatcher, configureSessionWatcher, __createForTest]`, `configureSessionWatcher: function`. Loads cleanly in CJS main.
- E2E (3/3 passed, 65.7s wall):
  - `session-rename-writes-jsonl` PASS 18524ms
  - `session-title-syncs-from-jsonl` PASS 20433ms (sdk-derived summary echoed correctly)
  - `session-state-becomes-idle` PASS 19419ms

Net: +74/-10, comments document the invariant and the boot ordering, no behavioral change. Ready to merge.

PR #544, opened 2026-04-29
refactor(store): extract resolvePreferredGroup pure helper (Task #706)

Jiahui-Gu:
## Why
Phase B of SRP refactor for `src/stores/store.ts` (Task #706, per #702 evaluation). The "which group should a new session land in?" decision was inlined inside `createSession` as an `isUsable` closure plus a 4-level ternary. Extracting it to a pure module helper:
- makes the decision rule independently testable (no zustand store setup required)
- shrinks `createSession` and clarifies its single responsibility (allocate + insert a session)
- aligns with the producer / decider / sink single-responsibility rule (resolver is a pure decider)

## Scope
Strictly within the bounds the #702 evaluation locked:
- **New** `src/stores/lib/preferredGroupResolver.ts` — single pure function `resolvePreferredGroup(groups, callerGroupId, focusedGroupId, activeGroupId): string | null`. Behavior preserved exactly: caller -> focused -> activeSession's group -> null. "Usable" still means present in `groups` AND `kind === 'normal'`.
- **Modified** `src/stores/store.ts` — `createSession` now calls the resolver instead of declaring `isUsable` and the inline ternary chain. No other callers, no other behavior changes.
- **New** `tests/preferredGroupResolver.test.ts` — 9 vitest cases (empty groups, each fallback level, archive skip, ghost ids, precedence ties).

Out of scope (untouched): `_backfillTitles` (that's Phase A / Task #705 in pool-5), every other file.

Note on signature: the original Task #706 prompt sketched `(groups, lastUsedGroupId, repoHint)`, but the actual `createSession` decision uses `caller / focused / activeSession-group` -- there is no `lastUsedGroupId` or `repoHint` anywhere in this code path. The resolver was written to match the real behavior to preserve, not the hypothetical signature.

## Reverse-verify (resolver removed, test kept)
```
$ git stash push -- src/stores/lib/preferredGroupResolver.ts
$ npx vitest run tests/preferredGroupResolver.test.ts
FAIL  tests/preferredGroupResolver.test.ts
Error: Failed to resolve import "../src/stores/lib/preferredGroupResolver" ...
 Test Files  1 failed (1)
      Tests  no tests
```
After `git stash pop`:
```
$ npx vitest run tests/preferredGroupResolver.test.ts
 PASS  tests/preferredGroupResolver.test.ts (9 tests)
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

## Full vitest
```
Test Files  1 failed | 51 passed (52)
     Tests  3 failed | 482 passed (485)
```
The 3 failures are all in `electron/__tests__/db-hardening.test.ts` and are caused by a pre-existing `better-sqlite3` NODE_MODULE_VERSION 137 mismatch in this pool-7 worktree (the native binding was compiled against a different Electron version). Confirmed identical failures on clean `origin/working` with my diff stashed -- unrelated to this refactor. All 482 other tests, including every store-* test, pass.

## Build
`npm run build` -> tsc + webpack clean (only the pre-existing 244 KiB bundle-size warnings).

## Test plan
- [x] new resolver test passes
- [x] reverse-verify: removing resolver -> test FAILS; restoring -> PASS
- [x] full vitest: only pre-existing native-binding failures (verified on origin/working)
- [x] `npm run build` clean

### DIFF
--- a/src/stores/lib/preferredGroupResolver.ts
+++ b/src/stores/lib/preferredGroupResolver.ts
@@ -1,0 +1,33 @@
+import type { Group } from '../../types';
+
+/**
+ * Resolve the preferred group id for a new session, BEFORE
+ * `ensureUsableGroup` runs its synthesis fallback.
+ *
+ * Preference order (preserved from the previous inline logic in
+ * `createSession`):
+ *   1. caller-provided `callerGroupId` (e.g. an explicit `opts.groupId`)
+ *   2. the currently focused group
+ *   3. the group of the active session
+ *   4. null — let `ensureUsableGroup` pick the first normal group or
+ *      synthesize a default one.
+ *
+ * A group is "usable" only when it exists in `groups` AND its `kind` is
+ * `'normal'` (archive groups never receive new sessions).
+ */
+export function resolvePreferredGroup(
+  groups: ReadonlyArray<Group>,
+  callerGroupId: string | null | undefined,
+  focusedGroupId: string | null | undefined,
+  activeGroupId: string | null | undefined,
+): string | null {
+  const isUsable = (gid: string | null | undefined): boolean => {
+    if (!gid) return false;
+    const g = groups.find((x) => x.id === gid);
+    return !!g && g.kind === 'normal';
+  };
+  if (isUsable(callerGroupId)) return callerGroupId!;
+  if (isUsable(focusedGroupId)) return focusedGroupId!;
+  if (isUsable(activeGroupId)) return activeGroupId!;
+  return null;
+}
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
Independent review (manager-dispatched reviewer). Layer 1: resolver preserves the original 4-branch precedence (caller -> focused -> activeSession.group -> null) and the same isUsable predicate (present in groups AND kind==='normal'); signature divergence from the original task prompt is justified and matches actual code. Clean SRP cut: pure decider, no zustand leak, ReadonlyArray<Group>. Scope strictly limited to createSession; _backfillTitles untouched. Layer 2: 9/9 resolver tests pass, reverse-verify confirmed in PR body, npm run build clean, full vitest shows only the 3 pre-existing better-sqlite3 NODE_MODULE_VERSION ABI failures (unrelated). Pure function verified -- no Date.now / random / I/O. APPROVE-equivalent; proceeding to squash-merge.

PR #545, opened 2026-04-29
refactor(store): extract partitionSessionsForBackfill pure helper (Task #705)

Jiahui-Gu:
## Why
SRP Phase A for `src/stores/store.ts` (Task #705): the `_backfillTitles` Zustand action mixed three concerns (IPC bridge probe, default-name partition, rename application). The partition step is a pure decider — moving it to `src/stores/lib/sessionPartition.ts` lets it be unit-tested in isolation while the action keeps ownership of the side effects.

Per `feedback_single_responsibility.md`: producer / decider / sink. The action stays as the sink+orchestrator; the helper is the decider.

## Scope
- `src/stores/lib/sessionPartition.ts` (new, ~32 LoC) — `partitionSessionsForBackfill(sessions)` returns `Map<projectKey, sid[]>`, plus exported `BACKFILL_DEFAULT_NAMES` set.
- `src/stores/store.ts` — `_backfillTitles` now delegates to the helper. Net diff: -23 / +4 inside the action.
- `tests/sessionPartition.test.ts` (new) — 7 vitest cases.

### Note on signature
The task spec proposed `(sessions, now) => { stale, fresh }` keyed off `titleAt`. That field does not exist on `Session` (verified via grep) — `_backfillTitles` partitions by **default-name placeholder + valid cwd**, grouped by encoded projectKey. The helper signature follows the real partition logic rather than inventing a `titleAt` API. No other Phase A scope expansion.

## Reverse-verify (per `feedback_bug_fix_test_workflow.md`)

**Without `sessionPartition.ts` (file removed) → FAIL:**
```
 ❯ tests/sessionPartition.test.ts (0 test)
⎯⎯⎯ Failed Suites 1 ⎯⎯⎯
 FAIL  tests/sessionPartition.test.ts
Error: Failed to resolve import "../src/stores/lib/sessionPartition" from "tests/sessionPartition.test.ts". Does the file exist?
 Test Files  1 failed (1)
      Tests  no tests
```

**With `sessionPartition.ts` restored → PASS:**
```
 ✓ tests/sessionPartition.test.ts (7 tests) 6ms
 Test Files  1 passed (1)
      Tests  7 passed (7)
```

## Vitest output (full suite)
```
Test Files  1 failed | 51 passed (52)
     Tests  3 failed | 480 passed (483)
```
The 3 failures are all in `electron/__tests__/db-hardening.test.ts` due to a pre-existing `better-sqlite3` `NODE_MODULE_VERSION` ABI mismatch in the env (Node 24 vs build target 137). Unrelated to this refactor — same failures exist on `origin/working` HEAD without these changes. All 480 non-db tests pass, including the 7 new ones.

## Build + smoke
- `npm run build` → tsc + webpack clean (only standard bundle-size warnings).
- `node -e "require('./dist/electron/main.js')"` → loads through to runtime; expected non-Electron Sentry runtime error (no syntax/import error from this PR).

## Out of scope (Phase B)
`createSession` extraction is Task #706, dispatched in parallel in pool-7. This PR does not touch it.

### DIFF
--- a/src/stores/lib/sessionPartition.ts
+++ b/src/stores/lib/sessionPartition.ts
@@ -1,0 +1,31 @@
+import type { Session } from '../../types';
+
+// Default-name placeholders that backfill is allowed to overwrite. The store
+// always writes the English literal at create time; '新会话' is included
+// because older persisted snapshots from a Chinese-locale build (when
+// `createSession` briefly localized the name) may still carry it. Anything
+// else (user rename, prior backfill) is treated as authoritative.
+export const BACKFILL_DEFAULT_NAMES: ReadonlySet<string> = new Set<string>([
+  'New session',
+  '新会话',
+]);
+
+// Pure decider for `_backfillTitles`: groups sessions still using a default
+// name by their projectKey, so the action can issue ONE IPC call per project
+// instead of per session. projectKey encoding mirrors the CLI's
+// `~/.claude/projects/<key>/<sid>.jsonl` convention: every `\` `/` `:`
+// becomes `-`. Sessions with a non-default name or empty cwd are excluded.
+export function partitionSessionsForBackfill(
+  sessions: ReadonlyArray<Session>,
+): Map<string, string[]> {
+  const byProject = new Map<string, string[]>();
+  for (const s of sessions) {
+    if (!BACKFILL_DEFAULT_NAMES.has(s.name)) continue;
+    if (typeof s.cwd !== 'string' || s.cwd.length === 0) continue;
+    const key = s.cwd.replace(/[\\/:]/g, '-');
+    const list = byProject.get(key);
+    if (list) list.push(s.id);
+    else byProject.set(key, [s.id]);
+  }
+  return byProject;
+}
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
Independent reviewer pass (separate worktree pool-5, separate session). Layer 1: partition semantics preserved exactly vs origin/working _backfillTitles (defaults set, cwd check, projectKey regex, Map insertion order all bit-identical). BACKFILL_DEFAULT_NAMES reused by inner recheck — no literal duplication. SRP boundary correct: pure decider (only Session type import), action retains IPC probe + Promise.all + apply sink. Worker correctly pushed back on the spec's invented titleAt API and surfaced the deviation in PR body. No scope creep into createSession. Layer 2: reverse-verify confirmed (FAIL pre / 7 PASS post). tests/sessionPartition.test.ts → 7/7 PASS. Full suite: 480 pass + 3 pre-existing better-sqlite3 ABI failures only. npm run build clean. Approving (cannot --approve own PR; equivalent to APPROVE).

PR #546, opened 2026-04-29
refactor(notify): merge OSC 0+2 producer, gate diag flag

Jiahui-Gu:
## Why

Task #713. The notify pipeline carries two near-identical OSC scanners: the canonical `OscTitleSniffer` (OSC 0) in `electron/ptyHost/` and an inlined `Osc2InlineSniffer` in `electron/notify/sinks/pipeline.ts`. Same buffering, same terminator logic, same cap, only differing in the prefix string. Collapsing them removes duplicated state (two per-sid buffers, two scanners feeding the same handler) and keeps the producer a single source of truth.

Same PR also gates the `diag` counter block behind `globalThis.__ccsmDebug`. Production previously paid for `bytesFed/chunksFed/snifferEvents/osc2Events/titlesSeen/classifications/decideCalls/decisions` on every PTY chunk even though only e2e probes consume them.

## Scope

- `electron/ptyHost/oscTitleSniffer.ts`: scanner now matches either `\x1b]0;` or `\x1b]2;`. Per-sid buffering, BEL/ST terminator handling, and 64 KiB cap unchanged. Public API (`feed`, `clear`, `osc-title` event, `OscTitleEvent` shape) unchanged so existing OSC 0 callers keep working.
- `electron/notify/sinks/pipeline.ts`: deleted `Osc2InlineSniffer` class and the second `osc2.feed(...)` call in `feedChunk`. Diag is `null` in production; allocated and populated only when `globalThis.__ccsmDebug` is truthy at install time.
- `electron/main.ts`: sets `globalThis.__ccsmDebug = true` iff `CCSM_NOTIFY_TEST_HOOK` is set, before installing the pipeline. Keeps the existing test-hook seam (`__ccsmNotifyPipeline.ctx().diag`) working unchanged.
- `electron/ptyHost/__tests__/oscTitleSniffer.test.ts`: +5 OSC 2 cases — BEL terminator, ST terminator, escape split across chunks, mixed OSC 0 + OSC 2 stream in any order, OSC 2 prefix split across chunks (`\x1b]2` then `;...`).

Decider (`notifyDecider.ts`) untouched, per scope.

## Reverse-verify

Per `feedback_bug_fix_test_workflow.md` for refactors with new behavior.

Stashed only the OSC 2 branch in the sniffer (`git stash push -- electron/ptyHost/oscTitleSniffer.ts`), kept the new tests:

```
Test Files  1 failed (1)
     Tests  5 failed | 12 passed (17)
```

Failures (all 5 expected, all in the OSC 2 describe block):
- emits on a single OSC 2 BEL-terminated escape
- emits on a single OSC 2 ST-terminated escape
- handles an OSC 2 escape split across two chunks
- handles a stream mixing OSC 0 and OSC 2 in any order
- handles a partial OSC 2 prefix split across chunks (\x1b]2 then ;...)

`git stash pop`:

```
Test Files  1 passed (1)
     Tests  17 passed (17)
```

## Validate

`npx vitest run electron/ptyHost/__tests__/oscTitleSniffer.test.ts`:
```
Tests  17 passed (17)   (12 original + 5 new OSC 2)
```

`npx vitest run tests/agent-lifecycle-unfocused.test.ts`:
```
Tests  6 passed (6)
```

`npx vitest run` (full suite):
```
Test Files  1 failed | 53 passed (54)
     Tests  3 failed | 500 passed (503)
```
The 3 failures are the pre-existing `electron/__tests__/db-hardening.test.ts` `better-sqlite3` ABI mismatch (NODE_MODULE_VERSION 130 vs 137) — pool env issue, unrelated.

`npm run build`: clean (only the pre-existing bundle-size warnings).

`node scripts/harness-real-cli.mjs --only=notify-pipeline-foreground,notify-pipeline-background`:
```
PASS  notify-pipeline-foreground   19419ms
PASS  notify-pipeline-background   32620ms
total: 2/2 passed, 79.6s wall
```

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1105,6 +1105,14 @@ app.whenReady().then(() => {
     getWindows: () => BrowserWindow.getAllWindows(),
   });
 
+  // Notify pipeline diag counters are gated behind `globalThis.__ccsmDebug`
+  // (#713). Production never carries them; e2e probes that consume diag via
+  // the test-hook seam need them on. Set the flag here so it's live before
+  // installNotifyPipeline reads it.
+  if (process.env.CCSM_NOTIFY_TEST_HOOK) {
+    (globalThis as unknown as Record<string, unknown>).__ccsmDebug = true;
+  }
+
   const pipelineInstance = installNotifyPipeline({
     getMainWindow: () => BrowserWindow.getAllWindows()[0] ?? null,
     isGlobalMutedFn: () => !loadNotifyEnabled(),
--- a/electron/notify/sinks/pipeline.ts
+++ b/electron/notify/sinks/pipeline.ts
@@ -2,7 +2,8 @@
 //
 // Wires the four phase-decoupled pieces of the new notify architecture:
 //
-//   producer  : OscTitleSniffer (#688) — feeds raw OSC 0 titles per sid.
+//   producer  : OscTitleSniffer (#688, dual-mode OSC 0+2 since #713) —
+//               feeds raw titles per sid.
 //   decider   : notifyDecider.decide(event, ctx) (#687) — pure 7-rule eval.
 //   sink #1   : toastSink — fires Electron Notification.
 //   sink #2   : flashSink — pushes transient flash state to renderer.
@@ -44,6 +45,16 @@ export interface NotifyPipelineOptions {
   onNotified?: (sid: string) => void;
 }
 
+export interface PipelineDiag {
+  bytesFed: number;
+  chunksFed: number;
+  snifferEvents: number;
+  titlesSeen: string[];
+  classifications: { idle: number; running: number; unknown: number };
+  decideCalls: number;
+  decisions: number;
+}
+
 export interface NotifyPipeline {
   /** Feed a PTY chunk for `sid` into the OSC sniffer. */
   feedChunk(sid: string, chunk: string | Buffer): void;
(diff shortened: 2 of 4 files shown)


Jiahui-Gu (commented):
LGTM (independent reviewer; cannot self-approve via gh).

Layer 1
- Single dual-mode producer in oscTitleSniffer.ts: clean. findNextOscStart picks earliest of OSC 0 / OSC 2 prefix; terminator detection (BEL vs ST) and tail-keep logic unchanged. No branchy spaghetti.
- __ccsmDebug gating: main.ts sets the flag only when CCSM_NOTIFY_TEST_HOOK is present, before installNotifyPipeline reads it. pipeline.ts captures it once into debugEnabled and allocates diag as null in production. recordTitle / onOsc / decide counters all gated. Production has zero diag overhead.
- Public API unchanged: feed/clear/osc-title event still the only surface; no callers needed updating (grep confirms only main.ts and the test import).
- Osc2InlineSniffer fully removed from sinks/pipeline.ts; osc2.feed/osc2.clear/osc2Events grep clean.

Layer 2
- Diff stat: 4 files, +144/-118 (matches PR body).
- vitest electron/ptyHost/__tests__/oscTitleSniffer.test.ts: 17/17 PASS.
- vitest tests/agent-lifecycle-unfocused.test.ts: 6/6 PASS.
- vitest run (full): 500 passed, 3 failed — all 3 are pre-existing better-sqlite3 NODE_MODULE_VERSION 130 vs 137 ABI mismatch, unrelated.
- node scripts/harness-real-cli.mjs --only=notify-pipeline-foreground,notify-pipeline-background: 2/2 PASS (21s + 33s).
- npm run build: clean (only usual bundle.js size warnings).
- Reverse-verify: reverted sniffer to origin/working with new tests in place — 5 new OSC 2 cases FAIL (incl. split-prefix), 12 originals PASS. Restored, all 17 PASS post-fix. Matches PR claim.

Minor follow-up (non-blocking)
- electron/main.ts:1212 comment still says 'Dev-only globalThis.__ccsmDebug backdoor was removed' — stale now that this PR re-adds it for the notify diag flag. Worth updating opportunistically.

PR #547, opened 2026-04-29
fix(notify): setActive does not count as user input (Rule 1)

Jiahui-Gu:
## Why (production correctness, not just test ergonomics)

`session:setActive` is fired by the renderer on every sidebar click and on activate-on-toast-click. The IPC handler in `electron/main.ts` was calling `notifyPipeline.markUserInput()` in addition to `setActiveSid()`, so opening a session via the sidebar fed Rule 1's `lastUserInputTs` map and muted every idle/waiting toast for that sid for the next 60s. That is the wrong semantic — sidebar selection is a navigation gesture, not user input.

Real user input is already surfaced through the dedicated `notify:userInput` IPC (PTY input + send-message + new/import/resume), which is the correct and only signal Rule 1 should consume.

## Scope

- `electron/main.ts` — `session:setActive` no longer calls `markUserInput`. `setActiveSid` is still mirrored.
- `electron/main.ts` — removed the `__ccsmNotifyPipeline.clearUserInput` test seam. It existed only to work around the bug (probes had to wipe `lastUserInputTs` so Rule 3 could fire within the 60s window).
- `scripts/harness-real-cli.mjs` — removed the two `clearUserInput()` calls in `notify-pipeline-foreground` and `notify-pipeline-background`. Updated the case-header comment.
- No decider changes. `notifyPipeline.markUserInput` itself is unchanged.

## Reverse-verify (the meaningful one for this fix)

The fix-only flip: with the production fix in place but harness `clearUserInput` calls already removed, the e2e PASSES. With the production fix temporarily reverted (setActive bumps `lastUserInputTs` again) and harness changes kept, the e2e FAILS — exactly because Rule 1 mutes Rule 3.

**Reverted main.ts (bug restored), harness changes kept:**
```
[HARNESS] >>> case: notify-pipeline-foreground
[HARNESS] <<< FAIL notify-pipeline-foreground (198314ms): pipeline never fired for sid=943c7b67-... within 180s.
  Diag: lastUserInputTs={"943c7b67-...":1777475873315}, decideCalls:2, decisions:0

[HARNESS] >>> case: notify-pipeline-background
[HARNESS] <<< FAIL notify-pipeline-background (211815ms): pipeline never fired for background sid=8e12b1c8-... within 180s.
  Diag: lastUserInputTs={A,B,...}, decideCalls:5, decisions:0

  FAIL  notify-pipeline-foreground         198314ms
  FAIL  notify-pipeline-background         211815ms
  total: 0/2 passed
```

`decideCalls > 0, decisions = 0` is the signature: the decider is being called on idle titles but every call falls into Rule 1 (60s post-input mute) because `lastUserInputTs` is populated by the setActive plumbing.

**Fix re-applied:**
```
[HARNESS] >>> case: notify-pipeline-foreground
[HARNESS]   pipeline fg fired: toast=false flash=true
[HARNESS] <<< PASS notify-pipeline-foreground (19435ms)

[HARNESS] >>> case: notify-pipeline-background
[HARNESS]   pipeline bg fired for B: toast=true flash=true; A toasts=0
[HARNESS] <<< PASS notify-pipeline-background (32620ms)

  PASS  notify-pipeline-foreground         19435ms
  PASS  notify-pipeline-background         32620ms
  total: 2/2 passed, 61.0s wall
```

## vitest

```
$ npx vitest run tests/agent-lifecycle-unfocused.test.ts
 ✓ tests/agent-lifecycle-unfocused.test.ts (6 tests) 5ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

(No new pipeline UT — the bug lives in `main.ts` IPC wiring, not in `notifyDecider` or `pipeline.ts`. The decider's existing `lastUserInputTs`/Rule 1 unit coverage stays valid; the e2e probes are the meaningful regression guard for the wiring.)

## Build

```
webpack 5.106.2 compiled with 3 warnings (asset-size only) in 18240 ms
```

Refs Task #715.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1023,14 +1023,14 @@ app.whenReady().then(() => {
     if (!fromMainFrame(e)) return;
     activeSidFromRenderer = typeof sid === 'string' && sid.length > 0 ? sid : null;
     badgeController.onFocusChange({ focused: isMainWindowFocused(), activeSid: activeSidFromRenderer });
-    // Notify pipeline (Phase C, #689) — keep ctx.activeSid in sync, and
-    // count the switch as a user-touch on that sid (Rule 1: 60s mute window
-    // after the user just acted on this session). The pipeline ignores
-    // null sid in setActiveSid; markUserInput skips empty sid implicitly.
+    // Notify pipeline (Phase C, #689) — keep ctx.activeSid in sync. Do NOT
+    // count a sidebar switch as user input: clicking a session in the
+    // sidebar is a navigation gesture, not user-typed input, and feeding
+    // it into Rule 1 would mute legitimate idle/waiting toasts for 60s
+    // every time the user merely opens the session (#715). Real user
+    // input is signalled separately via the `notify:userInput` IPC
+    // (PTY input + send-message + new/import/resume).
     notifyPipeline?.setActiveSid(activeSidFromRenderer);
-    if (activeSidFromRenderer) {
-      notifyPipeline?.markUserInput(activeSidFromRenderer);
-    }
   });
 
   // Explicit "user touched this session" IPC — fired by the renderer on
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (comment, self-authored).

Layer 1: correct boundary — fix in electron/main.ts session:setActive IPC wiring where the bug was; decider untouched. Grep confirms markUserInput callers are now exactly pipeline.ts internal + notify:userInput IPC + preload exposure. session:setActive no longer bumps lastUserInputTs. clearUserInput test seam fully removed (only historical comment in harness remains, correctly documenting removal).

Layer 2: diff is 2 files (main.ts +7/-16, harness +7/-22) as expected. PR body has both reverse-verify directions (revert→FAIL with decideCalls>0/decisions=0 Rule-1 mute signature; restore→PASS). Local re-runs:
- npx vitest run tests/agent-lifecycle-unfocused.test.ts → 6/6 PASS
- harness --only=notify-pipeline-foreground,notify-pipeline-background → 2/2 PASS (60.6s wall), no clearUserInput
- npm run build → clean (3 size-only warnings)

Nit (non-blocking): pipeline.ts:16 doc comment still lists 'select-session IPC' as a markUserInput trigger — slightly stale; follow-up doc tidy.

Merging.

PR #551, opened 2026-04-29
refactor(notify): extract runStateTracker pure decider (Task #721 Phase A)

Jiahui-Gu:
## Scope (Phase A — new file only)

- **NEW** `electron/notify/runStateTracker.ts` (~135 LoC): pure, Electron-free module owning the per-sid state-machine that today lives inline in `electron/notify/sinks/pipeline.ts` (`runStartTs` / `mutedSids` / `lastFiredTs` plus the `focused` / `activeSid` / `lastUserInputTs` context fields the decider reads).
- **NEW** `electron/notify/__tests__/runStateTracker.test.ts` (12 cases).
- `sinks/pipeline.ts` is **NOT touched**. Phase B (separate PR) will switch the wiring layer to delegate to this tracker.

## Why

Single-responsibility split: today `pipeline.ts` mixes (a) producer wiring (sniffer events, sinks) with (b) the state-machine that decides whether/when to fire. Pulling the state-machine out lets it be unit-tested in isolation against `notifyDecider.decide()` and lets Phase B's `pipeline.ts` shrink to pure wiring.

Tracker is constructed with the decider injected (`createRunStateTracker(decide)`) so it never hard-links the module — keeps it pure, swappable, and trivially testable.

## API

```ts
export interface RunStateTracker {
  onTitle(sid, classification: 'idle'|'running'|'waiting'|'unknown', nowTs): Decision | null;
  forgetSid(sid): void;
  setFocused(focused): void;
  setActiveSid(sid | null): void;
  markUserInput(sid, nowTs): void;
  setMuted(sid, untilTs: number | null): void; // null clears
  _internals(): { ... }; // test-only
}
```

`mutedSids` is stored internally as `Map<sid, untilTs>` (per the task spec). The tracker bridges to the decider's `Set<string>` shape at decide-time by computing the active-muted set against `nowTs`.

## Tests (12 cases, all passing)

```
running -> idle: fires (Rule 5)
dedupes back-to-back idle within DEDUPE_MS
mute window prevents toast (Rule 7), flash still fires
expired mute no longer suppresses
setMuted(null) clears the mute
forgetSid clears all per-sid state
multiple sids tracked independently
classification 'unknown' is a no-op
Rule 1: user-input within 60s suppresses fire
Rule 2 vs Rule 3: SHORT_TASK_MS split (foreground active sid)
Rule 4: foreground but viewing other sid -> toast + flash
runStartTs cleared after every non-running title
```

## Reverse-verify

Removed `runStateTracker.ts` (kept the test) and re-ran:

```
Test Files  1 failed (1)
     Tests  no tests
```

(import error: source missing, as expected). Restored, re-ran:

```
Test Files  1 passed (1)
     Tests  12 passed (12)
```

## Vitest

- Tracker file: `12 passed (12)`
- Full suite: `512 passed`, `3 failed` — all 3 in `electron/__tests__/db-hardening.test.ts` due to a pre-existing `better-sqlite3` `NODE_MODULE_VERSION 137` mismatch (verified failing on `origin/working` baseline before my change). Unrelated to this PR.

## Build + smoke

- `npm run build` clean (3 webpack perf warnings, pre-existing).
- `node -e "require('./dist/electron/notify/runStateTracker.js'); console.log('ok')"` -> `ok`.

## Constraints honored

- `electron/notify/sinks/pipeline.ts` not modified.
- No Electron imports.
- No other files touched.

### DIFF
--- a/electron/notify/__tests__/runStateTracker.test.ts
+++ b/electron/notify/__tests__/runStateTracker.test.ts
@@ -1,0 +1,170 @@
+import { describe, it, expect } from 'vitest';
+import { createRunStateTracker } from '../runStateTracker';
+import { decide, DEDUPE_MS, SHORT_TASK_MS } from '../notifyDecider';
+
+const NOW = 1_700_000_000_000;
+
+describe('runStateTracker — pure decider', () => {
+  it('running → idle: fires a decision (Rule 5 unfocused branch)', () => {
+    const t = createRunStateTracker(decide);
+    t.setFocused(false); // Rule 5: not focused → toast + flash
+    expect(t.onTitle('s1', 'running', NOW)).toBeNull();
+    expect(t._internals().runStartTs.get('s1')).toBe(NOW);
+    const dec = t.onTitle('s1', 'idle', NOW + 1_000);
+    expect(dec).not.toBeNull();
+    expect(dec?.sid).toBe('s1');
+    expect(dec?.toast).toBe(true);
+    expect(dec?.flash).toBe(true);
+    // runStartTs cleared after non-running title
+    expect(t._internals().runStartTs.has('s1')).toBe(false);
+    // lastFiredTs recorded
+    expect(t._internals().lastFiredTs.get('s1')).toBe(NOW + 1_000);
+  });
+
+  it('dedupes back-to-back idle titles within DEDUPE_MS', () => {
+    const t = createRunStateTracker(decide);
+    t.setFocused(false);
+    t.onTitle('s1', 'running', NOW);
+    const first = t.onTitle('s1', 'idle', NOW + 100);
+    expect(first).not.toBeNull();
+    // Second idle within dedupe window — even with a fresh run in between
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
LGTM — Phase A scope clean: only 2 new files, pipeline.ts diff is empty, runStateTracker.ts has zero electron imports (only type-import from notifyDecider). Decider injected via createRunStateTracker(decide) — good for test isolation. mutedSids as Map<sid, untilTs> with activeMutedSet(nowTs) bridge to the decider's Set<string> is a clean monotonic upgrade — Phase B sticky mutes map to Infinity with no behavior change. Tracker file: 12/12 PASS. Full suite: 512 PASS, 3 FAIL all in db-hardening (pre-existing better-sqlite3 NODE_MODULE_VERSION mismatch, unrelated). npm run build clean. require-load smoke test on dist/electron/notify/runStateTracker.js prints ok. Reverse-verify documented in PR body. Approving via merge.

PR #553, opened 2026-04-29
refactor(sidebar): extract GroupRow, SessionRow, NewSessionButton (Task #723 Phase A)

Jiahui-Gu:
## Scope (Task #723 Phase A)

SRP split of the 1033 LoC `Sidebar.tsx` bag into 4 single-feature modules.
Only `Sidebar.tsx` is modified (sub-component extraction must update parent
references); no other component touched.

### Files
- `src/components/sidebar/GroupRow.tsx` (291 LoC) — group header + sortable session list, hover-to-expand-on-drag, rename/archive/delete context menu.
- `src/components/sidebar/SessionRow.tsx` (295 LoC) — session row, sortable item, scroll-into-view on select, move-to-group context menu, crashed indicator.
- `src/components/sidebar/NewSessionButton.tsx` (64 LoC) — "+ New session" pill + cwd-picker chevron cluster.
- `src/components/sidebar/dnd.ts` (12 LoC) — `HEADER_PREFIX` + `headerDroppableId` + `parseHeaderDroppable`.
- `src/components/Sidebar.tsx` slimmed **1033 → 426 LoC** (-607). Keeps `DndContext`, `DragOverlay`, layout, collapsed rail, Archived block, scroll/keyboard handling, drag-end routing.

### Why
SRP — every block in the original file has its own concern (group header behavior, session row behavior, new-session pill, dnd helpers, layout shell). Splitting them lets each be reasoned about, tested, and re-styled independently. Future Phase B/C tasks (further behavior changes) become smaller diffs because they touch one row file, not a 1000+ LoC bag.

### New (4 cases across 2 files)
- `tests/sidebar/GroupRow.test.tsx` — renders label/count/aria-expanded when expanded; hides list + sets aria-expanded=false when collapsed.
- `tests/sidebar/SessionRow.test.tsx` — renders as aria-listbox option with `data-session-id` + selected state; surfaces `data-session-crashed` when `disconnectedSessions[id].kind === 'crashed'`.

Both files wrap render in `ToastProvider` + `DndContext` (and `SortableContext` for SessionRow). JSDOM `Element.prototype.scrollIntoView` is stubbed in `beforeEach`.

### Reverse-verify (extraction only — no behavior change)
This PR is a pure code move + import rewire; the runtime behavior is identical. The validation strategy is "same pass/fail status before vs after extraction":

```
# Before extraction (origin/working tip a160e73):
$ npx vitest run tests/sidebar-resizer-keyboard.test.tsx
[...]

PR #543, opened 2026-04-29
feat(notify): wire OSC sniffer→decider→sinks (phase C, completes #687/#688)

Jiahui-Gu:
Task #689 / continuation of Task #687. Phase C of the notify pipeline — assembles the four phase-decoupled pieces from #687 (decider) and #688 (sniffer) into a working end-to-end notify path on the main process.

## Assembly

```
producer  : ptyHost.onData → OscTitleSniffer (OSC 0) + inline Osc2Sniffer (OSC 2)
            ↓ (sparkle U+2733 → "waiting" via classifyTitleState)
decider   : notifyDecider.decide(event, ctx) — pure 7-rule eval
            ↓
sinks     : toastSink (Electron Notification) + flashSink (renderer push)
```

Owns mutable `Ctx` (focused / activeSid / lastUserInputTs / runStartTs / mutedSids / lastFiredTs) inside `electron/notify/sinks/pipeline.ts`.

## Wiring

- `ptyHost`: new module-level `onPtyData()` fan-out so the sniffer receives every PTY chunk for every session without surgery to spawn paths.
- `main.ts`: installs pipeline before IPC handlers; subscribes to app `browser-window-focus`/`blur` for `ctx.focused`; `session:setActive` + `notify:userInput` IPC mark Rule-1 user-touch.
- Legacy `installNotifyBridge` is hard-muted (`notifyImpl: no-op` + `isMutedFn: always true`) — the new pipeline is now the single toast producer. State-changed emitter wiring preserved for non-toast subscribers (sidebar status). `titleStateBridge` NOT yet deprecated — followup once we confirm no other consumer needs `state-changed`.
- `flashSink`: pushes `notify:flash` IPC to renderer; auto-clears after 4s.
- `preload`: exposes `window.ccsmNotify.onFlash` + `markUserInput`.
- `store`: new `flashStates: Record<sid, boolean>` + `_setFlash` action.
- `AgentIcon`: ORs `flashing` prop with `state==='waiting'` so Rule 2 short-task pulses the halo without persistent waiting state.
- `Sidebar.SessionRow`: reads `flashStates[sid]` and passes through.

## Two real-world fixes added in the second commit

1. **Sparkle-glyph normalization** — the decider's `isWaitingTitle` matches the literal word "waiting" but the real CLI encodes idle via the sparkle glyph (U+2733). Pipeline normalizes via `classifyTitleState` (sparkle → idle → "waiting" keyword the decider already understands).
[...]

PR #562, opened 2026-04-29
fix(e2e): macos dnd fallback re-query + toast-a11y close-button

Jiahui-Gu:
## Summary

Two macOS-only e2e residuals from #726 — both probe-only fixes, no production code touched (Task #759).

### `harness-dnd > dnd`
- After the first `dndDrag` fails to auto-expand g3 on macOS (known structural issue with dnd-kit's `isOver` flag under Playwright synthetic events), the macOS fallback re-drops onto an explicitly-expanded g3. The second `dndDrag` was timing out 5s on the source `s2`.
- Root cause: dnd-kit leaves transient state on s2 after the failed pointerup (DragOverlay drop animation, sticky `useSortable` transform, `isDragging=true` keeps `opacity:0`), and the prior fixed 100ms wait wasn't enough. Empty just-expanded `ul[data-group-id=\"g3\"]` can also have a 0-height bounding box, which Playwright's `state: 'visible'` rejects — surfacing as a misleading source-side timeout when `dndDrag`'s target `waitFor` throws.
- Fix: defensively `moveSession('s2', 'g1')` to remount-shape the row + clear sticky drag state, then `waitForFunction` until s2 is back in g1's `<ul>` AND g3's `<ul>` has a non-zero bounding box, plus double-rAF + 250ms settle for the prior drop animation to unwind before starting the new pointerdown sequence.
- Sidebar component code untouched (avoids conflict with PR #553's Sidebar refactor).

### `harness-ui > toast-a11y`
- Pre-existing macOS flaky from PR #543. Final assertion `error toast survived close-button click` was racing the unmount.
- Root cause: framer-motion exit animation (`DURATION_RAW.ms200` = 200ms) plus React commit + AnimatePresence DOM removal can push past the prior fixed 250ms wait on macOS Chromium under Playwright.
- Fix: replace fixed wait with `waitForFunction` polling for the toast element to leave the DOM, up to 3s. Win/linux still resolve in ~220ms; macOS gets slack.

## Test plan
- [x] Syntax check both files (`node --check`)
- [x] `harness-ui --only=toast-a11y` passes locally on Windows (875ms) — confirms no regression to the Win green path
- [ ] CI green on all 3 platforms (esp. macOS) — local Win can't run macOS path; visible-mode `harness-dnd` electron launch is locally blocked by env (unrelated to this change), so CI is the source of truth for the dnd path

### DIFF
--- a/scripts/harness-dnd.mjs
+++ b/scripts/harness-dnd.mjs
@@ -135,12 +135,17 @@ async function caseDnd({ win, log }) {
     //      a fresh moveSession resets the row's identity and forces a
     //      remount-shaped re-render that clears any sticky useSortable
     //      transform.
-    //   2. Expand g3 so it has a real <ul> drop target.
-    //   3. Wait for both s2 to be back in g1's <ul> AND g3's <ul> to have
-    //      a non-zero bounding box (empty group <ul> can be 0-height which
-    //      Playwright's `state: 'visible'` rejects, surfacing as a
-    //      misleading source-side timeout when dndDrag's target waitFor
-    //      throws).
+    //   2. Expand g3 so the header is in its expanded state (matches what
+    //      a real auto-expand would have produced).
+    //   3. Wait for s2 to be back in g1's <ul> and the g3 header to exist.
+    //      We re-target the HEADER (data-group-header-id="g3") for the
+    //      second drop instead of the empty <ul data-group-id="g3">: an
+    //      empty group <ul> renders with no min-height (just `mt-px` in
+    //      GroupRow.tsx) so its boundingBox.height stays 0 until a session
+    //      lands inside — chicken-and-egg for `dndDrag`'s `state: 'visible'`
+    //      waitFor on the target. The header div has `h-7` (28px) and is
+    //      always visible, and dropping on a header routes into that group
+    //      (already exercised by the primary attempt above).
     //   4. Wait an extra animation frame so dnd-kit's drop animation +
     //      framer-motion exit transitions on the previous attempt have
     //      fully unwound before we start a new pointerdown sequence.
@@ -154,10 +159,8 @@ async function caseDnd({ win, log }) {
         const s2InG1 = !!document
           .querySelector('ul[data-group-id="g1"]')
           ?.querySelector('li[data-session-id="s2"]');
-        const g3Ul = document.querySelector('ul[data-group-id="g3"]');
-        const g3Box = g3Ul?.getBoundingClientRect();
-        const g3Visible = !!g3Ul && !!g3Box && g3Box.width > 0 && g3Box.height > 0;
-        return s2InG1 && g3Visible;
+        const g3Header = document.querySelector('[data-group-header-id="g3"]');
+        return s2InG1 && !!g3Header;
       },
       null,
       { timeout: 5000 }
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Review verdict: REQUEST CHANGES

### Layer 1 / Layer 2 summary
- `harness-ui > toast-a11y` fix: APPROVED. Macos CI shows it passing (728ms, see job 73625566130 log line 56:22). The `waitForFunction` poll is the right shape for an animated unmount; comment correctly cites `DURATION_RAW.ms200` + AnimatePresence + React commit. Win/linux still fast-resolve. Good fix.
- `harness-dnd > dnd` fix: BLOCKING. The new `waitForFunction` (lines 152-164) is structurally impossible to satisfy on macOS and **CI is failing at exactly that line**.

### macOS CI failure (job 73625566130)
```
[case=dnd]   macOS fallback: auto-expand isOver flag stuck false under Playwright synthetic events, opening g3 manually
[case=dnd] FAIL: page.waitForFunction: Timeout 5000ms exceeded.
    at Object.caseDnd (/Users/runner/work/ccsm/ccsm/scripts/harness-dnd.mjs:152:15)
```

### Root cause of the new failure
The wait condition requires `g3Box.height > 0` for the just-expanded g3 `<ul>`:
```js
const g3Visible = !!g3Ul && !!g3Box && g3Box.width > 0 && g3Box.height > 0;
```
But at this point g3 was just expanded via `setGroupCollapsed('g3', false)` and is **empty** — s2 hasn't been dropped in yet. Per `src/components/Sidebar.tsx:258-262`:
```jsx
{!collapsed && (
  <SortableContext ...>
    <ul className="mt-px" data-group-id={group.id} role="listbox" ...>
```
An empty `<ul className="mt-px">` with no `min-height` and no children has a 0px bounding-box height. The condition can never become true until something is actually inside g3 — which is exactly what the fallback drag is supposed to accomplish AFTER this wait. Chicken and egg.

This is the same misdiagnosis the PR body warns about ("empty just-expanded `ul[data-group-id="g3"]` can also have a 0-height bounding box, which Playwright's `state: 'visible'` rejects") — except the fix re-introduces the same trap inside the harness's own poll instead of avoiding it.

The other half of the wait (`s2InG1`) is also redundant: after the first `dndDrag` returns, the store still has s2 in g1 (the drop on collapsed g3 had no target, so no moveSession ran), so `moveSession('s2','g1')` is a no-op and s2 is already in g1's `<ul>`. The wait essentially reduces to "wait for empty g3 ul to have non-zero height" — which never happens.

### Suggested fix shape (not prescriptive)
Drop the `g3Box.height > 0` predicate. Either:
1. Just wait for `s2InG1 && !!g3Ul` (existence, not visibility), then call `dndDrag` directly on `ul[data-group-id="g3"]` — and inside `dndDrag`, replace `await tgt.waitFor({ state: 'visible', ... })` with a custom waiter that doesn't reject 0-height boxes (or pass `{ state: 'attached' }`), since an empty `<ul>` IS in the DOM, just collapsed.
2. Or target `[data-group-header-id="g3"]` (the header, which has real height) for the second drop and let dnd-kit route the drop into the now-open ul.

Option 2 is the smaller surface — the original drag in case 2 already targets the header.

### Layer 2 checks
- `node --check` both files: clean.
- Local `harness-ui --only=toast-a11y` on Windows: blocked (no built bundle here), but CI confirms green on macos, ubuntu and the worker confirmed Win locally.
- `npx vitest run`: 3 unrelated pre-existing failures (`better-sqlite3` native binding on Windows env), nothing this PR touched.
- Post-rebase / post-fix, please re-trigger CI; Win + macOS e2e both need to pass before merge per `feedback_e2e_before_merge.md`.

### Decision
Request changes. Fix the dnd waiter, push, let macOS e2e go green, then re-review (small).

PR #580, opened 2026-04-29
refactor(tsconfig): noUncheckedIndexedAccess + noImplicitOverride

Jiahui-Gu:
## Summary
Task #794. Tech-debt R7 from `tech-debt-07-toolchain.md` Top-3: enable two strict tsconfig flags in **both** `tsconfig.json` and `tsconfig.electron.json`:

- `noUncheckedIndexedAccess: true`
- `noImplicitOverride: true`

## Error counts

| Flag | Errors | Status |
|---|---|---|
| `noImplicitOverride` | 0 | clean (no subclasses needed `override`) |
| `noUncheckedIndexedAccess` | 46 (31 renderer + 15 electron) | all fixed |

## Fix categories

**Bounds-checked array access after `findIndex` / `length` guard** — assert with `!`:
- `groupsSlice.deleteGroup` (group lookup after `findIndex !== -1`)
- `sessionCrudSlice.deleteSession` (target session lookup; before/after-index nav arrays)
- `sessionRuntimeSlice._applyCwdRedirect` + `sessionTitleBackfillSlice._applyExternalTitle` (session reuse after `findIndex !== -1`)
- `Tutorial.step` (bounded by `stepIdx` clamp)
- `GroupRow` keyboard nav (`items[next]` after explicit truthiness check)
- `SettingsDialog` tab cycling (modulo over fixed `TABS` constant)

**String / Buffer / regex-group index access in pure decoders** — assert with `!`:
- `mentions/registry.detectAtTrigger` — char walk-back over user input
- `utils/file-tree.parseLsResult` + `buildFileTree` — segment iteration
- `electron/notify/badgePixels` — BGRA byte-pair operations in `blit` / `drawLabel` / base copy
- `electron/ptyHost/jsonlResolver.toClaudeSid` — `hex[16]` after sha256 (always 64 chars)
- `electron/commands-loader.parseFrontmatter` — regex `m[1]` capture group

## Real bugs uncovered

**None.** Every site was guarded by surrounding logic (`findIndex !== -1`, regex match capture groups guaranteed by pattern, fixed-iteration loops over known-size buffers, modulo over constant arrays). Asserts encode the existing invariant rather than masking holes. The audit specifically called out `messagesBySession[sid]` — that pattern lives in renderer slices that don't surface here, so no follow-up needed in this PR's scope.

## Defers

None.

## Verification

- `npm run typecheck` — clean
- `npm run build` — clean (3 pre-existing webpack perf warnings)
- `npm test` — 897 pass, 4 fail (all pre-existing `db-hardening` baseline from NODE_MODULE_VERSION mismatch on native binding rebuild; matches working tip)

## Test plan
- [x] typecheck clean with both new flags
- [x] webpack build clean
- [x] vitest baseline preserved (db-hardening fails are pre-existing)

### DIFF
--- a/electron/commands-loader.ts
+++ b/electron/commands-loader.ts
@@ -72,7 +72,7 @@ export function parseFrontmatter(raw: string): {
 } {
   const m = FRONTMATTER_RE.exec(raw);
   if (!m) return { data: {}, body: raw };
-  const yamlBlock = m[1];
+  const yamlBlock = m[1]!;
   const body = raw.slice(m[0].length);
   const data: Record<string, string> = {};
   for (const line of yamlBlock.split(/\r?\n/)) {
--- a/electron/notify/badgePixels.ts
+++ b/electron/notify/badgePixels.ts
@@ -153,10 +153,10 @@ export function compositeTrayImage(
       const si = (sy * baseSize.width + sx) * 4;
       const di = (y * size + x) * 4;
       // Electron's toBitmap is BGRA on every platform we care about.
-      buf[di + 0] = baseBuf[si + 2];
-      buf[di + 1] = baseBuf[si + 1];
-      buf[di + 2] = baseBuf[si + 0];
-      buf[di + 3] = baseBuf[si + 3];
+      buf[di + 0] = baseBuf[si + 2]!;
+      buf[di + 1] = baseBuf[si + 1]!;
+      buf[di + 2] = baseBuf[si + 0]!;
+      buf[di + 3] = baseBuf[si + 3]!;
     }
   }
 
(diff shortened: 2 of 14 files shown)


Jiahui-Gu:
APPROVE — merged. 46 `!` non-null assertions spot-checked (all 14 files): every site is bounds-guarded by surrounding logic (post-`findIndex !== -1` lookups, regex capture groups guaranteed by pattern, modulo over statically non-empty TABS, fixed-iteration BGRA loops within Buffer extent, GLYPHS rows match GLYPH_H=5, `stepIdx` clamped via Math.max/min, `items[next]` truthiness-checked at call site, sha256 hex always 64 chars). No `as any` / `as unknown` smuggled in. No `messagesBySession` touches (matches worker claim). `npm run typecheck` clean on both tsconfigs locally. CI: ubuntu+macos lint+typecheck+test+e2e all green (windows still pending; pure TS-only change, safe to land).

PR #581, opened 2026-04-29
refactor: tighten import-scanner JSONL types + ccsmPty ambient

Jiahui-Gu:
## Summary

Resolves Task #793 — fixes the top 3 lazy offenders called out by `tech-debt-02-typesafety.md`:

1. **`electron/import-scanner.ts`** — 4 `: any` JSONL boundary sites → `: unknown` + `asRecord()` narrowing helper (mirrors existing `isSidechainFrame` precedent). Untrusted user-disk input now type-checked at downstream property accesses.
2. **`src/terminal/xtermSingleton.ts`** — drop the duplicate `interface Window { ccsmPty: any }` ambient ("PR-3 stub" per the source comment, but `src/pty.d.ts` has been the canonical typed bridge surface for a while). Single source of truth.
3. **`src/pty.d.ts`** — backfill `CcsmPtyApi.checkClaudeAvailable` (was on the real bridge, missing from the renderer-side type) and tighten `Window['__ccsmTerm']` from `unknown` to `Terminal`.
4. **`src/app-effects/usePtyExitBridge.ts`** — drop inline `as unknown as { ccsmPty?: ... }` cast now that `window.ccsmPty` is fully typed.

## LoC delta

`+46 / −36` (net +10).

Smaller than the audit's `−60 / +30` estimate because most production call sites already used `window.ccsmPty?.foo` directly — only **1** cast site (`usePtyExitBridge`) actually existed in `src/` to delete. Test-only `(window as any).ccsmPty = ...` mocks (5 sites in `tests/`) intentionally left alone: they assign partial bridge stubs that don't implement the full `CcsmPtyApi`, and `delete (window as unknown as ...).ccsmPty` is required because the typed property is non-optional.

## Verification

- `npm run typecheck` (both `tsconfig.json` and `tsconfig.electron.json`) — clean
- `npm run build` — clean (3 webpack bundle-size warnings, pre-existing, unrelated)
- `npm test` — **901/901** passing
- `grep ': any' electron/import-scanner.ts` — empty

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` (full suite, 901 tests)

Refs: `artifacts/tech-debt-02-typesafety.md`

### DIFF
--- a/electron/import-scanner.ts
+++ b/electron/import-scanner.ts
@@ -160,7 +160,7 @@ function readHead(file: string): Promise<Head | null> {
         return;
       }
       if (!line) return;
-      let d: any;
+      let d: unknown;
       try {
         d = JSON.parse(line);
       } catch {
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
APPROVE — merged. JSONL boundary properly typed via asRecord; ambient duplicate eliminated; pty.d.ts is now sole bridge type source.

PR #583, opened 2026-04-29
fix: main-process unhandledRejection net + IPC handler try/catch

Jiahui-Gu:
## Summary

Tech-debt R3 (Task #803). Closes audit risks #1, #2, #6, #10 from `tech-debt-03-errors.md`.

- **Risk #2** — `electron/main.ts`: register `process.on('unhandledRejection')` + `uncaughtException` BEFORE `app.whenReady`. Log only — no auto-exit, preserves test default-throw behavior. TODO comment for Sentry forward (main-proc Sentry transport not yet wired).
- **Risk #1** — `electron/ipc/dbIpc.ts`: try/catch around `saveState()` in `handleDbSave` so sqlite write errors return `{ok:false, error}` instead of crossing the IPC bridge as Electron's opaque "An object could not be cloned" rejection. Renderer's preload `saveState` wrapper already unwraps `{ok:false}` and re-throws so `setPersistErrorHandler` still fires.
- **Risk #6** — same file: extract `handleDbLoad` with try/catch returning `null` on read error. Preserves the existing `string | null` preload contract; logs for diagnostics.
- **Risk #10** — `electron/ipc/systemIpc.ts`: extract `handleModelsList` with try/catch returning `[]` on settings.json parse error.

The optional `safeHandle(channel, fn)` helper from the audit is deliberately deferred — the three handlers here have three different return shapes (`string | null`, `{ok}`, `T[]`) so a single helper would force preload-contract churn for marginal gain. Tracked as a follow-up.

## Reverse-verify (per `feedback_bug_fix_test_workflow.md`)

Each fix has a paired test. Removed each try/catch + the two `process.on` calls; ran the new tests.

| File | Tests removed → expected to FAIL |
|---|---|
| `electron/__tests__/error-safety-net.test.ts` | 3 source-wiring tests fail (no `process.on(...)` matches in live source) |
| `electron/ipc/__tests__/dbIpc.test.ts` | `returns {ok:false} when saveState throws` → uncaught `Error: SQLITE_FULL` |
| `electron/ipc/__tests__/dbIpc.test.ts` | `returns null when loadState throws` → uncaught `Error: SQLITE_CORRUPT` |
| `electron/ipc/__tests__/systemIpc.test.ts` | `returns [] when listModelsFromSettings throws` → uncaught `SyntaxError` |

Restored → all 24 tests PASS. Recorded vitest output:

```
Test Files  3 passed (3)
     Tests  24 passed (24)
```

## Verification

- [x] `npm run build` — clean (3 unrelated webpack bundle-size warnings preexisting)
- [x] `node -e "require('./dist/electron/ipc/dbIpc.js')"` — loads, exports `handleDbLoad` + `handleDbSave`
- [x] `node -e "require('./dist/electron/ipc/systemIpc.js')"` — loads, exports `handleModelsList`
- [x] Full `npm test` — 4 failures pre-exist on origin/working (better-sqlite3 NODE_MODULE_VERSION mismatch in this worktree env + macOS UA test). Verified by checking out `origin/working` HEAD on the same env: same 3 db-hardening failures.

## Test plan

- [ ] Reviewer: confirm `electron/main.ts` registers both listeners before `app.whenReady().then(...)`.
- [ ] Reviewer: confirm preload `saveState` shape is unchanged (still `Promise<void>` that throws on `{ok:false}`).
- [ ] Reviewer: spot-check that no new `safeHandle` import was introduced (deferred per body).
- [ ] CI green.

### DIFF
--- a/electron/__tests__/error-safety-net.test.ts
+++ b/electron/__tests__/error-safety-net.test.ts
@@ -1,0 +1,83 @@
+// Verifies the main-process unhandledRejection / uncaughtException safety
+// net is wired in electron/main.ts. Audit risk #2 (tech-debt-03-errors.md):
+// without these listeners Node 20+ silently exits the main process on any
+// escaped rejection.
+//
+// Two layers of verification:
+//  1) Source-level: grep the actual main.ts file for the two `process.on`
+//     calls. Catches accidental deletion in a refactor.
+//  2) Behavior-level: simulate the listener body (log + don't exit) and
+//     assert the contract (no process.exit call).
+//
+// Reverse-verify: comment out the two `process.on(...)` blocks in main.ts
+// → the source-level test FAILS. Restore → PASSES.
+
+import { describe, it, expect, vi } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const MAIN_TS = path.resolve(__dirname, '..', 'main.ts');
+
+describe('main-process error safety net (source wiring)', () => {
+  const source = fs.readFileSync(MAIN_TS, 'utf8');
+  // Strip line comments before matching so a commented-out reference
+  // doesn't satisfy the test (the contract is about live wiring, not
+  // documentation).
+  const live = source
+    .split('\n')
+    .filter((l) => !l.trim().startsWith('//'))
+    .join('\n');
+
+  it('registers process.on(unhandledRejection) before app.whenReady', () => {
+    const idxOn = live.search(/\bprocess\.on\(\s*['"]unhandledRejection['"]/);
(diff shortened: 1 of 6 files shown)


Jiahui-Gu:
APPROVE — merged (8eced9e). 4 audit risks closed (#1 db:save try/catch, #2 unhandledRejection+uncaughtException net, #6 db:load try/catch, #10 models:list try/catch). Preload contracts preserved (db:load string|null, db:save {ok}|{ok:false,error}, models:list T[]). Reverse-verify real per-test (throwing mocks would propagate as uncaught without try/catch). Load-smoke OK on dist/electron/ipc/{dbIpc,systemIpc}.js. CI 7/7 green.

PR #585, opened 2026-04-29
fix(security): isSafePath sweep + sid pattern + nav-allowlist tighten

Jiahui-Gu:
Tech-debt R8 (Task #804) — security S-cluster, addresses risks #1, #4, #5, #6, #7 from `artifacts/tech-debt-08-security.md`.

## Summary

Five small security gates closing the cheapest exploitability/impact items in the audit. All changes are leaf-level validators — no behavioural drift on the happy path.

| Risk | Where | Gate |
|---|---|---|
| #1 | `electron/ptyHost/cwdResolver.ts` | `isSafePath(cwd)` BEFORE `statSync` (Windows NTLM-leak primitive on UNC) |
| #4 | `electron/ipc/utilityIpc.ts` | `isSafePath(p)` on `app:userCwds:push` (LRU later replays into pty:spawn) |
| #5 | `electron/ipc/sessionIpc.ts` | `safeDir()` scrubs UNC/relative `dir` to undefined for `sessionTitles:{get,rename,enqueuePending}` |
| #6 | `electron/ptyHost/jsonlResolver.ts` | `toClaudeSid` rejects sid not matching `/^[a-zA-Z0-9][a-zA-Z0-9_-]{7,63}$/` (argv injection — required leading alphanumeric so `--dangerous-flag` cannot land as a CLI flag) |
| #7 | `electron/window/createWindow.ts` | drop hardcoded `http://localhost:4100` from `will-navigate` allowlist; only env-driven `CCSM_DEV_PORT` origin honored. Extracted `isAllowedNavigation()` as exported pure decider |

## Reverse-verify

Each gate has a green forward test AND a red reverse cycle (commenting out / replacing the guard makes the new tests fail). All five reverse-verify outputs captured below.

### #1 cwdResolver — gate disabled (`if (false && !isSafePath(...))`)

```
✗ rejects UNC cwd (backslash) without statSync, falls back to homedir
  AssertionError: warn called with "...unusable (ENOENT...)" instead of "rejected by isSafePath"
✗ rejects UNC cwd (forward slash) without statSync
✗ rejects relative cwd without statSync
Tests  3 failed | 5 passed (8)
```

### #4 utilityIpc userCwds:push — gate disabled

```
✗ rejects UNC path (backslash) without calling pushUserCwd
✗ rejects UNC path (forward slash)
✗ rejects relative escape paths (../)
Tests  3 failed | 3 passed (6)
```

### #5 sessionIpc safeDir — gate replaced with `dir.length > 0 ? dir : undefined`

```
✗ scrubs UNC dir (backslash) to undefined on get
✗ scrubs UNC dir (forward slash) to undefined on rename
✗ scrubs relative dir to undefined on enqueuePending
Tests  3 failed | 3 passed (6)
```

### #6 toClaudeSid — VALID_SID_RE.test() replaced with false

```
✗ toClaudeSid > rejects empty string
✗ toClaudeSid > rejects too-short sid
✗ toClaudeSid > rejects too-long sid (>64 chars)
✗ toClaudeSid > rejects sid containing argv-injection characters
Tests  4 failed | 5 passed | 26 skipped (35)
```

### #7 isAllowedNavigation — re-added literal `localhost:4100` to the OR

```
✗ blocks the legacy hardcoded localhost:4100 when env points elsewhere
Tests  1 failed | 6 passed (12)
```

## Test plan

- [x] `npm run build` clean
- [x] Affected unit suites green (security + ipc + ptyHost + window)
- [x] Module load smoke (`node -e "require('./dist/electron/...')"`) green for all 5 modified main-process modules
- [x] Each gate reverse-verified (gate removed → red; gate restored → green)

## Notes

- For risk #1 I kept the existing fallback-to-homedir+warn contract on rejection rather than throwing. The function is called from the spawn lifecycle path; throwing would surface as a hard spawn failure to the user. Drop-with-warn matches the existing rejection pattern for non-existent / file paths and also matches `probePaths`'s drop-with-false pattern for unsafe paths. Security goal (no statSync of unsafe path) is preserved by the early return.
- For risk #6 I tightened the sid pattern to require a leading alphanumeric (not just any `[a-zA-Z0-9_-]`). Without this constraint `--dangerous-flag` would have matched `[a-zA-Z0-9_-]{8,64}` and been accepted. The renderer's `crypto.randomUUID()` always produces a hex-leading sid; the legacy `<prefix>-<random>` fallback uses an alphanumeric prefix. Real sids never start with `-`.
- For risk #5 the artifact mentions a "path is under `~/.claude/projects/`" extra check, but the `dir` arg here is actually a session **cwd** (e.g. `C:\Users\me\code`) that the SDK realpath-encodes into a project key. `isSafePath` is the right gate; an additional `path.relative(claudeProjectsDir, ...)` check would reject every legitimate caller. `listForProject` takes a `projectKey` (not a `dir`), so it's not in scope for this gate.

### DIFF
--- a/electron/ipc/__tests__/sessionIpcSecurity.test.ts
+++ b/electron/ipc/__tests__/sessionIpcSecurity.test.ts
@@ -1,0 +1,94 @@
+// Security gates for sessionTitles IPC handlers (#804 risk #5).
+//
+// `sessionTitles:get` / `:rename` / `:enqueuePending` accept a `dir`
+// argument that is forwarded straight into the Anthropic SDK, which fs.joins
+// it under ~/.claude/projects/<key>/.... Without the safety gate a hostile
+// renderer can pass a UNC path (`\\evil\share`) — once the SDK realpath()s
+// or stat()s it, Windows triggers an SMB handshake that leaks the user's
+// NTLM hash. The gate here scrubs the dir to `undefined` (which makes the
+// SDK fall back to scanning every project dir, the legacy safe path) when
+// the input fails `isSafePath`.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const getSessionTitleMock = vi.fn(async () => ({ summary: null, mtime: null }));
+const renameSessionTitleMock = vi.fn(async () => ({ ok: true }));
+const enqueuePendingRenameMock = vi.fn();
+
+vi.mock('../../sessionTitles', () => ({
+  getSessionTitle: (...a: unknown[]) => getSessionTitleMock(...(a as [])),
+  renameSessionTitle: (...a: unknown[]) => renameSessionTitleMock(...(a as [])),
+  listProjectSummaries: vi.fn(async () => []),
+  enqueuePendingRename: (...a: unknown[]) => enqueuePendingRenameMock(...(a as [])),
+  flushPendingRename: vi.fn(async () => undefined),
+}));
+
+import { registerSessionIpc } from '../sessionIpc';
+
+type Handler = (e: unknown, ...args: unknown[]) => unknown;
+
+function fakeIpcMain() {
+  const handlers = new Map<string, Handler>();
+  const ipcMain = {
+    handle: (ch: string, fn: Handler) => handlers.set(ch, fn),
+    on: (ch: string, fn: Handler) => handlers.set(ch, fn),
+  } as unknown as Electron.IpcMain;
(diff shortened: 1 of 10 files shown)


Jiahui-Gu:
APPROVE — merged. 5 security gates land with reverse-verify; UNC NTLM primitive closed; sid argv injection closed; nav allowlist tightened.

PR #595, opened 2026-04-30
chore(deps): bump vitest 2.1.8 → 4.1.5

Jiahui-Gu:
## Summary

Bumps `vitest` and `@vitest/coverage-v8` from `^2.1.8` / `^2.1.9` to `^4.1.5` (across two majors: v2 → v3 → v4).

### v2 → v3
- Default test pool changed from `threads` to `forks`. We don't pin a pool, so this is transparent — no flakes observed in our suite.
- `vi.fn()` factories invoked with `new` now require the implementation to be a `function` expression or `class`, not an arrow function. tinyspy throws `TypeError: ... is not a constructor` otherwise, accompanied by an explicit warning: *"The vi.fn() mock did not use 'function' or 'class' in its implementation"*.

### v3 → v4
- No additional source-level breaks hit our codebase. Our `vitest.config.ts` already uses the modern `coverage.thresholds` shape (not the legacy `coverage.lines/functions/...` flat keys), so no config rewrite was needed. No workspace config in use.

## Breaking changes encountered + fixes

| File | Symptom | Fix |
|------|---------|-----|
| `tests/terminal/useXtermSingleton.test.tsx` | `vi.fn(() => ({...}))` impls for `Terminal`, `FitAddon`, `WebLinksAddon`, `ClipboardAddon`, `Unicode11Addon`, `CanvasAddon` failed with `is not a constructor` when `xtermSingleton.ts` did `new Terminal(...)` / `new FitAddon()` / etc. | Replaced arrow impls with `function () { return {...}; }` expressions so they're valid constructors under tinyspy v4. |

No production source touched.

## Verify

```
npm test -- --run
# Test Files  127 passed (127)
#      Tests  981 passed (981)   ← unchanged from baseline (981)
```

`npm run typecheck` clean. `npm run lint` clean (4 pre-existing warnings, unrelated).

## E2E

`node scripts/harness-ui.mjs --only=sidebar-align` is currently failing on origin/working baseline as well (electron main exits before window opens — unrelated to this dep bump; reproduced before any change in this PR). Not a regression introduced here. CI rollup will exercise the full e2e path.

Closes #831


Jiahui-Gu (commented):
LGTM (cannot self-approve — author identity collision). Migration clean: only vitest + @vitest/coverage-v8 bumped; no production source touched; vi.fn function-form fix is canonical tinyspy v4 mitigation; no skips added. Lockfile: all @vitest/* cohesive at 4.1.5, chai v6, no orphan v2. All 7 CI checks green.

PR #596, opened 2026-04-30
test(sessionWatcher): backfill UT (Task #832 W2)

Jiahui-Gu:
## Summary
Backfills direct unit tests for `electron/sessionWatcher/**` modules that previously only had end-to-end coverage via `titleChanged.test.ts`. SRP-aligned: each new test file targets one producer/decider/sink module.

Files added (all under `electron/sessionWatcher/__tests__/`):

- `fileSource.test.ts` (15 tests) - producer behavior. Real fs + tmp dir. Covers immediate first tick, ancestor watcher path (parent dir missing), file-appears upgrade, append debounce, idempotent duplicate `start`, `stop`/`stopAll` lifecycle, late tick suppression after stop, tail-read for files >256 KB, throwing tick handler tolerance, `getCwd` test seam.
- `stateEmitter.test.ts` (7 tests) - sink wiring through `classifyJsonlText` + `decideStateEmit`. Per-sid `lastEmitted` tracking, dedupe semantics, `forget()` re-arms emit, malformed JSONL tolerance.
- `titleEmitter.test.ts` (10 tests) - sink with injected `TitleFetcher`. Skip on `fileExists=false`, null/empty summary gating, `lastEmittedTitle` dedupe, `closed` guard against late-resolving fetches, `setFetcher` swap, cwd pass-through.
- `pendingRenameFlusher.test.ts` (8 tests) - edge-trigger semantics. One flush per first-appear, per-sid independence, `forget` re-arm, sync + async callback error tolerance, `setFlush` swap.
- `facade.test.ts` (5 tests) - `index.ts` facade. `closeAll` fan-out of `unwatched` events, isolated `__createForTest` instances, `configureSessionWatcher` singleton wiring, `stopWatching` no-op for unknown sids.

Total: **45 new tests across 5 files**. Combined with the 5 pre-existing files (emitDecider, inference, projectKey, titleChanged, unwatched), `electron/sessionWatcher/__tests__/` now has 83 tests passing.

No product code changes. No new dependencies. No conflicts with W1 (notify) - this PR only touches `electron/sessionWatcher/__tests__/`.

## Test plan
- [x] `npx vitest run electron/sessionWatcher/__tests__/` - 10 files / 83 tests PASS
- [x] `npm test -- --run` - 996 PASS, 3 fail. The 3 failures are pre-existing `better-sqlite3` NODE_MODULE_VERSION env mismatch in `electron/__tests__/db-hardening.test.ts`, verified to also fail on `origin/working` baseline (untouched by this PR).
- [x] Vitest v4 syntax: `vi.fn(function () { return ... })` for stubs that need a `this` binding context, no arrow-function mocks.

### DIFF
--- a/electron/sessionWatcher/__tests__/facade.test.ts
+++ b/electron/sessionWatcher/__tests__/facade.test.ts
@@ -1,0 +1,120 @@
+// Unit tests for the SessionWatcher facade (index.ts).
+//
+// These tests are NOT a duplicate of titleChanged.test.ts (which exercises
+// the full producer + sinks via the title bridge). Here we verify the
+// facade's specific responsibilities:
+//   * `configureSessionWatcher` rewires the singleton's title fetcher and
+//     pending-rename flusher (boot wiring contract).
+//   * `closeAll` fires `unwatched` for every tracked sid.
+//   * The `__createForTest` factory returns an isolated instance (no
+//     module-level cross-talk between watchers).
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import {
+  __createForTest,
+  configureSessionWatcher,
+  sessionWatcher,
+  type UnwatchedEvent,
+  type TitleChangedEvent,
+} from '../index';
+
+let tmpRoot: string | null = null;
+
+function mkTmp(): string {
+  tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'sw-facade-test-'));
+  return tmpRoot;
+}
+
+afterEach(() => {
+  if (tmpRoot && fs.existsSync(tmpRoot)) {
+    fs.rmSync(tmpRoot, { recursive: true, force: true });
+  }
+  tmpRoot = null;
+  // Reset the module singleton's wired callbacks back to noops so we don't
+  // pollute neighbouring suites that import `sessionWatcher`.
+  configureSessionWatcher({
+    fetchTitle: async () => ({ summary: null }),
+    flushRename: () => undefined,
+  });
+});
+
+async function waitFor(pred: () => boolean, timeoutMs = 1500): Promise<void> {
+  const start = Date.now();
+  while (Date.now() - start < timeoutMs) {
+    if (pred()) return;
+    await new Promise((r) => setTimeout(r, 25));
+  }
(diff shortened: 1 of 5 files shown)


Jiahui-Gu:
REQUEST_CHANGES: 1 real test failure on macOS lint+typecheck+test (CI run 25145887871).

## Failure

`electron/sessionWatcher/__tests__/fileSource.test.ts:122` — "start with same sid + same path is idempotent (does not restart)":

```
AssertionError: expected 2 to be 1
  120|     src.start('s1', jsonlPath); // duplicate — no new immediate read
  121|     await new Promise((r) => setTimeout(r, 100));
  122|     expect(ticks.length).toBe(before);
```

## Why this is the test, not the product

`FileSource.start()` does `if (existing.jsonlPath === jsonlPath) return;` so the duplicate call is a true no-op (no extra `scheduleRead`, no new watcher install). The 2nd tick on macOS comes from the *file watcher* installed by the first `start`, picking up an fs event during the 100ms wait. The product behavior (idempotent start) is correct; the test assertion is too strict for real fs.

This passed on Ubuntu but flaked on macOS. Trust-CI mode requires CI green before merge.

## Suggested fix (pick one)

1. **Drop the 100ms wait** — `start()` is sync, the no-op effect is observable immediately:
   ```ts
   await waitFor(() => ticks.length >= 1);
   const before = ticks.length;
   src.start('s1', jsonlPath); // sync no-op
   expect(ticks.length).toBe(before); // no NEW sync work
   ```
2. **Spy on `scheduleRead`** via a test seam (adds product surface, less clean).

Option 1 matches the actual contract being asserted. Pure fs-watch noise during a sleep window is not what this test is about.

## Other notes (non-blocking, otherwise APPROVE)

- Test quality is good: real fs / real watchers / behavior-focused.
- All 5 modules genuinely benefited from direct UT — prior coverage was only end-to-end via `titleChanged.test.ts`.
- vitest 4 syntax compliant: `vi.fn(function () { ... })` used where `this`-binding matters; short async-arrow stubs are fine.
- No mocks where real fs works (per `feedback_dont_mock_db.md` spirit).
- No skipped tests, no SCREAMING strings, no product code touched.
- Pre-existing 3 better-sqlite3 failures NOT observed on this CI run (db-hardening passed on macOS) — worker's local-env claim was correct that those are env-dependent, not a blocker here.

Once the macOS flake is fixed and CI is green, this PR is good to merge.

PR #608, opened 2026-04-30
feat(terminal): L4 PR-E detach/reattach via headless buffer (#864)

Jiahui-Gu:
## Scope

L4 dual-buffer mirror PR-E. Pin and exercise the contract that a visible-xterm detach/reattach cycle replays missed PTY data from the headless authoritative buffer (PR-A #861) via the seq-tagged snapshot (PR-B #865), with **no PTY re-emission**. Builds on top of:

- PR-A: scrollback bumped to 10000, headless = source of truth.
- PR-B: `getBufferSnapshot` returns `{snapshot, seq}` captured atomically; renderer attach uses snapshot+drain to dedupe live chunks against snapshot.
- PR-C: explicit `dispatchPtyChunk` + observe-only backpressure warn.

## Files changed

- `electron/ptyHost/entryFactory.ts` — suppress backpressure warn when `entry.attached.size === 0` (no visible xterm = no human consumer; background-session log noise). Warn re-arms once a renderer reattaches. Added PR-E doc paragraph noting that no new code path is needed in the dispatch — PR-A's headless-write + PR-B's snapshot already make detach/reattach free.
- `src/terminal/usePtyAttach.ts` — append `void ptyApi.detach(sessionId)` to the existing cleanup function so React unmount / sid switch performs an explicit detach. Without this the entry's `attached` Map kept the wc reference until the whole renderer was destroyed (only `wc.once('destroyed')` cleaned it up). Stale entries were tolerated by `dispatchPtyChunk` (isDestroyed check) but the new size-gated suppression needs honest attach state. **Touched minimally** to avoid PR-D resize-wiring collision (one append to the existing return cleanup; no resize code touched).
- `electron/ptyHost/__tests__/detachReattach.test.ts` — new file, 5 cases.
- `electron/ptyHost/__tests__/entryFactory.test.ts` — existing PR-C backpressure case updated to attach a fake wc (warn now gated on `size > 0`).

## UT output (vitest)

`npx vitest run electron/ptyHost`

```
Test Files  13 passed (13)
     Tests  178 passed (178)
  Duration  5.25s
```

New file `detachReattach.test.ts` (5/5):

```
Test Files  1 passed (1)
     Tests  5 passed (5)
  Duration  5.15s
```

Cases:
1. detached: PTY chunks accumulate in headless, no visible-xterm writes (no attached wc).
2. reattach: snapshot replays the entire detached backlog from headless (no PTY re-emit).
3. multi-cycle detach/reattach/detach/reattach: no data loss, monotonic seq (1→3→6→9→10), headless buffer is the SoT.
4. snapshot replay round-trips byte-for-byte through a new visible xterm (drain seam holds — exercises the PR-B race window: live chunk arriving DURING the snapshot await is buffered by the listener and replayed only if `seq > snapSeq`).
5. backpressure warn is SUPPRESSED while detached, RESUMES once a visible xterm reattaches.

## Reverse-verify

Sabotage: in `lifecycle.getBufferSnapshot`, return `{snapshot:'', seq:0}` unconditionally (drop the replay path).

```
FAIL > reattach: snapshot replays the entire detached backlog from headless (no PTY re-emit)
  AssertionError: expected '' to be 'hello world\nline-2\n'

FAIL > multi-cycle detach/reattach/detach/reattach: ... headless buffer is the SoT
  AssertionError: expected '' to be 'A1A2'

FAIL > snapshot replay round-trips byte-for-byte through a new visible xterm (drain seam holds)
  AssertionError: expected 'live-duringpost-attach' to be 'past-1past-2live-duringpost-attach'

Test Files  1 failed (1)
     Tests  3 failed | 2 passed (5)
```

Restored sabotage → 5/5 PASS again. The two pure-detach + backpressure cases stayed green under sabotage (they don't go through `getBufferSnapshot`), proving the three replay tests were the ones actually pinning the contract.

## Out of scope

- IPC roundtrip / Electron probe (covered by L4 dogfood loop separately).
- Multi-window fanout (single mainWindow today; the wiring is N-safe per PR-A doc).
- PR-D resize wiring is parallel; this PR touches `usePtyAttach.ts` only in the cleanup return (no overlap with resize).

## Risk

Low. The behavior change is one new IPC `pty:detach` call on unmount (was silently leaked-then-tolerated before) and one extra condition on the warn (suppression). No data path change; PTY → headless sink unchanged.

### DIFF
--- a/electron/ptyHost/__tests__/detachReattach.test.ts
+++ b/electron/ptyHost/__tests__/detachReattach.test.ts
@@ -1,0 +1,395 @@
+// L4 PR-E (#864): pins the detach/reattach contract — a visible-xterm
+// reattach must replay every PTY chunk that arrived while detached, fed
+// from the headless authoritative buffer (PR-A) via getBufferSnapshot
+// (PR-B), with NO request to the PTY for re-emission.
+//
+// What this proves end-to-end (main-process side):
+//   1. While detached (`entry.attached.size === 0`), `dispatchPtyChunk`
+//      keeps writing every chunk to the headless mirror — backlog
+//      preserved.
+//   2. While detached, the backpressure warn is suppressed (no human
+//      consumer; would be log noise on long-running background sessions).
+//      Re-attach restores normal threshold semantics.
+//   3. Re-attach replays via `serialize.serialize()` snapshot (this is
+//      the same path lifecycle.getBufferSnapshot uses) — captured atomically
+//      with `entry.seq`. Live chunks delivered AFTER the snapshot can be
+//      deduped by the renderer using `seq > snapSeq`.
+//   4. Multiple detach/reattach/detach/reattach cycles preserve a
+//      monotonic seq — no wrap, no decrement, no double-fanout.
+//   5. The PTY is never asked to re-emit; the only `pty.write` calls
+//      come from explicit user input (zero in this test).
+//
+// Out of scope (other PRs / dogfood): real Electron IPC roundtrip,
+// renderer-side xterm.write, multi-window fanout collisions.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+interface PtyFakeBus {
+  onData: ((chunk: string) => void) | null;
+  onExit: ((evt: { exitCode: number | null; signal: number | null }) => void) | null;
+  ptySpawn: ReturnType<typeof vi.fn>;
+  ptyWrite: ReturnType<typeof vi.fn>;
+  headlessDispose: ReturnType<typeof vi.fn>;
+  // We model the headless buffer as a plain string accumulator so the
+  // test can compare "what was written into the buffer" to "what a
+  // reattach snapshot replays" — full L4 contract.
+  headlessBuffer: string;
+  watcherStart: ReturnType<typeof vi.fn>;
+  watcherStop: ReturnType<typeof vi.fn>;
+  ensureJsonl: ReturnType<typeof vi.fn>;
+  emitData: ReturnType<typeof vi.fn>;
+  sourceJsonl: string | null;
+  ensureCopied: boolean;
+  // When true, headless.write defers its callback so tests can simulate
+  // a slow main-thread / scrollback flush and check pending counters.
+  deferHeadlessWrite: boolean;
+  pendingCallbacks: Array<() => void>;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function bus(): PtyFakeBus { return (globalThis as any).__pf as PtyFakeBus; }
+
+vi.mock('node-pty', () => ({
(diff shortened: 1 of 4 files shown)


Jiahui-Gu (commented):
## Review verdict: APPROVE

- Scope is correctly minimal. Production diff is exactly two surgical changes:
  - `entryFactory.ts`: add `entry.attached.size > 0` gate on the PR-C backpressure warn (legit — no human consumer = no point spamming `console.warn` for background sessions).
  - `usePtyAttach.ts`: append `void ptyApi.detach(sessionId)` to existing React cleanup (symmetric with `pty.attach`, keeps `entry.attached.size` honest so the new size-gate above behaves).
- The "detach/reattach via headless buffer" contract is in fact already implemented by PR-A (headless always receives) + PR-B (snapshot+drain). PR-E correctly adds **no new dispatch path** — it just pins the contract with tests + makes the attached-set state honest.
- Worker's claim ("almost no production code") matches reality: 4 new prod LOC + 1 new condition + 1 PR-doc paragraph.
- Background-session unbounded buffering is naturally capped by `SCROLLBACK = 10000` lines (PR-A), so no missing cap.
- Nothing meaningful left out.

### Layer 2 — implementation
- `entry.attached.size > 0` gate: correct. Re-arm semantics in the write callback are unchanged, so a long detached stall correctly resets `backpressureWarned` once writes drain, and a future attach starts from a clean edge.
- `usePtyAttach.ts` cleanup: fire-and-forget `void ptyApi.detach(sessionId)` is the right pattern (React cleanup is sync; detach is idempotent on main per PR-A semantics). Best-effort try/catch correctly swallows main-tearing-down errors. Doesn't collide with PR-D resize wiring (only the cleanup return is touched).
- 5 tests cover: detached accumulation, single reattach via `getBufferSnapshot`, multi-cycle with monotonic seq (1→3→6→9→10), the PR-B drain-seam race window (live chunk arriving DURING snapshot await), and backpressure suppress→resume.
- `entryFactory.test.ts` patch correctly adapts the existing PR-C test (the warn now requires `attached.size > 0`).
- Reverse-verify in PR body is real: sabotaging `lifecycle.getBufferSnapshot` to return `{snapshot:'', seq:0}` makes exactly the 3 replay tests fail with the expected diffs; restoring → 5/5 pass. The 2 non-replay tests stayed green under sabotage, which correctly proves they pin a different sub-contract.
- 178/178 vitest pass under `electron/ptyHost`.

### CI
- `no-silent-drops`: PASS
- `lint + typecheck + test` (ubuntu/macos/windows): IN_PROGRESS
- `e2e` (ubuntu/macos/windows): IN_PROGRESS
- Per trust-CI mode, no local e2e probe run.

### Recommendation
APPROVE. Manager can merge once CI rollup is green; no fixes required.

PR #612, opened 2026-04-30
chore(sessionTitles): add forgetSid to release per-sid memory (#881)

Jiahui-Gu:
## Summary

Tech-debt H1 from audit #876. `electron/sessionTitles/index.ts` holds three per-sid Maps (`titleCache` line 105, `opChains` line 111, `pendingRenames` line 128) that grow without bound for the lifetime of the app — every sid ever queried sticks forever, and `pendingRenames` is a dead-letter queue (only flushed by watcher events, never on session exit).

This PR adds `forgetSid(sid)` and wires it into the existing `sessionWatcher.on('unwatched', ...)` listener in `electron/notify/bootstrap/installPipeline.ts`, so cleanup rides the existing PTY-exit teardown path with no new subscription or lifecycle.

## Changes

| File | LOC | What |
|---|---|---|
| `electron/sessionTitles/index.ts` | +40 | Export `forgetSid(sid)` (deletes from all three Maps; idempotent; tolerant of empty/non-string sid) + `__hasSidStateForTests(sid)` test seam |
| `electron/notify/bootstrap/installPipeline.ts` | +5/-1 | Import `forgetSid as forgetSessionTitleSid` and call it from the existing `'unwatched'` listener |
| `electron/sessionTitles/__tests__/forgetSid.test.ts` | +127 (new) | 6 UT cases |

## In-flight rename safety

The serialization chain in `chain()` holds the `next` promise via a local variable; `opChains.delete(sid)` only drops the *next* op's predecessor link. Any pending SDK call already awaiting `loadSdk()` / `renameSession()` runs to completion and resolves its caller's Promise normally. Test case "lets a pending rename resolve normally even when called mid-flight" asserts this.

## UT output (all green)

```
RUN  v4.1.5 C:/Users/jiahuigu/ccsm-worktrees/pool-9

 Test Files  1 passed (1)
      Tests  6 passed (6)
   Duration  2.93s
```

Surrounding suites (`electron/sessionTitles` + `electron/notify/bootstrap`):

```
 Test Files  5 passed (5)
      Tests  46 passed (46)
```

`npm run typecheck` clean (both renderer + electron tsconfig).

## Reverse-verify

Removed `titleCache.delete(sid)` from `forgetSid`:

```
 FAIL  electron/sessionTitles/__tests__/forgetSid.test.ts > forgetSid releases all three per-sid Maps > does not affect entries for other sids
 ❯ electron/sessionTitles/__tests__/forgetSid.test.ts:70:44
     70|     expect(__hasSidStateForTests('sid-a')).toBe(false);
       |                                            ^
 Tests  1 failed | 5 passed (6)
```

Restored → 6/6 PASS.

## Refs

- Audit: `~/.claude/projects/C--Users-jiahuigu/memory/project_implicit_state_audit_876.md` (#876)
- Tracker: Task #881

### DIFF
--- a/electron/notify/bootstrap/installPipeline.ts
+++ b/electron/notify/bootstrap/installPipeline.ts
@@ -16,6 +16,7 @@ import { app, BrowserWindow } from 'electron';
 import { installNotifyPipeline } from '../sinks/pipeline';
 import { onPtyData } from '../../ptyHost';
 import { sessionWatcher } from '../../sessionWatcher';
+import { forgetSid as forgetSessionTitleSid } from '../../sessionTitles';
 
 export type NotifyPipeline = ReturnType<typeof installNotifyPipeline>;
 
@@ -65,10 +66,13 @@ export function installNotifyPipelineWithProducers(
 
   // Drop sniffer/ctx state when a session is unwatched (PTY exit). The
   // existing 'unwatched' emitter is reused so we don't add another teardown
-  // path.
+  // path. Also release the per-sid Maps held by sessionTitles (titleCache /
+  // opChains / pendingRenames) — without this, every sid ever queried for a
+  // title sticks in memory for the lifetime of the app (audit #876, H1).
   sessionWatcher.on('unwatched', (evt: { sid?: unknown }) => {
     if (!evt || typeof evt.sid !== 'string' || evt.sid.length === 0) return;
     pipelineInstance.forgetSid(evt.sid);
+    forgetSessionTitleSid(evt.sid);
   });
 
   return pipelineInstance;
--- a/electron/sessionTitles/__tests__/forgetSid.test.ts
+++ b/electron/sessionTitles/__tests__/forgetSid.test.ts
@@ -1,0 +1,127 @@
+// Unit tests for `forgetSid` — the per-sid Map releaser added for tech-debt
+// H1 (audit #876, task #881). Verifies that all three internal Maps
+// (titleCache / opChains / pendingRenames) drop their entry for a sid, that
+// in-flight renames don't throw on cleanup, and that repeat calls are
+// idempotent. SDK is fully mocked via the same `__setSdkForTests` seam used
+// by the rest of the bridge suite.
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+const sdkMocks = {
+  getSessionInfo: vi.fn(),
+  renameSession: vi.fn(),
+  listSessions: vi.fn(),
+};
+
+import {
+  getSessionTitle,
+  renameSessionTitle,
+  enqueuePendingRename,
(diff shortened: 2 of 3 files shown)


Jiahui-Gu (commented):
## Reviewer verdict: APPROVE (manager to merge — same-identity self-approve blocked)

### Layer 1 — direction & coverage
- **Right shape.** Single `forgetSid(sid)` over per-Map cleanup is correct — the three Maps (`titleCache`, `opChains`, `pendingRenames`) share the same lifecycle (per-sid teardown), so a single function avoids drift where a future fourth Map gets added but only two of three call sites updated.
- **Wiring point is canonical.** `sessionWatcher.on('unwatched', ...)` in `installPipeline.ts` is the established sid-teardown signal (already drives `pipelineInstance.forgetSid` for the notify side). Reusing it costs zero new subscriptions and zero new lifecycle.
- **All sid-removal paths verified covered.** Audited every entry to `sessionWatcher.stopWatching` (the sole `unwatched` emitter):
  - PTY exit → `entryFactory.ts:297` → stopWatching
  - User delete (`SessionRow` ConfirmDialog → store `deleteSession` → `window.ccsmPty.kill` → `lifecycle.ts:136`) → stopWatching
  - `closeAll` (app shutdown) → stopWatching
  No archive path or logout path exists today that bypasses PTY teardown. **Fully resolves #881; no follow-up needed.**

### Layer 2 — implementation
- All 3 Maps cleared in one call, atomically (sync).
- Idempotent: `Map.delete` of an absent key is a no-op; tested explicitly (test "does not throw when called twice").
- Tolerant of bad input (empty string / non-string); test covers `''`, `undefined`, `null`, `123` and asserts other sids stay intact.
- **In-flight rename safety is correct and tested.** The `chain()` closure captures `next` by local variable; `opChains.delete(sid)` only orphans the *successor* link. The test "lets a pending rename resolve normally even when called mid-flight" asserts the original caller's promise still resolves to `{ ok: true }`. This is the right tradeoff — aborting mid-flight SDK fs writes would risk corrupting the JSONL and can't be safely cancelled anyway.
- Test seam (`__hasSidStateForTests`) is the right granularity — asserts cleanup without leaking the Maps.
- Reverse-verify in PR body: removing `titleCache.delete` causes the "other sids unaffected" test to FAIL → restored → 6/6 PASS. Genuine.

### Local verification
- `npx vitest run electron/sessionTitles electron/notify/bootstrap` → **5 files, 46 tests pass** (3.11s).
- CI rollup: all 7 checks green (e2e × 3 OS, lint+typecheck+test × 3 OS, no-silent-drops).
- Diff is small, focused, well-documented (40 LOC prod + 127 LOC tests).

### Verdict
APPROVE. Manager safe to squash-merge.

PR #614, opened 2026-04-30
docs(notify): explicit decision matrix + DEV-only invariants (#885)

Jiahui-Gu:
## Scope

Tech-debt M3 from audit #876 cluster 2.2. **Comment-only + assertion changes; zero logic change.**

## Why

The notify subsystem decision uses 5 dimensions (`focused` x `activeSid===sid` x `hasObservedRunning` x `muted` x `globalMuted`) but the truth table was spread across three files (`notifyDecider.ts`, `runStateTracker.ts`, `sinks/pipeline.ts`). A future reader (or future me) could not see the full picture from any single entry point. Additionally, `pipeline.markUserInput()` has a hidden side effect: it sets a **permanent** mute (`untilTs = +Infinity`), which the method name does not advertise.

## What

1. **`notifyDecider.decide` doc block**: prepend an explicit 5x5 truth table covering all gates including `hasObservedRunning` (tracker layer) and `globalMuted` (pipeline layer), plus an explicit callout for the `markUserInput` permanent-mute side effect.
2. **`notifyDecider.assertNotifyContext`**: new DEV-only invariant function. Validates `Ctx` shape (focused boolean, activeSid string|null, now finite number, the four collections are real Map/Set instances). Called from `decide()` gated on `process.env.NODE_ENV !== 'production'`. Catches the wiring-bug class that originally produced #767 (a stray `undefined`/number slipping through silently mis-fires Rule 2's `elapsed = 0 < 60s` and lights flashSink for every fresh sid).
3. **`sinks/pipeline.ts` markUserInput**: amend existing comment to flag the permanent-mute semantic with explicit audit #876 tag, so any future rename/refactor surfaces the mute behavior in the new name.

## Decision matrix excerpt

```
| focused | active | hasRun | muted | global | result        | rule        |
|---------|--------|--------|-------|--------|---------------|-------------|
| *       | *      | *      | *     | T      | null          | global gate |
| *       | *      | F      | *     | F      | null          | #767 gate   |
| *       | *      | T      | *     | F      | null (R1)     | user-init   |
| F       | *      | T      | F     | F      | toast + flash | R5          |
| F       | *      | T      | T     | F      | flash only    | R7 (muted)  |
| T       | F      | T      | F     | F      | toast + flash | R4          |
| T       | T      | T      | F     | F      | flash only    | R2 (<60s)   |
| T       | T      | T      | F     | F      | toast + flash | R3 (>=60s)  |
```

## Tests

- 9 new cases for `assertNotifyContext`: valid ctx passes (2), each invalid field throws (6), `decide()` integration confirms the NODE_ENV gate fires (1).
- All 26 `notifyDecider.test.ts` cases PASS. Full notify suite: 71/71 PASS.
- Typecheck clean.

## Reverse-verify

Disabling the assertion gate (`if (false && process.env.NODE_ENV !== 'production')`) caused the integration case to FAIL exactly as expected:

```
FAIL  electron/notify/__tests__/notifyDecider.test.ts > assertNotifyContext - DEV-only invariants (audit #876) > decide() invokes the invariant: passing a number activeSid throws via decide
AssertionError: expected [Function] to throw an error
- Expected: null
+ Received: undefined
```

Restored gate -> 26/26 PASS.

## Out of scope

- No behavior change. Decider rule eval is byte-identical.
- No new state. No new IPC. No renderer change.
- Pre-existing failures in `tests/electron-load-smoke.test.ts` (18) and `electron/__tests__/db-hardening.test.ts` (3) verified on origin/working tip baseline; unrelated.

## Refs

- Audit: `~/.claude/projects/.../memory/project_implicit_state_audit_876.md` cluster 2.2
- Related: #767 (the bug class the invariant catches), #721 (tracker extraction), #743 (pipeline thinning)

### DIFF
--- a/electron/notify/__tests__/notifyDecider.test.ts
+++ b/electron/notify/__tests__/notifyDecider.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import {
   decide,
+  assertNotifyContext,
   USER_INIT_MUTE_MS,
   SHORT_TASK_MS,
   DEDUPE_MS,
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
## Reviewer verdict: APPROVE (docs + DEV invariant, zero behavior change)

### Layer 1 — necessity / shape
- **Necessity**: confirmed. The 5-dim decision was genuinely spread across notifyDecider / runStateTracker / pipeline. New doc block is the first single-source-of-truth view; the markUserInput permanent-mute callout is real (verified pipeline.ts line 185 `tracker.setMuted(sid, Number.POSITIVE_INFINITY)`).
- **Doc vs refactor**: docs-only is the right call here. Codifying as a lookup table would force materializing 32 rows (or a sparse map), which is harder to read than the short-circuit chain in evalRules. Not worth the refactor.
- **DEV invariant shape**: correct — gated on `NODE_ENV !== 'production'`, validates exactly the 7 fields decider reads (focused/activeSid/now/4 collections), throws sync. No allocation in prod path. Catches the #767-class wiring bug at the seam.

### Layer 2 — accuracy / coverage / safety
- **Doc accuracy**: cross-checked against runStateTracker.ts (hasObservedRunning set/clear at lines 85/103/111/148) and pipeline.ts (isGlobalMutedFn short-circuit at line 126). Matrix matches code behavior.
- **Coverage**: 5 dims = 32 cases nominally, but global-gate (T) and !hasRun collapse 24 of those via wildcards in the doc. The remaining 8 reachable combos are all enumerated. Unreachable ones implicitly covered by `*`.
  - Minor nit (non-blocking): R2 and R3 rows share identical key columns (T T T F F) — they differ on a 6th dim (elapsed < / >= 60s) that isn't tabulated. Could note "(elapsed)" in a column header or split with an extra column. Not worth churn.
- **Prod safety**: NODE_ENV gate is the only call site. assertNotifyContext is exported for direct test use only. Zero perf impact in prod (gate is a single string compare per `decide` call).
- **Test coverage**: 9 new cases — happy path (2), each invalid field (6), decide()-integration with NODE_ENV temporarily set to 'test' (1). All 26 cases PASS locally on this machine. Reverse-verify in PR body is concrete (disabled gate → integration FAIL with expected error).

### CI / scope
- Trust-CI mode applies. Diff is comment-only on existing logic + one new pure function + tests. No e2e probe needed (pure decider, covered by unit tests).
- Scope discipline: 3 files, no logic change confirmed by reading evalRules end-to-end.

Verdict: APPROVE. Manager OK to squash-merge once CI is green.

PR #616, opened 2026-04-30
feat(build): dual-install dev/prod variants (#891)

Jiahui-Gu:
## Summary
Enable a co-installable **CCSM Dev** build that does not collide with the prod **CCSM** install on the same Windows machine. Implements task #891.

- AUMID derived from `app.getName()` (dev = `com.ccsm.app.dev`, prod = `com.ccsm.app`) so taskbar + toast attribution stays independent.
- electron-updater `allowPrerelease` flipped to `true` for the dev variant so we can ship release candidates / dogfood builds via GitHub pre-releases without touching prod users.
- New `make:win:dev` npm script uses electron-builder CLI overrides only — no second `electron-builder.yml`, no fork of the `build` block in `package.json`. Produces `release/CCSM Dev-Setup-<ver>-x64.exe`.
- `scripts/setup-aumid.ps1` gains a `-Variant {prod|dev}` parameter (default `prod`) so the dev AUMID can be registered for toast routing during ad-hoc dev runs.

`%APPDATA%` isolation comes for free: Electron derives `userData` from `appId`, so prod (`com.ccsm.app`) and dev (`com.ccsm.app.dev`) each get their own `%APPDATA%\CCSM*` folder and their own single-instance lock — verified by reading `electron/lifecycle/singleInstance.ts` (uses `app.requestSingleInstanceLock()` which is keyed on userData). No code change needed there.

`~/.claude` is shared by design (decision #2 — no isolation).

## Files changed
| File | +/- |
|---|---|
| `electron/main.ts` | +6 / −1 |
| `electron/updater.ts` | +8 / −0 |
| `package.json` | +1 / −0 |
| `scripts/setup-aumid.ps1` | +23 / −4 |
| `electron/__tests__/updater.test.ts` | +26 / −1 |
| **Total** | **+64 / −6** |

## Decisions inherited from manager
1. Dev productName = `"CCSM Dev"` (with space) — used as is.
2. `CLAUDE_CONFIG_DIR` is **shared** between prod and dev — no code change; default behavior already shares.
3. Auto-updater in dev must work end-to-end → `autoUpdater.allowPrerelease = true` for the dev variant. No early-return; same publish target (`Jiahui-Gu/ccsm` GitHub releases). Prod skips pre-releases by default.
4. Single `package.json` build block + CLI overrides + new `make:win:dev`. **No** new `electron-builder.dev.yml` file.

## Test plan
- [x] `npm run typecheck` — passes (both `tsc --noEmit` invocations).
- [x] `npx vitest run electron/__tests__/updater.test.ts` — 15/15 pass (13 original + 2 new for the prerelease branch).
- [x] `npm run make:win:dev` — produced `release/CCSM Dev-Setup-0.1.0-x64.exe` (173 MB, signtool-signed).
- [x] Verified the unpacked exe metadata via `Get-Item 'CCSM Dev.exe').VersionInfo`: `ProductName: CCSM Dev`, `FileVersion: 0.1.0`.
- [x] Extracted bundled `package.json` from `app.asar`: `productName: "CCSM Dev"` correctly merged via `extraMetadata`. Source `package.json` is untouched (`git diff` shows only the new script line).
- [ ] Manual install of the dev exe on a machine that already has prod CCSM installed — **NOT done by worker** (would mutate the user's machine). Manager / reviewer should smoke-install once and confirm:
  - Both shortcuts coexist in Start Menu (`CCSM` + `CCSM Dev`).
  - `%APPDATA%\CCSM` and `%APPDATA%\CCSM Dev` are independent.
  - Both apps can run simultaneously (single-instance lock is per-appId).

## Follow-ups (not addressed in this PR)
- **Sentry release tagging** — both variants currently report under the same Sentry project / release tag. If we want to split dev crashes from prod crashes in Sentry, we should tag the release with the variant in `electron/sentry/init.ts` (e.g. `release: \`ccsm-dev@${version}\` vs `ccsm@${version}`). Flagged for a follow-up task.
- `make:mac:dev` / `make:linux:dev` analogues are not in scope for this PR — Windows-only per #891.
- `setup-aumid.ps1`'s `-Variant` change is opt-in; existing callers without the flag continue to register the prod AUMID (same default as before).

## Layer 1 (necessity)
Already validated by the evaluator subagent before dispatch. No cheaper path discovered during implementation — portable build still wouldn't satisfy the "auto-updater works end-to-end in dev" requirement, and a single env var can't carve a separate `userData` directory after the fact (it would orphan prod's data).

### DIFF
--- a/electron/__tests__/updater.test.ts
+++ b/electron/__tests__/updater.test.ts
@@ -25,6 +25,7 @@ const autoUpdaterEmitter = new EventEmitter();
 const state = {
   appIsPackaged: true,
   appVersion: '0.1.2',
+  appName: 'CCSM',
   checkForUpdatesImpl: (async () => ({ updateInfo: {} })) as () => Promise<unknown>,
   downloadUpdateImpl: (async () => undefined) as () => Promise<unknown>
 };
@@ -50,7 +51,8 @@ vi.mock('electron', () => {
     get isPackaged() {
       return state.appIsPackaged;
     },
-    getVersion: () => state.appVersion
+    getVersion: () => state.appVersion,
+    getName: () => state.appName
   };
   return { ipcMain, BrowserWindow, app };
 });
@@ -59,6 +61,7 @@ vi.mock('electron-updater', () => {
   const autoUpdater = {
     autoDownload: false,
     autoInstallOnAppQuit: false,
+    allowPrerelease: false,
     logger: null as unknown,
     on: (event: string, listener: Listener) => {
       autoUpdaterEmitter.on(event, listener);
@@ -79,6 +82,7 @@ function resetState() {
   quitAndInstallCalls.length = 0;
   state.appIsPackaged = true;
   state.appVersion = '0.1.2';
+  state.appName = 'CCSM';
   state.checkForUpdatesImpl = async () => ({ updateInfo: {} });
   state.downloadUpdateImpl = async () => undefined;
 }
@@ -118,6 +122,27 @@ describe('updater: IPC wiring', () => {
     const { autoUpdater } = await import('electron-updater');
     expect(autoUpdater.autoDownload).toBe(true);
     expect(autoUpdater.autoInstallOnAppQuit).toBe(true);
+  });
+
+  it('leaves allowPrerelease=false for the prod variant (#891)', async () => {
+    const { autoUpdater } = await import('electron-updater');
+    // Default reset state has appName='CCSM' so installUpdaterIpc must NOT
+    // flip allowPrerelease — prod users only see stable releases.
+    expect(autoUpdater.allowPrerelease).toBe(false);
+  });
+
+  it('flips allowPrerelease=true for the dev variant (#891)', async () => {
+    // Override appName before re-importing so the dual-install branch fires.
+    state.appName = 'CCSM Dev';
+    const mod = await import('../updater');
+    mod.__resetUpdaterForTests();
+    // Reset the mock field too — module-scoped mock object survives across
+    // freshModule() calls in beforeEach, so the prior installUpdaterIpc may
+    // have left it at its default. Force a known starting point.
+    const { autoUpdater } = await import('electron-updater');
+    autoUpdater.allowPrerelease = false;
+    mod.installUpdaterIpc();
+    expect(autoUpdater.allowPrerelease).toBe(true);
   });
 
   it('broadcasts status on update-available and fires update:available channel', () => {
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -169,7 +169,12 @@ app.whenReady().then(() => {
   // On Windows, set a stable AppUserModelID so the OS attributes the app to
   // its taskbar / Start Menu entry instead of generic "electron.exe".
   if (process.platform === 'win32') {
-    app.setAppUserModelId('com.ccsm.app');
+    // Dual-install (#891): the dev variant ships as productName "CCSM Dev"
+    // and must use a distinct AUMID so its taskbar / toast attribution
+    // doesn't collide with a co-installed prod build.
+    const isDevVariant = app.getName().includes('Dev');
+    const aumid = isDevVariant ? 'com.ccsm.app.dev' : 'com.ccsm.app';
+    app.setAppUserModelId(aumid);
   }
 
   initDb();
--- a/electron/updater.ts
+++ b/electron/updater.ts
@@ -98,6 +98,14 @@ export function installUpdaterIpc(): void {
   autoUpdater.autoInstallOnAppQuit = true;
   autoUpdater.logger = null;
 
+  // Dual-install (#891): the dev variant pulls GitHub pre-releases so we can
+  // ship release candidates / dogfood builds without affecting prod users
+  // (prod's autoUpdater leaves allowPrerelease=false, so it skips them).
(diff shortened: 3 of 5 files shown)


Jiahui-Gu (commented):
## Reviewer report (read-only, no approve — self-block)

### Verdict
LGTM with one nit. All Layer 1 + Layer 2 checks pass. New manager requirement (independent install dirs) is **already satisfied** by this PR — no extra commit needed.

### Layer 1
- **AUMID derivation** — `app.getName().includes('Dev')` works because `--config.extraMetadata.productName="CCSM Dev"` lands in the bundled `package.json` and Electron's `app.getName()` reads `name`/`productName` from there. Verified: prod build returns `'CCSM'` (untouched source `package.json`), dev build returns `'CCSM Dev'`. AUMIDs `com.ccsm.app` / `com.ccsm.app.dev` are disjoint.
- **`allowPrerelease = true`** — confirmed against `electron-updater` source: when set, the GitHub provider's version filter does NOT exclude releases marked `prerelease: true`. Same `Jiahui-Gu/ccsm` repo, dev sees both stable + RC tags, prod sees stable only. Correct semantics.
- **`--config.productName` AND `--config.extraMetadata.productName` both required** — confirmed not redundant:
  - top-level `productName` drives **build-time** macros: `${productName}` in `artifactName`, NSIS `APP_FILENAME` (install dir), `nsis.shortcutName` default, exe `VersionInfo.ProductName`.
  - `extraMetadata.productName` is merged into the bundled `package.json` so **runtime** `app.getName()` returns `"CCSM Dev"`.
  - Without the second one, `app.getName()` would still be `"CCSM"` and the AUMID branch would never fire.
- **`-Variant {prod|dev}`** — sensible shape. Default `prod` preserves back-compat with existing callers; explicit `-AppId`/`-ShortcutName` still win.

### Layer 2
- **Test coverage** — both branches asserted. `expect(autoUpdater.allowPrerelease).toBe(false)` for prod (line 131); `expect(autoUpdater.allowPrerelease).toBe(true)` for dev (line 145). Re-ran `npx vitest run electron/__tests__/updater.test.ts` locally → 15/15 pass.
- **No AUMID leaks** — grepped repo-wide for `com.ccsm.app` / `setAppUserModelId` / `appId`. Only hits: the (now branched) line in `electron/main.ts`, the `appId` in `package.json` build block (correctly overridden by `--config.appId=com.ccsm.app.dev`), docs/audit references (informational), and the test stub. Clean.
- **Single-instance lock** — `app.requestSingleInstanceLock()` keys on `userData`, which Electron derives from `appId`. Distinct appIds → distinct userData → independent locks. Both apps can run simultaneously. Worker's PR body call-out is accurate.
- **`app-update.yml`** — generated per-build by electron-builder from the same `publish: github → Jiahui-Gu/ccsm` block. Both variants will get one pointing at the same repo, which is exactly what we want (dev pulls pre-releases from there, prod pulls stables).

### New manager requirement: independent install dirs — VERIFIED, no fix needed
Confirmed by reading `node_modules/app-builder-lib/templates/nsis/multiUser.nsh` + `out/targets/targetUtil.js`:
- Per-user (`perMachine: false`, `oneClick: false`) install dir = `$LocalAppData\Programs\${APP_FILENAME}`.
- `APP_FILENAME` = `getWindowsInstallationDirName(appInfo, true)` → returns `productFilename` if it matches `[-_+0-9a-zA-Z .]+` (it does — space is allowed), otherwise `sanitizedName`.
- `productFilename = sanitizeFileName(productName)`.
- With `--config.productName="CCSM Dev"` (build-time, this PR), dev installer's `APP_FILENAME = "CCSM Dev"`.
- **Result**: dev → `%LOCALAPPDATA%\Programs\CCSM Dev\`, prod → `%LOCALAPPDATA%\Programs\CCSM\`. **No collision.**
- Uninstall registry key is also keyed off `appId` → separate uninstaller entries in Programs & Features.
- Inspected the built installer's `VersionInfo.ProductName` → `"CCSM Dev"`. Confirms top-level `productName` made it into the build, not just runtime `extraMetadata`.

### Risks (non-blocking)
1. **Brittle variant detection** — `app.getName().includes('Dev')` would also fire for any future productName containing "Dev" (e.g. "CCSM Developer Preview"). Low risk but worth tightening to an exact match (`=== 'CCSM Dev'`) or extracting a `const DEV_VARIANT_NAME = 'CCSM Dev'` shared between `main.ts` + `updater.ts`. Nit, not a merge blocker.
2. **Sentry release tagging** — worker correctly flagged as follow-up. **Not a blocker** — dev crashes will land in the same Sentry project under the same release tag, but stack frames + bundled paths still distinguish them in practice. File a follow-up task post-merge.
3. **Shortcut collision** — `nsis.shortcutName="CCSM Dev"` overrides for dev. Both `CCSM` + `CCSM Dev` Start Menu entries will coexist. No collision.

### Recommendation to manager
APPROVE-equivalent. Manager should:
1. Squash-merge after CI rollup green.
2. Smoke-install the dev exe (`release/CCSM Dev-Setup-0.1.0-x64.exe`, 173 MB, signtool-signed) on a machine with prod installed; confirm both shortcuts coexist + both `%APPDATA%\CCSM` and `%APPDATA%\CCSM Dev` directories appear.
3. Open follow-up tasks: (a) Sentry release tag split, (b) `make:mac:dev`/`make:linux:dev` analogues if/when needed.

PR #617, opened 2026-04-30
fix(lifecycle): complete deleteSession fan-out for badge/flash/crashed (#882)

Jiahui-Gu:
## Summary

Audit #876 H2: `deleteSession` left per-sid state in three sinks. This PR drains all three so a deleted session leaves no trace.

| Sink | Process | Leak path | Fix |
|---|---|---|---|
| `badgeStore.unread[id]` | main | toast fires → counter inflated forever; nothing called `clearSid` on session teardown | wire via existing `sessionWatcher.on('unwatched')` in `installPipeline.ts` (new optional `onUnwatchedSid` dep), `main.ts` passes `(sid) => badgeManager?.clearSid(sid)` |
| `flashStates[id]` | renderer | only cleared by main->renderer flashSink IPC, which fires on PTY exit; sessions with no PTY (or with stale flash from a previous run) leaked | `sessionCrudSlice.deleteSession` drains the entry directly |
| `disconnectedSessions[id]` | renderer | only cleared by `_applyPtyExit` on re-spawn — never on delete | same fix in `sessionCrudSlice.deleteSession` |

Pattern mirrors PR #881 (sessionTitles `forgetSid`). Single coherent fan-out expansion — not three unrelated cleanups.

## Files (5)
- `electron/main.ts` — pass `onUnwatchedSid` to drain badge.
- `electron/notify/bootstrap/installPipeline.ts` — accept `onUnwatchedSid`, invoke from existing unwatched handler.
- `electron/notify/bootstrap/__tests__/installPipeline.test.ts` — 3 new tests.
- `src/stores/slices/sessionCrudSlice.ts` — drain `flashStates[id]` + `disconnectedSessions[id]` in `deleteSession`. Identity-preserved when sid absent (no needless React re-renders); tolerates missing maps (slice composability).
- `tests/stores/slices/sessionCrudSlice.test.ts` — 4 new tests under `deleteSession fan-out (audit #876 H2)`.

## Test plan
- [x] Unit tests pass: `npx vitest run tests/stores/slices/sessionCrudSlice.test.ts electron/notify/bootstrap/__tests__/installPipeline.test.ts` -> **32 passed**.
- [x] Broader sanity: `npx vitest run electron/notify electron/sessionTitles tests/stores` -> **226 passed**.
- [x] Typecheck: `npx tsc --noEmit` -> clean.
- [x] Reverse-verify (cleanup disabled in both impl files, tests unchanged):
  ```
  Test Files  2 failed (2)
  Tests       3 failed | 29 passed (32)
  - flashStates[sid] for the deleted session: expected {b: true} got {a: true, b: true}
  - disconnectedSessions[sid] for the deleted session: expected {b:...} got {a:..., b:...}
  - onUnwatchedSid forwards for badge drain: expected vi.fn() called with ['s1'], called 0 times
  ```
  Restored:
  ```
  Test Files  2 passed (2)
  Tests       32 passed (32)
  ```
- [ ] CI green on PR.

Trust CI mode: not running e2e probe locally; the leak is a state-store assertion best covered by unit tests on the slice + the bootstrap wiring (the only observable signal of the fix is the Map size, which the unit test asserts directly).

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -259,6 +259,13 @@ app.whenReady().then(() => {
     onNotified: (sid) => {
       badgeManager?.incrementSid(sid);
     },
+    // audit #876 H2: drop the sid's unread counter when the session is
+    // unwatched (PTY exit / kill). Without this the badgeStore.unread map
+    // accumulated entries forever — every notified sid stayed counted for
+    // the app lifetime even after the session was deleted.
+    onUnwatchedSid: (sid) => {
+      badgeManager?.clearSid(sid);
+    },
   });
   // Hoist into the module-level holder so the IPC handlers above (registered
   // earlier in app.whenReady) can reach the pipeline. The handlers run later
--- a/electron/notify/bootstrap/__tests__/installPipeline.test.ts
+++ b/electron/notify/bootstrap/__tests__/installPipeline.test.ts
@@ -208,6 +208,49 @@ describe('installNotifyPipelineWithProducers', () => {
     expect(h.lastPipeline.current!.forgetSid).not.toHaveBeenCalled();
   });
 
+  // audit #876 H2: when a session is unwatched, the badge unread counter
+  // for that sid must be drained too. Without this fan-out the badge
+  // store accumulated entries forever — every notified sid stayed counted
+  // for the app lifetime even after the session was deleted.
+  it("sessionWatcher 'unwatched' forwards onUnwatchedSid for badge drain", () => {
+    const onUnwatchedSid = vi.fn();
+    installNotifyPipelineWithProducers({
+      getNameFn: () => null,
+      isGlobalMutedFn: () => false,
+      onNotified: () => {},
+      onUnwatchedSid,
+    });
+    h.watcherEmitter.emit('unwatched', { sid: 's1' });
+    expect(onUnwatchedSid).toHaveBeenCalledWith('s1');
+    expect(h.lastPipeline.current!.forgetSid).toHaveBeenCalledWith('s1');
+  });
+
+  it("sessionWatcher 'unwatched' does not invoke onUnwatchedSid for malformed payloads", () => {
+    const onUnwatchedSid = vi.fn();
+    installNotifyPipelineWithProducers({
+      getNameFn: () => null,
+      isGlobalMutedFn: () => false,
+      onNotified: () => {},
+      onUnwatchedSid,
+    });
+    h.watcherEmitter.emit('unwatched', undefined);
+    h.watcherEmitter.emit('unwatched', null);
+    h.watcherEmitter.emit('unwatched', {});
+    h.watcherEmitter.emit('unwatched', { sid: 123 });
+    h.watcherEmitter.emit('unwatched', { sid: '' });
+    expect(onUnwatchedSid).not.toHaveBeenCalled();
+  });
+
+  it('omitting onUnwatchedSid is allowed (optional dep)', () => {
+    installNotifyPipelineWithProducers({
+      getNameFn: () => null,
+      isGlobalMutedFn: () => false,
+      onNotified: () => {},
+    });
+    expect(() => h.watcherEmitter.emit('unwatched', { sid: 's1' })).not.toThrow();
+    expect(h.lastPipeline.current!.forgetSid).toHaveBeenCalledWith('s1');
+  });
+
   it('returns the live pipeline instance', () => {
     const ret = installNotifyPipelineWithProducers({
       getNameFn: () => null,
--- a/electron/notify/bootstrap/installPipeline.ts
+++ b/electron/notify/bootstrap/installPipeline.ts
@@ -28,6 +28,11 @@ export interface NotifyPipelineDeps {
   isGlobalMutedFn: () => boolean;
   /** Side-effect when a notification fires (badge bump). */
   onNotified: (sid: string) => void;
+  /** Side-effect when a session is unwatched (PTY exit / kill).
+   *  Used to drain badge unread counts so deleted/torn-down sessions
+   *  don't leave stale entries that inflate the aggregate badge total
+   *  forever (audit #876 H2). Optional for tests that don't care. */
+  onUnwatchedSid?: (sid: string) => void;
 }
 
 /** Construct the pipeline and wire the producer subscriptions. Returns the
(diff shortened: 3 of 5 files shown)


Jiahui-Gu (commented):
Reviewer report — verdict: APPROVE (manager merges; CI rollup pending on lint+typecheck+test ubuntu/windows).

## Layer 1 — direction & necessity
- **Asymmetry justified**, not a smell: badge state lives in main process (`badgeStore.unread` inside `electron/notify`); flash/disconnected live in renderer (`src/stores/slices/sessionRuntimeSlice`). Renderer cannot directly poke main-process state, and main has no `deleteSession` event — so:
  - badge drain rides the existing `sessionWatcher.on('unwatched')` listener (same pattern as #881 forgetSid). Right place.
  - flash + disconnected drain in `sessionCrudSlice.deleteSession`. Right place — `stopWatching` only fires on PTY exit/kill, so a session deleted **before any PTY spawn** would never trigger 'unwatched'. Renderer-side drain is necessary.
- One coherent fan-out, two mechanisms because the state lives in two processes. Consolidating them would require an extra IPC round-trip purely for symmetry — net loss.
- **No 4th sink missed.** Audited `src/stores/slices/types.ts`: only `flashStates` and `disconnectedSessions` are per-sid `Record<string, …>` in renderer state. Grepped all slices; no other `Record<string,` per-sid maps. The MEMORY-mentioned `messagesBySession` / `runningSessions` / `sessionStats` / `sessionContextUsage` are stale references — they don't exist in current code. Main-side per-sid maps in `sessionTitles` were already drained by #881.

## Layer 2 — implementation
- **Atomicity**: deleteSession builds a single `patch: Partial<RootStore>` and returns it from the same `set()` callback — one render, one commit.
- **Idempotency**: both drains guard on `s.flashStates && s.flashStates[id]` / same for disconnected. Calling deleteSession on a sid not in either map is a no-op.
- **Identity preservation verified** by the dedicated test `preserves identity when the sid was not present in either map` — when sid absent, `patch.flashStates` / `patch.disconnectedSessions` are not set, so the original references survive and React selectors don't re-run.
- **Defense for missing maps**: `tolerates missing maps` test covers harnesses without `sessionRuntimeSlice` composed.
- **Malformed-payload guard**: `installPipeline.ts` already filtered `!evt || typeof evt.sid !== 'string' || evt.sid.length === 0` before reaching `onUnwatchedSid`. Test `does not invoke onUnwatchedSid for malformed payloads` covers undefined/null/{}/numeric sid/empty string.
- **Optional-dep contract**: `omitting onUnwatchedSid is allowed` test confirms missing dep doesn't throw — keeps existing test harnesses green.

## Tests + reverse-verify
- 7 new tests as advertised: 4 in `tests/stores/slices/sessionCrudSlice.test.ts`, 3 in `electron/notify/bootstrap/__tests__/installPipeline.test.ts`. Locally: 32/32 pass (2.58s).
- Reverse-verify in PR body: with cleanup disabled, 3 fail with the right messages (flash not cleared, disconnected not cleared, onUnwatchedSid never called). With fix, 32 pass. Honest reverse-verify.
- Final tree clean (`git status` clean, `git log -1` matches PR HEAD `0807322`). No leftover state from the worker's mid-session hook reset.

## CI
- e2e ubuntu+macos SUCCESS, no-silent-drops SUCCESS. lint+typecheck+test ubuntu/windows + e2e windows still queued at review time. Manager: gate merge on full green rollup.

## Nits (non-blocking)
- One subtle gap not exercised by tests: a session that received toast notifications, then was deleted **without ever spawning a PTY (so no `stopWatching`)**, would still leak a badge entry. In practice such a session can't have a badge entry (badgeManager only increments via `onNotified`, which presumes a PTY chain). Safe by construction, but worth a comment in `main.ts` if a future change loosens that invariant. Not a merge blocker.

Verdict: APPROVE (no `--approve` flag per protocol). Manager merges after CI rollup turns fully green.

Jiahui-Gu (commented):
## Reviewer report (Task #882, audit #876 H2)

**Verdict: APPROVE-equivalent (manager merges).** No blockers. CI pending at review time.

### Layer 1 — direction & approach

**Asymmetry is correct, not inconsistent.** Process boundaries force the split:
- `badgeStore.unread` lives in **main** — only `sessionWatcher` events reach it cleanly. Wiring via the existing `sessionWatcher.on('unwatched')` in `installPipeline` mirrors PR #881's `forgetSid` pattern exactly.
- `flashStates` / `disconnectedSessions` live in **renderer Zustand** — `deleteSession` is the natural drain point. Trying to drain these from a main-side IPC would add an IPC round-trip + race window.

**Drain chain is complete end-to-end:** `renderer.deleteSession(id)` → `window.ccsmPty.kill(id)` (line 340) → `entryFactory` onExit handler → `sessionWatcher.stopWatching(sid)` → emits `'unwatched'` → handler calls `pipelineInstance.forgetSid(sid)` (which drains tracker maps + `flash.forget(sid)` on the main-side flashSink at `sinks/pipeline.ts:199`) + `forgetSessionTitleSid(sid)` + the new `deps.onUnwatchedSid(sid)` → `badgeManager.clearSid(sid)`.

**4th sink check (grepped):** No missed renderer per-sid stores. `src/stores/slices/types.ts` only exposes `flashStates` and `disconnectedSessions` as `Record<string, …>`. Main-side per-sid Maps (`runStateTracker`'s `mutedSids` / `lastFiredTs` / `lastUserInputTs` / `runStartTs`, `flashSink` timers + state, `notifyDecider` state) are all already drained by the existing `pipelineInstance.forgetSid` call. The only previously leaking sink in main was `badgeStore.unread`, which this PR now hooks.

**Edge case verified:** For sessions that never spawned a PTY, `window.ccsmPty.kill` is a no-op → no `unwatched` emit → badge isn't drained. But `badgeManager.incrementSid` only fires on a notify decision (which requires the run-state tracker to see PTY title events), so unread for that sid was never set in the first place. No leak.

### Layer 2 — implementation

- **Atomicity:** all three renderer drains happen in a single `set((s) => patch)` call → atomic Zustand commit. ✓
- **Identity preservation:** new map only allocated when `s.flashStates[id]` truthy → React selectors won't re-render on no-op deletes (test "preserves identity when the sid was not present" covers this). ✓
- **Composability:** `s.flashStates && s.flashStates[id]` guards tolerate slice harnesses missing the runtime slice (test "tolerates missing maps" covers). ✓
- **Test coverage:** 3 installPipeline tests (forwards, malformed-payload guard, optional-dep allowed) + 4 sessionCrudSlice tests (flash drain, disconnected drain, identity preservation, missing maps). Both touch real assertion targets, not log-spy proxies.
- **Reverse-verify in PR body:** 3 fail / 29 pass with cleanup disabled, 32/32 pass restored. Solid.
- **Worker hook-recovery:** final committed code on `0807322` is clean — diff matches PR body exactly (143/-1 across 5 files). No dangling artifacts.

### Nits (non-blocking)

- The `if (s.flashStates && s.flashStates[id])` guard pattern means an existing-but-`false` value (`flashStates[id] === false`) wouldn't be drained. In practice `_setFlash` deletes the key when `on=false` (line 64 of `sessionRuntimeSlice`), so a `false` entry never exists. Safe under current invariants; worth a comment if/when the invariant changes.
- No e2e probe added — justified by "leak is a Map-size assertion best covered by unit tests" (consistent with trust-CI mode + #881 precedent).

### Recommendation
Merge once CI rollup is green. No follow-ups needed.

PR #618, opened 2026-04-30
fix(lifecycle): dispose flashSink + listeners (#884)

Jiahui-Gu:
## Summary

Audit #876 cluster 1.14 + 3.8 — close 3 lifecycle leaks in the notify pipeline so app-shutdown / HMR re-install no longer leaks timers or app-level listeners.

### 1. `flashSink.timers` — pipeline.dispose did not clear per-sid setTimeout handles (#876 item 1.14)

`pipeline.dispose()` removed the sniffer listener but the flash sink kept its `Map<sid, NodeJS.Timeout>` populated. Every still-flashing sid leaked a closure over flashSink internals past disposal. Even with `unref()`, the closure pins the sink object until the timer fires.

**Fix:** added `flashSink.dispose()` that snapshots `Array.from(timers.keys())`, calls `clear(sid)` for each (clearTimeout + delete entry + send `on=false` IPC), and double-cleans `flashStates`. `pipeline.dispose()` now forwards to `flash.dispose()`.

### 2. `app.on('browser-window-focus' | 'browser-window-blur')` — never removed (#876 item 3.8)

`installNotifyPipelineWithProducers` registered focus/blur listeners on the Electron `app` and never removed them. App-lifetime in prod is fine; HMR / test re-installs leak.

**Fix:** named the handler closures (`onFocus`, `onBlur`), call `app.off(...)` in dispose.

### 3. `sessionWatcher.on('unwatched')` — never removed (#876 item 3.8)

Same pattern. The unwatched listener was an inline arrow function and could never be detached.

**Fix:** named closure `onUnwatched`, `sessionWatcher.off('unwatched', onUnwatched)` in dispose.

### Wiring

`installNotifyPipelineWithProducers` now returns `{ pipeline, dispose }` (was: just `pipeline`). `electron/main.ts` stores `notifyPipelineDispose` in a module-level holder and `electron/lifecycle/appLifecycle.ts` invokes it from the `before-quit` handler. `dispose` is idempotent (`disposed` flag) so duplicate calls are safe.

## Test plan

- [x] `electron/notify/sinks/__tests__/flashSink.test.ts` — 3 new tests on `dispose()`:
  - clears all pending timers + flash map + sends `on=false` for every active sid; advancing fake timers past `FLASH_DURATION_MS` produces no late sends (timer leak check)
  - idempotent (second call is a no-op)
  - safe when no active flashes
- [x] `electron/notify/sinks/__tests__/pipeline.test.ts` — 1 new test: `pipeline.dispose` clears flash sink state, `_peek()` empty, no late timer fires
- [x] `electron/notify/bootstrap/__tests__/installPipeline.test.ts` — 3 new tests:
  - `dispose` detaches focus, blur, unwatched listeners + forwards to `pipeline.dispose`
  - idempotent
  - post-dispose, focus/blur/unwatched events do not reach the pipeline
- [x] `tests/electron-load-smoke.test.ts` — all 19 require-load smokes pass post-build (CJS Electron-main compatibility)
- [x] `node -e "require('./dist/electron/notify/{bootstrap/installPipeline,sinks/flashSink,sinks/pipeline}.js'); require('./dist/electron/lifecycle/appLifecycle.js')"` — all four load cleanly

**Targeted suite: 53 tests pass (3 files + smoke).**

### Reverse-verify outputs (proving fix is load-bearing)

Removing `flash.dispose()` call from `pipeline.dispose()`:
```
FAIL  electron/notify/sinks/__tests__/pipeline.test.ts > dispose clears flash sink state (no leaked timers)
- Expected: {}
+ Received: { "s1": true, "s2": true }
```

Removing `app.off('browser-window-focus', onFocus)` from `installPipeline.dispose`:
```
FAIL  electron/notify/bootstrap/__tests__/installPipeline.test.ts > post-dispose, focus/blur events no longer reach the pipeline
expect(h.lastPipeline.current!.setFocused).not.toHaveBeenCalled()
Number of calls: 1
```

Both confirm: tests fail without the fix, pass with it.

## Out of scope

- Renderer `flashStates` / `disconnectedSessions` / `badgeStore.unread` cleanup on `deleteSession` — audit item #2, Task #882.
- `sessionTitles` per-sid Maps — audit item #1, Task #881 (already landed via 631596a).

## Constraints honored

- No touch to `sessionTitles` (#881) or `deleteSession` fan-out (#882).
- English only.
- Targeted to notify pipeline files; no overlap with #882's scope (renderer store sinks).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -41,12 +41,27 @@ jobs:
       # and inherits the postinstall-rebuilt native bindings (better-sqlite3,
       # node-pty) from the cached tree, which is what the Electron e2e
       # actually loads.
+      #
+      # CRITICAL: this cache key MUST be distinct from ci.yml's key. ci.yml
+      # runs `npm ci` with `ELECTRON_SKIP_BINARY_DOWNLOAD=1` (no electron
+      # binary) AND compiles native modules against Node 20's ABI
+      # (NODE_MODULE_VERSION 115). e2e needs the electron binary AND
+      # native modules built against Electron's ABI (NMV 145). Sharing a
+      # key meant whichever workflow won the race poisoned the other:
+      #   - ci.yml winning → e2e gets no electron binary + Node-20-ABI
+      #     `.node` files → main process throws ERR_DLOPEN_FAILED on
+      #     better-sqlite3 require during initDb() → BrowserWindow never
+      #     created → harness times out with the bare "no renderer window
+      #     appeared in time" error.
+      # The `e2e-` prefix isolates this workflow's cache entirely.
+      # (Task #913 healed the electron-binary half; Task #919 the
+      # native-module half.)
       - name: Cache node_modules
         id: nm_cache
         uses: actions/cache@v4
         with:
           path: node_modules
-          key: nm-${{ runner.os }}-${{ runner.arch }}-node20-${{ hashFiles('package-lock.json') }}
+          key: nm-e2e-${{ runner.os }}-${{ runner.arch }}-node20-${{ hashFiles('package-lock.json') }}
       - name: Cache Electron binary
         uses: actions/cache@v4
         with:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -69,6 +69,19 @@ jobs:
       # (Task #913)
       - name: Ensure Electron binary
         run: node node_modules/electron/install.js
+      # Rebuild native modules against Electron's ABI on every run. The
+      # node_modules cache is shared with ci.yml, which compiles
+      # better-sqlite3 / node-pty against Node 20's ABI (NODE_MODULE_VERSION
+      # 115). When the e2e job restores that cache and launches Electron
+      # (NODE_MODULE_VERSION 145), the main process throws
+      # "The module ... was compiled against a different Node.js version"
+      # during `initDb()` — the BrowserWindow is never created and the
+      # harness times out with the bare "no renderer window appeared in
+      # time" error. Running `@electron/rebuild` is a fast no-op when the
+      # `.node` files already match Electron's ABI and self-heals when
+      # ci.yml won the cache race. Diagnosed via Task #919.
+      - name: Ensure native modules built for Electron ABI
+        run: npx electron-rebuild
       - name: Build
         run: npm run build
       - name: Run e2e (Linux with xvfb)
--- a/electron/lifecycle/appLifecycle.ts
+++ b/electron/lifecycle/appLifecycle.ts
@@ -67,6 +67,12 @@ export interface LifecycleDeps {
    *  ptyHost. Idempotent; critical on Windows where conpty can otherwise
    *  leak the claude child past Electron quit. */
   killAllPtySessions: () => void;
+  /** Tear down the notify pipeline + its app-level listeners (focus/blur,
+   *  sessionWatcher 'unwatched') and any pending flash timers. Optional
+   *  because `before-quit` may fire before the pipeline is constructed
+   *  (e.g. early failure in app.whenReady). Idempotent on its own.
+   *  Audit #876 cluster 1.14 + 3.8 / Task #884. */
+  disposeNotifyPipeline?: () => void;
   /** Close the SQLite handle on a real quit. */
   closeDb: () => void;
   /** Spawn a fresh main window from the macOS dock-click `activate` path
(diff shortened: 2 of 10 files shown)


Jiahui-Gu (commented):
## Reviewer report — APPROVE

**Verdict:** Approve. Layer 1 (necessity, hook choice, idempotency) and Layer 2 (mechanics, wiring, tests) both clean.

### Layer 1
- **`before-quit` is the correct hook.** Fires before windows are destroyed, so `flashSink.dispose()`'s `on=false` IPC sends still reach surviving renderers. `will-quit` fires after `window-all-closed` — too late for IPC. Co-located with `killAllPtySessions` keeps cleanup ordering coherent.
- **Idempotency is sufficient.** `disposed` flag + the existing `isDestroyed()` guards in flashSink make a mid-fire race benign: even if a focus event is mid-flight when `dispose()` runs, `setFocused(true)` on a torn-down sniffer is harmless (sniffer already had `removeAllListeners` and pipeline still owns its own state). Worst case is an extra no-op state flip, no leak.
- **No other unhandled listeners in scope.** Grepped `app.on(` and `sessionWatcher.on(` across `electron/`. The remaining producers are app-lifetime singletons (`second-instance`, `window-all-closed`, `activate`, `before-quit`, `state-changed`/`title-changed` in `ipcRegistrar`). Notify pipeline cluster (#876 1.14 + 3.8) is fully covered by this PR.

### Layer 2
- **`flashSink.dispose()` sending `on=false` is correct, not wasteful.** Renderer mirror clears even if a window survives dispose (e.g. a rebuild scenario). For real shutdown, the existing `isDestroyed()` guard in `send()` short-circuits, so the cost is one snapshot loop — negligible.
- **Breaking shape change `{pipeline, dispose}` is fully wired.** Only production caller is `electron/main.ts` (line 262), updated correctly: `installed.pipeline` hoisted to `notifyPipeline`, `installed.dispose` to `notifyPipelineDispose`, then passed to `registerLifecycleHandlers` via the new optional `disposeNotifyPipeline` dep. Test file's 14 callers all either ignore the return value or use the new `.pipeline`/`.dispose` shape.
- **`appLifecycle` integration clean.** `disposeNotifyPipeline` is `optional` (correct — `before-quit` can fire before `whenReady` completes), guarded with try/catch, and called after `killAllPtySessions` (correct ordering: kill ptys first to stop emitting, then drain pipeline).
- **Tests strong.** 34 targeted tests pass locally. Reverse-verify outputs in PR body are credible and match the test assertions (`_peek()` empty + `setFocused not called`). Skipping a dedicated reverse-verify for `sessionWatcher.off` is acceptable — structurally identical to the focus/blur path and covered by the "every subscription gone" assertion.

### Risk check
- **Pre-existing `better-sqlite3` NODE_MODULE_VERSION 145 vs 137 mismatch** is environmental (Node 24 vs Electron 33 ABI), not introduced by this PR. Diff touches zero native-binding code.
- **No interaction with #881 (sessionTitles `forgetSid`)** — only consumed via the existing `onUnwatched` handler, which this PR refactors into a named closure (no semantic change).
- **No interaction with #882 (deleteSession fan-out)** — orthogonal scope (renderer stores vs main lifecycle).

### CI
Pending at review time. No reason to expect failures (smoke test + targeted suite green locally; lint-clean diff).

Recommend merge once CI is green.

Jiahui-Gu (commented):
Post-rebase review (Task #884, rebase onto working after #617 merge):

Verified clean merge of the two orthogonal additions in `electron/notify/bootstrap/installPipeline.ts`:

1. `onUnwatched` handler (lines 95-100) does all three: `pipelineInstance.forgetSid(evt.sid)` + `forgetSessionTitleSid(evt.sid)` + `deps.onUnwatchedSid?.(evt.sid)`. Nothing dropped from #617.
2. `dispose()` calls `sessionWatcher.off('unwatched', onUnwatched)` against the same named reference registered on line 101. Focus/blur listeners likewise tracked + detached.
3. `electron/main.ts` wires both: `onUnwatchedSid: (sid) => badgeManager?.clearSid(sid)` at line 272, and `disposeNotifyPipeline: () => notifyPipelineDispose?.()` at line 315.
4. Tests in `installPipeline.test.ts` cover both paths: badge-drain forwarding ("forwards onUnwatchedSid for badge drain", malformed payload guard, optional-dep omission), plus dispose detachment + idempotency + post-dispose silence.

CI rollup: lint+typecheck+test and e2e all pending (just kicked off post-rebase). Will report rollup once it settles. No code-level concerns blocking merge.

Not approving (self-block per workflow).

PR #620, opened 2026-04-30
chore(ui): remove redundant new-session entries + sidebar collapse (#894)

Jiahui-Gu:
## Summary
- Remove 4 of 6 redundant 'new session' entries (firstRun CTA, Tutorial, CollapsedRail, command palette items 'New session' + 'Toggle sidebar'). Keep only the sidebar `+` button + cwd chevron.
- Remove sidebar collapse feature entirely (state, persistence, Ctrl+B shortcut, CollapsedRail, AppShell conditional, command palette item, ShortcutOverlay row).
- Drop orphan `tutorialSeen` persisted flag (no remaining consumer).

## Why
User: "sidebar 不要收起功能，为什么老说有收起功能，肯定是有垃圾代码或者文档". Plus the firstRun CTA path was the source of the cold-start blank-screen bug (#852/#890): no active session → click central CTA → spawn → switch to TerminalPane on cold xterm → blank. Path no longer exists.

## Files deleted
- `src/components/Tutorial.tsx`
- `src/app-effects/useTutorialOverlay.ts`
- `src/components/sidebar/CollapsedRail.tsx`
- `tests/tutorial.test.tsx`
- `tests/app-effects/useTutorialOverlay.test.tsx`
- `tests/sidebar/CollapsedRail.test.tsx`

## i18n keys removed (en + zh)
- `sidebar.expandSidebarTooltip`, `sidebar.expandSidebarAria`, `sidebar.newSessionTooltip`, `sidebar.newSessionAria`
- `tutorial.*` (entire namespace: skip/back/next/done/stepXofY/goToStepAria/welcomeTitle/welcomeBody/sessionsTitle/sessionsBody/groupsTitle/groupsBody/startTitle/startBody/newSessionBtn/importSessionBtn)
- `commandPalette.cmdNewSession`, `commandPalette.cmdToggleSidebar`
- `shortcuts.actionNewSession`, `shortcuts.actionToggleSidebar`
- `firstRun.newSession`, `firstRun.importSession` (entire namespace)

Plus CSS tokens `--gradient-tutorial-welcome` and `--shadow-tutorial-card` (Tutorial-only).

## Tests
- `npm run typecheck`: PASS (renderer + electron tsconfigs)
- `npx vitest`: 1277/1277 pass (excluding pre-existing better-sqlite3 ABI mismatches in `electron/__tests__/db-hardening.test.ts` — env-only, unrelated)
- `node scripts/run-all-e2e.mjs --only=harness-ui`: 14/14 PASS

[...]

PR #623, opened 2026-04-30
fix(app): gate newSession on claudeAvailable to prevent blank-pane (#900, #852)

Jiahui-Gu:
## Bug (#852, per #896 audit Item 1)

User clicks the right-pane new-session trigger while `App.tsx`'s boot probe (the `window.ccsmPty.checkClaudeAvailable()` effect at lines 227-247) is still pending. Sequence:

1. `claudeAvailable` starts `undefined`.
2. Click new-session -> `createSession(null)` -> new sid becomes active.
3. Right-pane render branch falls through to the `[data-testid=\"claude-availability-probing\"]` spacer (a literal empty flex div) because `claudeAvailable` is still `undefined`.
4. No PTY is spawned. Pane stays blank, no user-visible explanation.

## Root cause

App.tsx render branch ladder (line ~346-354):

```tsx
) : claudeAvailable === false ? (
  <ClaudeMissingGuide ... />
) : claudeAvailable === true ? (
  <TerminalPane ... />
) : (
  // empty spacer when probe still pending
  <div className=\"flex-1 min-h-0\" data-testid=\"claude-availability-probing\" />
);
```

`newSession` (line ~263-271) had no awareness of probe state - clicking it during the brief boot window dropped the user on the empty branch.

## Fix

Two layers (belt-and-suspenders, same PR):

- **Gate**: `newSession` short-circuits when `claudeAvailableRef.current !== true`. Uses a ref mirror so the callback sees the latest probe value without re-rendering long-lived consumers (Sidebar, CommandPalette).
- **Visible affordance**: probing spacer now renders `t('claudeAvailability.probing')` ('Checking Claude CLI...' / '正在检测 Claude CLI...') instead of a blank div. Covers any pre-existing active session caught in the same boot window AND gives the user a reason for the pause. The `data-testid` is preserved so the existing harness probes that wait for it to disappear still work.

i18n key `claudeAvailability.probing` added to both `en.ts` and `zh.ts` (parity test pins this).

## Test plan

- [x] `tests/claude-probe-race-900.test.tsx` (new) - 3 contracts:
  - newSession is a no-op while boot probe is pending (deferred-promise stub)
  - newSession works once probe resolves true
  - Probing spacer surfaces the visible 'Checking Claude CLI...' affordance
- [x] **Reverse-verify** on pre-fix code (stash + rerun): 2/3 fail (the gate + visible-message contracts), 1/3 pass (the after-resolve test, expected - it's a strict superset).
- [x] **Post-fix**: 3/3 pass.
- [x] `npm run typecheck` clean.
- [x] `npm run lint` - 0 errors (92 pre-existing warnings unchanged).
- [x] `npx vitest run` - 1284 passed, 21 pre-existing better-sqlite3 ABI failures (NODE_MODULE_VERSION 145 vs 137, env issue, unrelated).
- [x] `tests/first-run-empty-state.test.tsx` + `tests/i18n-key-parity.test.ts` still pass.

### E2E harness rationale

Did NOT add a new harness-ui case. The race requires monkey-patching `window.ccsmPty.checkClaudeAvailable` via `addInitScript` before App boots (similar to `caseStartupPaintsBeforeHydrate`), then forcing reload, which adds ~30s and would need to be the last case (perturbs subsequent state). The vitest tests give deterministic, controlled-timing coverage of both the gate and the visible affordance - both are pure renderer-state assertions where jsdom + Chromium behave identically. Existing `terminal-pane-mounted` harness case still covers the `claudeAvailable=true` happy-path wiring.

## Files

- `src/App.tsx` - gate + visible spacer
- `src/i18n/locales/en.ts` + `zh.ts` - `claudeAvailability.probing` key
- `tests/claude-probe-race-900.test.tsx` - new test, 3 cases

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,6 +14,7 @@ import { DragRegion, WindowControls } from './components/WindowControls';
 import { InstallerCorruptBanner } from './components/InstallerCorruptBanner';
 import { useStore } from './stores/store';
 import { initI18n } from './i18n';
+import { useTranslation } from './i18n/useTranslation';
 import { usePreferences } from './store/preferences';
 import { useThemeEffect } from './app-effects/useThemeEffect';
 import { useLanguageEffect } from './app-effects/useLanguageEffect';
@@ -64,6 +65,7 @@ function AppEffectsBridge(): null {
 }
 
 export default function App() {
+  const { t } = useTranslation();
   const sessions = useStore((s) => s.sessions);
   const flashStates = useStore((s) => s.flashStates);
   const activeId = useStore((s) => s.activeId);
@@ -246,7 +248,27 @@ export default function App() {
   // We still blur the trigger element synchronously so repeated Enter
   // presses on the "New session" button do not spawn extra sessions
   // before xterm has had a chance to take focus.
+  // `claudeAvailableRef` mirrors the boot probe state for the gate below.
+  // We read it inside `newSession` instead of closing over the React state
+  // because the callback is also passed down to long-lived consumers
+  // (Sidebar, CommandPalette) and we want the latest probe value without
+  // re-rendering them when it flips.
+  const claudeAvailableRef = React.useRef(claudeAvailable);
+  React.useEffect(() => {
+    claudeAvailableRef.current = claudeAvailable;
+  }, [claudeAvailable]);
+
   const newSession = React.useCallback(() => {
+    // Bug #852 / Task #900: clicking "new session" while the boot probe
+    // (`claudeAvailable === undefined`) is still pending would swap the
+    // right pane to the blank `[data-testid="claude-availability-probing"]`
+    // spacer (no PTY spawn). Short-circuit until the probe resolves so the
+    // user can't strand themselves on a blank pane. The probing spacer
+    // below now also shows a visible "Checking Claude CLI…" affordance
+    // for any pre-existing active session caught in the same window.
+    if (claudeAvailableRef.current !== true) {
+      return;
+    }
     if (typeof document !== 'undefined') {
       const active = document.activeElement;
       if (active instanceof HTMLElement && active !== document.body) {
@@ -289,9 +311,19 @@ export default function App() {
   ) : claudeAvailable === true ? (
     <TerminalPane sessionId={active.id} cwd={active.cwd ?? ''} />
   ) : (
-    // Probing claude availability — render an empty flex spacer
-    // so the layout doesn't jump once the boot check resolves.
-    <div className="flex-1 min-h-0" data-testid="claude-availability-probing" />
+    // Probing claude availability — the right pane would otherwise be a
+    // blank flex spacer for the duration of the boot probe, which made the
+    // pane look broken when a user clicked "new session" before the probe
+    // resolved (Task #900 / bug #852). Render a low-key "Checking…" line
+    // instead so the user understands the pane is intentional. The
+    // outer testid is preserved so existing harness probes that wait for
+    // it to disappear still work.
+    <div
+      className="flex-1 min-h-0 flex items-center justify-center text-xs text-[var(--muted-fg)]"
+      data-testid="claude-availability-probing"
+    >
+      <span>{t('claudeAvailability.probing')}</span>
+    </div>
   );
 
   return (
--- a/src/i18n/locales/en.ts
+++ b/src/i18n/locales/en.ts
@@ -317,6 +317,12 @@ const en = {
     colShortcut: 'Shortcut',
     colAction: 'Action'
   },
+  // Right-pane affordance shown for the brief boot window between app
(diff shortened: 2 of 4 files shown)


Jiahui-Gu (commented):
## Verdict
APPROVE (posted as comment — self-approve blocked by GitHub same-identity)

## Layer 1
Right cause attacked at the right layer: the buttons live in Sidebar/CommandPalette without claudeAvailable awareness, so gating in the App-owned `newSession` callback is correct. Belt-and-suspenders pairing with the visible probing-spacer message addresses the silent-gate UX trade-off — user gets feedback ("Checking Claude CLI…") instead of a dead click.

## Layer 2
- ref mirror: pass (useRef + useEffect sync on `claudeAvailable`)
- gate placement: pass (early return at top of `newSession`, before any side effects)
- i18n symmetric: pass (`claudeAvailability.probing` in both en/zh, sentence case)
- test coverage 3/3: pass (pending no-op, post-resolve allows, visible-message contract on pre-existing active session)
- reverse-verify shown: confirmed (PR body: 2/3 fail pre-fix, 3/3 pass post-fix)
- preserves data-testid: pass (`data-testid="claude-availability-probing"` retained on outer div)
- never-resolves edge: out-of-scope here — `ClaudeMissingGuide` already covers `claudeAvailable === false`; the only failure mode left is probe IPC hanging forever, which is a separate boot-health concern (not introduced by this PR)
- lint/typecheck/vitest: pass (lint 0 errors / 92 pre-existing warnings; typecheck clean; new test 3/3 in 4.65s)

## Concerns
None blocking. Clean fix, well-commented, test pins both contracts.

Jiahui-Gu (commented):
## Verdict
APPROVE

Post-rebase review per feedback_rebase_always_reviewed.md. Rebase from a9fc110 onto current working tip (post-#620) is clean.

## Rebase-risk findings
- Gate semantics: PASS. `src/App.tsx:269` early-returns `claudeAvailableRef.current !== true` as the very first statement of the `newSession` callback, before the synchronous blur side effect (App.tsx:272-277) and before `createSession(null)` (App.tsx:278). State + ref + mirror useEffect all present (App.tsx:214, 256-259). Reachable from Sidebar `+` (NewSessionButton.tsx:30 → onCreateSession → App.tsx newSession) and the cwd chevron path goes through `createSession({cwd})` directly in Sidebar.tsx:155,166 — note: the chevron path is NOT gated, but that matches the pre-rebase design (chevron is a CWD-already-chosen path; same blank-pane risk technically applies but was out of scope per #900 description focused on the `+` button entry). Not a rebase regression.
- Probing-spacer + useTranslation: PASS. App.tsx:325 renders `t('claudeAvailability.probing')`, `data-testid="claude-availability-probing"` preserved (App.tsx:323), `useTranslation` import re-added (App.tsx:17), `const { t } = useTranslation()` re-added (App.tsx:68). i18n keys `claudeAvailability.probing` present in both en.ts:323 and zh.ts:306.
- Sidebar button text case: Scenario A-but-not-PR-introduced. `src/i18n/locales/en.ts:44` has `newSession: 'New Session'` (uppercase S), violating feedback_no_uppercase_ui_strings.md. However, `git blame` shows this string was introduced by c8b37f4f on 2026-04-22, well before PR #623. The test selector `/^New Session$/` correctly matches the live UI. PR #623 is consistent with the current UI; the casing violation is a pre-existing tech debt issue worth a separate followup task to fix the en.ts string + chase any other refs, NOT a blocker for this PR.
- firstRun grep-clean: PASS. No `firstRun.` i18n key references anywhere in src/ tests/ scripts/ docs/. Only hit is `scripts/dogfood-probe-current-ui.mjs:95-96` which references an unrelated `data-testid="first-run-empty"` selector.
- Other dropped symbols: NONE. Diff `git log -p HEAD~1..HEAD -- src/App.tsx` shows the rebase merge picked up only the two intended changes (useTranslation import + newSession gate + probing-spacer message). No other symbol drops.
- Lint: PASS (0 errors, 91 warnings — all pre-existing unused-disable + unused-var warnings, none in changed lines).
- Typecheck: PASS (`tsc --noEmit && tsc --noEmit -p tsconfig.electron.json` clean).
- Vitest 3/3: PASS (`tests/claude-probe-race-900.test.tsx` 3/3 in 4.63s).

## Concerns
- The uppercase `'New Session'` i18n string is a real violation but it predates this PR. Recommend a followup task to fix en.ts (and any zh.ts mirror — zh uses `'新会话'` which is fine) plus audit other sidebar.* keys for case violations. Filing this is out of scope for #623 since changing it now would force a test selector edit unrelated to the rebase fix.
- Cwd-chevron path in Sidebar.tsx:155,166 calls `createSession({cwd})` directly without going through the `newSession` gate. Same theoretical race, but: (a) chevron only opens after sidebar render so probe is more likely resolved by then, (b) #852 reproed via `+` button. Worth noting in the followup task list but not a rebase regression.

PR #624, opened 2026-04-30
test(e2e): alt-screen-fits-visible-viewport probe for L4 PR-F safety (#899)

Jiahui-Gu:
## Summary

Pins the visible-outcome contract that PR #611 (L4 PR-F, commit ac2017d) relied on when deleting PR #602's spawn-time PTY measurement hack.

PR #611's argument: the post-attach resize+replay at `src/terminal/usePtyAttach.ts:344-376` reflows alt-screen content from the headless source-of-truth buffer, so claude's full-screen panels (trust prompt, welcome card) render correctly at the visible viewport size without needing the spawn-time hack. No e2e test pinned this — silently re-introducing #602 / #852 was possible. Per #896 audit Item 4.

Adds `caseAltScreenFitsVisibleViewport` to `scripts/harness-real-cli.mjs` (per `feedback_e2e_prefer_harness.md`, prefer harness over standalone probe). The case:

1. Mints a fresh OS-tmp cwd outside HOME (so claude's resolver doesn't auto-trust via parent dir).
2. Spawns a session there, waits for claude's welcome / trust panel to render, then a 2s settle for post-attach fit + replay.
3. Reads the visible xterm buffer (rows 0..rows-1, no scrollback) plus live `term.cols` / `term.rows`.
4. Asserts:
   - **fit ran**: `view.cols !== 120 || view.rows !== 30` (xterm grew past spawn-time defaults).
   - **content reflowed**: longest run of `─` border chars on screen is within 30 cols of `view.cols` AND > 125 (i.e. clearly past the spawn-time 120-col width).

### Why welcome panel, not trust prompt

I originally targeted the trust prompt (`Do you trust the files in this folder?` / `1. Yes, proceed`). Empirically, claude v2.1.119 with a fresh isolated `.claude` config (the harness setup) does NOT show the trust prompt for arbitrary cwds — it goes directly to the welcome panel. Verified by dumping the full xterm buffer after spawn into a fresh OS-tmp cwd outside HOME.

Both panels exercise the same alt-screen-after-resize code path on the ccsm side (alt-screen mode + post-attach resize+replay), and the welcome panel is what users actually hit on first launch.

### Reverse-verify

Commented out the post-attach resize+replay block at `src/terminal/usePtyAttach.ts:344-376` (gated `if (fit && t4 && sid4 && false /* REVERSE-VERIFY */)`), rebuilt, ran case:

```
[HARNESS] >>> case: alt-screen-fits-visible-viewport
[...]

PR #626, opened 2026-04-30
ci(e2e): self-heal poisoned electron binary on cache hit (#913)

Jiahui-Gu:
## Summary
- Fixes PR #618 Win e2e CI failure (4 reruns, same error: `Electron failed to install correctly`).
- Root cause: `e2e.yml` and `ci.yml` share the cache key `nm-${runner.os}-${runner.arch}-node20-${hashFiles('package-lock.json')}`. `ci.yml` runs `npm ci` with `ELECTRON_SKIP_BINARY_DOWNLOAD=1`, so when ci.yml wins the cache-write race, the cached `node_modules/electron/` lacks `dist/electron(.exe)`, `path.txt`, and `dist/version`. e2e.yml restores it, hits the cache, skips `npm ci` (and therefore postinstall), and crashes at `electron.launch`.
- Fix: add an unconditional `node node_modules/electron/install.js` step in e2e.yml after install. Its `isInstalled()` short-circuit makes it a fast no-op on healthy caches; on poisoned caches it downloads the missing binary. `ELECTRON_SKIP_BINARY_DOWNLOAD` is intentionally NOT set on this step.

## Evidence
Failed run [25155457558](https://github.com/Jiahui-Gu/ccsm/actions/runs/25155457558/job/73747808071) (PR #618):
```
Cache hit for: nm-Windows-X64-node20-a7ddb6...
Cache not found for input keys: electron-Windows-a7ddb6...
[Install deps]  (skipped — cache hit)
electron.launch: Electron failed to install correctly...
```

Passing run [25158370860](https://github.com/Jiahui-Gu/ccsm/actions/runs/25158370860/job/73745990230) (PR #624):
```
Cache not found for input keys: nm-Windows-X64-node20-a7ddb6...
[Install deps] npm ci --legacy-peer-deps  (full postinstall ran)
```

Both PRs have unchanged `package-lock.json`, so they share a cache key — pass/fail purely depends on which workflow wrote the cache first.

## Test plan
- [ ] Win e2e job on this PR passes (cache will likely be hit, exercising the self-heal path).
- [ ] After merge, rerun PR #618 Win e2e — should pass.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -57,6 +57,18 @@ jobs:
       - name: Install deps
         if: steps.nm_cache.outputs.cache-hit != 'true'
         run: npm ci --legacy-peer-deps
+      # Ensure the Electron binary (`node_modules/electron/dist/`) is present.
+      # The node_modules cache is shared with ci.yml, which runs `npm ci` with
+      # ELECTRON_SKIP_BINARY_DOWNLOAD=1 — when ci.yml wins the race to write
+      # the cache, the restored node_modules has the electron JS shim but no
+      # `dist/electron(.exe)` and no `path.txt`/`dist/version`, causing
+      # `electron.launch` to error with "Electron failed to install correctly".
+      # `electron/install.js` short-circuits via `isInstalled()` when the
+      # binary is already present, so this is a fast no-op on healthy caches
+      # and self-heals poisoned caches. Must NOT inherit the env var above.
+      # (Task #913)
+      - name: Ensure Electron binary
+        run: node node_modules/electron/install.js
       - name: Build
         run: npm run build
       - name: Run e2e (Linux with xvfb)


Jiahui-Gu (commented):
## Verdict
APPROVE

## Layer 1
Real bug (PR #618 4x failures), surgical e2e.yml fix is correct blast-radius choice over editing ci.yml.

## Layer 2
- New step location: pass (after `Install deps`, before `Build` / `Run e2e`)
- ELECTRON_SKIP_BINARY_DOWNLOAD not set: pass (no env block on step; install.js exits early if env present, so explicitly NOT inheriting it is critical)
- Unconditional run: pass (no `if:` gate, runs on cache hit AND miss)
- isInstalled() safeguard real: pass — confirmed in `node_modules/electron/install.js`: `if (isInstalled()) process.exit(0);` checks `dist/version` matches package version (and further down, `path.txt` + executable existence). Fast no-op on healthy caches.
- CI evidence (25159701425 + 25159933573 green): confirmed — both runs all 3 OS green (ubuntu/macos/windows) on this branch. 25159933573 windows job 2m30s shows the self-heal path works.
- e2e.yml paths filter caveat: e2e.yml `on.pull_request.paths` is restricted to `electron/** | src/** | scripts/** | package.json | package-lock.json` — this PR only touches `.github/workflows/e2e.yml` so the PR-trigger e2e WILL NOT auto-run on push. Worker correctly used `workflow_dispatch` for both verification runs. Manager: rely on manual `gh workflow run` evidence above; future PRs that touch only workflows still need manual triggers (not blocking this PR).

## Followup tasks to file
- Drop `ELECTRON_SKIP_BINARY_DOWNLOAD=1` from `ci.yml` `npm ci` step as the root-cause fix (optional; current self-heal is robust). Cache key collision between ci.yml and e2e.yml remains, but is now harmless.
- Consider adding `.github/workflows/e2e.yml` to the `paths:` filter so workflow self-edits trigger their own validation.

## Concerns
None blocking. Comment block in the workflow is excellent — explains the racing-cache + ELECTRON_SKIP_BINARY_DOWNLOAD interaction so future maintainers won't "clean up" the seemingly redundant step.

PR #631, opened 2026-04-30
fix(ci): add npm run build to release verify job

Jiahui-Gu:
## Problem
v0.2.0 release attempt failed at workflow run [25165590364](https://github.com/Jiahui-Gu/ccsm/actions/runs/25165590364). The \`verify\` job ran \`npm test\` without first running \`npm run build\`, so \`tests/electron-load-smoke.test.ts\` failed because it requires \`dist/electron/**\` to exist. Build job was skipped \xe2\x86\x92 no installer artifact, no draft Release.

## Fix
Insert \`- run: npm run build\` between \`typecheck\` and \`test\` in the \`verify\` job. Order is now: install \xe2\x86\x92 lint \xe2\x86\x92 typecheck \xe2\x86\x92 build \xe2\x86\x92 test.

## Test plan
- [ ] CI green on this PR
- [ ] After merge to working \xe2\x86\x92 main \xe2\x86\x92 re-tag v0.2.0, release.yml verify passes and Windows installer is produced

Task #925.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,7 @@ jobs:
         run: npm rebuild better-sqlite3
       - run: npm run lint
       - run: npm run typecheck
+      - run: npm run build
       - run: npm test
 
   build:


Jiahui-Gu (commented):
## Review verdict: LGTM — recommend MERGE

### Layer 1 (necessity / direction): sound
- Root cause confirmed: `tests/electron-load-smoke.test.ts` (Task #764) hard-requires `dist/electron/**`. Without a prior build, the smoke `it.each` cases throw "Expected dist file missing" and the verify job fails — exactly what killed run 25165590364.
- `npm run build` is the right script. `package.json` defines it as `npm run clean && tsc -p tsconfig.electron.json && webpack --mode production && <touch>`, which is the only script that emits `dist/electron/**`. There is no separate `build:electron` script, so a narrower target is not an option.
- Loosening the smoke harness to be tolerant of missing dist is explicitly the wrong direction — its whole purpose (per `feedback_main_process_load_smoke_test.md`) is to catch PR #501-class ESM/CJS load regressions before they hit prod.
- Verify is the correct location. The downstream `build` job runs its own `npm run build` per-OS for packaging; we are not duplicating heavy work, just running it once on Linux to feed `npm test`. Splitting build into its own job would add cache-coordination complexity for no gain.
- Brings `release.yml` in line with `.github/workflows/ci.yml` (lines 86-90), which already has the same Build step with a comment citing Task #764. Pure parity fix.

### Layer 2 (implementation): clean
- One-line insertion at correct indentation, matches surrounding `- run:` step style.
- Step order is correct: checkout → setup-node → setup-python → npm ci → rebuild better-sqlite3 → lint → typecheck → **build** → test.
- Diff is scoped strictly to the `verify` job; no collateral edits.
- CI green on all 4 required checks (ubuntu/macos/windows verify + no-silent-drops). Base correctly `working`. `mergeable: MERGEABLE`.

### Recommendation
**MERGE.** No nits, no blockers. After merge, fast-forward `working` → `main`, retag `v0.2.0` (delete the failed tag first), and the release workflow should produce installers on this run.

PR #635, opened 2026-04-30
fix(build): upsize icon.png to 1024x1024 for macOS packaging

Jiahui-Gu:
## Summary
- v0.2.0 release run 25166890617 macOS step failed: `image /Users/runner/work/ccsm/ccsm/build/icon.png must be at least 512x512`. Linux + Windows electron-builder don't enforce this; macOS dmg/iconset packaging does.
- The procedural icon generator (`scripts/build-icon.cjs`) was writing the 256x256 entry from the `.ico` SIZES list as the standalone `build/icon.png`. Decouple the standalone PNG resolution from the `.ico` ladder and render it fresh at 1024x1024 (4x the macOS minimum, headroom for iconset generation).
- The `.ico` keeps the 16/24/32/48/64/128/256 multi-res ladder unchanged — 256 is the legacy ICO header maximum (the 1-byte width field encodes 256 as 0).

## Why script change instead of just replacing the binary
The icon is generated procedurally from a parameterized renderer (Task #929 setup notes "Find existing high-res source"; there is none — the source IS the script). Bumping `PNG_SIZE` keeps the regeneration story intact: re-run `npm run build:icon` any time the brand mark changes.

## Verify
- `file build/icon.png` -> `PNG image data, 1024 x 1024, 8-bit/color RGBA, non-interlaced` (was 256x256).
- `file build/icon.ico` -> still `MS Windows icon resource - 7 icons, 16x16 ...` (same SIZES, byte-size delta is PNG-encoder variance from re-rendering, visual output identical).
- `npm run build` -> green.
- `npm test` -> 1294 passed; 3 unrelated `better-sqlite3` `NODE_MODULE_VERSION` mismatch failures pre-exist on Win 11 worktree env, no DB code touched in this PR.

## Test plan
- [ ] CI green (Linux + Windows builds — these don't enforce >=512x512 so a regression would only show as a build-time failure, not a runtime one)
- [ ] Next release tag triggers `release.yml` and the macOS dmg step succeeds (real verification — cannot reproduce locally without macOS runner)

## Discipline
- All English; no `--no-verify`; PR base = `working`.


Jiahui-Gu (commented):
## Review verdict: APPROVE

- **Decoupling PNG_SIZE from the .ico SIZES ladder is the right call.** Adding 512 to the ICO ladder would (a) inflate the .ico for no Windows benefit (Win shell tops out at 256 in practice) and (b) still leave the PNG output at whatever the last-ladder-entry happens to be — coupling the two via `entries[entries.length - 1]` was the original sin. Naming the standalone PNG resolution as its own `PNG_SIZE` constant is the cleanest fix.
- **Procedural regeneration over a checked-in high-res source is correct here.** The repo already commits to "the source IS the script" for the brand mark (three callers all driven from the same parameterized renderer). Shipping a static high-res PNG would create a second source of truth that drifts the moment the mark changes.
- **1024 vs 512 is a sensible call.** macOS dmg only requires >=512, but iconset/Apple icon ladder goes up to 1024; rendering once at 1024 lets electron-builder downsample without ever upscaling. Cost is +21KB on disk (6.5KB -> 27.8KB), negligible.

### Layer 2 (clarity / safety)
- `scripts/build-icon.cjs:34-41` — `SIZES` and `PNG_SIZE` clearly separated with comments explaining why each ceiling exists (256 = legacy ICO header limit, 1024 = iconset headroom). Good.
- `scripts/build-icon.cjs:135-142` — the standalone PNG path now calls `renderCMark(PNG_SIZE).toPNG()` instead of reusing `entries[entries.length - 1].png`. Decoupling is complete; no leftover reference to the ladder for the PNG output.
- Header doc-comment at top of file (`scripts/build-icon.cjs:18-21`) updated to reflect the new resolution + macOS rationale.
- Console log line updated to print the actual `PNG_SIZE` rather than hardcoded `256x256`.
- **Third caller untouched.** Confirmed `scripts/screenshot-512-tray.cjs` and `scripts/screenshot-630-app-icon.cjs` (the other two `renderCMark`-style callers) are not in the diff. `git diff --stat` shows only `build/icon.ico`, `build/icon.png`, `scripts/build-icon.cjs`.
- **Binary artifacts verified:**
  - `file build/icon.png` -> `PNG image data, 1024 x 1024, 8-bit/color RGBA, non-interlaced`. Matches PR claim.
  - `build/icon.ico` parsed via ICO header: 7 frames at 16/24/32/48/64/128/256. ICO ladder is preserved bit-for-bit on the size dimension; the byte-size delta (12883 -> 11922) is just PNG-encoder variance from re-rendering, as the PR body notes.

### CI status (snapshot at review time)
- `no-silent-drops`: **pass**
- All other 6 jobs (`lint+typecheck+test` x 3 OS, `e2e` x 3 OS): **pending** (run id 25167882867 / 25167882813, just kicked off)
- Worker's locally-reported 3 `better-sqlite3` `NODE_MODULE_VERSION 145 vs 137` test failures are environmental (Win 11 worktree native-binding mismatch), not introduced by this PR — this PR touches zero JS/TS source, only a build script + two binary artifacts. CI will be the source of truth.

### Recommendation
**APPROVE-AND-MERGE once CI is green** (Linux + Windows builds + tests). Real macOS dmg verification only happens at the next release tag, but PR CI rules out regressions on the other two platforms and on the test suite. No code changes requested.

PR #661, opened 2026-05-01
feat(daemon): T22 marker corruption-treat-as-PRESENT

Jiahui-Gu:
## Summary

Implements the daemon-shutdown marker reader for v0.3 split-process upgrade flow.

**Spec ref**: `docs/superpowers/specs/v0.3-design.md` §6.4 ("Marker semantics" / rel-S-R8) and §6.1 R2 (marker-aware crash-loop skip).

> Marker corruption / partial write (rel-S-R8): if the marker file exists but is unreadable / malformed JSON, treat as PRESENT (conservative — prefer false-negative on crash-loop accounting over false-positive that bricks legitimate post-upgrade boots).

## Decision table (in `daemon/src/marker/reader.ts`)

| File state                              | Result                                  |
|-----------------------------------------|-----------------------------------------|
| ENOENT (no file)                        | `{ kind: 'absent' }`                    |
| Empty / whitespace-only                 | `{ kind: 'present', reason: 'empty' }`  |
| Unparseable JSON                        | `{ kind: 'present', reason: 'invalid-json' }` |
| Valid JSON missing required fields      | `{ kind: 'present', reason: 'missing-fields' }` |
| Valid JSON `{reason, version, ts}`      | `{ kind: 'present', payload }`          |
| Any other I/O error (EACCES/EISDIR/...) | `{ kind: 'present', reason: 'io-error' }` |

The `reason` field lets the consumer emit the spec-mandated `marker_corrupt` warn log without a second parse pass.

## Files

- `daemon/src/marker/reader.ts` — `readMarker(path)` + `DAEMON_SHUTDOWN_MARKER_FILENAME` constant + `ShutdownMarkerPayload` shape. Never throws.
- `daemon/src/marker/__tests__/reader.test.ts` — 14 cases covering every decision-table row, BOM tolerance, forward-compat extra fields, non-Error rejection paths, plus an invariant sweep.

## T13 dependency

T13 (runtimeRoot) is not merged yet (parallel worker). Reader is **path-agnostic** — only ships the spec-defined `daemon.shutdown` filename constant. Wiring layer (separate task) joins it under `runtimeRoot`/`dataRoot`. No block.

## Test plan

- [x] `npx vitest run daemon/src/marker/__tests__/reader.test.ts` — 14/14 pass
- [x] **Reverse-verify**: flipped ENOENT branch to return `present` and confirmed the absent-case test fails (1/14 fail), then restored.
- [ ] Wiring + e2e: deferred to the consumer task that integrates this with the §6.1 supervisor crash-loop counter (out of scope per spec — `[manager r9 lock]` "Marker unlink ownership = daemon" wires up in §6.4 producer side).

### DIFF
--- a/daemon/src/marker/__tests__/reader.test.ts
+++ b/daemon/src/marker/__tests__/reader.test.ts
@@ -1,0 +1,180 @@
+// T22 — marker reader: corruption-treat-as-PRESENT.
+// Spec ref: docs/superpowers/specs/v0.3-design.md §6.4 (Marker semantics
+// "rel-S-R8"): "if the marker file exists but is unreadable / malformed
+// JSON, treat as PRESENT".
+
+import { promises as fs } from 'node:fs';
+import { mkdtemp, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import {
+  DAEMON_SHUTDOWN_MARKER_FILENAME,
+  readMarker,
+  type MarkerReadResult,
+} from '../reader.js';
+
+let dir: string;
+let markerPath: string;
+
+beforeEach(async () => {
+  dir = await mkdtemp(join(tmpdir(), 'ccsm-marker-'));
+  markerPath = join(dir, DAEMON_SHUTDOWN_MARKER_FILENAME);
+});
+
+afterEach(async () => {
+  vi.restoreAllMocks();
+  await rm(dir, { recursive: true, force: true });
+});
+
+describe('readMarker — decision table', () => {
+  it('ENOENT (no file) -> absent', async () => {
+    const result = await readMarker(markerPath);
+    expect(result).toEqual<MarkerReadResult>({ kind: 'absent' });
+  });
+
+  it('empty file -> present (corruption: empty)', async () => {
+    await fs.writeFile(markerPath, '');
+    const result = await readMarker(markerPath);
+    expect(result).toEqual<MarkerReadResult>({
+      kind: 'present',
+      reason: 'empty',
+    });
+  });
+
+  it('whitespace-only file -> present (corruption: empty)', async () => {
+    await fs.writeFile(markerPath, '   \n\t  ');
+    const result = await readMarker(markerPath);
+    expect(result.kind).toBe('present');
+    if (result.kind === 'present') {
+      expect(result.reason).toBe('empty');
+      expect(result.payload).toBeUndefined();
+    }
+  });
+
+  it('partial / unparseable JSON -> present (corruption: invalid-json)', async () => {
+    // Realistic mid-write fragment — opening brace + half a key, no close.
+    await fs.writeFile(markerPath, '{"reason":"upg');
+    const result = await readMarker(markerPath);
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
**Independent review — APPROVE**

- Spec rule confirmed at `docs/superpowers/specs/v0.3-design.md:1660` (rel-S-R8): *"if the marker file exists but is unreadable / malformed JSON, treat as PRESENT (conservative — prefer false-negative on crash-loop accounting over false-positive that bricks legitimate post-upgrade boots). Log `marker_corrupt` warn and unlink as part of consumption."* Implementation matches verbatim.
- **Missing-fields → present**: defensible. Spec rationale ("prefer false-negative") generalizes — a partially-written marker with only `{reason}` is exactly the half-flushed case rel-S-R8 targets.
- **I/O error → present**: spec doesn't enumerate EACCES/EISDIR explicitly, but the rule "anything other than ENOENT = PRESENT" follows directly from the rationale (re-spawn-loop is worse than a one-boot crash-counter skip). Conservative read is correct.
- **`reason` corruption-flavour field**: not spec-mandated, but the spec mandates a `marker_corrupt` warn log (line 1660). Surfacing the flavour via a typed enum lets the caller emit that warn without a second parse pass. Clean add-on, well-justified.
- **T13 deferral**: reader is path-agnostic (only ships `DAEMON_SHUTDOWN_MARKER_FILENAME` constant). Wiring under `runtimeRoot`/`dataRoot` correctly deferred to consumer task. No block.
- **Single responsibility**: pure producer (read + decide), no logging / no unlink side-effect (per `[manager r9 lock]` — daemon owns unlink on producer side, this is the consumer-side reader). Correct split.

## Layer 2 (implementation)

- `npx vitest run daemon/src/marker/__tests__/reader.test.ts` — **14/14 pass** (verified locally in pool-8).
- `npx tsc --noEmit -p daemon/tsconfig.json` — **clean** (verified locally).
- **Reverse-verify reproduced**: flipped `ENOENT` branch from `{kind:'absent'}` to `{kind:'present', reason:'io-error'}` → exactly 1/14 fail (the ENOENT-absent test, line 35) → restored. Tests genuinely guard the contract.
- All 6 decision-table branches covered + BOM tolerance + forward-compat extra-fields + non-Error rejection + invariant sweep ("only ENOENT yields absent").
- Pure: no `console.*`, no `logger`, no fs writes, no unlink. Caller emits warn + does unlink per spec.
- `never throws` invariant: enforced by single try/catch around `readFile` + catch-all `'io-error'` branch + non-Error rejection test.
- New `daemon/src/marker/` directory — no overlap with existing `daemon/src/db/` (DB migration markers are unrelated).
- `isShutdownMarkerPayload` correctly `Number.isFinite(ts)` rules out `NaN`/`Infinity`.

## Nits (non-blocking)

- `MarkerCorruptionReason` collapses "missing field" and "wrong type" into `'missing-fields'`. Caller log won't distinguish them. Fine — both indicate corruption and spec only requires one `marker_corrupt` warn flavour.
- The `// spec §6.4 + §3` header comment references §3 but the actual rule lives at §6.4 line 1660. Minor doc nit.

## Verdict

**APPROVE.** Spec-faithful, pure, well-tested, reverse-verified. Ready to merge to `working` after CI green.

PR #663, opened 2026-05-01
feat(daemon): T8 hello interceptor

Jiahui-Gu:
## Summary

Adds `daemon/src/envelope/hello-interceptor.ts` — slot #0 of the L1 envelope interceptor pipeline (spec §3.4.1.f bullet 0 + §3.4.1.g handshake).

Pure decider; per-connection state is owned by the caller and injected via `ctx.state`. Returns one of `pass | reply | reject` so the caller (connect adapter / dispatcher) owns the side effects (writing the reply frame, `socket.destroy()` on rejection).

## Spec citations

- `docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md` §3.4.1.f bullet 0 — helloInterceptor runs FIRST; allowlist `["ccsm.v1/daemon.hello"]`; pre-handshake non-hello → `hello_required` + `socket.destroy()`. Short-circuits the chain so no `MIGRATION_PENDING` leak (r7 manager lock).
- Same fragment §3.4.1.g — 2-frame HMAC handshake: daemon proves possession of `daemon.secret` by computing `HMAC-SHA256(daemon.secret, decode(clientHelloNonce))` truncated to 16 bytes, base64url. Per-connection `clientHelloNonce` is single-use; second `daemon.hello` on same socket → `hello_replay`.
- Round-9 base64url lock: HMAC computed over the DECODED 16-byte nonce, NOT the ASCII string (regression test included).

## What's wired

- `decideHello(ctx, config)` — pure decision function over `{ rpcName, payload, state }` and `{ daemonSecret, bootNonce, minClient? }`.
- `createHelloState()` — caller calls per accepted socket; `{ handshakeComplete: false }` initially.
- `HELLO_METHOD`, `DAEMON_WIRE`, `DAEMON_ACCEPTED_WIRES`, `DAEMON_FRAME_VERSION`, `DAEMON_FEATURES`, `DAEMON_DEFAULTS` constants exported for the dispatcher to reuse.
- Compatibility check produces `{ compatible: false, reason }` for `wire-mismatch` / `version-mismatch` / `frame-version-mismatch`; reply still carries `helloNonceHmac` so client can rule out imposter (per spec).
- Reuses `hmac.ts` (`computeHmac`), `protocol-version.ts` (`DAEMON_PROTOCOL_VERSION`), `base64url.ts` (`decode`).

## Test summary (25 cases, all green)

- Pre-handshake gate: arbitrary RPC → `hello_required` + `destroySocket: true`; state stays `false`. Includes literal-compare guard (no normalisation of method name).
- Successful handshake: HMAC matches `computeHmac(secret, decode(nonce))`; full reply shape verified (protocol/defaults/bootNonce/frameVersion); flag flips; nonce recorded.
- Compatibility variants: `wire-mismatch`, `version-mismatch`, `frame-version-mismatch` → `compatible: false` + `reason`, but HMAC still surfaced.
- Schema rejection: null / non-object / missing nonce / wrong-length nonce / non-base64url nonce / non-integer protocol version / non-array frameVersions / non-string clientWire — all `schema_violation`, flag stays false.
- Post-handshake: arbitrary RPC → `pass`; second hello (fresh nonce) → `hello_replay`; exact-replay (same payload) → `hello_replay`.
- State isolation: connection A flip does not affect connection B.
- HMAC correctness regression: HMAC-over-bytes != HMAC-over-string; different secret → different HMAC.

## Verification

- `npx vitest run daemon/src/envelope/__tests__/hello-interceptor.test.ts` — 25/25 green (~5s).
- `npm run build:daemon` — clean.
- ESM smoke: `node --input-type=module -e "import('./daemon/dist-daemon/envelope/hello-interceptor.js')"` → exports `decideHello`, `createHelloState`, plus 6 constants.

## Wiring TODO (out of scope, follow-up tasks)

- T16 dispatcher slots `decideHello` at index 0 of `interceptors[]` and translates `{kind:'reject'}` into a wire-level reject frame + `socket.destroy()`.
- Per-connection `HelloConnectionState` allocated at socket-accept time and disposed at socket close.
- Pipeline-level test asserting `hello_required` short-circuits BEFORE `migrationGateInterceptor` (the actual r7 lock guarantee).

## Test plan

- [x] Unit tests pass (`vitest run daemon/src/envelope/__tests__/hello-interceptor.test.ts`)
- [x] Typecheck clean
- [x] Daemon build clean
- [x] ESM module loads under Node main process

### DIFF
--- a/daemon/src/envelope/__tests__/hello-interceptor.test.ts
+++ b/daemon/src/envelope/__tests__/hello-interceptor.test.ts
@@ -1,0 +1,366 @@
+import { describe, it, expect } from 'vitest';
+import { Buffer } from 'node:buffer';
+import { randomBytes } from 'node:crypto';
+import {
+  createHelloState,
+  decideHello,
+  DAEMON_ACCEPTED_WIRES,
+  DAEMON_FRAME_VERSION,
+  DAEMON_DEFAULTS,
+  DAEMON_FEATURES,
+  HELLO_METHOD,
+  type HelloConnectionState,
+  type HelloRequestPayload,
+} from '../hello-interceptor.js';
+import { computeHmac, generateNonce } from '../hmac.js';
+import { decode as base64urlDecode } from '../base64url.js';
+import { DAEMON_PROTOCOL_VERSION } from '../protocol-version.js';
+
+const SECRET = randomBytes(32);
+const BOOT_NONCE = '01HZZZBOOTNONCEFAKEULID00';
+const CONFIG = { daemonSecret: SECRET, bootNonce: BOOT_NONCE } as const;
+
+function makeHelloPayload(overrides: Partial<HelloRequestPayload> = {}): HelloRequestPayload {
+  return {
+    clientWire: 'v0.3-json-envelope',
+    clientProtocolVersion: DAEMON_PROTOCOL_VERSION,
+    clientFrameVersions: [0],
+    clientFeatures: ['binary-frames', 'hello'],
+    clientHelloNonce: generateNonce(),
+    ...overrides,
+  };
+}
+
+describe('decideHello — pre-handshake gate (spec §3.4.1.f slot #0)', () => {
+  it('rejects arbitrary RPC with hello_required + destroySocket=true', () => {
+    const state = createHelloState();
+    const decision = decideHello(
+      { rpcName: 'ccsm.v1/session.subscribe', state },
+      CONFIG,
+    );
+    expect(decision.kind).toBe('reject');
+    if (decision.kind === 'reject') {
+      expect(decision.code).toBe('hello_required');
+      expect(decision.destroySocket).toBe(true);
+      expect(decision.message).toContain('ccsm.v1/session.subscribe');
+    }
+    // State must NOT be flipped on rejection.
+    expect(state.handshakeComplete).toBe(false);
+    expect(state.consumedClientHelloNonce).toBeUndefined();
+  });
+
+  it.each([
+    'ccsm.v1/session.send',
+    'ccsm.v1/pty.write',
+    'daemon.shutdown',
+    '/healthz',
+    '',
+  ])('rejects %j with hello_required pre-handshake', (rpcName) => {
+    const state = createHelloState();
+    const decision = decideHello({ rpcName, state }, CONFIG);
+    expect(decision.kind).toBe('reject');
+    if (decision.kind === 'reject') {
+      expect(decision.code).toBe('hello_required');
+    }
+    expect(state.handshakeComplete).toBe(false);
+  });
+
+  it('does not normalise the hello method (literal compare)', () => {
+    const state = createHelloState();
+    // wrong case / whitespace must NOT be treated as the canonical hello.
+    for (const variant of ['CCSM.V1/DAEMON.HELLO', ' ccsm.v1/daemon.hello', 'ccsm.v1/daemon.hello ']) {
+      const decision = decideHello(
+        { rpcName: variant, payload: makeHelloPayload(), state },
+        CONFIG,
+      );
+      expect(decision.kind).toBe('reject');
+      if (decision.kind === 'reject') {
+        expect(decision.code).toBe('hello_required');
+      }
+    }
+    expect(state.handshakeComplete).toBe(false);
+  });
+});
+
+describe('decideHello — successful handshake', () => {
+  it('flips handshakeComplete and returns reply with HMAC + protocol block', () => {
+    const state = createHelloState();
+    const req = makeHelloPayload();
+    const decision = decideHello(
+      { rpcName: HELLO_METHOD, payload: req, state },
+      CONFIG,
+    );
+
+    expect(decision.kind).toBe('reply');
+    if (decision.kind !== 'reply') return;
+
+    expect(state.handshakeComplete).toBe(true);
+    expect(state.consumedClientHelloNonce).toBe(req.clientHelloNonce);
+
+    // Verify the HMAC matches what a client computing over the same nonce bytes
+    // would produce — this is the round-trip check the client performs via
+    // crypto.timingSafeEqual.
+    const nonceBytes = base64urlDecode(req.clientHelloNonce);
+    const expected = computeHmac(SECRET, nonceBytes);
+    expect(decision.payload.helloNonceHmac).toBe(expected);
+
+    // Spec-mandated reply fields.
+    expect(decision.payload.protocol).toEqual({
+      wire: 'v0.3-json-envelope',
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Reviewer report (Task #969 / T8 — independent review)

**Verdict: APPROVE** with note on task-brief vs spec divergence.

### Layer 1 — direction / necessity / better approach

#### Brief vs spec divergence (resolved in worker's favor)

The dispatch brief said *"daemon-side handler verifies HMAC"*. The worker reinterpreted to *"daemon **computes** HMAC, client verifies"*. **The worker is correct** and the brief was sloppy.

Spec citations from `docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md` §3.4.1.g:

- Round-5 manager lock (line 150): *"client issues nonce, daemon proves possession"*.
- Frame-2 schema (lines 170-193): daemon's reply payload contains `helloNonceHmac: "<base64-22>"` — the daemon's HMAC over the client-issued nonce. There is no daemon-verifies-anything field.
- Lines 195-196: *"The **client** computes the same HMAC over its own `clientHelloNonce` ... and compares against the daemon's `helloNonceHmac` field using `crypto.timingSafeEqual`. ... Reverse-direction proof (client proves to daemon) is unnecessary because peer-cred (§3.1.1) + ACL (§7.1) already authenticate the client."*
- Cross-frag rationale row (line 316): *"client issues nonce in frame 1, daemon proves possession in frame 2 (no third ack)"*.

**Reviewer confirms the task brief was the wrong direction**; spec §3.4.1.g unambiguously matches the worker's implementation. Manager should refine the task-brief format going forward to copy spec wording verbatim for security-critical handshakes.

#### Slot #0 enforcement (correct)

Spec §3.4.1.f bullet 0 (line 138) explicitly anchors the r7 manager lock: *"explicit ordering #0 so handshake-required check fires BEFORE migrationGateInterceptor; otherwise a non-hello RPC during migration would short-circuit with MIGRATION_PENDING and leak pre-handshake daemon state to an unverified peer."* The worker's PR description echoes this. Slot-#0 enforcement is the dispatcher's job (T16), not this module — out of scope here, correctly noted in PR body's "Wiring TODO".

#### HMAC over decoded 16-byte buffer (correct, with regression test)

Round-9 base64url lock (line 197) requires both sides decode the wire string to a 16-byte `Buffer` before HMAC compute / `timingSafeEqual`. `decideHello` calls `base64urlDecode(helloReq.clientHelloNonce)` on line 243 then `computeHmac(config.daemonSecret, nonceBytes)` on line 244. Test `computes HMAC over the DECODED nonce bytes, not the base64url string` (test file lines 326-347) is a real regression guard that asserts `computeHmac(SECRET, decodedNonceBytes) !== computeHmac(SECRET, asciiString)`.

#### `compatible: false` paths still surface `helloNonceHmac` (correct)

Spec §3.4.1.g line 208: *"On `compatible: false`, daemon does NOT close the socket immediately — the reply still carries `helloNonceHmac` (so client can rule out imposter) plus `reason`"*. Implementation (lines 246-262) builds the reply with `helloNonceHmac` first, then appends `compatible` + optional `reason`; tests `still replies (with HMAC) but compatible:false on wire mismatch` (line 122-141), version-mismatch (143-153), and frame-version-mismatch (155-165) all assert `helloNonceHmac.length > 0` on incompatible paths. Anti-imposter property preserved.

#### Single Responsibility (clean)

Per `feedback_single_responsibility`, `decideHello` is a pure decider:
- inputs: `{ rpcName, payload, state }` + `{ daemonSecret, bootNonce, minClient? }`
- outputs: `{ kind: 'pass' | 'reply' | 'reject' }`
- single side effect: in-place mutation of caller-injected `ctx.state` on a successful reply (lines 268-269), explicitly documented (lines 19-33, 263-269)

Caller owns the `Map<connId, HelloConnectionState>`, the `socket.destroy()` sink, and logging. No process state, env, or wall-clock reads. Defensible.

### Layer 2 — implementation

- **Test coverage (25/25 green locally, 2.53 s):** pre-handshake gate (any RPC → `hello_required`, including literal-compare guard against case/whitespace normalization), successful handshake (full reply shape verified incl. protocol/defaults/bootNonce/frameVersion), three compatibility variants, schema rejection (null/non-object/missing nonce/wrong-length nonce/non-base64url nonce/non-integer protocol version/non-array frame versions/non-string wire) — **all schema rejects assert `state.handshakeComplete` stays `false`** (test file line 181), the load-bearing invariant. Post-handshake: arbitrary RPC passes, second hello rejected `hello_replay`, exact-payload replay rejected. State isolation across two `HelloConnectionState` instances. HMAC correctness (over bytes, not string; different secret → different HMAC).
- **HMAC compute** reuses T4 `daemon/src/envelope/hmac.ts` (`computeHmac`, `HMAC_TAG_LENGTH`, `NONCE_BYTES`). No new crypto code in this PR — clean dependency on the audited primitive.
- **Constants exported for T16** (`HELLO_METHOD`, `DAEMON_WIRE`, `DAEMON_ACCEPTED_WIRES`, `DAEMON_FRAME_VERSION`, `DAEMON_FEATURES`, `DAEMON_DEFAULTS`) — the dispatcher will need these for the slot-#0 wiring.
- **CI**: no-silent-drops PASS; ubuntu/macos/windows lint+typecheck+test still queued at review time but typecheck + vitest + build:daemon clean per PR body and locally re-verified.
- **Out of scope, correctly deferred** (PR body "Wiring TODO"): T16 dispatcher slotting at index 0, per-connection state allocation at socket-accept, pipeline-level test asserting `hello_required` short-circuits BEFORE `migrationGateInterceptor`. The last item is the actual r7 lock guarantee and MUST land with T16 — flagging here so the T16 reviewer doesn't miss it.

### Minor observations (non-blocking)

1. `compatible: false` path also flips `state.handshakeComplete = true` (test line 140). Implementation comment (lines 263-269) explains: socket is about to be torn down by the caller and the per-connection nonce is single-use regardless. Reasonable, but T16 caller MUST destroy the socket after writing the incompatible reply — otherwise a `compatible: false` reply could leave a usable post-handshake connection. Worth a sentence in the T16 prompt.
2. `checkCompatibility` accepts `_minClient` as an unused parameter (line 397) — placeholder for future semver gating against `minClient: 'v0.3'`. Fine to leave for v0.3, but worth a TODO comment so a future worker doesn't think `minClient` is enforced.

Neither is a merge blocker.

---

Reviewer: opus 4.7 1M, independent pass per `feedback_independent_reviewer` + `feedback_reviewer_two_layers`. Brief-vs-spec divergence surfaced per `feedback_reviewer_flags_ambiguous_requirements`.

PR #666, opened 2026-05-01
feat(daemon/db): T35 MIGRATION_PENDING gate consumer

Jiahui-Gu:
## Summary

T35 — pure state holder for the SQLite migration lifecycle, consumed by T10 (`migrationGateInterceptor`) and by supervisor RPCs (`/healthz`, `/stats`, `/version`, `ping()`) that report `migrationState` as a diagnostic field per frag-8 §8.5.

## Spec refs

- `docs/superpowers/specs/v0.3-fragments/frag-8-sqlite-migration.md` §8.5 short-circuit scope (`migrationState: 'pending' | 'completed' | 'failed'` on supervisor RPCs)
- Same fragment §8.6 (data RPCs return `MIGRATION_PENDING` while pending/failed; daemon stays UP)
- `daemon/src/envelope/migration-gate-interceptor.ts` (T10 — pure decider that reads `isMigrationPending()`)
- `daemon/src/db/migration-events.ts` (T30 — wire event contracts; producers call `setMigrationState()` alongside emitting the event)
- `daemon/src/db/migrate-v02-to-v03.ts` (T29 — migration runner that drives state transitions)

## State contract

```ts
type MigrationState = 'idle' | 'pending' | 'completed' | 'failed';
isMigrationPending() === (state === 'pending' || state === 'failed');
```

- `idle` — boot default, no migration ever ran. Allows data RPCs.
- `pending` — runner in progress; T10 short-circuits data RPCs.
- `completed` — success; data RPCs proceed normally.
- `failed` — runner threw; data RPCs stay blocked so the renderer fatal-error modal (frag-8 §8.6) holds the surface.

Spec only surfaces 3 wire values; `idle → completed` collapse can happen at the supervisor-RPC bridge if needed.

## Layer 1

Confirmed no existing event bus in `daemon/src/`. Minimal subscribe/notify is appropriate per the spec's narrow consumer set (T10 + supervisor RPCs + Electron-main bridge for the modal). Pure state holder — no I/O, no event emission, no RPC dispatch.

## Test plan

- [x] 19 unit tests: initial state, every transition pair, `isMigrationPending()` boolean projection (T10 contract), no-op same-state re-emit, single + multi-subscriber notify, unsubscribe (idempotent + safe inside callback), listener error isolation via `queueMicrotask` rethrow.
- [x] `npx vitest run daemon/src/db/__tests__/migration-gate-consumer.test.ts` → 19/19 pass.
- [x] `npx tsc --noEmit -p daemon/tsconfig.json` clean.
- Pre-existing daemon test failures in `migrate-v02-to-v03.test.ts` and `schema/v0.3.test.ts` are an env-side `better-sqlite3` ABI mismatch (`NODE_MODULE_VERSION 145 vs 137`) — unrelated to this PR; the new module has zero DB I/O.

## Out of scope

- Wiring this consumer into supervisor RPC handlers (separate task that owns `/healthz` etc.)
- Wiring producers (T29 runner) to call `setMigrationState()` (separate task)
- Electron-main bridge that subscribes for the modal (frag-8 §8.6 / T33 modal driver task)

### DIFF
--- a/daemon/src/db/__tests__/migration-gate-consumer.test.ts
+++ b/daemon/src/db/__tests__/migration-gate-consumer.test.ts
@@ -1,0 +1,191 @@
+import { describe, expect, it, vi } from 'vitest';
+import { MigrationGateConsumer, type MigrationState } from '../migration-gate-consumer.js';
+
+describe('MigrationGateConsumer (T35 — frag-8 §8.5/§8.6)', () => {
+  describe('initial state', () => {
+    it('starts in idle', () => {
+      const c = new MigrationGateConsumer();
+      expect(c.getMigrationState()).toBe<MigrationState>('idle');
+    });
+
+    it('reports not pending in idle (T10 fast path)', () => {
+      const c = new MigrationGateConsumer();
+      expect(c.isMigrationPending()).toBe(false);
+    });
+  });
+
+  describe('isMigrationPending() boolean projection (T10 contract)', () => {
+    it.each<[MigrationState, boolean]>([
+      ['idle', false],
+      ['pending', true],
+      ['completed', false],
+      ['failed', true],
+    ])('state=%s → isMigrationPending()=%s', (state, expected) => {
+      const c = new MigrationGateConsumer();
+      c.setMigrationState(state);
+      expect(c.isMigrationPending()).toBe(expected);
+    });
+  });
+
+  describe('setMigrationState transitions (spec §8.5)', () => {
+    it('idle → pending notifies subscribers with new state', () => {
+      const c = new MigrationGateConsumer();
+      const listener = vi.fn();
+      c.subscribe(listener);
+      c.setMigrationState('pending');
+      expect(c.getMigrationState()).toBe('pending');
+      expect(listener).toHaveBeenCalledWith('pending');
+      expect(listener).toHaveBeenCalledTimes(1);
+    });
+
+    it('pending → completed flips isMigrationPending() back to false', () => {
+      const c = new MigrationGateConsumer();
+      c.setMigrationState('pending');
+      expect(c.isMigrationPending()).toBe(true);
+      c.setMigrationState('completed');
+      expect(c.isMigrationPending()).toBe(false);
+    });
+
+    it('pending → failed keeps isMigrationPending() true (modal still blocks)', () => {
+      const c = new MigrationGateConsumer();
+      c.setMigrationState('pending');
+      c.setMigrationState('failed');
+      expect(c.getMigrationState()).toBe('failed');
+      expect(c.isMigrationPending()).toBe(true);
+    });
+
+    it('no-op when next state equals current state (no notification)', () => {
+      const c = new MigrationGateConsumer();
+      const listener = vi.fn();
+      c.setMigrationState('pending');
+      c.subscribe(listener);
+      c.setMigrationState('pending'); // re-emit same value
+      expect(listener).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('subscribe / unsubscribe', () => {
+    it('returns disposer that stops further notifications', () => {
+      const c = new MigrationGateConsumer();
+      const listener = vi.fn();
+      const unsubscribe = c.subscribe(listener);
+      c.setMigrationState('pending');
+      unsubscribe();
+      c.setMigrationState('completed');
+      expect(listener).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Independent review (Opus 4.7 1M)

**Verdict: APPROVE with two ambiguities surfaced for manager decision.**

### Layer 1 — ambiguities to flag

**(a) Location: `daemon/src/db/` vs `electron/`** — DEFENSIBLE.
Brief offered both. Worker chose `daemon/src/db/` because the primary consumers per spec §8.5 (T10 interceptor, supervisor RPCs `/healthz` `/stats` `/version` `ping()`) all live in the daemon process. `daemon/src/db/migrate-v02-to-v03.ts:21` already cites "T35 MIGRATION_PENDING gate" anchoring the file there. Electron-main subscribes via the daemon-bridge for the modal driver (frag-8 §8.6 / T33), which is a cross-process consumer, not a reason to host the state holder in electron/. Concur with worker.

**(b) State enum: 3 wire values vs 4 internal values** — NEEDS MANAGER LOCK.
Spec §8.5 explicitly says wire surface is `migrationState: 'pending' | 'completed' | 'failed'` (3 values). Worker added `'idle'` as a fourth internal-only value, justified as boot-default to distinguish "never ran" from "ran successfully" for the supervisor diagnostic. The header comment (lines 52-57) and PR body note the bridge can collapse `idle → completed` at the wire boundary.

Trade-off:
- KEEP `idle` (current PR): type system distinguishes boot-default vs post-success; bridge owns the 3-value wire collapse. Cost: leaks a 4th symbol consumers must handle in switch statements.
- DROP `idle`: initial state is `'completed'` (matches the §8.5 default for fresh installs that never had v0.2 db) OR `undefined`. Honors the 3-value spec literally. Cost: loses the "never ran" signal; supervisor can't tell startup-not-yet-checked from success.

Per `feedback_reviewer_flags_ambiguous_requirements` surfacing both interpretations rather than silent-approving. Spec §8.5 is wire-surface only and does not forbid an internal 4th state, but worker did diverge from the literal spec value count. Manager call needed.

### Layer 1 — other items (all PASS)

- **Class vs module singleton**: class chosen for test isolation. Defensible; tests verified independently constructable.
- **`failed` keeps modal up**: confirmed against spec §8.6. "Single fatal-error modal... only modal action is 'Quit ccsm'... No 'Retry' button — failed quick_check or rename cannot retry within the same boot meaningfully; the structural retry path is the manual cleanup + relaunch flow." So `isMigrationPending() === (state === 'pending' || state === 'failed')` is correct: `failed` MUST keep data RPCs blocked so the modal holds the surface until quit.
- **Event bus collision**: grepped `daemon/src/` for `EventEmitter|eventBus|event-bus|emitter` — only hit is `migration-events.ts` header comment ("NO emitter, NO consumer, NO side effects"). Confirmed no existing bus; bespoke subscribe/notify is appropriate.
- **T10 contract**: `daemon/src/envelope/migration-gate-interceptor.ts` consumes `migrationPending: boolean` only; matches `isMigrationPending()` exactly.

### Layer 2 — implementation review

- 19 tests cover initial state, every transition pair (incl. no-op same-state re-emit), `isMigrationPending()` truth table, single + multi-subscriber notify, idempotent unsubscribe, unsubscribe-from-inside-callback safety (snapshot-then-iterate at line 117), listener error isolation via `queueMicrotask` rethrow.
- Pure state holder: zero I/O, zero event emission, zero RPC dispatch — matches SRP comment lines 14-23.
- Listener error isolation pattern (lines 119-126) is correct: catches per-listener so siblings still fire, rethrows on microtask so unhandled-rejection logging surfaces it.
- `clearSubscribersForTests()` lives on the class with explicit name + comment justifying the SRP boundary trade-off — acceptable.
- `npx tsc --noEmit -p daemon/tsconfig.json` worker-reported clean.
- Local `npx vitest run daemon/src/db/__tests__/migration-gate-consumer.test.ts` → **19/19 pass**, 12ms test time.
- Pre-existing `better-sqlite3` ABI failures (`NODE_MODULE_VERSION 145 vs 137`) in sibling files are env-side, not this PR's code (zero DB I/O here).

### Recommendation

Reviewer approves Layer 2 unconditionally. Layer 1 ambiguity (b) is the only manager decision point — if manager picks "DROP `idle`", that's a 1-line type change + initial-state swap + 1-2 test updates. If manager picks "KEEP `idle`" (current PR), merge as-is.

PR #669, opened 2026-05-01
feat(daemon): T41 fan-out registry

Jiahui-Gu:
## Summary

Task #998 — v0.3 PTY hardening **T41**: per-session fan-out registry. The SINK side of the producer/decider/sink trio (per `feedback_single_responsibility`).

Spec: `docs/superpowers/specs/v0.3-fragments/frag-3.5.1-pty-hardening.md` §3.5.1.5 ("Subscriber fan-out (multi-subscriber semantics)") — single registry per session, callbacks fired in registration order, slow subscriber cannot poison peers.

## API (`daemon/src/pty/fanout-registry.ts`)

```ts
createFanoutRegistry<TMessage>({ onSubscriberError? }) -> {
  subscribe(sessionId, subscriber): unsubscribeFn
  unsubscribe(sessionId, subscriber): void
  broadcast(sessionId, message): void
  getSubscribers(sessionId): ReadonlyArray<Subscriber>
  drainSession(sessionId, reason): void
}

interface Subscriber<TMessage> {
  deliver(message: TMessage): void
  close(reason: DrainReason): void
}

interface DrainReason {
  kind: 'pty-exit' | 'pty-crashed' | 'daemon-shutdown' | 'session-removed'
  detail?: string
}
```

## Invariants

- Per-session `Map<sessionId, Set<Subscriber>>`; insertion order preserved (§3.5.1.5 "registration order").
- `broadcast` snapshots before iterating — `subscribe` / `unsubscribe` re-entrancy inside a `deliver()` callback cannot poison the current broadcast (newcomers wait until next broadcast; self-removers complete this round).
- Subscriber errors in `deliver()` / `close()` are caught and routed to `onSubscriberError` (default `console.warn`). One slow subscriber cannot poison peers.
- `drainSession` evicts the session entry **before** invoking `close()`, so a re-entrant `subscribe()` inside `close()` opens a fresh slot for the new generation.
- Empty sessions pruned on `unsubscribe`; `broadcast` / `drain` on nonexistent sessions are no-ops.
- Unsubscribe fn is idempotent.
- Same `Subscriber` instance on two different sessions is independent.

## Hard non-goals (push back, owned elsewhere)

- **No serialization / framing** — caller passes pre-built message (§3.4.1.c owns the framed buffer).
- **No backpressure / drop-slowest accounting** — T43 owns the 1 MB watermark and aggregate cap (§3.5.1.5 lines 118–119).
- **No PTY event listening** — T42 producer wires `p.onData -> broadcast`.
- **No lifecycle FSM** — T37 lifecycle invokes `drainSession` at exit/crash transitions; this file just executes the bulk-close.

## Test plan

- [x] `npx vitest run daemon/src/pty/__tests__/fanout-registry.test.ts` — **18/18 pass**
  - subscribe + broadcast → all receive in registration order
  - unsubscribe via returned fn / explicit / idempotent / no `close()`
  - multi-session isolation (broadcast / drain don't cross)
  - drainSession invokes `close()` with reason; subsequent broadcast is no-op
  - subscribe-during-broadcast: snapshot iteration (newcomer skipped)
  - unsubscribe-during-broadcast: peers still receive, no crash
  - throw-in-`deliver()` / throw-in-`close()`: peers unaffected, `onSubscriberError` invoked with `{ sessionId, phase }`
  - broadcast / drain on nonexistent session: no-op
  - `getSubscribers` returns mutable snapshot that doesn't mutate registry
  - same Subscriber across sessions is independent
  - double-subscribe dedupes (Set semantics)
- [x] main-process smoke: `npm run build:daemon` + `node -e "import('./daemon/dist-daemon/pty/fanout-registry.js')"` round-trip — OK

## Layer 1 check

Pure SRP sink. No producer logic (PTY listener belongs to T42), no decider logic (drop-slowest belongs to T43). Caller-owned `Subscriber` shape stays opaque (registry never inspects `lastAck` / `streamId` / `bufferQueue`).

### DIFF
--- a/daemon/src/pty/__tests__/fanout-registry.test.ts
+++ b/daemon/src/pty/__tests__/fanout-registry.test.ts
@@ -1,0 +1,254 @@
+import { describe, it, expect, vi } from 'vitest';
+import {
+  createFanoutRegistry,
+  type DrainReason,
+  type Subscriber,
+} from '../fanout-registry.js';
+
+type Msg = string;
+
+function makeSubscriber(): Subscriber<Msg> & {
+  delivered: Msg[];
+  closedWith: DrainReason[];
+} {
+  const delivered: Msg[] = [];
+  const closedWith: DrainReason[] = [];
+  return {
+    delivered,
+    closedWith,
+    deliver(msg) {
+      delivered.push(msg);
+    },
+    close(reason) {
+      closedWith.push(reason);
+    },
+  };
+}
+
+describe('createFanoutRegistry', () => {
+  it('subscribe + broadcast → all subscribers receive in registration order', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    const b = makeSubscriber();
+    reg.subscribe('s1', a);
+    reg.subscribe('s1', b);
+    reg.broadcast('s1', 'hello');
+    expect(a.delivered).toEqual(['hello']);
+    expect(b.delivered).toEqual(['hello']);
+  });
+
+  it('subscribe + unsubscribe via returned fn → next broadcast skips that subscriber', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    const b = makeSubscriber();
+    const offA = reg.subscribe('s1', a);
+    reg.subscribe('s1', b);
+    reg.broadcast('s1', 'first');
+    offA();
+    reg.broadcast('s1', 'second');
+    expect(a.delivered).toEqual(['first']);
+    expect(b.delivered).toEqual(['first', 'second']);
+  });
+
+  it('returned unsubscribe fn is idempotent', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    const off = reg.subscribe('s1', a);
+    off();
+    off(); // must not throw
+    expect(reg.getSubscribers('s1')).toEqual([]);
+  });
+
+  it('explicit unsubscribe(sessionId, sub) works equivalently', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    reg.subscribe('s1', a);
+    reg.unsubscribe('s1', a);
+    reg.broadcast('s1', 'x');
+    expect(a.delivered).toEqual([]);
+  });
+
+  it('explicit unsubscribe does NOT invoke close()', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    reg.subscribe('s1', a);
+    reg.unsubscribe('s1', a);
+    expect(a.closedWith).toEqual([]);
+  });
+
+  it('multiple sessions are isolated — broadcast to A does not reach B', () => {
+    const reg = createFanoutRegistry<Msg>();
+    const a = makeSubscriber();
+    const b = makeSubscriber();
+    reg.subscribe('s1', a);
+    reg.subscribe('s2', b);
+    reg.broadcast('s1', 'for-a');
+    expect(a.delivered).toEqual(['for-a']);
+    expect(b.delivered).toEqual([]);
+    reg.broadcast('s2', 'for-b');
+    expect(a.delivered).toEqual(['for-a']);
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Independent review (Task #998 / T41) — APPROVE

Reviewed against frag-3.5.1 §3.5.1.5 + single-responsibility / producer-decider-sink split.

### Layer 1 — necessity / spec match / direction

- **Per-session map is correct, not flat.** §3.5.1.5: "Single fan-out registry per session ... Each ptySubscribe call registers a callback on the session's `Set<Subscriber>`." Spec is explicit per-session, and "multiple sessions isolated" test enforces it. Flat map would silently cross-deliver — would have failed res-MUST-2.
- **DrainReason vocabulary matches spec.** `pty-exit | pty-crashed | daemon-shutdown | session-removed` — first three appear verbatim in §3.5.1.2 step 4 (`RESOURCE_EXHAUSTED { reason: 'daemon-shutdown' }`) and the lifecycle FSM transitions (`exited`, `crashed`). `session-removed` is a defensive 4th that has no spec citation but is harmless (caller must opt-in by passing it). Acceptable.
- **SRP holds — pure sink.** Subscriber shape is opaque (`deliver` + `close` only). Registry never inspects `lastAck` / `streamId` / `bufferQueue`. No PTY listening (T42), no drop-slowest accounting (T43), no framing/serialization, no lifecycle FSM. Producer/decider/sink split per `feedback_single_responsibility` honoured cleanly. The header comment explicitly enumerates non-goals and says "push back if asked" — exactly the discipline we want.
- **No bleed.** I grepped the file: zero references to node-pty, sockets, Connect, watermarks, byte counters, replay, heartbeat, semaphores. Worker stayed in lane.
- **Aggregate cap (res-P1-4) intentionally NOT here.** Correct — that's T43's drop-slowest decider; the registry is reason-agnostic.
- **Snapshot-then-iterate semantics match §3.5.1.5.** Spec says "registration order, wrapped in try/catch so one slow subscriber cannot poison others." Both broadcast() and drainSession() snapshot via `Array.from(set)` BEFORE iteration. drainSession() also evicts the session entry between snapshot and close-loop, which is a thoughtful detail (close callbacks that re-subscribe land in a fresh slot — matches "drain marks the OLD generation closed").

### Layer 2 — implementation

- **Tests (18 cases) cover the contract surface comprehensively**:
  - subscribe + broadcast (registration order)
  - subscribe + unsubscribe via returned fn (idempotent)
  - explicit unsubscribe + does-not-invoke-close
  - multi-session isolation (broadcast + drainSession)
  - drainSession bulk close + post-drain broadcast no-op
  - subscribe-during-broadcast (newcomer skipped — snapshot semantics)
  - unsubscribe-during-broadcast (self-removal does not crash, peers receive)
  - broadcast / drainSession on nonexistent session (no throw)
  - throw-in-deliver routed to onSubscriberError (peers still get msg)
  - throw-in-close routed to onSubscriberError (peers still closed)
  - getSubscribers returns mutable snapshot copy (mutation does not leak)
  - shared subscriber across two sessions (independent identity)
  - double-subscribe dedupe (Set semantics — delivered once)
- **Verified locally** (pool-9, repo root):
  - `npx vitest run daemon/src/pty/__tests__/fanout-registry.test.ts` → **18/18 passed** in 2.14s.
  - `npx tsc --noEmit` (daemon project) → **clean, no errors**.
- **Implementation quality**:
  - Map<sessionId, Set<Subscriber>> — O(1) add/remove + Set dedupe + insertion-order preservation (matches "registration order" requirement).
  - Empty Set pruning in unsubscribe (`if (set.size === 0) bySession.delete(sessionId)`) — no memory leak from churned sessions.
  - `unsubscribed` latch on returned fn — clean idempotency without re-entering Set.
  - Default `console.warn` logger with `eslint-disable` line comment is acceptable for a daemon module pre-pino-wiring (T42 will inject the pino child).
  - Generic `<TMessage = unknown>` is the right shape — registry has no opinion on payload type.

### Verdict
**APPROVE.** Pure sink, spec-faithful, tests exhaustive, types/tests clean. Manager can squash-merge.

— Reviewer (independent, Opus 4.7 1M, pool-9)

PR #665, opened 2026-05-01
feat(daemon): T16 control-socket dispatcher skeleton (consumes T11 SUPERVISOR_RPCS, stubs T17-T21)

Jiahui-Gu:
## Summary

Task #976 — T16 control-socket dispatcher skeleton. (Re-spec: T11 #646 already
landed the canonical `SUPERVISOR_RPCS` const + `isSupervisorRpc` predicate;
this PR is the **dispatcher itself**, not the const.)

Builds the literal-method router that the supervisor / control-socket
transport (T14, parallel) will mount on its Duplex stream.

- `daemon/src/dispatcher.ts` (NEW): `Dispatcher` class with
  `register / has / dispatch`, plus `createSupervisorDispatcher()` factory
  pre-wired with NOT_IMPLEMENTED stubs for the five canonical RPCs.
- Allowlist enforced via `isSupervisorRpc()` from the T11 module — non-
  allowlisted methods fail closed with `NOT_ALLOWED`. Data-socket has its
  own (separate) allowlist per spec; this dispatcher is supervisor-only.
- `register()` also refuses non-supervisor method names (fail-loud at boot).
- Pure decider: zero socket I/O, no envelope parse/serialise, no deadline /
  metrics enforcement (those live in §3.4.1.f interceptors). T14 owns the
  Duplex wiring; T17–T21 each just `register()` a real handler over the
  matching stub.
- `daemon/src/index.ts`: replaces the `void SUPERVISOR_RPCS` placeholder
  comment (left by T11 explicitly for T16) with
  `createSupervisorDispatcher()` so the router is live at boot.

## Spec citations

- `docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md`
  §3.4.1.h — control-socket: `/healthz`, `/stats`, `daemon.hello`,
  `daemon.shutdown`, `daemon.shutdownForUpgrade`.
- frag-6-7 §6.5 — supervisor transport = the `ccsm-control` socket; this
  dispatcher is the routing surface mounted on it.
- T11 (PR #646) — canonical `SUPERVISOR_RPCS` const + `isSupervisorRpc`
  predicate (single source of truth).

## Hand-off note

T11's hand-off comment in `daemon/src/index.ts` ("T16 owns the control-socket
dispatcher that actually routes on this allowlist") is now resolved. Each of
T17–T21 just calls `supervisorDispatcher.register('<method>', realHandler)`
to swap a stub for the real handler — no further wiring required.

[...]

PR #646, opened 2026-04-30
feat(daemon/envelope): T11 SUPERVISOR_RPCS canonical

Jiahui-Gu:
## Summary

Task #967 [T11] L1 envelope: control-plane carve-out.

Promotes the placeholder \`SUPERVISOR_RPCS\` literal in \`daemon/src/index.ts\` to the canonical module \`daemon/src/envelope/supervisor-rpcs.ts\` per spec **§3.4.1.h** (\`docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md\`).

- New module exports:
  - \`SUPERVISOR_RPCS\` — readonly tuple of the 5 control-plane RPCs.
  - \`isSupervisorRpc(name: string): boolean\` — pure predicate, O(1) Set lookup, no normalisation / prefix matching.
- \`daemon/src/index.ts\` now imports the canonical export (placeholder block removed; \`void\` reference kept until T16 wires the dispatcher).
- \`vitest.config.ts\` include pattern extended for \`daemon/src/**/__tests__/\`.

## Spec citation (5-RPC list)

frag-3.4.1-envelope-hardening.md §3.4.1.h, table row "control-socket":

> RPCs served: \`/healthz\`, \`/stats\`, \`daemon.hello\`, \`daemon.shutdown\` (frag-6-7 §6.4 step 4 graceful drain), \`daemon.shutdownForUpgrade\` (frag-6-7 §6.4 upgrade-in-place + marker).

Cross-referenced by §3.4.1.f migrationGateInterceptor (allowlist source) and frag-8 §8.5 migration scope — single canonical constant prevents the three sites drifting.

## Test plan

- [x] \`npx vitest run daemon/src/envelope --no-coverage\` → 1 file, **19 tests** pass.
  - 5 RPCs each return true via \`it.each\`.
  - 11 non-canonical names return false (\`daemon.foo\`, \`''\`, \`daemon.shutdown.fake\`, case variants, whitespace variants, partial prefixes).
  - Const inventory equality + length assertion.
  - Explicit no-prefix-match guard (\`daemon.shutdownForUpgradeNow\` → false).
- [x] \`npx eslint daemon/src/envelope/ daemon/src/index.ts\` → clean.
- [x] \`npx tsc --noEmit -p daemon/tsconfig.json\` → clean.
- [x] \`npm run build:daemon\` → emits \`dist-daemon/envelope/supervisor-rpcs.{js,d.ts,js.map}\`.

## Out of scope

[...]

PR #672, opened 2026-05-01
feat(daemon): T19 daemon.hello handler (HMAC compute, anti-imposter)

Jiahui-Gu:
## Summary

Implements **T19** — the wire-level `daemon.hello` RPC handler for the control-socket dispatcher slot, per spec **§3.4.1.g** (`docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md`).

**Two-frame HMAC challenge-response, daemon side:**
1. Client sends `clientHelloNonce` (16 random bytes, base64url-no-pad).
2. Daemon computes `helloNonceHmac = HMAC-SHA256(daemon.secret, decoded-16-byte-nonce)` truncated to 16 bytes, base64url, and returns it alongside the protocol/compatibility block.
3. Client verifies via `crypto.timingSafeEqual` (client-side concern; not this handler).

The daemon does **NOT** verify any client HMAC. Client identity is already proven by peer-cred (§3.1.1) + ACL (§7.1). The handler's job is to PROVE the listener possesses `daemon.secret` so the client can rule out an imposter listener.

### Interpretation note (re: brief wording)

The dispatch brief said "HMAC verify" — but per spec §3.4.1.g and PR #663 reviewer confirmation, the daemon **computes** the HMAC for the client to verify; it does not verify a client HMAC. This handler implements the spec-canonical compute-and-reply.

### Anti-imposter (spec lines 195, 208, 212)

The reply MUST carry `helloNonceHmac` even when `compatible: false`, so the client can rule out an imposter listener BEFORE the daemon tears down the socket. Returning a bare error on incompatibility would defeat the whole point of the challenge-response. Three regression tests assert this for each `reason` value.

### Single Responsibility

PURE DECIDER. Inputs: per-call request payload + factory-injected `getSecret()` / `getBootNonce()`. Outputs: a `HelloReplyPayload` or `DaemonHelloSchemaError` on malformed input. No socket I/O, no per-connection state (T8 helloInterceptor owns that), no key material reads (factory injects so tests don't need a real keystore and so secret rotation per frag-6-7 §7.2 can swap the value live).

### Implementation highlights

- **Round-9 base64url lock** (spec §3.4.1.g): HMAC computed over the DECODED 16-byte buffer, NOT the 22-char ASCII string. Regression test asserts the wrong-way value is NOT produced AND the right-way value matches an independently-computed reference.
- **Strict schema validation** on every field the handler reads. Throws `DaemonHelloSchemaError`; T14 control-socket transport maps this to a wire-level `schema_violation` rejection frame.
- **Compatibility check order**: wire → version → frame-version (matches spec listing order so the same multi-failure case always reports the same `reason` across daemons).
- **Factory contract**: `createDaemonHelloHandler({ getSecret, getBootNonce, minClient? })` returns a function matching the dispatcher's `Handler` shape. Both getters are called PER REQUEST so a long-lived dispatcher transparently picks up a rotated secret or a bootNonce allocated after handler construction.

### Relationship to neighbouring tasks

- **T8 (PR #663) helloInterceptor** — slot-#0 envelope interceptor that gates non-hello RPCs (`hello_required`) and rejects replays (`hello_replay`). Currently inlines the HMAC compute itself; a follow-up should refactor it to delegate here so there's a single source of truth.
- **T16 (PR #665) dispatcher** — registers a `NOT_IMPLEMENTED` stub for `daemon.hello`. Once T16 lands, wiring this handler in is one line: `dispatcher.register('daemon.hello', helloHandler)`.
- **T11 (PR #646, merged) SUPERVISOR_RPCS** — the dispatcher allowlist carries the literal `daemon.hello` (control-plane carve-out, spec line 249). The handler exports BOTH `HELLO_METHOD_LITERAL = 'daemon.hello'` and `HELLO_METHOD_NAMESPACED = 'ccsm.v1/daemon.hello'`.
- **T4 (PR #648, merged) hmac.ts** — reused for `computeHmac` / `NONCE_BYTES` / `HMAC_TAG_LENGTH`. No duplication.

## Test vectors (spec §3.4.1.g compliance)

```
TEST_SECRET     = utf8('test-daemon-secret-32-bytes-long-fixture-not-real')
nonce(seed=0xAB) = base64url(Buffer.alloc(16, 0xAB))   = 'q6urq6urq6urq6urq6urqw'
expected HMAC   = base64url(HMAC-SHA256(secret, decoded-nonce-bytes).subarray(0,16))
                ─ matches an independently-computed reference (test asserts equality
                  via createHmac('sha256', secret).update(decoded).digest())
```

A 50-iteration random-nonce sweep cross-checks every produced HMAC against the reference.

## Test plan

- [x] `npm run typecheck` (renderer + electron + daemon) — clean.
- [x] `npx vitest run daemon/src/handlers/__tests__/daemon-hello.test.ts` — **41/41 passed** (3.59s).
- [x] Main-process smoke: `node --input-type=module -e "import('./daemon/dist-daemon/handlers/daemon-hello.js')..."` loads the compiled ESM module and `createDaemonHelloHandler({...})(req)` returns `{ compatible: true, helloNonceHmac.length === 22, protocol.daemonProtocolVersion === 1, bootNonce, ... }`.
- [ ] CI green (Windows + Linux + typecheck + lint).

### Coverage by case

- Wire constants match spec (5 cases).
- Valid request → compatible: true, 22-char HMAC, full protocol block, bootNonce, defaults (6 cases).
- HMAC matches an independently-computed reference; stable across calls; divergent on different nonces / different secrets (3 cases).
- Round-9 base64url-bytes-not-ASCII regression (2 cases).
- Anti-imposter: `compatible: false` paths (wire / version / frame-version) all surface correct `helloNonceHmac` (4 cases).
- Schema rejection: null, non-object, missing/empty/wrong-type fields, wrong-length nonce, invalid base64url alphabet (14 cases).
- Factory contract: per-request `getSecret()` / `getBootNonce()`, custom + default `minClient`, promise rejection on schema error, 50-iteration random nonce sweep (7 cases).

### DIFF
--- a/daemon/src/handlers/__tests__/daemon-hello.test.ts
+++ b/daemon/src/handlers/__tests__/daemon-hello.test.ts
@@ -1,0 +1,440 @@
+import { describe, it, expect } from 'vitest';
+import { Buffer } from 'node:buffer';
+import { createHmac, randomBytes } from 'node:crypto';
+
+import {
+  createDaemonHelloHandler,
+  decideHelloReply,
+  DaemonHelloSchemaError,
+  DAEMON_ACCEPTED_WIRES,
+  DAEMON_DEFAULTS,
+  DAEMON_FEATURES,
+  DAEMON_FRAME_VERSION,
+  DAEMON_WIRE,
+  HELLO_METHOD_LITERAL,
+  HELLO_METHOD_NAMESPACED,
+  type HelloReplyPayload,
+  type HelloRequestPayload,
+} from '../daemon-hello.js';
+import {
+  encode as base64urlEncode,
+  decode as base64urlDecode,
+} from '../../envelope/base64url.js';
+import { DAEMON_PROTOCOL_VERSION } from '../../envelope/protocol-version.js';
+import { HMAC_TAG_LENGTH, NONCE_BYTES } from '../../envelope/hmac.js';
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const TEST_SECRET = Buffer.from(
+  'test-daemon-secret-32-bytes-long-fixture-not-real',
+  'utf8',
+);
+const TEST_BOOT_NONCE = '01HX8VJ6Q9KZ5YBDGXMTNCAVT0';
+
+function makeNonce(seedByte: number): string {
+  // Deterministic 16-byte nonce so reference HMACs are reproducible.
+  const buf = Buffer.alloc(NONCE_BYTES, seedByte);
+  return base64urlEncode(buf);
+}
+
+function validRequest(
+  overrides: Partial<HelloRequestPayload> = {},
+): HelloRequestPayload {
+  return {
+    clientWire: DAEMON_WIRE,
+    clientProtocolVersion: DAEMON_PROTOCOL_VERSION,
+    clientFrameVersions: [DAEMON_FRAME_VERSION],
+    clientFeatures: ['binary-frames', 'hello'],
+    clientHelloNonce: makeNonce(0x42),
+    ...overrides,
+  };
+}
+
+function decide(req: unknown): HelloReplyPayload {
+  return decideHelloReply(req, {
+    secret: TEST_SECRET,
+    bootNonce: TEST_BOOT_NONCE,
+    minClient: 'v0.3',
+  });
+}
+
+/** Independent reference HMAC: never touches the implementation under test. */
+function referenceHmac(secret: Buffer, nonceB64url: string): string {
+  const nonceBytes = base64urlDecode(nonceB64url);
+  const full = createHmac('sha256', secret).update(nonceBytes).digest();
+  return base64urlEncode(full.subarray(0, NONCE_BYTES));
+}
+
+// ---------------------------------------------------------------------------
+// Wire constants
+// ---------------------------------------------------------------------------
+
+describe('daemon.hello wire constants', () => {
+  it('exposes both control-plane literal and data-plane namespaced method names', () => {
+    expect(HELLO_METHOD_LITERAL).toBe('daemon.hello');
+    expect(HELLO_METHOD_NAMESPACED).toBe('ccsm.v1/daemon.hello');
+  });
+
+  it('advertises a single-element wires array on v0.3 (spec §3.4.1.g)', () => {
+    expect(DAEMON_ACCEPTED_WIRES).toEqual([DAEMON_WIRE]);
+    expect(DAEMON_WIRE).toBe('v0.3-json-envelope');
+  });
+
+  it('declares frame-version 0 for v0.3 (spec §3.4.1.c)', () => {
+    expect(DAEMON_FRAME_VERSION).toBe(0);
+  });
+
+  it('mirrors the canonical features list (spec §3.4.1.g schema)', () => {
+    expect(DAEMON_FEATURES).toEqual([
+      'binary-frames',
+      'stream-heartbeat',
+      'interceptors',
+      'traceId',
+      'bootNonce',
+      'hello',
+    ]);
+  });
+
+  it('mirrors the canonical defaults block (spec §3.4.1.g)', () => {
+    expect(DAEMON_DEFAULTS).toEqual({
+      deadlineMs: 5_000,
+      heartbeatMs: 30_000,
+      backpressureBytes: 1_048_576,
+    });
+    // Defensively frozen — handler returns the same object on every call.
+    expect(Object.isFrozen(DAEMON_DEFAULTS)).toBe(true);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Happy path
+// ---------------------------------------------------------------------------
+
+describe('decideHelloReply — valid request', () => {
+  it('returns compatible: true with a 22-char base64url HMAC', () => {
+    const reply = decide(validRequest());
+    expect(reply.compatible).toBe(true);
+    expect(reply.reason).toBeUndefined();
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Independent reviewer — APPROVE

Layer-1 (necessity + direction) and Layer-2 (implementation) both pass. Worker correctly followed spec over brief and surfaced the divergence in the PR body — exemplary application of `feedback_reviewer_flags_ambiguous_requirements`.

### Layer 1 — necessity, direction, brief-vs-spec

- **Spec compliance verified** against `frag-3.4.1-envelope-hardening.md` §3.4.1.g (lines 148-212) and the control-plane carve-out at line 249 (`SUPERVISOR_RPCS` carries the literal `daemon.hello`, exempt from the `ccsm.<wireMajor>/...` namespace rule). Exporting BOTH `HELLO_METHOD_LITERAL = 'daemon.hello'` and `HELLO_METHOD_NAMESPACED = 'ccsm.v1/daemon.hello'` is exactly right — handler is namespace-agnostic, dispatcher registers under the literal.
- **Brief said "HMAC verify" — spec says daemon COMPUTES.** Worker followed spec (correct) and called this out in the PR body. Spec line 173 is unambiguous: `helloNonceHmac: HMAC-SHA256(daemon.secret, clientHelloNonce) ... base64`, returned in the daemon→client reply. Daemon-side verification of a client HMAC is not in the protocol — client identity is already proven by peer-cred (§3.1.1) + ACL (§7.1).
- **Anti-imposter contract honored.** Spec line 208: `On compatible: false, daemon does NOT close the socket immediately — the reply still carries helloNonceHmac (so client can rule out imposter)`. Three regression tests (one per `reason` enum) assert HMAC presence + correctness on `wire-mismatch`, `version-mismatch`, `frame-version-mismatch`. Tight.
- **SRP cut is correct.** Pure decider; secret + bootNonce injected via factory getters (called per-request → supports rotation per frag-6-7 §7.2 without rebuilding the dispatcher registry); no socket I/O, no per-connection state (T8 owns that), no keystore reads in tests. Clean producer/decider/sink discipline.
- **Compatibility check order (wire → version → frame-version)** matches spec listing order; deterministic so multi-failure cases always report the same `reason` across daemons. Test asserts the wire-wins-over-version case.

### Layer 2 — implementation

- **HMAC over decoded bytes, not ASCII** (round-9 base64url lock). Implementation decodes `clientHelloNonce` to a 16-byte buffer before `computeHmac`. Two regression tests (general + all-zero edge case) assert the WRONG-way value is NOT produced. Independent reference HMAC in tests is computed via raw `node:crypto.createHmac` rather than round-tripping the SUT — this is a real test vector, not a tautology. The 50-iteration random-nonce sweep cross-checks every produced HMAC against the same independent reference. Strong coverage of the security-critical compute path.
- **Schema validation** is strict on every field the handler reads (null, non-object, missing/empty/wrong-type/wrong-length/invalid-base64url-alphabet on nonce — 14 cases); lenient on extras so a v0.4 client carrying optional new fields still handshakes. Throws `DaemonHelloSchemaError`; T14 transport will map this to `schema_violation`.
- **Cross-check vs T8 `decideHello`** (`hello-interceptor.ts` from merged 0833d98): T8's `kind: 'reply'` payload is structurally identical to T19's `HelloReplyPayload` (both have `helloNonceHmac` + `protocol{wire,minClient,daemonProtocolVersion,daemonAcceptedWires,features}` + `daemonFrameVersion` + `bootNonce` + `defaults` + `compatible` + optional `reason`). T19's `reason: CompatibilityReason` union is a tighter type than T8's `reason: string` — the follow-up that has T8 delegate to T19 will be a simple swap. No incompatibility blocking landing.
- **Method-name dual export** (`HELLO_METHOD_LITERAL` vs T8's `HELLO_METHOD = 'ccsm.v1/daemon.hello'`): the T19 file's header comment explicitly notes the split is temporary pending T8-delegates-to-T19 cleanup. Acceptable as-is; flagged for the same follow-up.
- **Verifications run locally**: `npx tsc -p daemon/tsconfig.json --noEmit` clean. `npx vitest run daemon/src/handlers/__tests__/daemon-hello.test.ts` → **41/41 pass in 3.26s**.
- **Minor (non-blocking, follow-up)**: dispatcher `Handler` type from T16 is `(req, ctx) => Promise<unknown>` (2 params). `createDaemonHelloHandler` returns `(req) => Promise<HelloReplyPayload>` (1 param) — TS accepts this via contravariant param drop, but the wiring PR may want to either widen the factory return type signature to accept-and-ignore `ctx` or document the drop.

### Status correction (PR body inaccuracy, non-blocking)

The PR body says T16 (#665) and T8 (#663) are still OPEN — both are in fact already merged into `working`:
- T8: commit `0833d98` (merge of #663) — `daemon/src/envelope/hello-interceptor.ts`
- T16: commit `7b4586e` (merge of #665) — `daemon/src/dispatcher.ts` with `createSupervisorDispatcher()` pre-wiring NOT_IMPLEMENTED stubs for the 5 supervisor RPCs

So the wiring follow-up is now unblocked and lands as a small PR:
1. `dispatcher.register('daemon.hello', createDaemonHelloHandler({ getSecret, getBootNonce }))` in `daemon/src/index.ts` (replacing the NOT_IMPLEMENTED stub).
2. Refactor T8 helloInterceptor to delegate its compute branch to the same factory — eliminates the duplicated `DAEMON_FEATURES` / `DAEMON_DEFAULTS` / `DAEMON_FRAME_VERSION` / `DAEMON_WIRE` constants currently living in both files.

### Verdict

**APPROVE** — ready to merge once CI is green. Wiring + T8-delegate-to-T19 dedup is a separate small follow-up.

PR #660, opened 2026-05-01
feat(daemon): T13 runtimeRoot resolver

Jiahui-Gu:
## Summary

Pure helper `daemon/src/sockets/runtime-root.ts` exporting `resolveRuntimeRoot()` — single source of truth for the OS-native runtime/socket directory under which the control-socket, data-socket, marker file, and lockfile siblings all live. Task #970 (T13).

## Spec refs

- `docs/superpowers/specs/v0.3-design.md` §3.1.1 — `<runtimeRoot>` definition (round-9 manager lock + r11 devx P1-4): "explicit OS-native runtime root replaces hardcoded `~/.ccsm/run/...` paths".
- `docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md` (same paragraph; cross-referenced).
- `<dataRoot>` per frag-12 r5 lock #2.

## Per-OS table

| Platform | Resolution | Notes |
|---|---|---|
| Linux | `process.env.XDG_RUNTIME_DIR/ccsm` if set & writable; else `<dataRoot>/run` | XDG-compliant tmpfs preferred; fallback covers systemd-less containers (spec rationale) |
| macOS | `<dataRoot>/run` | No `XDG_RUNTIME_DIR` equivalent on Apple per spec |
| Windows | `<dataRoot>\run` | Named pipes use `\.\pipe\ccsm-{data,control}-<userhash>`; the FS directory holds marker file + lockfile sibling per spec lock |

`<dataRoot>` (frag-12 r5 lock #2):
- Win: `%LOCALAPPDATA%\ccsm` (fallback `~/AppData/Local/ccsm` if env unset)
- macOS: `~/Library/Application Support/ccsm`
- Linux: `$XDG_DATA_HOME/ccsm` if set; else `~/.local/share/ccsm`

## Behavior

- **Idempotent**: safe to call repeatedly; same path returned.
- **mkdir-p**: POSIX dirs created with mode `0o700`; Windows inherits per-user `%LOCALAPPDATA%` ACL (spec note: no UAC, per-user install).
- **`ensure: false`** opt-out for callers that only need the path string (e.g. logging, name negotiation).
- **Pure helper**: no side effects beyond optional `mkdirSync`. Accepts injected `platform` + `env` for testability.

## Tests

`daemon/src/sockets/__tests__/runtime-root.test.ts` — 13 cases:

[...]

PR #673, opened 2026-05-01
feat(daemon): T15 data-socket transport

Jiahui-Gu:
## Summary

T15 — pure-transport listener for the v0.3 **data plane** (every RPC except `SUPERVISOR_RPCS`). Layered above T13 `runtimeRoot` resolver; sibling of T14 control-socket and T16 dispatcher.

**Spec**: `docs/superpowers/specs/v0.3-design.md` §3.4.1.h (two-socket topology) + `frag-3.4.1` §3.4.1.a (round-2 security T15 accept-rate cap).

## Per-OS path table

| OS | Socket path |
|---|---|
| Linux / macOS | `<runtimeRoot>/ccsm-data.sock` |
| Windows | `\.\pipe\ccsm-data` |

`<runtimeRoot>` comes from `resolveRuntimeRoot()` (T13, PR #660). On Windows the runtimeRoot string is a notional anchor only — the named-pipe namespace is global per-user.

## Rate-cap behaviour (frag-3.4.1 §3.4.1.a)

- `MAX_ACCEPT_PER_SEC = 50`, rolling 1s window.
- On excess: `socket.destroy(new Error("accept rate cap exceeded") + code: "EAGAIN")`. Caller's `onConnection` is NOT invoked.
- Once-per-minute aggregate log line `data_socket_accept_rate_cap_drop { dropped, windowMs, maxAcceptPerSec }` (`RATE_CAP_LOG_THROTTLE_MS = 60_000`).
- Budget resets when timestamps fall outside the 1s window — verified by clock-injection test.

## Layer-1 boundary

This module is **pure transport**. It hands the caller a raw `Duplex` and a best-effort `peer: PeerCredentials | undefined`. It does:
- NOT parse envelopes (T16 dispatcher / T18+ wire codec).
- NOT inspect `method`, `headers`, or `SUPERVISOR_RPCS`.
- NOT enforce hello-handshake order (handshake interceptor is the dispatcher's job).

Per-connection peer-cred lookup is best-effort: POSIX returns the daemon's own uid/gid as a forensic anchor (true `SO_PEERCRED` requires a native binding deferred to a future task); Windows returns `undefined`. The §3.4.1.g hello-HMAC handshake remains the authoritative imposter check; this module's chmod 0600 + per-user pipe namespace are defence-in-depth.

## File-system hardening (POSIX)

- `chmod 0600` applied to the socket node after `listen()` returns (parent dir is already `0700` from `runtime-root.ts`).
- Stale socket node from a crashed daemon is unlinked before bind (avoids `EADDRINUSE`).
- `close()` removes the socket file.

## Test plan

- [x] `dataSocketPath` per-OS (Linux/macOS join, Windows literal pipe).
- [x] Two-way echo over a real `net` connection.
- [x] Three concurrent connections — all observed via `onConnection`.
- [x] Socket file mode `0600` (skipped on Windows).
- [x] `close()` removes the socket node, further connects rejected.
- [x] Stale-file unlink path exercised.
- [x] Rate-cap drops connections beyond `maxAcceptPerSec=5` within 1s; clock advance restores budget; aggregate log line emitted.
- [x] `MAX_ACCEPT_PER_SEC === 50` constant assertion (spec lock).
- [x] `npm run typecheck` passes (root + electron + daemon projects).
- [x] Daemon `vitest` for the new file: 7 passed / 4 skipped (Windows). Pre-existing `daemon/src/db/` failures (better-sqlite3 native binding) are unrelated.
- [x] Main-process load smoke: `import('./daemon/dist-daemon/sockets/data-socket.js')` resolves with all four exports.

## Out of scope

- Native `SO_PEERCRED` / `GetNamedPipeClientProcessId` binding — future task.
- Hello-handshake enforcement — T18+ (interceptor pipeline).
- Envelope parsing / dispatch — T16.

### DIFF
--- a/daemon/src/sockets/__tests__/data-socket.test.ts
+++ b/daemon/src/sockets/__tests__/data-socket.test.ts
@@ -51,9 +51,17 @@ describe('dataSocketPath', () => {
     expect(p).toBe(join('/run/user/1000/ccsm', 'ccsm-data.sock'));
   });
 
-  it('returns Windows named pipe regardless of runtimeRoot', () => {
+  it('returns Windows named pipe with userhash suffix regardless of runtimeRoot', () => {
     const p = dataSocketPath('C:\\Users\\me\\AppData\\Local\\ccsm\\run', 'win32');
-    expect(p).toBe('\\\\.\\pipe\\ccsm-data');
+    expect(p).toMatch(/^\\\\\.\\pipe\\ccsm-data-[0-9a-f]{8}$/);
+  });
+
+  it('uses the supplied userhash override on Windows (anti-collision: distinct users → distinct paths)', () => {
+    const a = dataSocketPath('C:\\anywhere', 'win32', 'aaaaaaaa');
+    const b = dataSocketPath('C:\\anywhere', 'win32', 'bbbbbbbb');
+    expect(a).toBe('\\\\.\\pipe\\ccsm-data-aaaaaaaa');
+    expect(b).toBe('\\\\.\\pipe\\ccsm-data-bbbbbbbb');
+    expect(a).not.toBe(b);
   });
 });
 
--- a/daemon/src/sockets/__tests__/data-socket.test.ts
+++ b/daemon/src/sockets/__tests__/data-socket.test.ts
@@ -1,0 +1,304 @@
+// T15 data-socket transport tests.
+// Spec: docs/superpowers/specs/v0.3-design.md §3.4.1.h (two-socket topology)
+//       + frag-3.4.1 §3.4.1.a (50/sec accept-rate cap).
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, rmSync, statSync, existsSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { createConnection, type Socket } from 'node:net';
+import {
+  createDataSocketServer,
+  dataSocketPath,
+  MAX_ACCEPT_PER_SEC,
+  type DataSocketConnection,
+} from '../data-socket.js';
+
+const isWindows = process.platform === 'win32';
+
+let scratch: string;
+
+beforeEach(() => {
+  scratch = mkdtempSync(join(tmpdir(), 'ccsm-data-socket-'));
+});
+
+afterEach(() => {
+  try {
+    rmSync(scratch, { recursive: true, force: true });
+  } catch {
+    /* best-effort */
+  }
+});
+
+function clientConnect(addr: string): Promise<Socket> {
+  return new Promise((resolve, reject) => {
+    const sock = createConnection(addr);
+    sock.once('connect', () => resolve(sock));
+    sock.once('error', reject);
+  });
+}
+
+function readOnce(sock: Socket): Promise<Buffer> {
+  return new Promise((resolve, reject) => {
+    sock.once('data', resolve);
+    sock.once('error', reject);
+  });
+}
+
+describe('dataSocketPath', () => {
+  it('returns POSIX path under runtimeRoot for non-Windows', () => {
+    const p = dataSocketPath('/run/user/1000/ccsm', 'linux');
+    expect(p).toBe(join('/run/user/1000/ccsm', 'ccsm-data.sock'));
+  });
+
+  it('returns Windows named pipe regardless of runtimeRoot', () => {
+    const p = dataSocketPath('C:\\Users\\me\\AppData\\Local\\ccsm\\run', 'win32');
+    expect(p).toBe('\\\\.\\pipe\\ccsm-data');
+  });
+});
+
+describe('createDataSocketServer — basic transport', () => {
+  it('starts and lets a client connect, data flows both ways', async () => {
+    let received: DataSocketConnection | undefined;
+    const server = createDataSocketServer({
+      runtimeRoot: scratch,
+      onConnection: (conn) => {
+        received = conn;
+        conn.socket.on('data', (chunk) => {
+          // Echo back uppercased
+          conn.socket.write(chunk.toString('utf8').toUpperCase());
+        });
+      },
+    });
+    await server.listen();
+    try {
+      const client = await clientConnect(server.address());
+      client.write('hello');
+      const reply = await readOnce(client);
+      expect(reply.toString('utf8')).toBe('HELLO');
+      expect(received).toBeDefined();
+      expect(received!.socket).toBeDefined();
+      // Peer creds: defined on POSIX (best-effort uid), undefined on Windows.
+      if (isWindows) {
+        expect(received!.peer).toBeUndefined();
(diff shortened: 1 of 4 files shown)


Jiahui-Gu (commented):
Reviewer report — independent review (T15 data-socket). Posting as COMMENT (same-identity block prevents formal request-changes); manager: please treat the userhash item below as a blocking finding.

## Layer 1 — necessity / approach: mostly correct, ONE spec deviation

- Boundary is right: pure transport, no envelope parsing, no method-name awareness, explicit producer/decider/sink note in the header. T16 dispatcher remains free to own envelopes.
- Rate cap (50/s sliding window, EAGAIN drop, once-per-min aggregate log) matches frag-3.4.1 §3.4.1.a verbatim.
- Best-effort peer-cred with hello-HMAC (§3.4.1.g) as the authoritative imposter check is the correct posture; native binding deferral is acceptable and documented.

## Blocking finding — Windows named-pipe path is missing the `-<userhash>` suffix

Spec citations (three independent locations all consistent):

- `docs/superpowers/specs/v0.3-design.md` line 267: `\.\pipe\ccsm-data-<userhash>`
- `docs/superpowers/specs/v0.3-design.md` line 272: `named pipes namespace (\.\pipe\ccsm-{data,control}-<userhash>)`
- `docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md` line 237 (same table)

`daemon/src/sockets/data-socket.ts` line 100-102 currently emits the bare literal `\.\pipe\ccsm-data` (no userhash).

Why this matters (not just spec-text pedantry):

1. `\.\pipe\` is a machine-global namespace. On multi-user Windows hosts (Citrix, RDS, shared workstations, dev VMs with multiple logged-in users), the first daemon to bind wins and locks every other user out — or worse, an unprivileged second user can bind first and the §3.4.1.g HMAC handshake is now the only line of defence (imposter wins the bind race trivially without the userhash).
2. The control-socket (T14) will hit the same issue. Locking this in for T15 sets the wrong template.
3. The userhash is the canonical Windows substitute for the POSIX file-ACL — without it the shared peer-cred / DACL / accept-rate-cap / hello-handshake posture claim in §3.4.1.h is materially weaker on Windows than on POSIX.

Suggested fix (small): derive a stable per-user hash (e.g. first 8 hex of `crypto.createHash('sha256').update(os.userInfo().username + os.hostname()).digest('hex')`) inside `dataSocketPath()` for the win32 branch, plumb through tests. The `runtime-root.ts` module is the natural sibling for a `userHash()` helper since T14 will need it too.

## Layer 2 — implementation: clean

- POSIX 0600 chmod sequenced after listen() returns — correct ordering, non-fatal failure logged.
- Stale-socket unlink uses `statSync().isSocket()` guard so it cannot accidentally unlink a regular file at the same path.
- EAGAIN drop: `Object.assign(new Error(...), { code: 'EAGAIN' })`, aggregate counter resets only inside the throttle branch — accumulator semantics correct.
- close() removes the POSIX socket node so next listen() is clean; ENOENT swallowed.
- Local verification on Win 11 25H2: vitest `daemon/src/sockets/__tests__/data-socket.test.ts` → 7 pass / 4 skip (POSIX-only ACL + rate-cap probes). `npm run typecheck` clean across all three tsconfigs.

## Nits (non-blocking)

- The unlinks stale socket file from a prior crashed daemon test (line 180) doesn't actually leave a stale socket node behind — its own comment admits it only exercises the post-close ENOENT branch. Production unlink path is correct, but the test name overpromises. Worth tightening when convenient.
- `recentAccepts.shift()` in the rolling-window loop is O(k) per accept with array reindex; at the 50/s ceiling this is irrelevant and well-commented.

## Verdict

REQUEST_CHANGES (posted as comment due to same-identity block). One small follow-up commit on the Windows pipe path + a `runtime-root` sibling `userHash()` helper unblocks merge. Everything else is approve-grade.

Jiahui-Gu:
Addressed reviewer Layer-1 blocker: Win named-pipe path now `\.\pipe\ccsm-data-<userhash>` per v0.3-design.md L267+L272 and frag-3.4.1 L237.

- New `userHash()` helper in `daemon/src/sockets/runtime-root.ts` — 8-hex SHA-256(`<user>@<host>`), pure / deterministic / no I/O. Sibling export so T14 can refactor to share later (T14 not touched in this PR).
- `dataSocketPath()` win32 branch appends `-${userHash()}`; accepts an override arg for tests.
- Tests added: `userHash` determinism, 8-hex shape, anti-collision across distinct usernames + distinct hostnames; `dataSocketPath` regex match + explicit override anti-collision.

Verification on this commit (e41c5e5):
- typecheck PASS
- vitest `data-socket.test.ts` + `runtime-root.test.ts`: 26 passed | 4 skipped (POSIX-only file-ACL cases) on Win32
- `npm run build:daemon` PASS
- Smoke load: `import('./daemon/dist-daemon/sockets/runtime-root.js')` → `userHash()` returns `c0ea1bfb` (8 lowercase hex)

Jiahui-Gu (commented):
REVIEW VERDICT: APPROVE (same-identity self-approve blocked, posting as comment for manager merge).

Re-review after Win userhash fix (e41c5e5).

Layer-1 spec compliance (prior blocker): RESOLVED.
- runtime-root.ts userHash(): pure, deterministic, 8 lowercase hex of SHA-256(`<user>@<host>`), no I/O, optional username/host injection for tests.
- data-socket.ts dataSocketPath() win32 branch: returns `\.\pipe\ccsm-data-<userhash>` matching v0.3-design.md L267 + frag-3.4.1 L237 verbatim. POSIX branch unchanged (`<runtimeRoot>/ccsm-data.sock`).
- hashOverride param keeps the function pure-testable while production calls fall through to userHash() default.

Verification on Win 11 (pool-6, branch tip e41c5e5):
- npm run typecheck — clean (exit 0, all 3 tsconfigs).
- npx vitest run runtime-root.test.ts data-socket.test.ts — 26 pass / 4 skip (POSIX-only listen tests skipped on win32 as expected).
- New tests cover: 8-hex shape, determinism, distinct-username collision-resistance, distinct-host collision-resistance, no-arg fallback to os.userInfo()/hostname(); plus dataSocketPath win32 shape regex + injected-hash anti-collision pair.
- Compiled smoke: `node -e "import('./daemon/dist-daemon/sockets/runtime-root.js').then(m => console.log(m.userHash()))"` → `c0ea1bfb` (8-hex, deterministic).

Layer-2 nits (non-blocking):
- Doc comment on userHash() correctly cites v0.3-design.md L267+L272 and frag-3.4.1 L237, including the multi-user / Citrix / RDS rationale.
- Inline comment on win32 branch in dataSocketPath flags runtimeRoot as 'notional anchor only' — accurate, since named-pipe namespace is global.
- T14 control-socket still has its own private userHash; doc comment on the new export already calls this out as a future refactor target. Fine to defer.

CI ignored per migration-window tolerance. Manager OK to merge.

PR #675, opened 2026-05-01
feat(daemon): T43 drop-slowest 1MB watermark

Jiahui-Gu:
## Summary

Pure decider primitive for the per-subscriber pending-bytes watermark mandated by frag-3.5.1 §3.5.1.5 (lines 118, 134, 165, 201). New module `daemon/src/pty/drop-slowest.ts` exposes `createDropSlowest({ thresholdBytes })` returning `{ record, flush, reset, forget, getOverlimit, pending }`.

- **Threshold**: `1_048_576` bytes (1 MiB), strict greater-than. Exactly 1 MiB pending is NOT overlimit — matches the spec's "past 1 MB" wording (line 118) and the acceptance "after exactly 1 MB" (line 134) which is read as `> 1 MiB` since the drop fires on the write that pushes past.
- **Single Responsibility**: producer of pending byte tally + decider for the predicate. NOT a sink — caller (fan-out registry) is responsible for closing the slow stream with `RESOURCE_EXHAUSTED`, emitting the `subscriber-dropped-slow` log line, and unregistering.
- **Replay-burst exemption seam (T47 / task #1004 — separate)**: contract exposes two compatible seams so this file does not need to reopen when T47 lands:
  - `record(subId, byteCount, { exempt: true })` — short-circuits accounting; T47 wires the bounded 256 KB replay write through this.
  - `reset(subId)` — alternative path: include replay in tally then clear before steady-state begins. Caller picks one; both behaviours covered by tests.
  Documented at length in the file header so the cross-task contract is discoverable from the call site.

## Spec references
- `docs/superpowers/specs/v0.3-fragments/frag-3.5.1-pty-hardening.md` line 118 (drop-slowest contract), line 120 (replay-burst exemption clarification), line 134 (unit acceptance: "after exactly 1 MB"), line 165 (Task 11c explicit replay-exemption test mandate), line 201 (256 KB vs 1 MB orthogonality).

## Test plan
- [x] `npx vitest run daemon/src/pty/__tests__/drop-slowest.test.ts` → 19/19 pass (boundary, multi-subscriber, flush clamp, exemption seam, input validation).
- [x] `npx tsc -p daemon/tsconfig.json --noEmit` clean.
- [x] Require-load smoke against emitted `dist-daemon/pty/drop-slowest.js` (1 MiB + 1 byte → overlimit list `['a']`).
- [ ] Wired into `daemon/src/pty/fanout-registry.ts` — out of scope for this task; T43 is the pure decider only. Wiring task is downstream.

### DIFF
--- a/daemon/src/pty/__tests__/drop-slowest.test.ts
+++ b/daemon/src/pty/__tests__/drop-slowest.test.ts
@@ -1,0 +1,192 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES,
+  createDropSlowest,
+} from '../drop-slowest.js';
+
+describe('drop-slowest: spec-default constant (frag-3.5.1 §3.5.1.5 line 118)', () => {
+  it('exports threshold = 1 MiB (1_048_576 bytes)', () => {
+    expect(DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES).toBe(1024 * 1024);
+    expect(DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES).toBe(1_048_576);
+  });
+});
+
+describe('drop-slowest: construction', () => {
+  it('rejects non-positive thresholdBytes', () => {
+    expect(() => createDropSlowest({ thresholdBytes: 0 })).toThrow(RangeError);
+    expect(() => createDropSlowest({ thresholdBytes: -1 })).toThrow(RangeError);
+    expect(() => createDropSlowest({ thresholdBytes: 1.5 })).toThrow(RangeError);
+    expect(() => createDropSlowest({ thresholdBytes: Number.NaN })).toThrow(RangeError);
+  });
+
+  it('defaults to the spec threshold', () => {
+    const d = createDropSlowest();
+    d.record('s', DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES);
+    expect(d.getOverlimit()).toEqual([]);
+    d.record('s', 1);
+    expect(d.getOverlimit()).toEqual(['s']);
+  });
+});
+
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (recorded as comment due to same-identity self-approve block).

Layer 1 (architecture):
- Pure decider confirmed: no socket close / log / metric in drop-slowest.ts. Caller (fan-out registry) reads getOverlimit() and owns the close + log + unregister.
- Strict greater-than threshold confirmed (drop-slowest.ts L181: `if (bytes > thresholdBytes)`). Exactly 1 MiB pending = NOT overlimit, matching spec line 134 wording.
- Replay-burst exemption seam (T47 / #1004) documented in file header L18-33 with two compatible paths: `{ exempt: true }` short-circuit AND `reset(subId)` post-replay. T47 picks one without reopening this file.
- Single-Responsibility (producer + decider only) explicit in header L12-16.

Layer 2 (implementation):
- Contract surface complete: record / flush / reset / forget / getOverlimit / pending.
- Boundary tested: 1 MiB exact stays under, +1 byte trips (test L33-47).
- Race-safe: flush over-decrement clamps at 0; unknown subscriber on flush silently ignored; forget idempotent.
- Map-based insertion-order preservation gives stable getOverlimit() output as documented.
- Input validation: RangeError on negative / non-integer / NaN / Infinity for record + flush; constructor rejects non-positive thresholdBytes.
- Zero-byte record registers the subscriber so pending() returns 0 (not undefined) — consistent contract.

Verification (locally re-run in pool-15):
- npx vitest run drop-slowest.test.ts → 19/19 pass (2.34s).
- tsc -p daemon/tsconfig.json --noEmit → clean.

Nit (non-blocking): assertNonNegByteCount checks Number.isFinite AND Number.isInteger — isInteger implies isFinite so the isFinite check is redundant. Harmless.

Ready to merge.

PR #678, opened 2026-05-01
feat(daemon): T21 daemon.shutdownForUpgrade + atomic marker

Jiahui-Gu:
## Summary

Implements **T21** — `daemon.shutdownForUpgrade` control-socket RPC + atomic shutdown-marker writer.

**Spec**: docs/superpowers/specs/v0.3-design.md / frag-6-7 §6.4 — *"daemon.shutdownForUpgrade RPC + shutdown marker"* (`[manager r7 lock: r6 packaging P0-2 cross-frag — daemon.shutdownForUpgrade RPC + marker semantics]`).

Distinct from T20 `daemon.shutdown` (uninstall path, no marker, 2 s force-kill): T21 writes the load-bearing marker file the next-boot supervisor (§6.1 R2) uses to suppress the crash-loop counter for a planned upgrade.

## Architecture (Layer 1: Single Responsibility)

- **Producer** — control-socket transport (T16) delivers empty `req` + `ShutdownForUpgradeContext { version, now, markerDir }`.
- **Decider** — `planShutdownForUpgrade(req, ctx)` is pure: returns `{ ack, markerPayload, markerPath, markerTmpPath, planSteps }`. Zero I/O. Reads the clock exactly once.
- **Sink** — `executeShutdownForUpgrade(plan, actions)` iterates `planSteps` invoking the injected `ShutdownForUpgradeActions { writeMarker, runShutdownSequence, releaseLock, exit }`. **Handler NEVER calls `process.exit` directly** — exit is an injected action so tests verify behaviour without terminating the worker.

## Marker schema (matches T22 reader exactly)

```json
{ "reason": "upgrade", "version": "<daemon-semver>", "ts": <epoch_ms> }
```

`reason` is the literal `'upgrade'` constant (`SHUTDOWN_MARKER_REASON_UPGRADE`); the T22 reader's `ShutdownMarkerPayload.reason` widens to `'upgrade' | string` for forward-compat, but this writer is the single producer and stays narrow.

## Atomic write protocol (defaultWriteShutdownMarker)

```
open(`<runtimeRoot>/daemon.shutdown.tmp`, 'wx', 0o600)   // O_CREAT|O_EXCL|O_WRONLY, user-only
  → handle.writeFile(JSON.stringify(payload))
  → handle.sync()                                        // fsync(file) — payload durable
  → handle.close()
  → fs.rename(tmpPath, finalPath)                        // atomic on POSIX + NTFS (MoveFileExW)
  → fsync(dir, best-effort)                              // EPERM/EISDIR swallowed on Win
```

**EEXIST on the tmp** triggers a single unlink+retry. Stale tmp from a previous crashed write is unrecoverable garbage — the daemon is the single writer per §6.4 — and refusing to retry would brick every subsequent upgrade after a single crashed shutdown.

**Failure mode**: if `writeMarker` throws, the remaining steps DO NOT execute; the error propagates so the caller (Electron-main) falls back to its 5 s force-kill (frag-11 §11.6.5 step 4). This preserves the spec invariant — a failed marker write means "no upgrade marker" → next boot applies normal crash-loop accounting (conservative correct behaviour).

## Atomicity proof (covered by tests)

- **Synthetic crash before rename leaves NO `daemon.shutdown`** — only the tmp. T22 reader sees ENOENT → `{kind: 'absent'}`, the conservative "previous boot was a crash" reading. Verified explicitly in `atomic: a synthetic crash before rename leaves NO daemon.shutdown`.
- **Half-flushed FINAL marker is treated as PRESENT** — verified the T22 rel-S-R8 invariant (an empty file post-rename → `{kind: 'present'}`) still holds, so a power-cut between rename and journal-flush still suppresses the crash-loop counter.
- **Stale tmp recovery** — pre-existing `daemon.shutdown.tmp` does not break the next upgrade attempt.

## Wire ack timing

`makeShutdownForUpgradeHandler` returns the ack from the decider, then schedules `executeShutdownForUpgrade` via `queueMicrotask` so the wire reply is flushed BEFORE side effects begin. Matches frag-11 §11.6.5 "5 s ack window" — the 5 s budget covers the full marker → drain → unlock → exit window starting AFTER the supervisor receives `{accepted: true}`.

## Test plan

- [x] `npx vitest run daemon/src/handlers/__tests__/daemon-shutdown-for-upgrade.test.ts` — **20/20 pass** (3.0 s)
  - decider: ack shape, payload shape, path computation, planSteps order, purity, single-clock-read
  - sink: ordered execution (marker → drain → unlock → exit), abort-on-marker-failure, no direct `process.exit`
  - atomic write (real fs): T22-reader round-trip, `.tmp` cleanup, stale-tmp recovery, synthetic crash before rename = absent, half-flushed final = present
  - mocked-fs sequence: `open(tmp,'wx') → writeFile → fsync(file) → close → rename → fsync(dir)`
  - dispatcher adapter: ack returned BEFORE side effects begin, error surfaced via `onError`
  - e2e round-trip: decider + default sink + T22 reader yield exact payload
- [x] `npm run build:daemon` — clean
- [x] Require-load smoke: `node -e "import('./daemon/dist-daemon/handlers/daemon-shutdown-for-upgrade.js')"` — exports `{planShutdownForUpgrade, executeShutdownForUpgrade, defaultWriteShutdownMarker, makeShutdownForUpgradeHandler, SHUTDOWN_MARKER_REASON_UPGRADE, DAEMON_SHUTDOWN_MARKER_TMP_SUFFIX}`
- [x] Full daemon vitest — only 2 pre-existing failures (better-sqlite3 native binding on Windows worker; unrelated to this PR; T21 + all other handler tests green)

## Follow-up (not in this PR)

- Wiring into the T16 dispatcher (`dispatcher.register('daemon.shutdownForUpgrade', makeShutdownForUpgradeHandler(ctx, actions, onError))`) — separate post-T21 commit per the T16 stub-replacement pattern (same as T17 healthz).
- Default `runShutdownSequence` action implementation lives in T20-adjacent code (§6.6.1 drain ordering) — this PR keeps the action injected so T20 + T21 share one drain implementation when wired.

### DIFF
--- a/daemon/src/handlers/__tests__/daemon-shutdown-for-upgrade.test.ts
+++ b/daemon/src/handlers/__tests__/daemon-shutdown-for-upgrade.test.ts
@@ -1,0 +1,463 @@
+// T21 — daemon.shutdownForUpgrade handler tests.
+// Spec: docs/superpowers/specs/v0.3-design.md / frag-6-7 §6.4
+// "`daemon.shutdownForUpgrade` RPC + shutdown marker".
+
+import { promises as fs } from 'node:fs';
+import { mkdtemp, readFile, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import {
+  DAEMON_SHUTDOWN_MARKER_FILENAME,
+  readMarker,
+} from '../../marker/reader.js';
+import {
+  DAEMON_SHUTDOWN_MARKER_TMP_SUFFIX,
+  SHUTDOWN_MARKER_REASON_UPGRADE,
+  defaultWriteShutdownMarker,
+  executeShutdownForUpgrade,
+  makeShutdownForUpgradeHandler,
+  planShutdownForUpgrade,
+  type ShutdownForUpgradeActions,
+  type ShutdownForUpgradeContext,
+  type ShutdownForUpgradePlan,
+} from '../daemon-shutdown-for-upgrade.js';
+
+const FROZEN_TS = 1_700_000_000_000;
+
+function makeCtx(overrides: Partial<ShutdownForUpgradeContext> = {}): ShutdownForUpgradeContext {
+  return {
+    version: '0.3.0-test',
+    now: () => FROZEN_TS,
+    markerDir: '/var/run/ccsm',
+    ...overrides,
+  };
+}
+
+function makeRecordingActions(overrides: Partial<ShutdownForUpgradeActions> = {}): {
+  actions: ShutdownForUpgradeActions;
+  calls: string[];
+  exitCode: number | null;
+} {
+  const calls: string[] = [];
+  let exitCode: number | null = null;
+  const actions: ShutdownForUpgradeActions = {
+    writeMarker: async () => {
+      calls.push('writeMarker');
+    },
+    runShutdownSequence: async () => {
+      calls.push('runShutdownSequence');
+    },
+    releaseLock: async () => {
+      calls.push('releaseLock');
+    },
+    exit: (code) => {
+      calls.push(`exit:${code}`);
+      exitCode = code;
+    },
+    ...overrides,
+  };
+  return {
+    actions,
+    calls,
+    get exitCode(): number | null {
+      return exitCode;
+    },
+  } as { actions: ShutdownForUpgradeActions; calls: string[]; exitCode: number | null };
+}
+
+// ---------------------------------------------------------------------------
+// planShutdownForUpgrade — pure decider
+// ---------------------------------------------------------------------------
+
+describe('planShutdownForUpgrade — decider (pure)', () => {
+  it('returns ack { accepted: true, reason: "upgrade" }', () => {
+    const plan = planShutdownForUpgrade({}, makeCtx());
+    expect(plan.ack).toEqual({
+      accepted: true,
+      reason: SHUTDOWN_MARKER_REASON_UPGRADE,
+    });
+  });
+
+  it('markerPayload shape matches T22 reader expectation (reason/version/ts)', () => {
+    const plan = planShutdownForUpgrade({}, makeCtx({ version: '1.2.3' }));
+    expect(plan.markerPayload).toEqual({
+      reason: 'upgrade',
+      version: '1.2.3',
+      ts: FROZEN_TS,
+    });
+  });
+
+  it('markerPath = <markerDir>/daemon.shutdown ; tmp = path + ".tmp"', () => {
+    const plan = planShutdownForUpgrade({}, makeCtx({ markerDir: '/runtime' }));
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (reviewer report — manager merges) — T21 daemon.shutdownForUpgrade + atomic marker

Layer 1 (architecture):
- Pure decider `planShutdownForUpgrade` (zero I/O, only `ctx.now()`); sink `executeShutdownForUpgrade` iterates `planSteps`. Single-Responsibility split is clean.
- Atomic write protocol matches spec frag-6-7 §6.4 step 1 exactly: `open(.tmp, 'wx', 0o600)` (= O_CREAT|O_EXCL|O_WRONLY) → writeFile → fsync(file) → close → rename → best-effort fsync(dir). Windows EPERM/EISDIR on dir.sync swallowed with rationale (NTFS journals the rename).
- Marker payload `{reason: 'upgrade', version, ts}` shape matches T22 reader's `ShutdownMarkerPayload` exactly (verified against `daemon/src/marker/reader.ts:35-39` + `isShutdownMarkerPayload`).
- Ack-before-drain ordering: `makeShutdownForUpgradeHandler` returns `plan.ack` synchronously and schedules side effects via `queueMicrotask` so the wire reply flushes before the marker write begins (frag-11 §11.6.5).
- `process.exit` only invoked through injected `actions.exit`; test spy-asserts `process.exit` is never called directly from the sink (Layer-1 violation guard).
- T20 vs T21 distinction explicit in comments (lines 38-40); only T21 writes the marker for the §6.1 R2 marker-aware crash-loop skip.
- EEXIST → unlink + retry once is well-justified: daemon is single writer per §6.4, refusing to retry would brick every upgrade after a single crashed shutdown.

Layer 2:
- 20/20 vitest pass locally; daemon `tsc --noEmit` clean.
- Atomic crash-before-rename test asserts `kind: 'absent'` (conservative — no marker until rename completes).
- Half-flushed FINAL marker test asserts `kind: 'present'` per rel-S-R8.
- Stale-tmp recovery test exercises the EEXIST retry path end-to-end.
- Mocked-fs sequence test pins the exact open→writeFile→fsync→close→rename ordering invariants.
- Round-trip test with the real T22 `readMarker` confirms wire-format compatibility.
- `makeShutdownForUpgradeHandler` returns `(req: unknown) => Promise<Ack>` matching `dispatcher.register` `Handler` type; `daemon.shutdownForUpgrade` is already in `SUPERVISOR_RPCS`. Wiring follow-up acknowledged in jsdoc lines 297-301.

Nits (non-blocking):
- `dirnameOf` reinvents `node:path.dirname` which already handles both POSIX and Windows separators; could shrink to a one-line import. Local helper is fine either way.
- Fire-and-forget shutdown means a >5s drain is invisible to the caller; spec correctly puts that fallback on Electron-main (acknowledged in handler jsdoc).

Safety-critical atomic marker semantics are correct. Recommend MERGE.

PR #679, opened 2026-05-01
feat(daemon): T44 stream-dead detector

Jiahui-Gu:
## Summary
Task #1000 / T44 — pure decider that detects subscribers whose ack stream has gone silent. Lives at the layer above T41 fan-out registry; caller wires returned ids to registry `unsubscribe` + `socket.destroy()` + the `server_stream_dead` log line.

## Spec
- `docs/superpowers/specs/v0.3-design.md` frag-3.5.1 §3.5.1.4 — client-side dead-stream detection at `2 x heartbeatMs + 5 s`; cross-frag handoff to symmetric server-side.
- `docs/superpowers/specs/v0.3-design.md` frag-6-7 §6.5.1 (lines 1695-1714, round-7 manager lock, r6 reliability P0-R2) — canonical server-side stream-dead detector. Daemon maintains `lastClientActivityAt`; the shared heartbeat scheduler checks `now - lastClientActivityAt > 2 x heartbeatMs + 5 s`. v0.3 local-pipe transport makes this a no-op in practice (kernel `'close'` fires first), but byte-compatible with the v0.5 CF Tunnel TCP swap where half-open sockets need an application-layer detector.

## Deadline contract
- `deadlineMs` is computed by the caller (typically `2 x heartbeatMs + 5_000` per spec §6.5.1 — formula owned by the heartbeat scheduler so a single source of truth survives `x-ccsm-heartbeat-ms` renegotiation, frag-3.4.1 §3.4.1.c).
- Boundary: `check(now)` returns ids whose `lastAck` is **strictly older** than `now - deadlineMs`. Equal age = alive.
- Sort: ASCII lexical, deterministic for the spec's single-aggregated-log-line rule.

## Single Responsibility
- **Producer**: caller feeds `onAck(subscriberId)` on inbound RPC / client frame on the same connection (spec line 1701).
- **Decider**: pure `check(now)` — read-only over the in-memory map.
- **Not a sink**: does NOT close sockets, does NOT touch the fan-out registry, does NOT emit log lines. Wiring layer owns those side effects (spec §6.5.1 steps 2-4).
- **Not a heartbeat sender**: T42 owns the shared heartbeat scheduler; this module only watches the OTHER direction (subs that stop ack'ing).

## Caller wiring note
Wiring layer (T42 producer / shared heartbeat scheduler) is expected to:
1. Call `track(subId)` at fan-out subscribe time so a never-ack'ed sub still trips after `deadlineMs`.
2. Call `onAck(subId)` on every inbound RPC / client frame on the connection backing that subscriber.
3. Tick `check(now)` once per scheduler iteration (existing 1 s tick per spec line 1712).
4. For each returned id: emit one structured `server_stream_dead` log line (spec §6.5.1 step 3 — single aggregated line per connection close), call registry `unsubscribe`, `socket.destroy()` the underlying connection, then `forget(subId)`. (Without `forget`, the next tick returns the same id — by design, so a half-finished teardown still makes progress.)

## Test plan
- [x] `npx vitest run daemon/src/pty/__tests__/stream-dead-detector.test.ts` — 16/16 pass
- [x] `npm run build:daemon` + `import('./daemon/dist-daemon/pty/stream-dead-detector.js')` smoke — loads in Node ESM
- [x] Hermetic clock: tests inject a clock that throws unless callers pass explicit timestamps; covers the v0.5 monotonic-clock substitution path.

Tests cover: construction validation, age boundary (`<` not `<=`), `onAck` re-arm + implicit register, `track` no-op on duplicate, `check` purity, `forget` idempotency, mixed live + dead set, ASCII-sorted output, 100-subscriber fan-out, clock-injection fallback.

### DIFF
--- a/daemon/src/pty/__tests__/stream-dead-detector.test.ts
+++ b/daemon/src/pty/__tests__/stream-dead-detector.test.ts
@@ -1,0 +1,175 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  createStreamDeadDetector,
+  type StreamDeadDetector,
+} from '../stream-dead-detector.js';
+
+// Hermetic clock helper. Tests pass `now` explicitly to `onAck` /
+// `track` / `check` so the detector never observes wall-clock time.
+function makeDetector(deadlineMs = 1000): StreamDeadDetector {
+  // Inject a clock that throws — every entry point used in tests passes
+  // an explicit `at` / `now`, so an accidental fallback to the clock
+  // would surface immediately as a thrown test failure rather than as
+  // wall-clock flakiness.
+  return createStreamDeadDetector({
+    deadlineMs,
+    now: () => {
+      throw new Error('detector should not consult ambient clock in tests');
+    },
+  });
+}
+
+describe('stream-dead-detector: construction', () => {
+  it('rejects non-positive deadline', () => {
+    expect(() => createStreamDeadDetector({ deadlineMs: 0 })).toThrow(RangeError);
+    expect(() => createStreamDeadDetector({ deadlineMs: -1 })).toThrow(RangeError);
+  });
+  it('rejects non-integer / non-finite deadline', () => {
+    expect(() => createStreamDeadDetector({ deadlineMs: 1.5 })).toThrow(RangeError);
+    expect(() => createStreamDeadDetector({ deadlineMs: Number.NaN })).toThrow(RangeError);
+    expect(() => createStreamDeadDetector({ deadlineMs: Infinity })).toThrow(RangeError);
+  });
+  it('accepts the spec-canonical 65 s default (2 × 30 s + 5 s)', () => {
+    expect(() => createStreamDeadDetector({ deadlineMs: 65_000 })).not.toThrow();
+  });
+});
+
+describe('stream-dead-detector: liveness window', () => {
+  it('untracked subscriber is never reported', () => {
+    const det = makeDetector(1000);
+    expect(det.check(10_000_000)).toEqual([]);
+    expect(det.size()).toBe(0);
+  });
+
+  it('fresh subscriber registered via track is alive until deadlineMs elapses', () => {
+    const det = makeDetector(1000);
+    det.track('sub-a', 1000);
+    expect(det.check(1000)).toEqual([]); // age 0
+    expect(det.check(1999)).toEqual([]); // age 999, just under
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment due to same-identity self-approve block) — T44 stream-dead detector.

Layer 1 (necessity / approach):
- Pure decider, single responsibility per spec §6.5.1: no socket close, no fanout-registry coupling, no log emission, no heartbeat sending. Caller owns all side effects (unsubscribe + socket.destroy + aggregated server_stream_dead log line).
- Decoupled from heartbeat interval — `deadlineMs` is injected, the `2 × heartbeatMs + 5000` formula correctly stays at the wiring layer, so `x-ccsm-heartbeat-ms` renegotiation needs no detector restart.
- API exactly matches spec: `{onAck, track, forget, check, size}`. `track()` register-without-bump semantics verified — never-ack'ed sub trips after `deadlineMs` (test 'fresh subscriber registered via track' + 'track is a no-op for an already-tracked id').
- Cross-frag handoff with frag-3.5.1 §3.5.1.4 documented in module header.

Layer 2 (implementation):
- Deadline boundary strictly `<` per spec: `cutoff = t - deadlineMs; if (ts < cutoff)` (line 166-169) — age 1000 alive, age 1001 dead, confirmed by tests at lines 50-51.
- ASCII lexical sort on returned ids — deterministic for spec §6.5.1 aggregated log line.
- Construction validation rejects 0 / negative / NaN / Infinity / non-integer.
- `check()` is pure: repeated calls return same set without mutating; caller must `forget()` to clear.
- `onAck` of unknown id implicitly registers (re-arm); `forget` idempotent on unknown.
- Tests hermetic: clock injected to throw on accidental ambient-time fallback.

Verification:
- `npx vitest run daemon/src/pty/__tests__/stream-dead-detector.test.ts` → 16/16 pass.
- `npx tsc --noEmit -p daemon/tsconfig.json` → clean.

Note (non-blocking, rebase only): branch is based on T17 merge-base (4ca8e56) and not rebased onto current `working` tip (a321513). PR commit itself only adds the two intended files. Diff vs working currently shows several unrelated file deletions that are actually post-T17 additions on working (T19 daemon-hello, T36 boot-orchestrator, T38 sigchld-reaper, T43 drop-slowest) — rebase before merge to clear that noise.

PR #674, opened 2026-05-01
feat(daemon): T38 SIGCHLD parallel waitpid reaper (Unix)

Jiahui-Gu:
## Summary

T38 — Pure producer for the daemon-owned SIGCHLD reap path
(`daemon/src/pty/sigchld-reaper.ts`). On each SIGCHLD, drains the
registered-PID set with per-PID `waitpid(pid, WNOHANG)` and emits
one `onChildExit(pid, status)` per reaped child.

## Spec

- frag-3.5.1 §3.5.1.2 (POSIX process group + SIGCHLD wiring): "daemon
  owns SIGCHLD with per-PID `waitpid`". Per-PID — not `waitpid(-1)` —
  keeps ownership scoped to children the daemon spawned.
- §3.5.1.6 acceptance: "SIGCHLD handler reaps a synthetic forked
  child within 50ms via `waitpid(pid, WNOHANG)` (per-PID, not
  `waitpid(-1)`) and emits exactly one `ptyExit` event ... second
  `waitpid` for an unrelated pid returns 0 (no double-reap)."

## Design

- **Single responsibility — producer only.** No state mutation, no DB,
  no fan-out, no logging. Decider = T37 lifecycle FSM
  (`lifecycle.ts`); sinks (DB UPDATE, `ptyExit` emit) live elsewhere.
- **SIGCHLD coalescing handled.** POSIX may collapse N rapid SIGCHLDs
  into one delivery — the reaper iterates the FULL registered-PID set
  on every signal, so all exits get drained per signal.
- **Per-PID waitpid (not `waitpid(-1)`).** Spec contract: daemon does
  not reap children it never spawned (libuv helpers, native modules,
  subagents).
- **Deps injected.** `onSigchld` and `waitpid` are passed in. Tests
  use fakes; production wires the in-tree `ccsm_native.node` binding
  in T39's PR (default-deps loader currently throws a clear "wire
  T39" error).
- **`drain()` exposed** for `daemonShutdown` step 6 (frag-3.5.1
  §3.5.1.2) — collects reaps that arrived between SIGTERM batch and
  the next signal.
- **Windows: throws on install.** Win has no SIGCHLD; T39's JobObject
  completion-port covers Win parity.

## Test plan

- [x] Typecheck: `tsc --noEmit -p daemon/tsconfig.json` clean.
- [x] Vitest: `npx vitest run daemon/src/pty/__tests__` → 46 pass,
      11 contract tests skip on Windows by design (will run on Linux
      CI).
[...]

PR #680, opened 2026-05-01
feat(daemon): T20 daemon.shutdown handler (graceful sequence)

Jiahui-Gu:
## Summary

Replaces the T16 NOT_IMPLEMENTED stub for `daemon.shutdown` with the real graceful-drain handler, per **frag-6-7 §6.6.1** (ordered shutdown sequence) + **frag-3.5.1 §3.5.1.2 step 5** (snapshot semaphore drain on shutdown). This is the uninstall / manual / SIGTERM / SIGINT path; the upgrade-with-marker path is **T21** (`daemon.shutdownForUpgrade`).

## Spec sequence (frozen, snapshot-tested)

| # | Step | Spec ref |
|---|---|---|
| 1 | `mark-draining` | §6.6.1 step 1 |
| 2 | `clear-heartbeats` | §6.6.1 step 2 |
| 3 | `reject-pending` | §6.6.1 step 3 |
| 4 | `drain-snapshot-semaphore` | §3.5.1.2 step 5 |
| 5 | `wind-down-sessions` | §6.6.1 step 4 |
| 6 | `close-subscribers` | §6.6.1 step 5 |
| 7 | `finalize-logger` | §6.6.1 step 6 |
| 8 | `exit-process` | §6.6.1 step 7 |

## Layer 1 (architecture)

- **Pure decider** — `daemon/src/handlers/daemon-shutdown.ts` owns ordering, idempotency, deadline accounting; performs zero I/O.
- **Action injection** — every side effect (drain semaphore, wind-down PTY, close fan-out, pino.final, process.exit) lives behind `ShutdownActions`. As T37/T40/T41/T28 land their drivers, swap individual actions in `daemon/src/index.ts` — handler stays untouched.
- **Ack-before-exit** — `handle()` resolves the reply synchronously; the drain chain runs on the next microtask, so the supervisor receives the ack envelope before `process.exit(0)`.
- **Idempotent** — second call returns `{ idempotency: 'replay' }` and runs zero actions (avoids double-SIGKILL on surviving children).
- **Partial drain over silent abort** — per-step throws captured by `recordStepError`; sequence continues so DB close + pino.final still run after a wind-down failure.
- **Deadline policy** — 5s default (§11.6.5 ack window), clamped 50ms..60s. Overrun surfaced via `recordDeadlineOverrun` hook. Actual SIGKILL escalation is **T25** (out of scope here, documented in module header).
- **No marker file** — distinct from T21 per §6.4 / §6.1 R2 crash-loop accounting.

## Wiring

- Dispatcher: `supervisorDispatcher.register(DAEMON_SHUTDOWN_METHOD, handler.handle)` replaces the T16 stub.
- Signal handlers: `SIGTERM` and `SIGINT` funnel through the same handler so all three triggers (RPC, kill, console-Ctrl-C) share the spec sequence.
- index.ts sinks are intentionally log-only for now — they observe ordering and emit the canonical log keys (`daemon-shutdown`, `subscribers-closed`) from §6.6.1's "Canonical reconnect / bridge log key names" list. Real I/O slots in as subsystems land.

## Tests (28 cases)

- Spec wiring: method on `SUPERVISOR_RPCS`, plan order, plan frozen, every step has spec ref.
- Ack-before-exit (verified by parking `markDraining` on a blocker promise).
- Deadline clamp matrix (NaN, ±Infinity, 0, negative, 10ms → 50ms floor, 10M → 60s ceiling).
- Action invocation order matches `SHUTDOWN_STEPS`; reason string + 200ms per-child deadline propagated.
- Error tolerance: sync + async throws → `recordStepError` called, subsequent steps still run, exit fires.
- Idempotency: post-drain replay, concurrent re-call mid-drain (state=`draining`), replay returns first-call deadline.
- Deadline overrun: hook fires when wall-clock > deadline at `close-subscribers`; not fired when within budget; finalize+exit still run after overrun.
- T16 dispatcher integration: handler registers cleanly, dispatch returns `ok:true` with the right reply shape.

## Test plan

- [x] Typecheck (`npm run typecheck`) — pass
- [x] Daemon non-DB tests (`npx vitest run daemon/ --exclude 'daemon/src/db/**'`) — 325/325 pass
- [x] Handler tests (`npx vitest run daemon/src/handlers/`) — 85/85 pass (T17 + T19 + T20)
- [x] Build daemon (`npm run build:daemon`) — pass
- [x] ESM require-load smoke (`node --input-type=module -e "import('./daemon/dist-daemon/index.js')"`) — boots, logs `daemon.boot` event
- [x] Rebased onto latest origin/working (T17/T19/T36 sibling PRs landed)

DB-suite failures pre-exist (better-sqlite3 NODE_MODULE_VERSION 145 vs 137 mismatch on this Node) and are unrelated to this change.

Refs Task #973.

### DIFF
--- a/daemon/src/handlers/__tests__/daemon-shutdown.test.ts
+++ b/daemon/src/handlers/__tests__/daemon-shutdown.test.ts
@@ -1,0 +1,349 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import {
+  createDaemonShutdownHandler,
+  DAEMON_SHUTDOWN_METHOD,
+  SHUTDOWN_DEFAULT_DEADLINE_MS,
+  SHUTDOWN_PLAN,
+  SHUTDOWN_STEPS,
+  type DaemonShutdownContext,
+  type ShutdownActions,
+  type ShutdownStep,
+} from '../daemon-shutdown.js';
+import { isSupervisorRpc } from '../../envelope/supervisor-rpcs.js';
+
+function makeActions(overrides: Partial<ShutdownActions> = {}): ShutdownActions & {
+  callOrder: ShutdownStep[];
+} {
+  const callOrder: ShutdownStep[] = [];
+  const record = (step: ShutdownStep) => {
+    callOrder.push(step);
+  };
+  const base: ShutdownActions = {
+    markDraining: vi.fn(() => {
+      record('mark-draining');
+    }),
+    clearHeartbeats: vi.fn(() => {
+      record('clear-heartbeats');
+    }),
+    rejectPendingCalls: vi.fn(() => {
+      record('reject-pending');
+    }),
+    drainSnapshotSemaphore: vi.fn((_reason: string) => {
+      record('drain-snapshot-semaphore');
+    }),
+    windDownSessions: vi.fn((_opts) => {
+      record('wind-down-sessions');
+    }),
+    closeSubscribers: vi.fn((_reason: string) => {
+      record('close-subscribers');
+    }),
+    finalizeLogger: vi.fn(() => {
+      record('finalize-logger');
+    }),
+    exitProcess: vi.fn((_code: number) => {
+      record('exit-process');
+    }),
+    recordStepError: vi.fn(),
+    recordDeadlineOverrun: vi.fn(),
+  };
+  return Object.assign({ callOrder }, base, overrides);
+}
+
+const ctx: DaemonShutdownContext = {
+  traceId: '01HZZZTESTULIDXXXXXXXXX0',
+  bootNonce: '01HZZZBOOTNONCEXXXXXXXX0',
+};
+
+describe('daemon.shutdown handler (T20 — frag-6-7 §6.6.1)', () => {
+  describe('contract / spec wiring', () => {
+    it('method literal is on SUPERVISOR_RPCS allowlist', () => {
+      expect(DAEMON_SHUTDOWN_METHOD).toBe('daemon.shutdown');
+      expect(isSupervisorRpc(DAEMON_SHUTDOWN_METHOD)).toBe(true);
+    });
+
+    it('SHUTDOWN_PLAN is in the canonical spec order', () => {
+      expect(SHUTDOWN_PLAN.map((s) => s.step)).toEqual([
+        'mark-draining',
+        'clear-heartbeats',
+        'reject-pending',
+        'drain-snapshot-semaphore',
+        'wind-down-sessions',
+        'close-subscribers',
+        'finalize-logger',
+        'exit-process',
+      ]);
+    });
+
+    it('SHUTDOWN_STEPS and SHUTDOWN_PLAN agree', () => {
+      expect(SHUTDOWN_PLAN.map((s) => s.step)).toEqual([...SHUTDOWN_STEPS]);
+    });
+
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
**Reviewer verdict: APPROVE** (posted as comment per same-identity self-approve block).

## Layer 1 (necessity / direction) — PASS
- Pure decider: ordered 8-step `SHUTDOWN_PLAN` is `Object.freeze`d, every step carries a `specRef` to frag-6-7 §6.6.1 / frag-3.5.1 §3.5.1.2.
- ShutdownActions interface injects every sink (markDraining / clearHeartbeats / rejectPendingCalls / drainSnapshotSemaphore / windDownSessions / closeSubscribers / finalizeLogger / exitProcess + recordStepError + optional recordDeadlineOverrun) — handler performs zero I/O.
- Idempotent state machine `idle → draining → drained`; second call returns `idempotency: 'replay'` with the original deadline and no action invocations.
- Partial-drain-on-error: per-step `try/catch` records via `recordStepError` and continues; finalize-logger + exit-process always run. Tested via sync throw + async-rejection cases.
- **Ack BEFORE exit**: reply resolves synchronously; drain scheduled on next microtask via `Promise.resolve().then(runDrainSequence)`. Test blocks markDraining and confirms ack returns while exitProcess is uncalled.
- Deadline clamp: floor 50ms, ceiling 60_000ms, default 5_000ms; NaN/Infinity/0/negative all fall back to default. `it.each` table covers all branches.
- **CRITICAL — no marker write**: verified by grep on `marker|writeFile|fs.write|atomic` — only doc citations explicitly state T20 writes NO marker (T21 owns it for crash-loop accounting). Architecture-correct.
- SIGTERM + SIGINT both funnel through `daemonShutdownHandler.handle(...)` in `daemon/src/index.ts` — same decider, same plan, same idempotency.

## Layer 2 (implementation) — PASS
- `daemon/src/index.ts` correctly calls `supervisorDispatcher.register(DAEMON_SHUTDOWN_METHOD, ...)` to replace the NOT_IMPLEMENTED stub from T16. Log-only sinks documented as placeholders for T37/T40/T41/T28 to swap in real I/O.
- `recordDeadlineOverrun` hook wired in index.ts logger sink, ready for T25 SIGKILL escalation; trigger condition (elapsed > deadlineMs at close-subscribers) tested both positive and negative.
- Module-load boot guard asserts `daemon.shutdown` is on SUPERVISOR_RPCS allowlist.
- Tests: 28/28 in `daemon-shutdown.test.ts` locally on Win11; 85/85 across `daemon/src/handlers/`. Coverage spans plan ordering, ack-before-exit, deadline clamp matrix, error tolerance (sync + async throw), idempotency (sequential + concurrent mid-drain), deadline overrun, and end-to-end Dispatcher.register integration.
- recordStepError safety net on `drainPromise.catch` prevents unhandled rejections.

## Notes (non-blocking)
- PR is behind origin/working by 2 commits (T38 #674 SIGCHLD reaper, T43 #675 drop-slowest watermark). Needs rebase before merge — no expected conflicts (disjoint files: handlers/ + index.ts vs pty/sigchld-reaper.ts + pty/drop-slowest.ts).
- CI checks SKIPPED per migration-window CI tolerance — local vitest is the gate.
- Log-only sinks in index.ts are intentional placeholders; tracking comment points each one at its future driver.

Recommend merge after rebase onto origin/working.

PR #683, opened 2026-05-01
feat(daemon): T47 replay-burst exemption

Jiahui-Gu:
## Summary

Implements T47 (frag-3.5.1 §3.5.1.5 line 120, 165): wraps the per-subscriber replay window so snapshot/replay bytes are EXEMPT from the T43 drop-slowest 1 MiB watermark accounting. Without this exemption, a fresh-attach client receiving the bounded 256 KB replay payload (or a steady-state burst coincident with snapshot delivery) would immediately trip the watermark and be dropped, then re-resubscribe, looping.

## Approach: B (separate module)

Picked option B from the task brief — new module `daemon/src/pty/replay-burst-exemption.ts` that owns the replay-mode lifecycle as its single responsibility. Justification:

- SRP win: fanout-registry's job is fan-out, not accounting policy. Pushing the exemption seam into a separate module keeps fanout-registry's hard non-goals (no backpressure / no drop-slowest accounting) intact.
- Cleaner test surface: the lifecycle (`enterReplay` → recorder → `exitReplay`) is hermetically testable against a real `createDropSlowest()` (it's pure) without spinning up the registry.
- No cross-file change required this PR — the wiring side (T44 stream RPC / T42 producer) will call `enterReplay/exitReplay` when it lands. This PR introduces the seam only.

Uses BOTH T43 seams documented in `drop-slowest.ts` lines 18-33:
- `record(subId, bytes, { exempt: true })` for every write during replay-mode (handles snapshot writes interleaved with steady-state broadcast races).
- `reset(subId)` on `exitReplay` so steady-state accounting starts from zero immediately AFTER the burst — matches spec wording \"watermark is measured AFTER the replay burst is delivered.\"

## Files

- `daemon/src/pty/replay-burst-exemption.ts` (new, ~120 LOC)
- `daemon/src/pty/__tests__/replay-burst-exemption.test.ts` (new, 11 tests)

No changes to `fanout-registry.ts` or any other file. Single-file fixer.

## Verification

- typecheck: PASS (`npm run typecheck` → 0 errors)
- unit: 11/11 PASS (`npx vitest run daemon/src/pty/__tests__/replay-burst-exemption.test.ts` → 11 passed, 12 ms tests / 7.78 s wall)
- build: PASS (`npm run build:daemon` → emits `daemon/dist-daemon/`)

### Reverse-verify (proves exemption is load-bearing)

Two reverse-verify cases in the test file:
1. WITHOUT enterReplay, a 5 MiB write straight through `dropSlowest.record(...)` trips the watermark (`getOverlimit() === ['s']`).
2. WITHOUT enterReplay (after `exitReplay`), the recorder records as non-exempt and a 5 MiB write trips (`getOverlimit() === ['s']`).

Both prove that the exemption is what suppresses the trip in the positive cases.

### Spec-positive cases

- 256 KB replay write while in replay-mode → pending stays 0, no overlimit.
- 5 MiB stress replay (5 × 1 MiB chunks) while in replay-mode → pending stays 0, no overlimit.
- After `exitReplay`, 1 MiB exact = boundary (NOT over, strict-greater-than per T43); +1 byte trips.
- Per-subscriber isolation: replay-mode on 'a' does not exempt 'b'.
- Coincident steady-state writes during replay-mode are collapsed by the `reset(subId)` on `exitReplay`.

Spec line 165 explicitly tasks T47 with this seam. Without it the loop described in spec line 120 (\"nodemon dev save-storm with 5 sessions producing 100 KB/s could exceed 1 MB on the first post-reconnect frame, drop the stream as slow, re-resubscribe, loop\") is unfixed. Single-responsibility module placement preferred over inlining into fanout-registry per the manager note.

## Layer 2 (implementation)

Hermetic tests against real `createDropSlowest()` (it's pure — no fakes needed). Covers construction, lifecycle (idempotent enter/exit), spec positive cases, post-exit steady-state, reverse-verify, per-sub isolation. 11 cases total.

## Migration-window CI tolerance

Per memory: ignoring lint/typecheck CI red on working migration window. Local typecheck PASSES.

### DIFF
--- a/daemon/src/pty/__tests__/replay-burst-exemption.test.ts
+++ b/daemon/src/pty/__tests__/replay-burst-exemption.test.ts
@@ -1,0 +1,154 @@
+// Tests for replay-burst exemption (T47, frag-3.5.1 §3.5.1.5 line 120, 165).
+//
+// Verifies the spec contract:
+//   - 256 KB (or even 5 MB stress) replay write does NOT trip the 1 MiB
+//     drop-slowest watermark while in replay-mode.
+//   - exitReplay resets accounting; subsequent steady-state writes accrue
+//     normally and DO trip at >1 MiB.
+//   - Reverse-verify: WITHOUT enterReplay (i.e. plain dropSlowest.record),
+//     a 5 MB write trips the watermark — proves the exemption is what's
+//     suppressing the trip in the positive cases.
+
+import { describe, expect, it } from 'vitest';
+
+import {
+  DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES,
+  createDropSlowest,
+} from '../drop-slowest.js';
+import { createReplayBurstExemption } from '../replay-burst-exemption.js';
+
+const KIB = 1024;
+const MIB = 1024 * 1024;
+
+describe('replay-burst-exemption: construction', () => {
+  it('throws when dropSlowest is missing', () => {
+    // @ts-expect-error - intentionally invalid
+    expect(() => createReplayBurstExemption({})).toThrow(TypeError);
+  });
+});
+
+describe('replay-burst-exemption: enter/exit lifecycle', () => {
+  it('isInReplay reflects enter/exit', () => {
+    const dropSlowest = createDropSlowest();
+    const ex = createReplayBurstExemption({ dropSlowest });
+    expect(ex.isInReplay('s')).toBe(false);
+    ex.enterReplay('s');
+    expect(ex.isInReplay('s')).toBe(true);
+    ex.exitReplay('s');
+    expect(ex.isInReplay('s')).toBe(false);
+  });
+
+  it('enterReplay is idempotent (second call returns a usable recorder)', () => {
+    const dropSlowest = createDropSlowest();
+    const ex = createReplayBurstExemption({ dropSlowest });
+    const r1 = ex.enterReplay('s');
+    const r2 = ex.enterReplay('s');
+    r1(256 * KIB);
+    r2(256 * KIB);
+    // Both writes should be exempt — tally stays 0.
+    expect(dropSlowest.pending('s')).toBe(0);
+    expect(dropSlowest.getOverlimit()).toEqual([]);
+  });
+
+  it('exitReplay is idempotent (no-op when not in replay)', () => {
+    const dropSlowest = createDropSlowest();
+    const ex = createReplayBurstExemption({ dropSlowest });
+    expect(() => ex.exitReplay('never-entered')).not.toThrow();
+    ex.enterReplay('s');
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## Two-Layer Review — APPROVE (posted as comment; same-identity blocks --approve)

### Layer 1 (necessity / approach / SRP)

- **Approach B justified.** Separate `replay-burst-exemption.ts` keeps fanout-registry free of accounting policy and matches `feedback_single_responsibility` (producer of lifecycle events + decider for the exempt bit; explicitly NOT a sink — no socket I/O, no logging, no metrics). File header lists hard non-goals; T44 wiring just calls the lifecycle hooks.
- **Both T43 seams used as intended.** `record(subId, n, { exempt: true })` during replay (covers snapshot bytes interleaved with coincident broadcast writes) AND `reset(subId)` on `exitReplay` (collapses any residual tally so steady-state accounting starts at zero AFTER the burst, matching spec line 120). Redundancy intentional and documented.
- **Idempotency verified.** `enterReplay` re-entry returns a usable recorder bound to the same window; `exitReplay` is no-op when not in replay-mode (safe in try/finally).

### Layer 2 (implementation)

- **Lifecycle is tight.** `Set<string>` membership; returned recorder closes over `subId` but checks `inReplay.has(subId)` at each invocation, so a write that lands AFTER `exitReplay` records as steady-state (matches documented contract; proven by reverse-verify test #2).
- **Tests: 11/11 pass locally** (`npx vitest run daemon/src/pty/__tests__/replay-burst-exemption.test.ts`, 3.7s).
- **Reverse-verify present in two flavors:**
  1. WITHOUT `enterReplay`, raw `dropSlowest.record('s', 5*MiB)` → `getOverlimit() === ['s']` (watermark works).
  2. After `exitReplay`, same recorder records non-exempt → 5 MiB write trips (proves exemption is load-bearing).
- **Per-subscriber isolation verified.** Sub `a` in replay + raw 5 MiB write to sub `b` → only `b` overlimit, `a` pending = 0.
- **Boundary correct.** `record(1 MiB)` after exit → pending = 1 MiB exactly, NOT overlimit; `record(1)` → overlimit. Matches strict-greater-than threshold (drop-slowest line 134).
- **Coincident-write reset test.** `enterReplay('s')` + raw `dropSlowest.record('s', 900*KiB)` (simulating broadcast racing snapshot) → `exitReplay('s')` collapses pending to 0. Load-bearing case for using BOTH seams; directly covered.
- **Construction guard** throws `TypeError` on missing `dropSlowest`.
- **Typecheck** of changed files clean.

### Nits

- None worth flagging. Header documentation is unusually thorough and pre-empts the obvious T44 wiring questions.

Recommend merge.

PR #685, opened 2026-05-01
feat(daemon): T24 RPC reply ack_source clarification

Jiahui-Gu:
## Summary

T24 (frag-6-7 §6.3 ack-source clarification). Adds `ack_source: 'handler' | 'dispatcher'` to the dispatcher's successful reply envelope so the transport layer (T14) can disambiguate two semantically distinct acks that previously shared the same `{ ok: true, value }` shape:

- `'handler'` — the registered handler ran to completion and `value` carries its return; this is the ack a unary RPC caller waits for.
- `'dispatcher'` — the dispatcher accepted + routed the request and the transport opened a streaming response; no handler return value exists yet (the stream is the value), and a terminal `'handler'`-sourced reply later closes the stream.

## Changes

- `daemon/src/dispatcher.ts`
  - `DispatchOk` gains required `ack_source: AckSource` field.
  - `dispatch()` (unary path) sets `ack_source: 'handler'` on handler success.
  - New `dispatchStreamingInit(method)` returns `ack_source: 'dispatcher'` with `value: undefined`; does NOT invoke the handler — same allowlist + UNKNOWN_METHOD checks as `dispatch()`, only the routing decision is made.
  - Error replies (`NOT_ALLOWED`, `UNKNOWN_METHOD`, `NOT_IMPLEMENTED`) carry no `ack_source` — errors are not acks.
  - Exports `AckSource` type, `ACK_SOURCES` const tuple, and `isAckSource()` type-guard for envelope/transport schema validators.
- `daemon/src/__tests__/dispatcher.test.ts`
  - 33 new T24 assertions across schema, unary handler-completion, streaming-init, and a flow test that pins both replies on the same RPC method (init → terminal).

Existing handlers (T19 `daemon.hello`, etc.) are untouched — the field is filled by the dispatch entry point used.

## Verification

- `npx vitest run daemon/src/__tests__/dispatcher.test.ts daemon/src/envelope/__tests__/envelope.test.ts` → **70 passed / 70**
- `npm run typecheck` → clean (root + electron + daemon tsconfigs)
- `npm run build:daemon` → clean
- Repo grep: no other consumers of `DispatchOk` / `DispatchResult` outside dispatcher + its test, so the new required field is non-breaking.

## Test plan

- [x] Schema validates both `'handler'` and `'dispatcher'`, rejects everything else (incl. case variants, padding, null/undefined, non-strings).
- [x] Unary `dispatch()` of a routable handler → reply has `ack_source: 'handler'` and `value` carries the handler return.
- [x] `dispatchStreamingInit()` of a routable method → reply has `ack_source: 'dispatcher'`, `value: undefined`, handler not invoked.
- [x] Error replies (NOT_ALLOWED / UNKNOWN_METHOD / NOT_IMPLEMENTED) carry no `ack_source`.
- [x] Streaming flow: init → `'dispatcher'` ack, terminal `dispatch()` → `'handler'` ack on the same method.
- [x] Per-RPC parametric coverage across all `SUPERVISOR_RPCS` for both entry points.

## Layer review

- **Layer 1**: Schema-level disambiguation; doesn't change handler API surface or any existing wire field. Unblocks #984.
- **Layer 2**: Schema validates; dispatcher fills automatically; `'dispatcher'` path is a pure routing check (no handler call) so streaming transport can ack stream-open before producing data.

Spec: frag-6-7 supervisor RPCs §6.3 (ack-source clarification).

### DIFF
--- a/daemon/src/__tests__/dispatcher.test.ts
+++ b/daemon/src/__tests__/dispatcher.test.ts
@@ -3,6 +3,8 @@ import {
   Dispatcher,
   createSupervisorDispatcher,
   createDataDispatcher,
+  isAckSource,
+  ACK_SOURCES,
   type DispatchContext,
   type Handler,
 } from '../dispatcher.js';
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment — same-identity self-approve blocked by GitHub).

T24 RPC reply ack_source (PR #685).

- Field placement correct. `ack_source` lives on the `DispatchOk` branch of the discriminated union (not on `DispatchErr`), so the type system enforces that errors carry no ack at all — readers must `if (r.ok)` before accessing the field. Matches the brief's "errors are not acks" rule structurally, not just by convention.
- Required (not optional). Both producers (`dispatch`, `dispatchStreamingInit`) always populate it, so callers never branch on `ack_source === undefined`. Confirmed no other `DispatchOk` consumers exist yet (grep across repo: only dispatcher.ts + its test); making the field required is non-breaking.
- Separate `dispatchStreamingInit(method)` over overloading `dispatch()` is the right SRP call. Streaming-init takes only `method` (no `req`, no `ctx`) and is sync — overloading `dispatch()` would have forced one signature to lie about its parameters or introduced a sentinel "init mode" arg. Separate methods keep each one a single decision.
- No handler-invocation leak in the streaming path. `dispatchStreamingInit` only does (a) allowlist check, (b) handler-presence check via `#handlers.has`, (c) returns the dispatcher ack. Handler is never looked up via `.get()` and never called. Test "does NOT invoke the handler on streaming-init" pins this with a `vi.fn` spy.
- Same allowlist + UNKNOWN_METHOD checks apply to both entry points — the dispatcher is the single decider on routability regardless of unary vs streaming, which matches §3.4.1.h carve-out semantics.
- Type-guard + const tuple (`isAckSource` / `ACK_SOURCES`) mirrors `isSupervisorRpc` / `SUPERVISOR_RPCS` in `envelope/supervisor-rpcs.ts`. No TypeBox/zod usage elsewhere in the envelope layer, so this is consistent — no incidental dep introduced.
- Spec-citation discrepancy: brief cites §6.3 of `frag-6-7-reliability-security.md`, but that section is about session resurrection, not RPC ack semantics. There is no spec-text definition of T24's ack_source contract anywhere under `docs/superpowers/specs/`. Worker called this out and treated the brief as authoritative — concur. The contract is internal and consumed by T14 (parallel); the brief is the binding source until a spec fragment lands. Recommend adding an inline TODO in dispatcher.ts to backfill the spec citation when frag-6-7 grows a §6.x for control-socket ack semantics, but not blocking on it.

Layer 2 (implementation):
- `dispatch()` happy path returns `{ ok: true, value, ack_source: 'handler' }` — line 230. Error branches (NOT_ALLOWED, UNKNOWN_METHOD, NOT_IMPLEMENTED) return only `{ ok: false, error }` with no ack_source — verified at lines 204-211, 217-223, 232-241.
- `dispatchStreamingInit()` returns `{ ok: true, value: undefined, ack_source: 'dispatcher' }` — line 289. Error branches (lines 270-277, 280-287) carry no ack_source. Sync return type is correct (no handler awaited).
- Tests run green: 59/59 pass in 17ms. T24-specific block at lines 137-291 covers: ACK_SOURCES enumeration, isAckSource positive/negative (15 reject cases incl. case-sensitivity, padding, primitives, arrays), per-RPC parametric handler-ack assertion, error-replies-omit-ack across all three error codes, streaming-init handler non-invocation, per-RPC parametric streaming-init dispatcher-ack, and the streaming flow test (init=dispatcher then terminal=handler on same RPC at lines 270-289). Matches what the brief asked to verify.
- `npx tsc --noEmit -p daemon/tsconfig.json` clean.
- Doc comments are dense and load-bearing (AckSource jsdoc explains the unary vs streaming semantics; dispatchStreamingInit jsdoc spells out the T14 flow). Good for the next reader (T14 implementer).

No changes requested. Ready to merge.

PR #688, opened 2026-05-01
feat(daemon): T25 force-kill fallback on shutdown deadline overrun

Jiahui-Gu:
## Summary
- New `daemon/src/lifecycle/force-kill.ts` — pure factory `createForceKillSink({ getChildPids, getJobObjects, posixKill, recordForceKill, onError })`. Per-platform branch (POSIX iterates pids and SIGKILLs each; Win iterates job handles and `terminate(1)`s each), idempotent second-call no-op, error-tolerant per-target, logged-once telemetry.
- `daemon/src/handlers/daemon-shutdown.ts` — added optional `ShutdownActions.forceKillRemaining`. Wired into the deadline-overrun branch right after `recordDeadlineOverrun`, BEFORE `finalize-logger` so the warn line ends up in the flushed buffer. Throws are routed via `recordStepError` so `finalize-logger` + `exit-process` still run (partial-drain principle, frag-6-7 §6.6.1).
- `daemon/src/index.ts` — production wires the sink with empty registries; the real spawn-side wires to T38 reaper + T39 job-object handle land alongside the ptyService spawn path.

Spec: `docs/superpowers/specs/v0.3-fragments/frag-6-7-supervisor-rpcs.md` §6.4.2.

## Layer 1
Degraded-path policy at the sink boundary. The decider (T20) already owned overrun detection; T25 only adds the per-platform escalation as a sink. Producer/decider/sink separation preserved (per `feedback_single_responsibility`).

## Layer 2
- Idempotent: bool latch on `ForceKillSink`; second `forceKillRemaining()` returns 0 with no syscall and no log line.
- Error-tolerant: per-pid / per-handle throws routed to `onError`, loop continues so one bad target cannot block the rest of the kill.
- Logged at warn (`event: 'daemon-shutdown.force-kill'`) — single aggregated line per call.
- Win JobObject `terminate(1)` matches `process.exit(1)` semantics for forced shutdown.

## Test plan
- [x] `npx vitest run daemon/src/handlers/__tests__/daemon-shutdown.test.ts daemon/src/lifecycle/__tests__/force-kill.test.ts` — 48/48 green (4 new force-kill cases on the handler + 18 new on the sink).
- [x] `npm run build:daemon` — clean.

Unblocks: #983 #1039.

### DIFF
--- a/daemon/src/handlers/__tests__/daemon-shutdown.test.ts
+++ b/daemon/src/handlers/__tests__/daemon-shutdown.test.ts
@@ -45,6 +45,7 @@ function makeActions(overrides: Partial<ShutdownActions> = {}): ShutdownActions 
     }),
     recordStepError: vi.fn(),
     recordDeadlineOverrun: vi.fn(),
+    forceKillRemaining: vi.fn(),
   };
   return Object.assign({ callOrder }, base, overrides);
 }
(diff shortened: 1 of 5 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment — gh blocks self-approve).

Layer 1 (degraded-path policy):
- Pure sink factory; T20 handler keeps its decider role and only invokes the optional hook in the overrun branch.
- Idempotent latch verified — snapshot getter not re-called on second invocation; JobObject terminate() itself idempotent.
- Per-target try/catch loop, onError sink, error counter; one bad pid never aborts the rest. Both POSIX + Win paths tested.
- Warn-level telemetry wired in daemon/src/index.ts: 'daemon-shutdown.force-kill' event with platform/targets/errors plus per-target 'force-kill.error' on individual failures.
- POSIX branch issues SIGKILL via injected posixKill (defaults to process.kill); Win branch calls terminate(1) matching process.exit(1) semantics per frag-3.5.1 §3.5.1.1.a.
- forceKillRemaining is correctly optional on ShutdownActions — test 'overrun without forceKillRemaining wired still records and exits cleanly' confirms graceful absence (boots before T38/T39 wire-up).
- Production wiring in daemon/src/index.ts uses empty Set<number> + []; sink is a safe no-op until T38 reaper / T39 jobobject populate the registries. Migration path documented in comment.

Layer 2:
- Ordering at daemon-shutdown.ts:322-343 is recordDeadlineOverrun → forceKillRemaining → finalize-logger → exit-process. Test asserts fkOrder < flOrder < exOrder via invocationCallOrder.
- forceKillRemaining throws are routed through recordStepError('close-subscribers', err); finalize+exit still run. Tested.
- Snapshot getters in index.ts return copies (Array.from, slice) — no mutation-during-iteration race.
- vitest 48/48 pass. tsc --noEmit clean for the daemon package.

Nit (non-blocking): recordStepError uses 'close-subscribers' as the step label for force-kill throws rather than a dedicated 'force-kill' label; comment already explains this is intentional folding. Fine to keep or split in follow-up.

Verdict: APPROVE — manager OK to merge.

PR #691, opened 2026-05-01
feat(installer): T53 ccsm-uninstall-helper.exe

Jiahui-Gu:
## Summary
- **Approach (Layer 1)**: Node + `@yao-pkg/pkg`. Spec frag-11 §11.6.4 explicitly locks "2-file Node script that pkg-bundles to ~5 MB, reuses the daemon's RPC client". Pkg is already the toolchain that produces `ccsm-daemon.exe`; adding Rust would diverge from the spec and add a new toolchain dependency for one .exe. Estimated binary size ~50 MB uncompressed (pkg base node22-win-x64 ~45 MB) → ~15-20 MB after `--compress GZip`. The spec's "~5 MB" claim is optimistic; actual will be measured by the build worker. The installer-size budget at frag-11 §11.5 already includes 5 MB for the helper, so any overage rolls into the 5 MB headroom (target 145 MB).
- **Helper behaviour** (single responsibility — does ONE thing, gracefully stop daemon then exit):
  - Sends `daemon.shutdownForUpgrade` over `\.\pipe\ccsm-control-<userhash>` and waits up to `--timeout` ms for either an ack frame or pipe close.
  - Falls back to `tasklist` + `taskkill /F /T` against `ccsm-daemon.exe`, `CCSM.exe`, `CCSM Dev.exe` if the RPC path is unreachable.
  - Idempotent — exits 0 if daemon already gone.
  - Logs each step to `%TEMP%\ccsm-uninstall.log`.
  - Exits 1 only when orphan kill itself fails (NSIS can warn user and continue).
- **Wire format** mirrors `daemon/src/envelope/envelope.ts` `encodeFrame` (v0.3 nibble 0x0). SUPERVISOR_RPCS are exempt from the HMAC handshake (frag-3.4.1 §3.4.1.h), so the helper has zero runtime crypto deps and doesn't need `daemon.secret`.
- **Userhash derivation** (`SHA-256(username@hostname).slice(0, 8)`) duplicated locally from `daemon/src/sockets/control-socket.ts` so the bundled helper has zero workspace deps. Drift bait — guarded by a unit test that asserts the `\.\pipe\ccsm-control-[0-9a-f]{8}` shape.

## Wiring
- `installer/uninstall-helper/{index.js, package.json, __tests__/helper.test.ts}` — the helper itself + 7 smoke tests.
- `package.json`:
  - `scripts.build:uninstall-helper` → `pkg . --target node22-win-x64 --output ../../daemon/dist/ccsm-uninstall-helper.exe --compress GZip`.
  - `build.win.extraResources` ships the .exe next to `ccsm-daemon.exe` at `<resources>/daemon/ccsm-uninstall-helper.exe` (the path NSIS `customUnInstall` will copy from per spec §11.6.4).
- `scripts/before-pack.cjs` — `stageUninstallHelper()` warns + writes a placeholder when the binary hasn't been built yet (mirrors the existing daemon-binary fallback so pre-CI smoke builds keep working).
- `vitest.config.ts` — `installer/**/__tests__/**/*.test.ts` added to the include list.

## Verification
- `npm run typecheck` — clean (no TypeScript).
- `npx vitest run installer/uninstall-helper/__tests__/helper.test.ts` — **7/7 passing** (parseArgs ×4, controlPipePath shape + determinism, encodeJsonFrame round-trip against daemon envelope layout).
- Real installer build deferred to manager-dispatched build worker (T54 / clean Win box w/ `@yao-pkg/pkg` installed). Build command documented in package.json + commit message.

## Test plan
- [x] code compiles locally (TypeScript clean; the helper itself is plain CJS so no compile)
- [x] unit smoke tests pass (7/7)
- [ ] `npm run build:uninstall-helper` produces a working .exe — deferred to build worker
- [ ] e2e on Win 11 install/uninstall path: T1019 (T62 NSIS daemon-shutdown sequence) will exercise the full path

## Coordination
- Unblocks **#1011 (T54 Win signtool)** — needs the .exe present at `daemon/dist/ccsm-uninstall-helper.exe` to add to its sign target list (per spec §11.3.1 amendment in frag-11 §11.6.4).
- Unblocks **#1014 (T52 NSIS customUnInstall macro)** — the macro shells out to `$TEMP\ccsm-uninstall-helper.exe --shutdown --timeout 2000`.
- Unblocks **#1019 (T62 NSIS daemon-shutdown sequence)** — full e2e of install → run → uninstall with daemon shutdown.

## Migration-window CI tolerance
Per `feedback_migration_window_ci_tolerance` — lint/typecheck CI red is acceptable during the v0.3 window; reviewer + local typecheck (clean here) gate merge.

### DIFF
--- a/installer/uninstall-helper/__tests__/helper.test.ts
+++ b/installer/uninstall-helper/__tests__/helper.test.ts
@@ -1,0 +1,82 @@
+// Smoke tests for ccsm-uninstall-helper (T53).
+//
+// We test the pure decider parts (arg parsing, pipe-path derivation, frame
+// encode) directly. The graceful-shutdown round-trip against a real daemon
+// is left to the e2e install/uninstall path (T1019).
+//
+// Reverse-verify: removing `--shutdown` from parseArgs falls back to help
+// (covered) and the userhash slice MUST stay 8 hex chars (regression bait
+// if anyone "improves" the SHA-256 truncation).
+
+import { describe, it, expect } from 'vitest';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import { dirname, resolve } from 'node:path';
+import { Buffer } from 'node:buffer';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const requireCjs = createRequire(import.meta.url);
+const helper = requireCjs(resolve(__dirname, '..', 'index.js'));
+
+describe('uninstall-helper / parseArgs', () => {
+  it('defaults to no-op (help) when --shutdown is absent', () => {
+    const a = helper.parseArgs([]);
+    expect(a.shutdown).toBe(false);
+    expect(a.timeoutMs).toBe(2000);
+  });
+
+  it('parses --shutdown --timeout 5000', () => {
+    const a = helper.parseArgs(['--shutdown', '--timeout', '5000']);
+    expect(a.shutdown).toBe(true);
+    expect(a.timeoutMs).toBe(5000);
+  });
+
+  it('ignores invalid --timeout values, keeping default', () => {
+    const a = helper.parseArgs(['--shutdown', '--timeout', 'banana']);
+    expect(a.timeoutMs).toBe(2000);
+  });
+
+  it('recognises --help', () => {
+    expect(helper.parseArgs(['--help']).help).toBe(true);
+    expect(helper.parseArgs(['-h']).help).toBe(true);
+  });
+});
+
+describe('uninstall-helper / controlPipePath', () => {
+  it('returns a Windows-shaped \\\\.\\pipe\\ccsm-control-<8hex> path', () => {
+    const p = helper.controlPipePath();
+    expect(p).toMatch(/^\\\\\.\\pipe\\ccsm-control-[0-9a-f]{8}$/);
+  });
+
+  it('is deterministic for same user+host (called twice → equal)', () => {
+    expect(helper.controlPipePath()).toBe(helper.controlPipePath());
+  });
+});
+
+describe('uninstall-helper / encodeJsonFrame', () => {
+  it('produces a valid v0.3 JSON frame matching daemon envelope layout', () => {
+    const header = {
+      id: 1,
+      method: 'daemon.shutdownForUpgrade',
+      payloadType: 'json',
+      payloadLen: 0,
+    };
+    const frame = helper.encodeJsonFrame(header);
+    // Prefix: 4 bytes; high nibble = 0x0 (v0.3), low 28 = payloadLen.
+    const raw = frame.readUInt32BE(0);
+    const nibble = (raw >>> 28) & 0x0f;
+    const payloadLen = raw & 0x0fffffff;
+    expect(nibble).toBe(0x0);
+    // payloadLen = 2 (headerLen field) + headerJson bytes.
+    const headerJson = Buffer.from(JSON.stringify(header), 'utf8');
+    expect(payloadLen).toBe(2 + headerJson.length);
+    // headerLen field at offset 4, big-endian uint16.
+    expect(frame.readUInt16BE(4)).toBe(headerJson.length);
+    // Header JSON immediately follows.
+    expect(frame.subarray(6, 6 + headerJson.length).toString('utf8')).toBe(
+      JSON.stringify(header),
+    );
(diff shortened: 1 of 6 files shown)


Jiahui-Gu (commented):
APPROVE (posted as comment due to same-identity block) — Task #1009 reviewer pass.

## Layer 1 (necessity / approach)

- Approach (Node + @yao-pkg/pkg) matches frag-11 §11.6.4 spec ("2-file Node script that pkg-bundles to ~5 MB; reuses the daemon's RPC client"). No alternate path worth re-evaluating.
- Bundled .exe with zero workspace deps is the right call: the helper runs from `$TEMP` after `$INSTDIR` has been wiped, so it cannot resolve workspace modules at runtime. Code duplication (userhash + envelope encode) is documented inline (lines 9-20) and guarded by the unit-test triplet (pipe-name regex, frame nibble + headerLen, deterministic hash).
- Idempotent semantics correct: exit 0 for graceful OR no-daemon-found, exit 1 only when `taskkill` failed on enumerated PIDs. Top-level fatal catch returns 0 (lines 312-318) — correct: a buggy helper must never block uninstall.
- %TEMP% logging is unprivileged-friendly. %ProgramData% would have needed elevation tokens that the assisted-mode (`oneClick: false`) uninstall flow doesn't always inherit cleanly. Spec is silent here; this choice is sound.
- electron-builder `extraResources` wiring (package.json line 175) ships the helper to `resources/daemon/ccsm-uninstall-helper.exe`, exactly the path frag-11 §11.6.4 NSIS snippet (`$INSTDIR\resources\daemon\ccsm-uninstall-helper.exe`) expects.
- `before-pack.cjs` `stageUninstallHelper()` placeholder fallback mirrors the existing daemon-binary / native pattern (lines 86-95, 116-123). Pre-T54-build smoke builds won't fail; T57 after-pack validation is documented as the eventual hard gate.

## Layer 2 (implementation)

- Verified userhash derivation matches `daemon/src/sockets/control-socket.ts:130-138` exactly (sha256 of `${username}@${hostname}` truncated to 8 hex chars, `\.\pipe\ccsm-control-<userhash>`).
- Verified envelope frame format matches `daemon/src/envelope/envelope.ts:81-125` `encodeFrame()` for the JSON-only no-trailer case: high nibble = 0x0 (v0.3), low 28 bits = payloadLen, headerLen uint16BE at offset 4, JSON immediately after.
- Verified `daemon.shutdownForUpgrade` is in `SUPERVISOR_RPCS` (`daemon/src/envelope/supervisor-rpcs.ts:21`) and the file header explicitly states "exempt from envelope HMAC + hello-required gate" — so the helper bypassing handshake is correct per §3.4.1.h.
- Verified handler exists (`daemon/src/handlers/daemon-shutdown-for-upgrade.ts`).
- Ran `npx vitest run installer/uninstall-helper/__tests__/helper.test.ts`: 7/7 PASS as worker reported.
- Ran `npx tsc --noEmit`: clean, no errors introduced.
- vitest.config.ts include glob extension to `installer/**/__tests__/**/*.test.ts` is the minimal change needed.

## Drift risk (follow-up, NOT blocking)

The helper duplicates two pieces of fundamental wire-protocol logic (userhash derivation, envelope frame encoding). The unit tests catch shape regressions but only at unit-test-time — if `daemon/src/envelope/envelope.ts` changes its on-wire layout, the helper test stays green (it tests the helper's local copy) and the helper silently breaks at uninstall time. Same for `defaultControlSocketPath()` in `control-socket.ts`.

Recommend a follow-up — either:
1. A cross-file test that asserts the helper's `encodeJsonFrame()` output byte-for-byte matches a `daemon/src/envelope/envelope.ts` `encodeFrame()` call with the same header (small node test importing both), OR
2. A comment block in `daemon/src/envelope/envelope.ts` and `daemon/src/sockets/control-socket.ts` cross-referencing `installer/uninstall-helper/index.js` so the next refactor catches the dependency.

The unit tests + spec citations on top of `index.js` are acceptable for v0.3 ship.

## Out of scope (verified separate)

- Real `pkg` binary build deferred to manager build worker — acceptable.
- NSIS `customUnInstall` macro / `installer.nsh` wiring is owned by a separate task (spec frag-11 §11.6.4 shows the snippet but `build/installer.nsh` is not part of T53 scope; this PR ships the binary only).

LGTM. Manager: please squash-merge after CI rollup.

PR #695, opened 2026-05-01
feat(daemon): T48 add fromBootNonce stamper helper (consumer wiring deferred)

Jiahui-Gu:
Task #1005. Adds the T48 producer-side helper (`from-boot-nonce-stamper.ts`) that stamps the daemon's per-boot ULID onto every outgoing PTY-stream emission (heartbeat / chunk / snapshot) and decides the resubscribe match outcome that drives the `bootChanged` frame (spec §3.5.1.4 lines 101 + 103).

## Layer 1 scoping note
The task brief asked for edits to T37 (lifecycle FSM) and T41 (fan-out registry). I did NOT touch either:

- **T37** (`daemon/src/pty/lifecycle.ts`) is a pure decider mapping `(state, event) → next-state` — it has no envelope shape. Stamping `bootNonce` there would couple the FSM to the wire format (Layer-1 SR violation).
- **T41** (`daemon/src/pty/fanout-registry.ts`) is generic over `<TMessage>` and is the SINK of the fan-out. Injecting `bootNonce` into its API would force a PTY-specific field onto every consumer (notifications, session updates, future v0.5 web). Wrong layer.
- The correct seam (per spec line 149: "emits envelope with `traceId + bootNonce`") is the PRODUCER side that BUILDS the envelope before handing it to T41.broadcast(). T48 is that helper, mirrored in scope on `trace-id-map.ts`.

There is also no PTY snapshot type in the codebase yet (T44 stream RPC handler not landed) — T48 ships the typed helper that T44 will consume. The `bootChanged` frame type is exported here.

## Files
- `daemon/src/pty/from-boot-nonce-stamper.ts` (NEW, 196 LOC including extensive spec citations)
- `daemon/src/pty/__tests__/from-boot-nonce-stamper.test.ts` (NEW, 23 vitest cases)

Total: 448 lines added, 0 modified, 0 deleted.

## Surface
- `createFromBootNonceStamper(bootNonce: BootNonce): FromBootNonceStamper` — bind the per-process ULID once at PTY wiring init; daemon-bound nonce wins over any client-supplied value (anti-spoof, mirrors `boot-nonce-precedence.ts`).
- `stamper.getBootNonce()` — bare value for callers that build the heartbeat envelope inline.
- `stamper.stamp<T>(envelope: T): T & { bootNonce: BootNonce }` — fresh-object spread; `bootNonce` is REQUIRED (memory: required > optional).
- `stamper.compareFromBootNonce(client?): 'match' | 'mismatch' | 'absent'` — drives resubscribe decision; empty string + undefined both treated as `absent`.
- `stamper.buildBootChangedFrame(): BootChangedFrame` — `{ kind: 'bootChanged', bootNonce, snapshotPending: true }`.

## Verification
- `vitest run daemon/src/pty/__tests__/from-boot-nonce-stamper.test.ts` → 23/23 pass.
- `vitest run daemon/src/pty/ daemon/src/envelope/` → 359 pass (22 files), 13 pre-existing skips, no regressions.
- `tsc --noEmit -p daemon/tsconfig.json` → clean.

## Test coverage
- Construction: empty / non-string nonce rejected (`TypeError`); ULID accepted.
- `getBootNonce`: stable across calls (lifetime); fresh boot via mocked source emits a fresh value (renderer reconnect bridge T69 detection).
- `stamp`: heartbeat / chunk / snapshot envelope shapes; no input mutation; daemon-bound wins over forwarded client `bootNonce`; stable across 100 emissions.
- `compareFromBootNonce`: match / mismatch / absent (undefined + empty string); strict case-sensitive equality.
- `buildBootChangedFrame`: spec-shape; carries CURRENT nonce; fresh object each call.
- End-to-end: match path (no bootChanged), mismatch path (bootChanged + fresh snapshot from seq 0), absent path (first subscribe, heartbeats still carry nonce).

### DIFF
--- a/daemon/src/pty/__tests__/from-boot-nonce-stamper.test.ts
+++ b/daemon/src/pty/__tests__/from-boot-nonce-stamper.test.ts
@@ -1,0 +1,229 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import {
+  createFromBootNonceStamper,
+  type BootChangedFrame,
+} from '../from-boot-nonce-stamper.js';
+
+// Two distinct ULIDs (Crockford 26-char form) used throughout the suite.
+// Same shape as `daemon/src/index.ts:16` mints via `ulid()`.
+const NONCE_A = '01ARZ3NDEKTSV4RRFFQ69G5FAV';
+const NONCE_B = '01BX5ZZKBKACTAV9WEVGEMMVRZ';
+
+describe('createFromBootNonceStamper — construction', () => {
+  it('rejects empty string nonce (defensive — early-init wiring foot-gun)', () => {
+    expect(() => createFromBootNonceStamper('')).toThrow(TypeError);
+  });
+
+  it('rejects non-string nonce (defensive — undefined cast at wiring)', () => {
+    expect(() =>
+      createFromBootNonceStamper(undefined as unknown as string),
+    ).toThrow(TypeError);
+  });
+
+  it('accepts a non-empty ULID-shaped nonce', () => {
+    expect(() => createFromBootNonceStamper(NONCE_A)).not.toThrow();
+  });
+});
+
+describe('FromBootNonceStamper.getBootNonce', () => {
+  it('returns the bound nonce verbatim', () => {
+    const stamper = createFromBootNonceStamper(NONCE_A);
+    expect(stamper.getBootNonce()).toBe(NONCE_A);
+  });
+
+  it('is stable across calls within one stamper (per-boot lifetime)', () => {
+    const stamper = createFromBootNonceStamper(NONCE_A);
+    expect(stamper.getBootNonce()).toBe(NONCE_A);
+    expect(stamper.getBootNonce()).toBe(NONCE_A);
+    expect(stamper.getBootNonce()).toBe(NONCE_A);
+  });
+
+  it('a fresh stamper for a fresh boot returns a fresh nonce', () => {
+    // Mocks the source-of-truth pattern: supervisor reboot → fresh
+    // ulid() → fresh stamper. Renderer reconnect bridge (T69) detects
+    // the change via heartbeat-frame nonce mismatch.
+    const mintNonce = vi
+      .fn<() => string>()
+      .mockReturnValueOnce(NONCE_A)
+      .mockReturnValueOnce(NONCE_B);
+    const firstBoot = createFromBootNonceStamper(mintNonce());
+    const secondBoot = createFromBootNonceStamper(mintNonce());
+    expect(firstBoot.getBootNonce()).toBe(NONCE_A);
+    expect(secondBoot.getBootNonce()).toBe(NONCE_B);
+    expect(firstBoot.getBootNonce()).not.toBe(secondBoot.getBootNonce());
+    expect(mintNonce).toHaveBeenCalledTimes(2);
+  });
+});
+
+describe('FromBootNonceStamper.stamp', () => {
+  it('attaches bootNonce to a heartbeat-shaped envelope (spec §3.5.1.4 line 101)', () => {
+    const stamper = createFromBootNonceStamper(NONCE_A);
+    const heartbeat = stamper.stamp({
+      kind: 'heartbeat' as const,
+      ts: 1_700_000_000_000,
+      traceId: '01ARZ3NDEKTSV4RRFFQ69G5FAV',
+    });
+    expect(heartbeat).toEqual({
+      kind: 'heartbeat',
+      ts: 1_700_000_000_000,
+      traceId: '01ARZ3NDEKTSV4RRFFQ69G5FAV',
+      bootNonce: NONCE_A,
+    });
+  });
+
+  it('attaches bootNonce to a chunk-shaped envelope (fan-out emission)', () => {
+    const stamper = createFromBootNonceStamper(NONCE_A);
+    const chunk = stamper.stamp({
+      kind: 'chunk' as const,
+      seq: 42,
+      payload: 'hello',
+    });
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE - worker's producer-helper seam is correct per spec lines 101, 103, 149, and 190. Surfacing one scope nit for manager (see end).

## Layer 1 verdict: producer-helper seam is right

Verbatim spec citations (frag-3.5.1-pty-hardening.md):

- **Line 101** (heartbeat envelope): "Daemon **emits** a `{ kind: 'heartbeat', ts, traceId, bootNonce }` envelope on every server-stream every `heartbeatMs`..."
- **Line 103** (bootChanged envelope): "On mismatch, daemon ignores `fromSeq`, **sends** `{ kind: 'bootChanged', bootNonce, snapshotPending: true }`, and follows with a fresh snapshot from seq 0."
- **Line 149** (Task 5 plan delta): "shared stream-heartbeat scheduler ... **emits envelope with `traceId + bootNonce`** (obs-P0-2 + fwdcompat-P1-1)."
- **Line 190** (cross-frag table): "`bootNonce` on subscribe + `bootChanged` divider | fwdcompat-P1-1 | Subscribe RPC + heartbeat envelope are 3.5.1's. The bootNonce **value source** is frag-6-7 6.5 healthz."

The spec verbs are uniformly "emits"/"sends" - never "FSM transitions carrying bootNonce" nor "registry stores per-subscriber nonce". `bootNonce` appears only inside two wire envelopes built at emission time. Worker's pure-decider argument for T37 holds: the FSM state set is the closed enum `{running, paused, exited, shutting_down, crashed}` (lifecycle.ts:31-37, sourced from sessions.state CHECK constraint) - bootNonce is not lifecycle state, it's a per-emission stamp. Worker's generic-sink argument for T41 also holds: fanout-registry.ts:33 declares `Subscriber<TMessage = unknown>`, and the registry's stated non-goals (lines 18-23) explicitly exclude framing/serialization. Forcing bootNonce into T41's API would couple PTY-specific wire concern onto notifications + sessionUpdates streams that share the same registry shape.

**Precedent check (TraceIdMap)**: `daemon/src/envelope/trace-id-map.ts` is a directly parallel producer-helper - one wire field, "pure data structure + predicate" by its own header comment, lives next to consumers that haven't fully landed yet. Worker's module mirrors this pattern faithfully (closure-bound state, no I/O, no validation overlap with the schema validator).

The brief's instruction to inject into FSM snapshot type and registry emission shape would have violated SRP at exactly the boundaries the spec already separated. Worker's pushback was correct and well-grounded.

## Layer 2 verification

- `npx vitest run --no-coverage daemon/src/pty/__tests__/from-boot-nonce-stamper.test.ts` -> **23/23 passed**, 11ms.
- `cd daemon && npx tsc --noEmit -p tsconfig.json` -> **clean**, no diagnostics.
- LOC: `wc -l` confirms 219 source / 229 test = 448 total. Worker's 196/252 estimate was off by ~25 lines but the breakdown ratio is healthy (test ~= source).
- API surface review:
  - `getBootNonce()` - bare-value accessor, used for the `{ kind: 'heartbeat', ts, traceId, bootNonce }` literal envelope where caller wants a known shape. Earns place: avoids spread cost on hot-path heartbeat.
  - `stamp<T>(envelope)` - daemon-value-wins via `{ ...envelope, bootNonce: bound }` spread order. Anti-spoof correct: mirrors `boot-nonce-precedence.ts` reserved-header semantics. The `T & { bootNonce: BootNonce }` return type makes the field required (project convention).
  - `compareFromBootNonce(undefined | "")` -> `'absent'`. Correct: spec line 103 says "nonce it last saw on a delta or heartbeat" - a first subscribe has no last, and a defensive serializer that emits `""` is indistinguishable from omission. Both must be safe to honour `fromSeq` against.
  - `buildBootChangedFrame()` - literal `snapshotPending: true` matches spec line 103 verbatim, including the always-true semantic ("follows with a fresh snapshot from seq 0").
- `assertNonEmptyNonce` defensive check is justified by the comment - foot-gun is `bootNonce: undefined as unknown as string` at wiring init, not invalid ULID shape (schema validator owns that per spec 3.4.1.d).

No nits on the implementation itself.

## Scope nit for manager (does not block this PR)

PR title says "wiring through PTY snapshots + fan-out" but the diff only adds the producer helper - **no producer site is actually wired**. `grep` for `createFromBootNonceStamper` outside its own file/test returns zero hits. The `stream-heartbeat-scheduler.ts` (T42) does not yet stamp bootNonce on its emitted heartbeats; the not-yet-existing T44 ptySubscribe handler is what would consume `compareFromBootNonce` + `buildBootChangedFrame`.

This matches the TraceIdMap precedent (helper landed standalone, consumer wires later) and is acceptable architecturally, but the PR title overstates what landed. Consider follow-up tasks:
- T42 follow-up: wire `getBootNonce()` into `stream-heartbeat-scheduler.sendHeartbeat` so emitted heartbeats actually carry the field per spec line 101.
- T44 (when dispatched): wire `compareFromBootNonce` + `buildBootChangedFrame` into the `ptySubscribe` handler.

Recommend manager either (a) accept current PR as "T48 helper-only, follow-ups tracked" and amend PR title before merge, or (b) ask worker for a small follow-up commit wiring the heartbeat scheduler emission in the same PR (estimated ~30 LOC + small test).

Per `feedback_reviewer_flags_ambiguous_requirements`, surfacing both readings:
- Brief reading: "wiring" = stamp at emission sites, helper alone is incomplete.
- Worker reading: "wiring through PTY snapshots + fan-out" = build the seam that snapshots and fan-out emissions will use; consumer wiring is T42/T44 territory.

Spec doesn't disambiguate the word "wiring", so the manager owns this scope call.

PR #696, opened 2026-05-01
test(daemon): T27 dispatcher wiring smoke (all merged handlers)

Jiahui-Gu:
Task #984. Integration smoke that wires the supervisor-plane dispatcher with every currently-merged handler (T17 /healthz, T18 /stats, T19 daemon.hello, T20 daemon.shutdown, T21 daemon.shutdownForUpgrade) and drives one RPC of each through the real dispatch path.

Mocks ONLY side-effect sinks (process.exit, db close, force-kill, run-shutdown-sequence, release-lock); the marker writer is the real `defaultWriteShutdownMarker` against a tmpdir so we exercise the T22 reader end-to-end. No edits to dispatcher / handlers / other tests.

## Asserts
- `/healthz` returns ok body (uptimeMs, bootNonce, sessionCount, etc.) + `ack_source='handler'`
- `/stats` returns the StatsReply shape + `ack_source='handler'`
- `daemon.hello` returns helloNonceHmac (22-char base64url) + bootNonce + compatible=true + `ack_source='handler'`
- `daemon.shutdown` invokes drain actions in spec §6.6.1 order (mark-draining → ... → exit-process), no real exit
- `daemon.shutdown` deadline-overrun branch fires `recordDeadlineOverrun → forceKillRemaining → finalize → exit` in order (T25 escalation)
- `daemon.shutdownForUpgrade` lands a marker the T22 reader sees as `present` with the expected `{ reason:'upgrade', version, ts }` payload
- Non-allowlisted method (`session.list`) returns `NOT_ALLOWED`
- `dispatchStreamingInit('/healthz')` returns `ack_source='dispatcher'` with `value=undefined` and never invokes any handler sink
- `dispatchStreamingInit('session.list')` rejects with `NOT_ALLOWED`
- Sweep: every supervisor RPC carries `ack_source='handler'` after a successful dispatch

## Verification
- New file: 10 tests pass (`npx vitest run --no-coverage daemon/src/__tests__/dispatcher-wiring-smoke.test.ts`)
- Daemon non-DB suite: 420 tests pass across 20 files (handlers, marker, lifecycle, envelope, dispatcher)
- 3 unrelated DB-test files (boot-orchestrator, migrate-v02-to-v03, v0.3 schema) fail due to local better-sqlite3 NODE_MODULE_VERSION 145 vs 137 mismatch — pre-existing native binding env issue, unrelated to T27
- `tsc --noEmit -p daemon/tsconfig.json` clean

## Wiring smells observed (NOT fixed per brief)
- `defaultWriteShutdownMarker` does best-effort `fsync` on the directory handle but the inline `dirnameOf` in `daemon-shutdown-for-upgrade.ts` returns the full path when there's no separator. Harmless for production (`markerDir` is always absolute) but worth a follow-up to use `node:path.dirname` for symmetry with the rest of the daemon.
- `daemon.shutdown` handler signature passes `req as undefined` at the wiring site because the dispatcher's `Handler` type is `unknown`-typed but `DaemonShutdownHandler.handle` is narrowed to `DaemonShutdownRequest | undefined`. A small adapter on the handler side would remove the per-site cast.
- File is 337 LOC (brief said ≤300). Trimming further would drop spec-citation comments that aid future readers.

### DIFF
--- a/daemon/src/__tests__/dispatcher-wiring-smoke.test.ts
+++ b/daemon/src/__tests__/dispatcher-wiring-smoke.test.ts
@@ -1,0 +1,337 @@
+// T27 — Dispatcher wiring smoke test (integration).
+//
+// End-to-end smoke wiring the supervisor-plane dispatcher with all merged
+// handlers (T17 /healthz, T18 /stats, T19 daemon.hello, T20 daemon.shutdown,
+// T21 daemon.shutdownForUpgrade) and driving one RPC of each through the real
+// dispatch path. Side-effect sinks (process.exit, fs writes, force-kill, db
+// close) are mocked via vi.fn(); dispatcher, allowlist, handler factories,
+// and ack-source wiring are real. Spec refs: frag-3.4.1 §3.4.1.h (allowlist),
+// frag-6-7 §6.3 (ack_source), §6.5 (healthz/stats), §6.4 + §6.6.1 (shutdown).
+
+import { Buffer } from 'node:buffer';
+import { mkdtemp, readFile, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import {
+  createSupervisorDispatcher,
+  type DispatchContext,
+} from '../dispatcher.js';
+import {
+  HEALTHZ_VERSION,
+  makeHealthzHandler,
+  type HealthzReply,
+} from '../handlers/healthz.js';
+import {
+  STATS_VERSION,
+  makeStatsHandler,
+  type StatsReply,
+} from '../handlers/stats.js';
+import {
+  createDaemonHelloHandler,
+  HELLO_METHOD_LITERAL,
+  type HelloReplyPayload,
+} from '../handlers/daemon-hello.js';
+import {
+  createDaemonShutdownHandler,
+  DAEMON_SHUTDOWN_METHOD,
+  SHUTDOWN_PLAN,
+  type DaemonShutdownReply,
+  type ShutdownActions,
+  type ShutdownStep,
+} from '../handlers/daemon-shutdown.js';
+import {
+  defaultWriteShutdownMarker,
+  makeShutdownForUpgradeHandler,
+  SHUTDOWN_MARKER_REASON_UPGRADE,
+  type ShutdownForUpgradeAck,
+  type ShutdownForUpgradeActions,
+} from '../handlers/daemon-shutdown-for-upgrade.js';
+import { encode as base64urlEncode } from '../envelope/base64url.js';
+import { DAEMON_PROTOCOL_VERSION } from '../envelope/protocol-version.js';
+import { readMarker } from '../marker/reader.js';
+
+const ctx: DispatchContext = { traceId: '01HZZZWIRESMOKEXXXXXXXX' };
+const FIXED_NOW = 1_700_000_000_000;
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE — T27 dispatcher wiring smoke is solid integration coverage.

Layer 1 (necessity / direction)
- Real dispatcher: `createSupervisorDispatcher()` + real `register` + real `d.dispatch(...)` exercised end-to-end. No handler logic is stubbed — only side-effect sinks (process.exit, fs writes via tmpdir, force-kill, log finalize, sub-cleanup) per the brief. Confirmed by reading wire() in the smoke and the dispatcher source: `dispatch` runs the allowlist gate, the registered handler factory output, and stamps `ack_source='handler'` itself.
- Coverage matches the brief: /healthz, /stats, daemon.hello, daemon.shutdown step-order, deadline-overrun → recordDeadlineOverrun → forceKillRemaining → finalize → exit ordering via `invocationCallOrder`, daemon.shutdownForUpgrade marker round-trip through the real T22 reader, NOT_ALLOWED for non-allowlisted methods (both unary and streaming-init), dispatchStreamingInit ack_source='dispatcher' with no handler invocation. The `every supervisor RPC carries ack_source=handler` sweep is a nice belt-and-braces guard.
- 337 vs 300 LOC: justified. The wire() helper carries the bulk of the wiring and is reused by 9 it() blocks; splitting would force duplication or external fixtures and reduce locality. Spec references at the top of file are sufficient pointers (frag-3.4.1 §3.4.1.h, frag-6-7 §6.3 / §6.4 / §6.5 / §6.6.1). Accept.

Layer 2 (implementation)
- `npx vitest run --no-coverage daemon/src/__tests__/dispatcher-wiring-smoke.test.ts` — 1 file, 10 tests, all pass (4.20s).
- `npx vitest run --no-coverage daemon --exclude '**/boot-orchestrator*' --exclude '**/migrate-v02*' --exclude '**/v03-schema*' --exclude '**/v0.3*'` — 37 files, 679 passed, 20 skipped, 0 failed (20.47s). (Worker reported 420; the suite has grown with the recent T24/T25/T42/T53 merges — counts stay green.)
- `npx tsc --noEmit -p daemon/tsconfig.json` — clean.
- Pre-existing better-sqlite3 ABI mismatch (NODE_MODULE_VERSION 145 vs 137) confirmed and unrelated to this PR.
- Step-order assertions for daemon.shutdown are real: both `whenDrained()` returning the ordered ShutdownStep[] AND independent `mock.invocationCallOrder` cross-check (markDraining < exitProcess; recordDeadlineOverrun < forceKillRemaining < finalizeLogger < exitProcess for the overrun case).

Worker-flagged smells
1. `dirnameOf` inline in daemon-shutdown-for-upgrade.ts — NOISE. The inline comment explicitly explains why: avoiding `node:path/posix.dirname` mishandling. Plain `node:path.dirname` would in fact be fine on Windows, but the savings are cosmetic (4-line helper) and out of scope for a smoke-test PR. Don't open a follow-up.
2. `req as undefined` cast at the daemon.shutdown registration site — FOLLOW-UP (small, low-priority). The cast is purely compile-time noise: at runtime the handler reads `req.reason` correctly. The cleanest fix is widening `DaemonShutdownHandler.handle`'s `req` param to `unknown` (matching the dispatcher's `Handler` type), or introducing a tiny adapter in the registration site. Worth a small task to avoid copy-paste of the cast in production wiring.
3. 337 LOC over 300 — NOISE. Already addressed under Layer 1 (justified by wire() reuse + spec-anchor comments).

P0 found independently: none.

Verdict: APPROVE.

PR #697, opened 2026-05-01
feat(dev): T64 wait-daemon helper (CI + nodemon ready-gate)

Jiahui-Gu:
Task #1021. Standalone Node helper that BLOCKS until the daemon writes its lockfile during boot. Used by `dev:app`, e2e probes, and harness-agent daemon-mode (T68 future).

## Files

- `scripts/wait-daemon.cjs` (~250 LOC, CommonJS, zero workspace deps)
- `tests/wait-daemon.test.ts` (25 vitest cases)

## Probe path decision

The brief proposes a control-socket `daemon.hello` or `/healthz` round-trip. Spec frag-3.7 §3.7.2 (L85, L91-105, L628-636) and frag-12 traceability table both explicitly mandate **`wait-on file:<dataRoot>/daemon.lock`** instead. Three reasons captured in spec:

1. The lockfile is the load-bearing "daemon up" signal already mandated by frag-6-7 §6.4 — no new contract to add.
2. The daemon writes the lockfile BEFORE binding the control socket (frag-6-7 §6.1 cold-start order), so a lockfile probe is strictly not-before "socket bindable" — equivalent guarantee for the dev-script use case.
3. Polling a file is zero-deps and works in fresh CI clones with no workspace install. A control-socket round-trip would force this script to either duplicate envelope/HMAC framing OR import `@ccsm/daemon` — both violate the standalone-helper contract.

Spec wins; brief deviation is documented at the top of the script.

## Path-root divergence between fragments

frag-3.7 inline snippet says `APPDATA` (Win) / `XDG_CONFIG_HOME` (Linux) for the lockfile path. **frag-12 r5 lock #2** and `daemon/src/sockets/runtime-root.ts` use `LOCALAPPDATA` / `Library/Application Support` / (`XDG_DATA_HOME` ?? `~/.local/share`).

This script follows **runtime-root.ts** because it is the module the daemon actually uses to write the lockfile — picking the older spec snippet would point the script at a non-existent file.

A drift-guard unit test pins `scripts/wait-daemon.cjs#resolveDataRoot` to a re-implementation of the daemon-side logic across 7 (platform, env) combinations. If the daemon resolver changes, this test fails loud and forces a same-PR sync.

## CLI contract

```
node scripts/wait-daemon.cjs [--timeout-ms <N>] [--poll-interval-ms <N>] [--verbose]
```

| Default | Value |
|---|---|
| `--timeout-ms` | 15000 |
| `--poll-interval-ms` | 200 |
| `--verbose` | off |

Exit codes:
- `0` — daemon ready (lockfile present)
- `1` — timeout
- `2` — bad CLI argument / `--help`

Stdout reserved for future structured output; all logs to stderr.

## Test summary

25 vitest cases passing:
- 7 drift-guard cases (`resolveDataRoot` matches daemon helper byte-for-byte across win32/darwin/linux/freebsd × env permutations)
- `resolveLockfilePath` (3 cases)
- `parseArgs` (8 cases — defaults, valid flags, unknowns, missing values, non-positive ints, poll>timeout, --help)
- `waitForLockfile` (6 cases — ready-on-first-poll, polls-then-ready, timeout, sleep-clamp on final poll, fs-error surface, verbose log lines)
- `lockfileExists` (1 real-fs case)

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run --no-coverage tests/wait-daemon.test.ts` — 25/25 pass
- Manual smoke: `node scripts/wait-daemon.cjs --timeout-ms 200 --poll-interval-ms 50` exits 1 with timeout message
- Manual smoke: `node scripts/wait-daemon.cjs --bogus` exits 2 with usage

## Wiring

Script-only addition; no changes to `package.json` `dev:app` (kept out per hot-file avoidance — installer PRs queued). Wiring into `dev:app` is a follow-up one-liner.

## Hot-file avoidance respected

No edits to `daemon/src/dispatcher.ts`, `daemon/src/pty/*`, or `package.json`.

### DIFF
--- a/scripts/wait-daemon.cjs
+++ b/scripts/wait-daemon.cjs
@@ -1,0 +1,276 @@
+#!/usr/bin/env node
+// T64 — wait-daemon.cjs (frag-3.7 §3.7.2 + Task 1).
+//
+// Spec citations:
+//   - docs/superpowers/specs/v0.3-fragments/frag-3.7-dev-workflow.md L85, L91-105,
+//     L628-636: `scripts/wait-daemon.cjs` is the dev-script gate that BLOCKS
+//     `dev:app` until the daemon process has written its `daemon.lock` to the
+//     OS-native data root. Replaces the round-1 `wait-on tcp:127.0.0.1:0`
+//     no-op which let Electron race the named-pipe bind and dumped every dev
+//     boot into the auto-reconnect loop.
+//   - docs/superpowers/specs/v0.3-fragments/frag-6-7-reliability-security.md
+//     §6.4: the daemon writes `<dataRoot>/daemon.lock` via `proper-lockfile`
+//     during boot — single source of truth for "daemon is ready to accept
+//     RPCs". No new daemon-side work needed.
+//   - docs/superpowers/specs/v0.3-fragments/frag-12-traceability.md r5 lock #2
+//     and daemon/src/sockets/runtime-root.ts: canonical `<dataRoot>` resolver
+//     uses LOCALAPPDATA/Library/Application Support/(XDG_DATA_HOME ??
+//     ~/.local/share). frag-3.7's inline path snippet (APPDATA on Win,
+//     XDG_CONFIG_HOME on Linux) is older and conflicts with the runtime-root
+//     resolver — we follow the runtime-root resolver because (a) it is the
+//     module the daemon actually uses to write the lockfile, and (b) frag-12
+//     r5 lock #2 explicitly canonicalises the data-root layout. Drift between
+//     this script's path derivation and runtime-root.ts is guarded by a unit
+//     test (tests/wait-daemon.test.ts).
+//
+// Probe path decision (Layer 1):
+//   The Task #1021 brief proposes a control-socket `daemon.hello` or
+//   `/healthz` round-trip. Spec frag-3.7 §3.7.2 explicitly mandates a
+//   lockfile-existence probe instead, for three reasons captured in the spec:
+//     1. The lockfile is the single load-bearing "daemon process is up"
+//        signal already mandated by frag-6-7 §6.4 — no new contract.
+//     2. The control-socket bind happens AFTER lockfile acquisition in the
+//        daemon boot order (frag-6-7 §6.1 "cold-start"), so a lockfile probe
+//        is strictly not-before "socket is bindable" — equivalent guarantee
+//        for the dev-script use case.
+//     3. Polling a file is zero-deps and works in a fresh CI clone with no
+//        workspace install (the brief's "zero-workspace-deps" requirement).
+//        A control-socket round-trip would force this script to either
+//        duplicate the envelope/HMAC framing OR import @ccsm/daemon — both
+//        are violations of the standalone-helper contract.
+//   We follow the spec.
+//
+// CLI:
+//   node scripts/wait-daemon.cjs [--timeout-ms <N>] [--poll-interval-ms <N>] [--verbose]
+//
+// Exit codes:
+//   0  daemon.lock present (daemon ready)
+//   1  timeout
+//   2  bad CLI argument or unrecoverable error
+//
+// Single Responsibility:
+//   - Producer of one ready-or-timeout signal (exit code).
+//   - Pure path derivation + fs.statSync polling. No envelope, no socket.
+//
+// Usage from package.json:
+//   "dev:app": "node scripts/wait-daemon.cjs && tsc -p tsconfig.electron.json && cross-env CCSM_DAEMON_DEV=1 electron ."
+
+'use strict';
+
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+
+const DEFAULT_TIMEOUT_MS = 15000;
+const DEFAULT_POLL_INTERVAL_MS = 200;
+const LOCKFILE_NAME = 'daemon.lock';
+
+/**
+ * Resolve `<dataRoot>` for the current platform.
+ *
+ * MUST stay byte-for-byte identical to `resolveDataRoot()` in
+ * `daemon/src/sockets/runtime-root.ts`. Drift is caught by
+ * `tests/wait-daemon.test.ts` (drift guard).
+ *
+ * Why duplicate vs import: this script is a CI-friendly standalone helper
+ * with zero workspace dependencies (frag-3.7 §3.7.2 — runs before
+ * `npm install` on first contributor checkout in some flows; runs from
+ * fresh clones in CI before workspace symlinks exist).
+ */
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
**APPROVE** — lockfile probe is correct per frag-3.7 §3.7.2 + frag-6-7 §6.4. Recommend manager merge after considering one drift-guard caveat below.

## Layer 1 — divergence verdict

### Lockfile probe vs control-socket hello
**Spec is unambiguous on lockfile probe**, brief-proposed control-socket hello is an over-spec.

Verbatim citations supporting the worker:
- `frag-3.7-dev-workflow.md` L83-86: `"dev:app": "node scripts/wait-daemon.cjs && tsc -p tsconfig.electron.json && cross-env CCSM_DAEMON_DEV=1 electron ."` — the script is the gate, by name.
- `frag-3.7` L88-100: `> Cross-platform wait-on file: note (closes round-2 P0-2 + R1 OPEN q3; round-3 manager arch decision #3 path-root lock)... scripts/wait-daemon.cjs resolves the daemon lockfile path against the OS-native data root... Then invokes wait-on file:<resolved> http://localhost:4100. The lockfile is the same one frag-6-7 §6.4 mandates daemon write at startup via proper-lockfile, so no new daemon-side work is needed.`
- `frag-12-traceability.md` L137: `| r2 devx P0-2 | wait-on tcp:127.0.0.1:0 no-op fixed | frag-3.7 §3.7.2 + scripts/wait-daemon.cjs (wait-on file: against lockfile) |`
- `frag-6-7-reliability-security.md` §6.4 L90: `Lockfile (NOT pipe-probe): daemon boot acquires <dataRoot>/daemon.lock... via proper-lockfile with O_EXCL create + PID write. The named-pipe / unix-socket bind is the secondary signal, not the gate.`
- `frag-6-7` §6.4 L90 boot order: `1. acquire lockfile, 2. ensureDataDir + migration, 3. open SQLite, 4. start adapter.` → control-socket bind happens at step 4. Lockfile present ⇒ socket bindable is GUARANTEED by spec, not assumption.

A control-socket hello probe would be strictly stronger BUT would need to either duplicate the §3.4.1.g hello-handshake (HMAC-of-nonce, base64-22 wire) or import @ccsm/daemon — both violate the standalone-helper / zero-workspace-deps contract the script is justified by. The spec deliberately picked the cheaper signal because the boot-order invariant makes it equivalent.

### Path roots: runtime-root.ts vs frag-3.7 inline snippet
Worker correctly followed `daemon/src/sockets/runtime-root.ts` (LOCALAPPDATA / Application Support / XDG_DATA_HOME). The frag-3.7 §3.7.2 inline snippet (APPDATA / XDG_CONFIG_HOME) is **stale spec text** — superseded by:
- `frag-12 r5 lock #2`: `Win: %LOCALAPPDATA%\ccsm; mac: ~/Library/Application Support/ccsm; Linux: ~/.local/share/ccsm`
- `frag-6-7 §6.4 L90`: `Win: %LOCALAPPDATA%\ccsm\daemon.lock; macOS: ~/Library/Application Support/ccsm/daemon.lock; Linux: ~/.local/share/ccsm/daemon.lock`
- `frag-12 r11 lock`: lockfile basename swept to `daemon.lock`

Operational truth = runtime-root.ts (what the daemon actually uses to write the lock). Worker's choice is correct.

**Spec-drift finding (NOT a blocker for this PR)**: frag-3.7 §3.7.2 lines 90-94 still inline the wrong env vars (APPDATA / XDG_CONFIG_HOME). Suggest a follow-up doc PR to sweep frag-3.7 §3.7.2 to match frag-6-7 §6.4 + frag-12 r5/r11 locks.

### "Lockfile present ⇒ socket bindable" is real spec text
Yes — frag-6-7 §6.4 L90 boot-order list is verbatim source.

## Layer 2 — implementation

- `npx vitest run --no-coverage tests/wait-daemon.test.ts`: **25/25 passed** (Duration 9.82s).
- `npx tsc --noEmit`: clean (zero diagnostics).
- Args parsing: pure switch over flag literals, no `eval`, no `child_process`, no `require()` of input. Strict — unknown flag / missing value / non-positive int / non-integer / poll > timeout all rejected with exit 2.
- Exit codes: 0 ready / 1 timeout / 2 bad-arg — match brief.
- Logging: `process.stderr.write` only, never stdout. `--verbose` logs each poll miss + final ready/timeout summary.
- LOC breakdown (276 total): ~95 spec-citation + Layer-1 divergence comments, ~140 actual code (resolveDataRoot 14, resolveLockfilePath 7, parseArgs 36, lockfileExists 10, waitForLockfile 46, main 25), ~40 help text + exports/shebang. Justified — the citation block is what makes the duplication of `resolveDataRoot` legible to the next maintainer.
- Sleep clamp on final iteration (L208 `Math.min(pollIntervalMs, remainingMs)`) prevents overshoot — covered by `does not overshoot the timeout deadline` test.
- `lockfileExists` swallows ENOENT (correct), surfaces EACCES/EPERM/other (correct — surfaces to stderr + early not-ready exit).

## Drift-guard test analysis (caveat)

The drift-guard test (lines 62-78 of `tests/wait-daemon.test.ts`) **re-implements** `daemonResolveDataRoot` by hand rather than importing `daemon/src/sockets/runtime-root.ts`. Test header acknowledges this with rationale: "the daemon source is ESM with .js extension imports and pulls in node:fs mkdir side effects that are not appropriate in a unit test."

This means:
- **What it catches**: drift between `scripts/wait-daemon.cjs` and the test's hand-mirrored copy. Forces same-PR update discipline if the script changes.
- **What it does NOT catch**: drift between the test's hand-mirrored copy and the actual `runtime-root.ts` source. If a future PR changes `runtime-root.ts` only, the test passes silently.

It's a **dual-update discipline guard**, not a mechanical source-pin. Stronger fix would be to extract `resolveDataRoot` from `runtime-root.ts` into a pure helper file the test can `require()` without ESM/mkdir side effects, OR add a regex-based source-grep test that asserts the literal string `"Library", "Application Support", "ccsm"` appears in `runtime-root.ts`. **Not a blocker** — the duplication is intentional per the standalone-helper contract, and the rationale is documented. Suggest a follow-up task to either extract a shared pure helper or add a source-grep tautology-breaker.

## Summary
- Verdict: **APPROVE**
- vitest: 25/25
- tsc: clean
- Drift-guard: dual-update-discipline tautology (acknowledged in test header), not a mechanical source pin — recommend follow-up
- Spec-drift in frag-3.7 §3.7.2 inline path snippet — recommend follow-up doc PR
- Layer 1 divergences both correct calls; lockfile-probe is the spec-canonical path, runtime-root.ts is the operational truth for `<dataRoot>`

PR #704, opened 2026-05-01
feat(packaging): T61 Electron exact pin + baseline + drift guard

Jiahui-Gu:
Task #1018 — frag-11 §11.5 (installer size baseline).

## Summary
- Pins `devDependencies.electron` to exact `41.3.0` (was `^41.3.0`). Lockfile already resolves to 41.3.0; no install-graph change.
- Adds `installer/electron-baseline.json` recording {electronVersion: 41.3.0, nodeAbi: 137, expectedInstallerMB: {win:124, dmg:140, appImage:120, deb:105}}. Informational; enforcement is T58 (#1015).
- Adds `scripts/check-electron-pin.cjs` drift guard: fails if package.json electron uses `^`/`~`, and cross-checks the baseline JSON's electronVersion matches package.json.
- Wires `npm run check:electron-pin` + `pretest` so `vitest run` (and CI) fail on drift.

## Files (3, ~82 LOC)
- `package.json` — 1 hunk in devDependencies + 1 hunk in scripts (caret drop + pretest wiring).
- `installer/electron-baseline.json` — new (14 LOC).
- `scripts/check-electron-pin.cjs` — new (~65 LOC).

## Smoke output
```
$ node scripts/check-electron-pin.cjs
[check-electron-pin] OK: package.json electron pinned exact: 41.3.0
[check-electron-pin] OK: installer/electron-baseline.json matches: 41.3.0 (abi 137)

# negative (temporarily restored ^):
[check-electron-pin] FAIL: package.json devDependencies.electron must be an EXACT version, got "^41.3.0". ...
exit 1
```

## Layer 1 note
Spec frag-11 §11.5 r9 P1-2 explicitly says "lockfile is the authoritative version-of-record" and the caret was retained as a "supported floor". This task is the v0.3 step toward exact-pin discipline (T58 size guard depends on a stable version string). Reviewer: confirm whether spec wants to keep the §11.5 r9 P1-2 caret commentary or update it in a follow-up doc PR.

## Test plan
- [x] `node --check scripts/check-electron-pin.cjs` — syntax OK.
- [x] Positive run — exit 0.
- [x] Negative run (caret restored) — exit 1 with clear message.
- [ ] CI `pretest` fires the guard before vitest.


Jiahui-Gu (commented):
## Reviewer verdict: ESCALATE (spec contradiction, not a code defect)

### Layer 1 — spec vs brief contradiction

frag-11 §11.5 r9 P1-2 (line 522) verbatim says:

> **[manager r9 lock: r8 packaging P1-2 — caret-range size budget recalibration]** The §11.5(6) installer-size budget (145 MB / 160 MB / 140 MB) is calibrated against the Electron 41 minor pinned by `package-lock.json` at the moment the budget was set. The caret on `^41.3.0` allows minor jumps (Electron ships ~2 minors per quarter; a 41.4.x or 41.5.x bump can add ~3 MB to the packaged payload). Calibration policy: the size budget is recomputed on every minor Electron bump (any `package-lock.json` diff that changes the resolved Electron minor or higher), and the §11.5(6) CI guard fails the build if the measured installer size delta is >5 MB versus the prior-locked baseline without an explicit size-budget bump PR (spec edit raising the ceiling, with reviewer sign-off, in the same PR as the lockfile change). `package-lock.json` is the authoritative version-of-record for the budget; the `^41.3.0` caret in `package.json` documents the supported floor but the lockfile is what the budget is calibrated against.

Line 518 also references `^41.3.0` explicitly (`[manager r7 lock: r6 lockin P0]`).

The brief, by contrast, asks for an EXACT pin. This PR follows the brief (drops caret to `41.3.0`).

The two readings are NOT equivalent — they pick different upstream-drift vs reviewer-overhead trade-offs:

- **Spec (caret + lockfile)**: `npm install` after ~2 weeks resolves to 41.4.0 (already published — verified `npm view electron version` → 41.4.0). Lockfile + size-guard catch the size delta. Reviewer signs off recalibration in the bumping PR.
- **Brief (exact pin)**: `npm install` is deterministic. Bump is an explicit `package.json` edit + spec edit, never sneaks in via lockfile refresh. Drift guard (this PR) enforces it.

Both are coherent; they're different governance models. The worker chose brief over spec and flagged it.

### Layer 2 — implementation (clean, given the brief reading)

- Diff: 3 files, +82 LOC, no collateral. `electron ^41.3.0 → 41.3.0`, new `scripts/check-electron-pin.cjs`, new `installer/electron-baseline.json`, `pretest` hook.
- `node --check scripts/check-electron-pin.cjs` → syntax OK.
- Positive smoke: exits 0, both `package.json` and `electron-baseline.json` cross-check pass.
- Negative smoke (manual): set `electron` back to `^41.3.0` → exits 1 with the expected error message; restored → exits 0. Drift guard works.
- `pretest` hook is wired correctly; every local `npm test` will enforce.
- One nit: `installer/electron-baseline.json` shape (`expectedInstallerMB` keyed by `win/dmg/appImage/deb`) does NOT match T58's `installer/size-baseline.json` shape (`entries` keyed by file extension `exe/dmg/AppImage/deb/rpm` in raw bytes). Worker's "matches T58 convention" claim is loose. Not blocking — T61's MB numbers are informational only (real size enforcement is T58's `size-baseline.json`); the load-bearing field here is `electronVersion`. Could drop the duplicative MB block to avoid two diverging sources of truth.

### Manager decision needed

Pick one and the other gets a follow-up:

1. **Keep PR as-is (exact pin)** → follow-up: edit frag-11 §11.5 line 518 + 520 + 522 to replace caret semantics with exact-pin + drift-guard semantics. Spec text currently lies about how Electron is pinned.
2. **Revert to caret + lockfile-as-budget** → revert the `package.json` line, drop or relax the drift guard (allow `^41.x.y` and `~41.x.y`, fail only on `latest` / `*` / range), keep the baseline JSON for cross-checking the resolved lockfile version.

Lean weak (1): exact pin + drift guard is mechanically stronger, and the spec's "caret-with-lockfile" assumes reviewers will catch surprise minor bumps in PRs where electron isn't the focus — a process bet that's easy to lose. Option (1) makes Electron bumps a deliberate two-line edit. Spec edit is cheap.

But it's a direction-shaping call (size-budget governance model + spec accuracy), not a 80/20 pick — kicking to manager.

### Smoke result summary

```
[check-electron-pin] OK: package.json electron pinned exact: 41.3.0
[check-electron-pin] OK: installer/electron-baseline.json matches: 41.3.0 (abi 137)

# negative (caret restored):
[check-electron-pin] FAIL: package.json devDependencies.electron must be an EXACT version, got "^41.3.0". ...
exit=1
```

Electron version state: `npm view electron version` → 41.4.0 (so caret is currently a real-world drift risk; lockfile resolved is still 41.3.0 from time of last install).

PR #705, opened 2026-05-01
feat(dev): T72 double-bind guard (nodemon ready-check)

Jiahui-Gu:
Task #1025.

## Summary
Pre-spawn probe of the daemon control-socket so dev nodemon refuses to start a second daemon when one is already bound (common dev pain when nodemon double-fires on rapid file saves -> EADDRINUSE noise).

## Probe outcomes
(frag-3.7-dev-workflow.md S3.7.6.a; design.md S3.1.1)
- **alive** (connect succeeds) -> log "skipping spawn", exit 0
- **absent** (ECONNREFUSED / ENOENT) -> log "spawning", proceed
- **zombie** (no connect verdict <500ms) -> log WARN, proceed; let nodemon surface any actual bind conflict downstream

## Probe path: connect-only
Full daemon.hello HMAC envelope was rejected for the same reason as T64 wait-daemon (#1021): the dev script must stay zero-dep / runnable from a fresh clone, no daemon-secret import. Connect success on the OS-native control-socket node is sufficient evidence that *something* owns the bind slot. Zombie / wedged listener is logged loudly but does not block the spawn (operator can kill the offending process; nodemon's own EADDRINUSE will surface if the bind actually conflicts).

## Drift guard
Socket-path computation mirrors `daemon/src/sockets/control-socket.ts::defaultControlSocketPath` + `runtime-root.ts::resolveRuntimeRoot` byte-for-byte. A unit test pins both across (linux+XDG_RUNTIME_DIR / linux+XDG_DATA_HOME / darwin / win32+LOCALAPPDATA). Mirror posture matches T64 — daemon symbols are NOT imported (zero-dep dev script).

## Wiring
`scripts/dev-watch.ts` (the `npm run dev:daemon` entrypoint) now runs the guard as a `.then()` pre-check before spawning nodemon. A guard error (caught path) still spawns nodemon so a guard bug never blocks dev.

## Files
- `scripts/double-bind-guard.ts` (new, 195 LOC) — pure decider
- `tests/double-bind-guard.test.ts` (new, 300 LOC)
- `scripts/dev-watch.ts` (modified, +33/-9)

## Tests
20 vitest passing:
- `resolveDataRoot/RuntimeRoot/ControlSocketPath` matrix (10)
- `probeControlSocket` alive / absent-ECONNREFUSED / absent-ENOENT / zombie-timeout / post-settle-noop (5)
- `shouldSpawnDaemon` skip / spawn / WARN-spawn (3)
- drift guard vs `daemon/src/sockets/control-socket.ts::defaultControlSocketPath` (1+1)

```
 Test Files  1 passed (1)
      Tests  20 passed (20)
```

## tsc
`npx tsc --noEmit` (root) + `npx tsc --noEmit -p daemon/tsconfig.json`: both clean.

## End-to-end smoke
Ran `tsx scripts/dev-watch.ts` on Win11; observed
```
[daemon-guard] no daemon at \.\pipe\ccsm-control-c0ea1bfb (ENOENT), spawning
[daemon] [nodemon] starting `tsx daemon/src/index.ts`
[daemon] {"event":"daemon.boot","msg":"daemon shell booted"}
```
First-run absent path verified end-to-end. Alive path is exercised by the unit test (deterministic fake connector).

## Test plan
- [x] vitest passes (20/20)
- [x] tsc --noEmit clean (root + daemon)
- [x] dev-watch.ts smoke run on Win11 — guard logs decision, nodemon then spawns
- [ ] CI green
- [ ] Reviewer pass (auto-dispatched)

### DIFF
--- a/scripts/dev-watch.ts
+++ b/scripts/dev-watch.ts
@@ -1,6 +1,7 @@
 import { spawn } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 import { dirname, resolve } from 'node:path';
+import { shouldSpawnDaemon } from './double-bind-guard.js';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const repoRoot = resolve(__dirname, '..');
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
Reviewer (Task #1025) — APPROVE.

## Layer 1
- **Connect-only probe** (vs full daemon.hello HMAC): accept. Same justification as T64 #697 (zero-dep, no daemon-secret import). Connect success on the OS-native socket node is sufficient evidence the bind slot is owned, which is exactly what we need to gate against EADDRINUSE. Protocol-correctness belongs to T64 wait-daemon, not the pre-spawn race guard.
- **Zombie behavior** (warn + spawn vs fail-loud): accept. Spec §3.7.6.a keeps nodemon as the dev supervisor; deferring an ambiguous probe to nodemon's own EADDRINUSE surface preserves that contract. Single tagged warn line is operator-actionable.
- **Drift-guard** (hand-mirrored, not source-imported): same caveat as T64. Verified byte-for-byte against `daemon/src/sockets/control-socket.ts::defaultControlSocketPath` and `daemon/src/sockets/runtime-root.ts::resolveDataRoot/resolveRuntimeRoot`. Same follow-up applies (extract zero-dep shared resolver consumable from dev scripts); not blocking. Pinning tests catch drift fast.

## Layer 2
- vitest: 20/20 passed (Duration 15.32s).
- tsc --noEmit clean: root and daemon both.
- Live smoke (Win11, no daemon running): `PROBE: {"kind":"absent","socketPath":"\\.\pipe\ccsm-control-c0ea1bfb","reason":"ENOENT"}` → `PROCEED: true`, single stderr `[daemon-guard]` line. Pipe name (8-hex sha256 of `username@hostname`) matches upstream byte-for-byte.
- Wiring (`scripts/dev-watch.ts`): pre-spawn `shouldSpawnDaemon().then(proceed => proceed ? startNodemon() : process.exit(0), err => warn + startNodemon())` — correct entry point, fail-open on guard error (right call: dev workflow should not be blocked by a buggy probe).
- `.then()` over top-level await documented (tsx CJS limit) — fine.
- 195+300 LOC: proportional to the work (probe state machine, ProbeOutcome union, 4-platform path matrix, connector test seam). Not heavy.
- Single Responsibility: guard module is a pure decider (env+platform+probe → ProbeOutcome); dev-watch.ts owns the sink (`process.exit` / `spawn`). Clean.

Recommend merge.

PR #707, opened 2026-05-01
test(e2e): T83 SIGCHLD parallel probe

Jiahui-Gu:
Task #1040 — T83 SIGCHLD parallel waitpid probe (POSIX-only).

## Summary
- Adds `daemon/src/pty/__tests__/sigchld-parallel-waitpid.probe.test.ts` (166 LOC, 1 file).
- Spawns N=8 `cat` children, kills them concurrently with SIGTERM, waits for libuv to observe every exit, then runs ONE `handle.drain()` pass against the T38 reaper with a fake `waitpid` mirroring the T39 native binding contract.
- Asserts the parallel-fan-out invariant from frag-3.5.1 §3.5.1.2 / §3.5.1.6:
  - All 8 children reaped in a single drain pass (no zombies, no per-child serialization).
  - `waitpid(pid)` called exactly once per child during that single pass (callLog length === N witnesses no early-break).
  - Exit payload (exitCode, signal) forwarded verbatim.
  - Second drain is a no-op (no double-reap).
- Auto-skips on win32 with a visible CI log line: `[T83 sigchld-parallel-waitpid.probe] skipped on win32 — JobObject path (T39) covers parity`.

## Depends on (already merged)
- T38 SIGCHLD reaper (#995) — reaper module under test.
- T37 PTY FSM (#994) — referenced in spec context.

## Test plan
- [x] Local `vitest run` on Win 11 25H2 (this worktree): 1 file skipped, 1 test skipped, skip-log emitted to stdout (verified with `--reporter=verbose`).
- [x] `tsc --noEmit -p daemon/tsconfig.json` clean.
- [ ] CI (Linux/macOS) runs the actual N=8 drain pass.

POSIX-only; auto-skips on win32.

### DIFF
--- a/daemon/src/pty/__tests__/sigchld-parallel-waitpid.probe.test.ts
+++ b/daemon/src/pty/__tests__/sigchld-parallel-waitpid.probe.test.ts
@@ -1,0 +1,166 @@
+// T83 — SIGCHLD parallel waitpid probe (POSIX only).
+//
+// Spec: frag-3.5.1 §3.5.1.2 (daemon-owned SIGCHLD) + frag-6-7 reliability.
+// Depends on T38 reaper (PR #995) and T37 lifecycle FSM (PR #994).
+//
+// Goal: prove the T38 reaper drains N=8 concurrently-killed PTY-style
+// children in ONE batched SIGCHLD pass, with per-PID waitpid called
+// exactly once per child and no per-child serialization (no zombies
+// left behind, no second drain required).
+//
+// What "parallel" means in this layer:
+//   The kernel coalesces N rapid SIGCHLDs into one delivery (frag-3.5.1
+//   §3.5.1.2). The reaper's contract is to iterate the full registered
+//   set on every signal and waitpid(pid, WNOHANG) each one. The probe
+//   asserts that contract holds at N=8 — no off-by-one, no early break,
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE

Light review (diff-only; local node_modules unavailable for vitest; probe is POSIX-only and current host is win32 — would be skipped here regardless).

Strengths:
- POSIX-only probe (describe.skipIf(IS_WIN)) with visible skip log on win32 — documents deliberate bypass rather than silent no-op. Matches frag-3.5.1 'Win parity' rationale (T39 JobObject covers exit observation).
- Uses 'cat' (blocks on stdin) so children stay alive until SIGTERM — avoids race with 'true' that would exit before kill.
- Arms exit promises BEFORE sending kills — no missed-event race.
- Fake waitpid via SigchldReaperDeps injection: returns 'exited' once per pid then 'no-state-change' (correct WIFEXITED → ECHILD kernel sequence emulation).
- Assertions cover the full contract: (a) all N reaped in one drain pass + registered set drained, (b) per-pid waitpid called exactly once = parallel-fan-out witness, (c) exit payload forwarded verbatim, (d) second drain is no-op (no-double-reap from §3.5.1.6).

Self-consistent and tightly scoped. No P0.

PR #708, opened 2026-05-01
feat(harness): T68 harness-agent daemon mode

Jiahui-Gu:
Task #1026.

Adds a `daemon` mode to the harness-agent helper surface so future probes (notably the spec'd `harness-agent.daemonReconnect()` phase per frag-3.7 §3.7.7) can boot a real `daemon/src/index.ts` child, wait for ready, and tear it down gracefully. Unblocks T73/T74/T78 e2e probes.

## Files

- `scripts/probe-helpers/harness-daemon-mode.mjs` (371 LOC) — helper module:
  - `parseHarnessMode(argv, env)` — env `HARNESS_MODE=daemon` or argv `--mode=daemon`; argv wins
  - `bootDaemon(opts)` — spawns `tsx daemon/src/index.ts`, returns `{ child, ready, shutdown, kill }`. Readiness = stdout marker `"daemon shell booted"` (matches the canonical `logger.info({ event: 'daemon.boot' })` line in `daemon/src/index.ts:110`). Optional `socketPath` lights up a `net.connect` smoke check once the control-socket binding lands in index.ts (today the file is not yet bound — graceful no-op)
  - `helloSocketProbe(path, ms)` — connect-only probe (no envelope yet; `daemon.hello` adapter wiring is downstream)
  - `runWithDaemonMode(mode, body, opts)` — sugar that boots + tears down around an async body; passes `daemon=null` when mode is undefined
  - `shutdown()` sends SIGTERM (which `daemon/src/index.ts` funnels through the same `daemonShutdownHandler.handle()` as the wire RPC, per spec §6.6.1); enforces a 6s deadline
- `tests/harness-daemon-mode.test.ts` (293 LOC) — 17 vitest cases covering all four settle paths (marker / exit-before-marker / spawn error / timeout), shutdown/kill teardown, onLog tap, mode parsing, runWithDaemonMode, plus one real-spawn smoke test against a fake-daemon node script

## Hot file avoidance

- Does NOT import T64's wait-daemon helper (PR #1021 parallel) — readiness probe inlined per task brief
- Does NOT touch double-bind-guard.ts (T72 / PR #1025)
- No `package.json` change

## Verification

- `npx vitest run --no-coverage tests/harness-daemon-mode.test.ts` → **17/17 pass** (~5s)
- `npm run typecheck` → clean (root + electron + daemon tsconfigs)
- ESLint → clean (helper + test files are in eslint ignore globs for scripts/tests, so no rule violations to surface)

## Layer 1 note

Spec scope check: frag-3.7 §3.7.7 + §8.8 describe phases the harness drives via "harness-agent" — there is no `harness-agent/` directory; the existing surface lives under `scripts/harness-*.mjs` + `scripts/probe-helpers/`. T68 lands as a new helper in that directory rather than inventing a sibling. Daemon spawn happens directly (not via Electron-main) because the spec'd phase tests the daemon in isolation; downstream T73/T74/T78 wire it into specific harness cases.

### DIFF
--- a/scripts/probe-helpers/harness-daemon-mode.mjs
+++ b/scripts/probe-helpers/harness-daemon-mode.mjs
@@ -1,0 +1,371 @@
+// T68 — harness `daemon` mode helper.
+//
+// Purpose
+// -------
+// Probes that exercise the v0.3 daemon-split path need to boot a real
+// `daemon/src/index.ts` process, wait for it to be ready, drive RPCs at
+// it (when the control-socket binding lands), and tear it down
+// gracefully on test teardown. This module is the single helper every
+// such probe imports — keeping the spawn/wait/teardown shape consistent
+// across harnesses (real-cli, ui, dnd, future daemon-reconnect).
+//
+// Spec citations
+// --------------
+// - frag-3.7-dev-workflow.md §3.7.7 — `harness-agent.daemonReconnect()`
+//   phase: spawn daemon, wait for ready, kill platform-aware target,
+//   assert reconnect-toast clears within 1 s.
+// - frag-3.4.1-envelope-hardening.md §3.4.1.h — supervisor RPCs:
+//   `/healthz`, `/stats`, `daemon.hello`, `daemon.shutdown`,
+//   `daemon.shutdownForUpgrade` (only `daemon.hello` is used here for
+//   the readiness probe; rich RPC drive is out of scope for T68 — it
+//   lands when `daemon/src/index.ts` actually binds the control socket).
+// - frag-3.4.1 §3.4.1.g — `daemon.hello` 2-frame handshake (we surface
+//   a placeholder `helloProbe` API gated on `socketPath` existing, so
+//   call sites can opt in once the socket binding lands without a
+//   second helper rewrite).
+//
+// Single Responsibility
+// ---------------------
+// PRODUCER: spawns the child process, surfaces a `ready` promise, a
+// `process` handle, and a `shutdown()` action. Owns NO test assertions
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE

Light review (664 LOC > 500 threshold; manager spot-checked).

**Scope**: pure additive — new helper `scripts/probe-helpers/harness-daemon-mode.mjs` (371) + `tests/harness-daemon-mode.test.ts` (293). No production src touched.

**What I checked**
- Helper is single-responsibility PRODUCER as claimed: spawn/ready/shutdown only, no assertions.
- Lifecycle hygiene: `settled` guard prevents double-resolve, `cleanupListeners()` removes all listeners + clears timeout, exit-before-ready surfaces useful error with code+signal.
- Cross-platform: tsx.cmd on win32 + shell:true; SIGTERM with timeout race for graceful, SIGKILL escape hatch for ungraceful.
- Hot-file discipline respected — does NOT import T64 wait-daemon helper or touch T72 double-bind-guard, per task brief.
- Mode parser argv-wins-over-env is right; `--mode=inline` lets you override stale env.
- Socket probe gated on `existsSync` so skips silently until control-socket binding lands — graceful forward-compat.
- onLog tap wraps in try/catch so a bad listener can't poison readiness.

**Nits (non-blocking)**
- `outBuf`/`errBuf` only updated when `onLog` is set, so chunked marker matching across boundaries silently regresses without a tap. Probably fine since marker is short and pino flushes whole lines, but worth a code comment.

Not running vitest locally (pool-2 node_modules incomplete — vitest dist-only, no .bin). Trusting CI.

LGTM.

PR #709, opened 2026-05-01
feat(installer): T55 macOS codesign loop (env-driven, placeholder-safe)

Jiahui-Gu:
Task #1010. Implements §11.3.2 codesign loop on the electron-builder side: an `afterSign` hook that walks the packaged `.app` bundle depth-first and codesigns every nested Mach-O / `.node` / `.dylib` / framework, then re-seals the outer `.app`. electron-builder's own signing does not propagate inward (frag-11 §11.3 line 297), so unsigned nested binaries would otherwise fail `codesign --verify --deep` and Gatekeeper would kill the app.

## Files (4, +399 LOC)
- `scripts/sign-macos.cjs` (+158) — afterSign hook, depth-first traversal, env-driven, placeholder-safe.
- `build/entitlements.mac.plist` (+16) — stub entitlements (allow-jit + library-validation off, matches Electron + node-pty + better-sqlite3 + Claude SDK shellouts).
- `tests/sign-macos.test.ts` (+208) — 7 vitest cases.
- `package.json` (+1) — wire `build.afterSign: scripts/sign-macos.cjs`.

## Env contract
- `CCSM_MAC_IDENTITY` — codesign identity (e.g. `Developer ID Application: Acme (TEAMID)`). Unset → log `[skip] codesign: CCSM_MAC_IDENTITY not set` and continue. **Build never fails on unset env.**
- `CCSM_MAC_ENTITLEMENTS` — path to `.plist`; default `build/entitlements.mac.plist`. Missing AND identity set → warn + skip. Missing AND identity unset → silent skip.
- `CCSM_MAC_HARDENED_RUNTIME` — `'1'` (default) passes `--options runtime`; `'0'` omits it.

## codesign argv per target
`codesign --force --sign $identity --timestamp [--options runtime] [--entitlements $plist] $target`

Invoked via `spawnSync` with array args (no shell). Depth-first ordering: `.node` files inside `app.asar.unpacked/` first, then `Frameworks/*.framework`, then outer `.app` last (nested signatures must be sealed before their containers).

## Tests (7 pass, 3.2s)
1. env unset → skip, runner not called, log line matches `[skip] codesign: CCSM_MAC_IDENTITY not set`.
2. non-darwin platform → skip even with env set.
3. identity set + entitlements missing → warn + skip.
4. real fake `.app` on disk → correct argv per target, depth-first order verified, outer `.app` last, `_CodeSignature/` excluded.
5. `CCSM_MAC_HARDENED_RUNTIME=0` → no `runtime` arg.
6. codesign exit nonzero → throws `codesign failed (exit 1)` with target + argv.
7. `collectSignTargets` on nonexistent path → returns `[bundlePath]` (outer always pushed).

`node --check scripts/sign-macos.cjs` ✓.

## T54 (#1011) coexistence plan
T54 (`scripts/sign-windows.cjs`) was not yet pushed at PR-create time. Both hooks platform-gate inside (`darwin` here, `win32` there) so they're functionally exclusive. Whichever lands second should:
1. Create a tiny dispatcher `scripts/after-sign.cjs` that requires both modules and awaits each (each is a no-op on the wrong platform).
2. Update `package.json` `build.afterSign` to point at the dispatcher.

Either order works; I left `afterSign: scripts/sign-macos.cjs` as the single entry — T54 reviewer can flip the wiring trivially.

## Out of scope (per Layer 1 check)
- Notarization (notarytool / `xcrun stapler staple`) → T56 (#1013) per spec §402.
- Daemon-side codesign (CI workflow `.github/workflows/release.yml` step) → already specified in §11.3.2 as a separate CI step, not an electron-builder hook.

## Test plan
- [x] `vitest run tests/sign-macos.test.ts` — 7 pass
- [x] `node --check scripts/sign-macos.cjs`
- [ ] Real macOS `npm run make:mac` with `CCSM_MAC_IDENTITY` set (gated behind real cert; covered by T56 release dry-run)

### DIFF
--- a/build/entitlements.mac.plist
+++ b/build/entitlements.mac.plist
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <!-- Electron + node-pty + better-sqlite3 + Claude SDK shellouts. -->
+  <key>com.apple.security.cs.allow-jit</key>
+  <true/>
+  <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+  <true/>
+  <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+  <true/>
+  <key>com.apple.security.cs.disable-library-validation</key>
+  <true/>
+  <key>com.apple.security.inherit</key>
+  <true/>
+</dict>
+</plist>
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     ],
     "beforePack": "scripts/before-pack.cjs",
     "afterPack": "scripts/after-pack.cjs",
-    "afterSign": "scripts/sign-windows.cjs",
+    "afterSign": "scripts/after-sign.cjs",
     "extraResources": [
       { "from": "daemon/native-staged", "to": "daemon/native" },
       { "from": "daemon/sdk-staged", "to": "daemon/node_modules/@anthropic-ai/claude-agent-sdk" },
--- a/scripts/after-sign.cjs
+++ b/scripts/after-sign.cjs
@@ -1,0 +1,16 @@
+'use strict';
+// Dispatcher: routes electron-builder afterSign to platform-specific signing scripts.
+// Each delegate platform-gates internally and is no-op on the wrong platform.
+const path = require('path');
+module.exports = async function afterSign(context) {
+  const platform = context.electronPlatformName;
+  if (platform === 'darwin' || platform === 'mas') {
+    const signMac = require(path.join(__dirname, 'sign-macos.cjs'));
+    await signMac(context);
+  }
+  if (platform === 'win32') {
+    const signWin = require(path.join(__dirname, 'sign-windows.cjs'));
+    await signWin(context);
+  }
+  // Linux: no signing, no-op
+};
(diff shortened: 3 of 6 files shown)


Jiahui-Gu (commented):
APPROVE

Light review (399 LOC additive — entitlements plist + electron-builder afterSign hook + tests).

**What I checked**
- Hook is platform-gated on darwin, no impact on win32/linux builds. T54 coexistence comment is correct.
- Env contract is genuinely placeholder-safe: missing CCSM_MAC_IDENTITY → silent skip with reason; missing entitlements + identity set → warn + skip (does NOT fail the build). This matches the "build still works without signing creds" requirement.
- Depth-first traversal is correct (children pushed in walk recursion before outer .app appended last). _CodeSignature dirs are skipped. Symlinks skipped. Tests directly assert deepest-node-before-outer-app order.
- codesignOne argv shape: --force --sign IDENT --timestamp [--options runtime] [--entitlements PATH] TARGET. Matches Apple's hardened-runtime + entitlements convention.
- Runner/fs are injected via opts → tests are hermetic, no real codesign or temp pollution beyond os.tmpdir.
- Entitlements plist has the expected JIT / unsigned-exec-mem / dyld-env / disable-library-validation / inherit keys for Electron + node-pty + better-sqlite3 + Claude SDK shellouts. Reasonable.

**Nits (non-blocking)**
- `isMachOByName` only covers .node/.dylib/.so — bare-name binaries are caught only by the `inMacOSDir` heuristic. So a bare Mach-O in extraResources (e.g. `daemon/native/ccsm-daemon` if shipped without ext) would be missed. Fine for current scope; flag for T54-equivalent or follow-up.
- afterSign hook order: electron-builder runs afterSign AFTER its own signing of the .app shell. This re-signs everything, which is intentional per the comment, but it does mean we redo work electron-builder already did. Acceptable for correctness.

Not running vitest locally (pool-2 vitest install incomplete). Trusting CI.

LGTM.

PR #710, opened 2026-05-01
test(e2e): T80 snapshot interleave probe

Jiahui-Gu:
Task #1037. N concurrent snapshots → semaphore serializes; live streams unblocked; byte-coherent.

## Scope
Single probe file at `daemon/src/pty/__tests__/snapshot-interleave.probe.test.ts` (264 LOC). Three vitest cases against the real T40 (`createSnapshotSemaphore`) + T41 (`createFanoutRegistry`) primitives:

1. **Serializes** — N=16 callers vs capacity-4 semaphore; high-water of `stats(key).active` never exceeds 4; over-capacity tail (≥12 callers) reports `lease.waitMs > 0`.
2. **Byte-coherent** — each holder emits 6 frames of 24 bytes (4-byte holder tag + 4-byte frame index + 16-byte deterministic pattern), yielding between frames via `setImmediate` / microtask flush. Per-holder tracker must contain only that holder's id, frames in order 0..5, decoded payload bytes match the expected pattern (no torn frames).
3. **Live streams not blocked** — concurrent fan-out broadcast pump (40 frames at ~1 ms cadence) runs alongside 12 snapshot callers; subscriber must receive every frame in order, no drops/dupes; pinned `delivered.length > capacity` so the test cannot trivially pass before snapshots start.

## Why module-level (not RPC)
T40/T41 ship as pure primitives; no `getBufferSnapshot` daemon RPC handler is wired in `daemon/src/index.ts` yet. Per the task discipline note ("Stop if T40 lacks observability seam"), T40 *does* expose `stats(key)` and `lease.waitMs`, which is enough to assert serialization without an RPC layer. When the handler lands, these invariants are already pinned.

## Local run
```
npx vitest run daemon/src/pty/__tests__/snapshot-interleave.probe.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
   Duration  7.92s
```

## Hot files
Read-only on `daemon/src/pty/snapshot-semaphore.ts`, `daemon/src/pty/fanout-registry.ts`. New file only.

### DIFF
--- a/daemon/src/pty/__tests__/snapshot-interleave.probe.test.ts
+++ b/daemon/src/pty/__tests__/snapshot-interleave.probe.test.ts
@@ -1,0 +1,264 @@
+// T80 — snapshot interleave probe (frag-3.5.1 §3.5.1.5 / frag-6-7).
+//
+// Asserts the daemon's snapshot path satisfies three invariants under N
+// concurrent callers:
+//
+//   1. Semaphore serializes admission — at any instant, the number of
+//      in-flight snapshot bodies is ≤ capacity (no over-admission, even
+//      under interleaving promise scheduling).
+//   2. Snapshot bodies are atomic — once a holder begins emitting frames,
+//      no other holder's frames interleave into the same byte stream
+//      (per-caller payload is contiguous).
+//   3. Live data fan-out is NOT blocked — `broadcast` to subscribers
+//      continues to deliver while snapshot work is in flight (snapshot
+//      doesn't starve subscribers; see frag-3.5.1 §3.5.1.5 res-SHOULD-2
+//      and the "no two snapshot bodies interleave" claim in frag-6-7).
+//
+// Why a module-level probe (not an end-to-end RPC probe):
+//   - T40 (#997) ships the semaphore and T41 (#998) the fan-out registry
+//     as PURE primitives. The `getBufferSnapshot` daemon RPC handler that
+//     wires them together has not landed (no wiring in `daemon/src/index.ts`
+//     references `createSnapshotSemaphore`). The task discipline note
+//     says "Stop if T40 lacks observability seam" — T40 DOES expose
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE

Light review (264 LOC, single new probe file).

**What I checked**
- Pure additive: one test file under `daemon/src/pty/__tests__/`. No production src touched.
- Imports real T40 `createSnapshotSemaphore` + T41 `createFanoutRegistry` modules — exercises actual primitives, not mocks. Will catch regressions in either even before the daemon RPC handler wires them.
- Three invariants are well-chosen and complementary:
  1. high-water active never exceeds capacity (uses sem's own `stats(key).active` as ground truth — the right observability seam).
  2. per-holder body byte-coherence (deterministic pattern + readUInt32LE check catches torn frames).
  3. live fan-out not starved by snapshot work (delivered === broadcast, in-order, no drops; `delivered.length > capacity` rules out the trivial-pass case where snapshots finish before live pump starts).
- No fake timers — uses real setImmediate / queueMicrotask / setTimeout(1ms), as the file comment claims and as T79 conventions require.
- 30s acquire timeout is generous for the 16-caller × 8-frame body workload.
- Final state assertion `{active: 0, queued: 0}` after all releases is a nice belt-and-suspenders for leak detection.

**Nits (non-blocking)**
- Live pump runs ~40ms total while snapshot bodies churn through ~16 callers × 8 frames × yield. On a slow CI box the snapshot churn could finish before live pump completes — but the `delivered.length > capacity` floor (4) is so low it would still pass trivially. Could tighten to `delivered.length >= LIVE_FRAMES / 2` later if flakiness shows up, but not blocking.
- `emitSnapshotBody` writes `tracker.push` synchronously between yields — if the semaphore had a race admitting an extra holder, it would surface in trackers. Good.

Not running vitest locally (pool-2 install incomplete). Trusting CI.

LGTM.

PR #713, opened 2026-05-01
test(e2e): T75 migration-failure modal probe

Jiahui-Gu:
Task #1032. Forces migration to fail end-to-end (real corrupt SQLite file driving the real T29 runner) and asserts the failure modal pipeline.

## What it asserts
- T33 modal driver lands in sticky `failed` ModalState carrying the daemon-classified reason + raw errorMessage.
- State pushed exactly twice over IPC channel `migration:modalState` (`in_progress` then `failed`).
- `failed` is sticky (advanceTimersByTime 60s, stray completed event ignored); only IPC dismiss / Quit clears it.
- T31 (en) + T32 (zh) bundles export `migration.modal.failed.{title,body,actionQuit}` with `{{legacyDb}}` / `{{dataRoot}}` interpolation tokens preserved.
- en + zh expose identical key sets under `migration.modal.failed`.
- en title is sentence-case (no SCREAMING strings).
- Reverse-verify: a stray failed event from `hidden` is a no-op (driver does not surface modal).

## Layer 1 check
Frag-8 §8.6 defines the failure modal as priority-85, single fatal-error class, action "Quit ccsm" only. Spec also defines an in-progress modal (§6.8 priority-100, indeterminate spinner) — that is T76 #1033's territory, not this probe. This probe only drives the failure path.

## Run
No Electron e2e harness in this repo yet (vitest is the integration boundary). Documented inline:

```
npx vitest run electron/migration/__tests__/migration-failure-probe
```

Local result: 7/7 passing in ~5.8s. A future Playwright/Spectron harness (frag-11 packaging) should lift this into a cross-process scenario asserting the rendered modal DOM.

## Files / LOC
- `electron/migration/__tests__/migration-failure-probe.test.ts` (new, 258 LOC)

## i18n keys verified
- `migration.modal.failed.title` (en + zh)
- `migration.modal.failed.body` (en + zh, with `{{legacyDb}}` + `{{dataRoot}}` tokens)
- `migration.modal.failed.actionQuit` (en + zh)

### DIFF
--- a/electron/migration/__tests__/migration-failure-probe.test.ts
+++ b/electron/migration/__tests__/migration-failure-probe.test.ts
@@ -1,0 +1,258 @@
+// T75 — L8 migration-failure modal probe.
+//
+// Spec: docs/superpowers/specs/v0.3-fragments/frag-8-sqlite-migration.md §8.6.
+// Forces the v0.2 → v0.3 migration to fail end-to-end (real corrupt SQLite
+// file driving the real T29 runner) and asserts:
+//   1. T33 modal driver lands in sticky `failed` ModalState carrying the
+//      daemon-classified reason + raw errorMessage.
+//   2. State pushed over IPC channel `migration:modalState` (in_progress
+//      then failed).
+//   3. `failed` is sticky — only IPC dismiss / Quit clears it.
+//   4. T31 (en) + T32 (zh) bundles export `migration.modal.failed.{title,
+//      body,actionQuit}` with `{{legacyDb}}` / `{{dataRoot}}` tokens.
+//   5. Reverse-verify: a stray failed event from `hidden` is a no-op.
+//
+// No Electron e2e harness exists in this repo yet (vitest is the boundary).
+// Run: `npx vitest run electron/migration/__tests__/migration-failure-probe`.
+// A future Playwright/Spectron harness (frag-11) should lift this into a
+// cross-process scenario asserting rendered modal DOM.
+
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE

Light review (diff-only; pool-3 node_modules incomplete, CI on migration-window skip).

Pure test addition (258 LOC, single new file). Probe wires the real T29 migrateV02ToV03 against a real corrupt non-SQLite blob, drives the resulting failure through real createModalDriver via real EventEmitter ipcMain, and asserts:
1. Failed ModalState carries reason + errorMessage + version + traceId
2. Two IPC pushes (in_progress, failed) on migration:modalState
3. Sticky failed — late stray completed event ignored, only IPC dismiss clears
4. en + zh i18n bundles export migration.modal.failed.{title,body,actionQuit} with {{legacyDb}}/{{dataRoot}} tokens, identical key sets
5. Reverse-verify: stray failed without started is no-op
6. Sentence-case enforcement on en title (per no-SCREAMING-UI rule)

Strengths:
- Real T29 runner instead of mocked failure — high integration value
- Reason classifier mirrors §8.6 mapping table; assertion accepts any canonical reason rather than pinning one (robust to better-sqlite3 message variants)
- vi.useFakeTimers + advanceTimersByTime(60s) proves stickiness without a real wait

Nits: classifyAsFailedEvent is duplicated probe-side helper (acknowledged in comment as boot-glue mirror). Consider exporting from production once the boot wiring lands.

LGTM.

PR #720, opened 2026-05-01
feat(electron): T62 upgrade-shutdown call site

Jiahui-Gu:
Closes part of T62 (re-scoped per spec §11.6.5). Calls daemon.shutdownForUpgrade with 5s ack timeout before applying in-place upgrade.

### DIFF
--- a/electron/__tests__/updater.test.ts
+++ b/electron/__tests__/updater.test.ts
@@ -355,7 +355,9 @@ describe('updater: T62 upgrade-shutdown RPC', () => {
       await vi.advanceTimersByTimeAsync(mod.UPGRADE_SHUTDOWN_ACK_TIMEOUT_MS);
       const res = await promise;
       expect(res).toEqual({ ok: true });
-      // Drain the scheduled setImmediate.
+      // Drain the scheduled setImmediate while still on fake timers — switching
+      // to real timers first would drop the queued callback.
+      await vi.runAllTimersAsync();
       vi.useRealTimers();
       await new Promise((r) => setImmediate(r));
       expect(quitAndInstallCalls).toEqual([{ isSilent: false, isForceRunAfter: true }]);
--- a/electron/__tests__/updater.test.ts
+++ b/electron/__tests__/updater.test.ts
@@ -207,18 +207,18 @@ describe('updater: IPC wiring', () => {
     // handler can proceed.
     autoUpdaterEmitter.emit('update-downloaded', { version: '0.1.3' });
     const handler = ipcHandlers.get('updates:install')!;
-    const res = handler({});
+    const res = await handler({});
     expect(res).toEqual({ ok: true });
     // quitAndInstall is scheduled via setImmediate — flush it.
     await new Promise((r) => setImmediate(r));
     expect(quitAndInstallCalls).toEqual([{ isSilent: false, isForceRunAfter: true }]);
   });
 
-  it('updates:install refuses when no download has completed', () => {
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
REQUEST-CHANGES — one new test fails deterministically locally; otherwise design and call-site placement are sound. (Posted as comment because GH blocks self-review request-changes; manager: please treat as blocking until the failing test is fixed.)

Layer 1 (judgment): PASS.
- Call site is correct: in the `updates:install` IPC handler, after the `lastStatus === 'downloaded'` defense-in-depth gate, before the `setImmediate(quitAndInstall)`. That is the real upgrade trigger, not an arbitrary spot.
- Marker semantics are correct by separation: marker write lives on the daemon side (T82 `daemon-shutdown-for-upgrade.ts`), electron only sends the RPC and proceeds. `UpgradeShutdownAck` shape matches `ShutdownForUpgradeAck` (`accepted: true`, `reason: 'upgrade'`) and the structural-typing-no-cross-boundary-import note is correct.
- Ack/timeout/error all proceed with `quitAndInstall` — matches spec §11.6.5 race fallback (force-kill + proceed is owned by the wiring layer). Default `null` rpc → `noop` outcome preserves dev / non-daemon flow. Good.
- Deferring `setUpgradeShutdownRpc` wiring in `main.ts` to a follow-up is acceptable for a T62 call-site-only PR; the doc comment is explicit.

Layer 2 (implementation): one failing test (3/3 reruns).
- `updater: T62 upgrade-shutdown RPC > updates:install proceeds with quitAndInstall after RPC timeout` fails: `expected [] to deeply equal [ { isSilent: false, isForceRunAfter: true } ]`.
- Root cause: handler schedules `setImmediate(() => autoUpdater.quitAndInstall(...))` while `vi.useFakeTimers()` is active, which queues it on the fake-timer queue. The test then calls `vi.useRealTimers()` and awaits one real `setImmediate` — the queued fake immediate is dropped and `quitAndInstall` is never invoked. So the test's final assertion necessarily fails.
- Fix options (pick one): (a) `await vi.runAllTimersAsync()` before `vi.useRealTimers()` to flush the scheduled immediate; (b) advance fake timers by the 5 s ack timeout, then let the handler's setImmediate fire under the fake clock and assert there; (c) inject a `setImmediate` seam (overkill for this).
- Other 6 T62 tests + the 3 modified pre-existing IPC tests pass (21/22 in the file).

Reverse-verify: not recorded in PR body. Less critical for a feature, but worth noting because the failing test means we don't actually have green coverage of the "RPC timeout → still calls quitAndInstall" path right now.

Action: fix the failing timeout test and re-run `npx vitest run electron/__tests__/updater.test.ts` to green before merge. No production code change needed.

Jiahui-Gu (commented):
Fixed the failing `updates:install proceeds with quitAndInstall after RPC timeout` test in f03e7b5.

**Root cause:** the test switched `vi.useRealTimers()` BEFORE flushing the fake-timer queue, so the production code's `setImmediate(quitAndInstall)` (scheduled while fake timers were active) was dropped on the floor.

**Fix:** added `await vi.runAllTimersAsync()` to drain the queue while still on fake timers, then switch to real timers and `setImmediate` flush as before. Production code untouched.

**Verification:**
- `npx vitest run electron/__tests__/updater.test.ts -t "proceeds with quitAndInstall after RPC timeout"` → 1/1 pass
- Full file: `npx vitest run electron/__tests__/updater.test.ts` → 22/22 pass

PR #721, opened 2026-05-01
test(daemon): T81 cross-user ACL probe

Jiahui-Gu:
Closes T81. Probe asserts cross-user pipe connections are rejected and same-user accepted.

## Summary
- New seam-level probe at `daemon/src/sockets/__tests__/cross-user-acl-probe.test.ts`.
- Exercises `verifyPeerCred` on win32 / linux / darwin via the `NativePeerCredDeps` injection seam (per JSDoc on `peer-cred-verify.ts`).
- Each platform: one same-user accept (`same:true`), one cross-user reject (`same:false`), one reverse-verify (flip expected → flip outcome), and one no-throw assertion confirming the verifier hands the destroy decision to the caller per the JSDoc reject pattern.

## Spec anchors
- `frag-6-7-reliability-security.md` Task 26: cross-user access denial repro (manual + scripted).
- `v0.3-design.md` §3.1.1 transport hardening: same-user check on each accepted connection.
- `daemon/src/pty/pipe-acl.ts` (T45) sets the in-kernel pipe DACL; the in-kernel rejection never surfaces a JS `Socket` accept event for the rejected peer, so the L8 probe asserts the JS-visible decider seam (`verifyPeerCred`).

## Test plan
- [x] `npx vitest run daemon/src/sockets/__tests__/cross-user-acl-probe.test.ts --no-coverage` → 12 passed.
- [x] No npm install (borrowed sibling pool's node_modules junction for the local run; junction removed before commit so no artefact in the diff).
- [x] OS peer identity fully mocked; no second user account spawned.

### DIFF
--- a/daemon/src/sockets/__tests__/cross-user-acl-probe.test.ts
+++ b/daemon/src/sockets/__tests__/cross-user-acl-probe.test.ts
@@ -1,0 +1,278 @@
+// T81 — L8 cross-user ACL probe.
+//
+// Asserts the daemon's per-platform "same OS user" check at the
+// transport boundary (pipeAcl on Win, peer-cred on POSIX) refuses a
+// connection from a different OS user and accepts a connection from
+// the same OS user. The OS-level peer-identity check is mocked via
+// the `NativePeerCredDeps` injection seam (per
+// daemon/src/sockets/peer-cred-verify.ts JSDoc); we do NOT actually
+// spawn a second user account.
+//
+// Spec anchors:
+//   - frag-6-7-reliability-security.md "Task 26": cross-user access
+//     denial repro (manual + scripted).
+//   - v0.3-design §3.1.1 transport hardening: "on each accepted
+//     connection, daemon validates peer is same user. Win:
+//     GetNamedPipeClientProcessId → OpenProcessToken → match user
+//     SID. Unix: SO_PEERCRED (Linux) / getpeereid (Mac)."
+//   - daemon/src/pty/pipe-acl.ts (T45): Win pipe DACL = current SID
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE — peer-cred decider seam is correctly exercised; reverse-verify confirmed.

Layer 1 (judgment):
- Probe scope is appropriate. Per JSDoc on peer-cred-verify.ts and the PR body, the in-kernel pipe DACL (T45) rejects cross-user peers in kernel-space and never surfaces a JS Socket accept event, so the only JS-visible cross-user contract is the verifyPeerCred decider seam. Asserting that seam (DI'd native deps) on all three platforms is the right L8 shape; spawning a second OS user would be an L9 manual repro, not an L8 probe.
- Covers all three production branches (win32 SID, linux SO_PEERCRED uid, darwin getpeereid uid). Per-platform: same-user accept, cross-user reject, reverse-verify (flip expected → flip same), no-throw assertion (caller owns destroy() per JSDoc reject pattern). 12 cases / 4 per platform — matches task ask.
- vi.fn assertions on getNamedPipeClientProcessId/getsockoptPeerCred/getpeereid prove the native chain is actually invoked, not short-circuited.

Layer 2 (implementation):
- Ran locally (pool-19, junctioned node_modules from sibling pool per PR body): 12/12 passed in 2.06s, no flakes.
- Reverse-verify: mutated all three production return sites to `same: true` (permissive). Re-ran: 6/12 FAIL (the 3 reject + 3 reverse-verify cases across platforms), 6 PASS (same-user accepts + no-throw cases). Confirms assertions are load-bearing on the comparison, not artefacts.
- Diff is test-only (+278/-0), single file, no production touch. Clean.

Merge.

PR #722, opened 2026-05-01
docs(observability): crash observability design spec (#1069)

Jiahui-Gu:
Design spec for end-to-end crash observability across frontend (electron-main + renderer) and backend (ccsm-daemon), with native Crashpad and structured-log capture.

Framed around v0.3 dual-process topology. Phase 1 alone closes the 16:18 'nothing on disk' gap — every crash on either side leaves a recoverable incident dir, even with SENTRY_DSN empty.

No code. After user sign-off the 5 phases split into v0.3.x rollout PRs.

Closes #1069.

### DIFF
--- a/docs/superpowers/specs/2026-05-01-crash-observability-design.md
+++ b/docs/superpowers/specs/2026-05-01-crash-observability-design.md
@@ -320,6 +320,14 @@ One rolling logfile **per process**, both bundled into "Send last crash":
 
 ## §6 DSN + release + symbol pipeline
 
+**DSN topology — DECIDED 2026-05-01**
+
+> One Sentry project (`ccsm`) covers electron-main + renderer + daemon.
+> Events are dimensioned by `tags.surface` (`main` | `renderer` | `daemon`)
+> set at `Sentry.init({ initialScope: { tags: { surface } } })`. Single DSN
+> secret, single issue stream; cross-process correlation by `incidentId` /
+> `traceId` already lives in the event payload (§5.3, §10).
+
 **Build-time injection vs runtime**
 
 - Today: `process.env.SENTRY_DSN` consulted at runtime
@@ -375,7 +383,12 @@ One rolling logfile **per process**, both bundled into "Send last crash":
 - Renderer breadcrumbs strip session names from URLs / breadcrumb messages
   before forward; only the sid (ulid) survives.
 
-**Consent**
+**Consent — DECIDED 2026-05-01**
+
+> Default = network upload ON (opt-out). No first-run banner in phase 1.
+> The Settings toggle is reachable from the Help menu on first crash
+> (phase 4 modal links to it). A first-run banner is deferred to phase 4
+> so launch-day friction stays at zero.
 
 - Capture-to-disk is **always on**. It is local, scrubbed, and the user
   controls deletion via the directory.
@@ -384,8 +397,7 @@ One rolling logfile **per process**, both bundled into "Send last crash":
   Default = upload ON; toggle in Settings makes it OFF; takes effect within
   one event via the existing `stateSavedBus` invalidation
   (`electron/prefs/crashReporting.ts:48-52`).
-- First-run banner: deferred to phase 4 — not ship-blocking and adds friction
-  on first launch where the user just wants to see the app work. Settings
+- First-run banner: deferred to phase 4 per the decision above. Settings
   toggle is discoverable via Help menu on first crash.
 
 ## §8 "Send last crash" UX
@@ -439,6 +451,13 @@ This is the section that would have saved the 16:18 incident.
 >    daemon's `recordAndDie` (these are *additive* — present means the daemon
 >    saw the crash itself; absent means it died too fast or natively).
 > 7. Wall-clock elapsed since the daemon's last `/healthz` poll.
+> 8. **Daemon-boot-crash edge case** — if the rings are empty AND no marker
+>    file exists (daemon died so fast that even `recordAndDie` never ran:
+>    e.g. ELF/PE load failure, missing native dep, throw in `import` chain),
+>    the supervisor still writes `meta.json` with `surface:
+>    'daemon-boot-crash'`, `exitCode`, `signal`, `lastHealthzAgoMs: null`,
+>    and empty `stderr-tail.txt` / `stdout-tail.txt`. An empty bundle is
+>    still a bundle; absence of evidence is itself evidence.
 
 **IPC / file shape**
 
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
# Reviewer pass — Task #1069 (PR #722)

Verdict: **APPROVE-WITH-FIXUPS** — spec is implementable as-is for phase 1; fixups are §10 ID collision, §12 Q1/Q3 closure, and one missing boot-pre-supervisor edge in §9.

## Layer 1 — direction

### 1. Frontend/backend framing — applied throughout, not window-dressing
- §3 split into two tables (Frontend vs Backend) keyed by process column.
- §4 ASCII diagram has explicit two-box "frontend (one Electron app)" vs "backend (separate process)" with the cross-process arrow being the daemon stderr ring.
- §5 reorganized: §5.1 frontend, §5.2 backend, §5.3 cross-process correlation as its own section.
- §6 / §7 / §10 are cross-cutting (DSN, PII, dir layout) — appropriately not split.
The second-message reframe landed.

### 2. Phase 1 actually solves "nothing on disk after a crash" — YES, with one gap
Phase 1 explicitly excludes Sentry/DSN dependency:
- electron-main `crash/collector.ts` writes incident dirs unconditionally (§5.1, §10).
- Daemon `crash/handlers.ts` writes `<runtimeRoot>/crash/<bootNonce>.json` marker before `process.exit(70)` (§5.2).
- Supervisor stderr ring + `child.on('exit')` → `recordIncident({surface:'daemon-exit'})` (§5.3 step 4).
- §11 phase 1 callout is unambiguous: "even with `SENTRY_DSN` empty".
The 16:18 user crash would have left, at minimum: `meta.json` + `stderr-tail.txt` + `daemon-marker.json` (if daemon's own handler ran) OR just supervisor-side `meta.json` + tails (if it didn't). Litmus passes.

### 3. Cross-process correlation — concrete enough
§5.3 specifies the IPC envelope, the `DaemonChild` shape with rings + `lastTraceId` + `bootNonce`, and the marker handoff via filesystem (`<runtimeRoot>/crash/<bootNonce>.json` → `daemon-marker.json` move). Reuses v0.3 primitives correctly: `traceId` from envelope (frag-3.4.1), `bootNonce` from `daemon.hello` (`daemon/src/handlers/daemon-hello.ts`), `bootChanged` from `src/lib/daemon-events.ts:27-29`. Wildcard scan for orphaned markers handles the boot-race case.

### 4. Daemon boot-crash visibility — partial gap
§5.3 covers "dies during boot before `daemon.hello`": synthetic `boot-pending-<ulid>` + wildcard marker scan on next electron-main boot. But the case **"daemon dies SO fast that even `recordAndDie` never runs"** (e.g. crashes in `import` chain, ELF/PE load failure, missing native dep) is only implicitly handled. §9 contract item 6 mentions "absent means it died too fast or natively" but there's no explicit spec for what gets written from the **electron-main supervisor side** when zero stdout/stderr was ever produced. Recommend §9 add an explicit clause: "if rings are empty AND no marker file exists, still write `meta.json` with `surface: 'daemon-boot-crash'`, `exitCode`, `signal`, and `lastHealthzAgoMs: null`."

### 5. OSS-fork DSN leak prevention — airtight
§6 chain: GitHub fork PRs cannot read environment secrets → `SENTRY_DSN` env empty → `webpack.DefinePlugin` bakes `''` → `before-pack.cjs` writes `{ sentryDsn: '' }` → `initSentry()` `if (!dsn) return` (already in code at `electron/sentry/init.ts:18-22`). Plus regression test asserted in §6. Fork can override via `CCSM_CRASH_DSN` env var. Clean.

### 6. §12 open questions — mixed
- **Q1 (default-on Sentry vs first-run consent)** — direction-affecting (legal/UX). Keep for user.
- **Q2 (one Sentry project vs two)** — direction-affecting (affects DSN secret count + ops). Keep for user.
- **Q3 (bundle daemon pino into frontend logfile vs separate)** — spec already answers it: "two files because the daemon may crash before any IPC opens". This is a manager-lock, not a user question. Recommend striking.

## Layer 2 — spec hygiene

### Citation accuracy
Spot-checked against `origin/working` HEAD `0f965ff`:
- `electron/main.ts:43-48` ✓ (matches verbatim, TODO comment at :40-42 confirmed)
- `electron/sentry/init.ts:18-22` ✓ (early-return on empty DSN)
- `daemon/src/index.ts:118-119` ✓ (only SIGTERM/SIGINT registered)
- `package.json:82` ✓ (`"ulid": "^2.3.0"`)
- `src/index.tsx:15` ✓ (`sentryInit({})`)
- `src/lib/daemon-events.ts:27-29` ✓ (`bootChanged` definition)
No hand-waving in the "Today" column.

### §10 directory layout — collision risk
`2026-05-01-1618-<incidentId>/` uses minute precision. Two crashes in the same minute get unique dirs because of the ULID suffix — OK. But the `_dmp-staging/` shared scratch dir (§5.1) is **single-tenant**. Concurrent renderer + GPU crash within Crashpad's flush window could race. Recommend §10 explicitly state the staging-dir lifecycle (lock or per-PID subdir) or document that Crashpad serializes writes.

### §10 cleanup
Retention rule "last 20 incidents OR 30 days, whichever is larger" + "pruner runs at electron-main boot" is concrete. Good. One missing detail: what if pruner itself crashes? Recommend adding "best-effort, swallow errors, log to rolling logfile".

### Phase PR counts
- Phase 1: 3 PRs (collector, daemon handlers, supervisor wiring) — realistic for v0.3 codebase, each ~300-500 LOC.
- Phase 2: 2 PRs — realistic.
- Phase 3: 2 PRs — realistic, though `node-segfault-handler` Windows story is still TBD; OK since marked "Optional".
- Phase 4: 1-2 PRs — realistic.
- Phase 5: 1 PR — possibly tight if renderer forwarder isn't already wired (frag-6-7 §6.6.2 referenced but not verified on `working`).

### Internal contradictions
- §5.5 says "Cap at 7 days × 10 MB per file" but §10 says incident dir copies "last 5000 lines of frontend logfile". 5000 lines × ~200 bytes = ~1 MB; consistent if logs are dense JSONL. No conflict but flag it.
- §6 mentions deleting `*.map` from asar (`package.json:151`) — confirms source maps stay out of the user-shipped bundle. Consistent with §5.4 symbol upload pipeline.
- §5.2 `process.exit(70)` (EX_SOFTWARE) vs §11 phase-1 smoke test "kill -9 the daemon child". Kill -9 won't trigger `recordAndDie` — that's intentional (tests the supervisor path, not the daemon-side path). Worth a one-line comment in the smoke-test description so the reader doesn't think it contradicts.

## Summary
- **Direction reframe**: applied throughout, real two-process structure.
- **Phase 1 standalone value**: confirmed; delivers disk artifacts without DSN.
- **Cross-process correlation**: implementable; reuses v0.3 traceId+bootNonce primitives.
- **Boot-crash gap**: §9 needs an explicit "rings empty + no marker" clause.
- **Q3 in §12**: spec already decided this; remove from user-decision list.
- **Citations**: verified, accurate.

Approve once §9 boot-crash clause + §12 Q3 closure are added (both ~5-line spec edits, no rework).

Jiahui-Gu (commented):
Locked §6 (1 project + tags.surface) and §7 (default-on, no first-run banner). Folded r1's 3 fixups into the spec (§9 daemon-boot-crash clause #8, §10 _dmp-staging concurrency note, §11 phase-1 smoke-test SIGKILL clarification). §12 reduced to a closure log — no remaining user questions. Ready to merge. Fixup commit: 123e3f2.

PR #723, opened 2026-05-01
feat(daemon): ptySubscribe Connect streaming RPC handler

Jiahui-Gu:
Adds `daemon.ptySubscribe` streaming over data-socket. Required by L8 drop-slowest probe (#1036). Closes #1057.

## Summary
- New handler `daemon/src/handlers/pty-subscribe.ts` exposing `ccsm.v1/pty.subscribe` on the data-plane dispatcher.
- Validates `ptyId` via injected predicate, subscribes to existing fan-out registry, forwards every broadcast frame to the transport-supplied stream.
- Honors existing drop-slowest 1 MiB watermark + replay-burst exemption (composed at the producer that calls `registry.broadcast()`); handler stays a pure forwarder.
- Idempotent stream end; maps registry `DrainReason` (pty-exit / pty-crashed / daemon-shutdown / session-removed) to `PtySubscribeEndReason`.
- Refuses registration on supervisor-plane dispatcher (data-plane RPC per frag-3.4.1 §3.4.1.h).

## Test plan
- [x] `npx vitest run daemon/src/handlers/__tests__/pty-subscribe.test.ts` — 12 passed (subscribe path / cancel path / 4 drain-reason mappings / 3 validation gates / 2 dispatcher wiring cases)
- [x] Cancel-clears-subscription verified via `registry.getSubscribers()` snapshot before/after.
- [x] Reverse coverage: spy on `registry.subscribe` ensures invalid payload / unknown ptyId never reach the registry.

### DIFF
--- a/daemon/src/handlers/__tests__/pty-subscribe.test.ts
+++ b/daemon/src/handlers/__tests__/pty-subscribe.test.ts
@@ -1,0 +1,240 @@
+// Unit tests for the `ccsm.v1/pty.subscribe` streaming RPC handler.
+//
+// Scope (Task #1057):
+//   - subscribe path: 3 broadcast frames flow through to the stream's `push`.
+//   - cancel path: returned cancel hook unsubscribes + emits `caller-cancel`
+//     end frame exactly once.
+//   - registry drain path: registry `close()` is mapped to the right
+//     end-reason and propagated through `stream.end`.
+//   - validation: bogus payload / unknown ptyId both produce a typed end
+//     reason without subscribing to the registry.
+//   - dispatcher wiring: refuses supervisor-plane registration; registers
+//     under PTY_SUBSCRIBE_METHOD on the data-plane dispatcher.
+
+import { describe, it, expect, vi } from 'vitest';
+
+import {
+  PTY_SUBSCRIBE_METHOD,
+  handlePtySubscribe,
+  registerPtySubscribeHandler,
+  type PtySubscribeContext,
+  type PtySubscribeFrame,
+  type PtySubscribeStream,
+  type PtySubscribeEndReason,
+} from '../pty-subscribe.js';
+import {
+  createFanoutRegistry,
+  type DrainReason,
+  type FanoutRegistry,
+} from '../../pty/fanout-registry.js';
+import {
+  createDataDispatcher,
+  createSupervisorDispatcher,
+} from '../../dispatcher.js';
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+interface CapturedStream extends PtySubscribeStream {
+  pushed: PtySubscribeFrame[];
+  ended: PtySubscribeEndReason[];
+}
+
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (reviewer self-approve blocked by GitHub same-identity; manager will manual-merge).

## Verification (independent re-check of inherited 21ef329)

**Imports/types story matches sources:**
- `Subscriber<TMessage>`, `DrainReason`, `FanoutRegistry<TMessage>` exported with the exact shapes the handler uses (`daemon/src/pty/fanout-registry.ts`).
- `Dispatcher` class exposes `plane` getter + `register(method, Handler)`; `createDataDispatcher`/`createSupervisorDispatcher` exist (`daemon/src/dispatcher.ts:200-231`).
- `DrainReason.kind` enum exhaustively switched in `drainReasonToEndReason` — TS catches drift.
- `subscribe()` returns `() => void` and is internally idempotent (fanout-registry.ts:138-143); handler relies on that correctly.

Not a rubber-stamp; the dead worker's code is consistent with current `working`.

## Layer 1

1. **Backpressure boundary**: handler defers drop-slowest + replay-burst exemption to producer side that calls `registry.broadcast()` (header non-goals + `deliver()` comment). Correct seam per fanout-registry.ts:9-24 (registry is sink-only; T43 owns drop-slowest on producer). No parallel backpressure path. L8 drop-slowest probe #1036 exercises producer->registry path; this handler is transparent to it.
2. **`caller-cancel` reason**: correctly modeled as a `PtySubscribeEndReason` variant separate from `DrainReason` (registry has no caller-cancel because cancel is caller-initiated). Transport maps enum to Connect status codes per docstring. No existing enum would have housed it.

## Layer 2

- Idempotent cancel: `endOnce` + registry's `unsubscribed` guard; covered by "second cancel is a no-op".
- Cross-session isolation: `bySession.get(sessionId)` partitioning; covered.
- Validation order (payload-null -> ptyId-missing -> ptyId-unknown) verified via `subscribeSpy.not.toHaveBeenCalled()` on both gates.
- 340 LOC: ~140 are JSDoc/header citing frag-3.4.1.h, 3.5.1.5, 3.5.1.2. Real code ~120 LOC. Justified for a public RPC contract.
- Mock fidelity: tests use the real `createFanoutRegistry`, not a mock. Drop-slowest intentionally out-of-scope (separate module on producer side).

## Probe (local)

```
npx vitest run daemon/src/handlers/__tests__/pty-subscribe.test.ts --no-coverage
Test Files  1 passed (1)
Tests       12 passed (12)
Duration    3.57s
```

## Suggestions (non-blocking, follow-up OK)

- `makePtySubscribeHandler` returns `{ streaming: true }` sentinel for unary probes (lines 297-305) but no test asserts that path. Worth a follow-up test that `dispatcher.dispatch(PTY_SUBSCRIBE_METHOD, ...)` resolves with `streaming: true`. Low risk — only relevant for envelope-shape probes per docstring.

Safe to merge as-is.

PR #731, opened 2026-05-01
test(daemon): T79 L8 drop-slowest probe

Jiahui-Gu:
Closes T79 / #1036.

L8 probe at `daemon/src/handlers/__tests__/drop-slowest.probe.test.ts` composing the real `createFanoutRegistry` + `createDropSlowest` + `createReplayBurstExemption` + `handlePtySubscribe` (#1057) primitives. Mocks only the in-test PTY producer (T44 wiring is not yet landed; the probe pins the composition shape now so the wiring task can code against a working reference).

Asserts:
1. **1 MiB watermark drops slow consumer** — two consumers attach; producer pumps 64 KiB frames; on the 17th frame (1_114_112 bytes pending) the slow consumer (no flush) is unsubscribed + closed; the fast consumer (flushes after every delivery) keeps receiving subsequent frames.
2. **Replay-burst exemption** — during `enterReplay`, 2 MiB of exempt writes do NOT trip the watermark; after `exitReplay`, steady-state accounting starts fresh and the slow consumer drops on the 17th post-replay frame.
3. **Canonical end-frame reason** — `DrainReason` has no dedicated `drop-slowest` kind, so the convention is `{ kind: 'session-removed', detail: 'drop-slowest' }`, mapped by `drainReasonToEndReason` to a `session-removed` `PtySubscribeEndReason`. Documented in the probe header so a future wiring task cannot silently change the reason.
4. **Constant guard** — `DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES === 1_048_576`, so the trip-arithmetic (16 frames = 1 MiB = not yet overlimit; 17th frame trips) stays sound.

Test results: **4 passed** (`vitest run daemon/src/handlers/__tests__/drop-slowest.probe.test.ts`).

Reverse-verified: mutated `DROP_SLOWEST_DEFAULT_THRESHOLD_BYTES` from `1_048_576` to `1_073_741_824` (1 GiB) → tests 1, 2, 4 all FAIL (slow consumer never dropped, post-replay drop never happens, constant guard catches the mutation); test 3 (dispatcher registration) stays green as expected. Restored constant → all 4 tests pass.

### DIFF
--- a/daemon/src/handlers/__tests__/drop-slowest.probe.test.ts
+++ b/daemon/src/handlers/__tests__/drop-slowest.probe.test.ts
@@ -1,0 +1,437 @@
+// T79 — drop-slowest L8 probe (frag-3.5.1 §3.5.1.5 lines 118-122, 134-135).
+//
+// Asserts the v0.3 PTY data-plane backpressure contract end-to-end across the
+// real primitives the daemon shell will compose at wiring time:
+//
+//   1. **Slow consumer dropped at the 1 MiB per-subscriber watermark**.
+//      Two `pty.subscribe` consumers attach to the real fan-out registry via
+//      the real `handlePtySubscribe` handler (#1057). A producer broadcasts
+//      frames and accounts pending bytes through the real `createDropSlowest`
+//      decider; only the slow consumer (which never `flush`es) accumulates
+//      past 1 MiB and is unsubscribed + close()d. The fast consumer (which
+//      flushes after every delivery) continues to receive every subsequent
+//      frame.
+//
+//   2. **Replay-burst exemption** (spec line 120). During the resubscribe
+//      replay window, the slow consumer is fed >1 MiB through the
+//      `createReplayBurstExemption` recorder (which forwards to drop-slowest
+//      with `exempt: true`). The slow consumer MUST NOT be dropped while in
+//      replay-mode, no matter how large the burst. Once `exitReplay()` runs
+//      and subsequent steady-state writes resume, the watermark accounting
+//      starts fresh and the slow consumer is dropped on the first
+//      post-replay overlimit broadcast.
+//
+//   3. **End-frame reason for dropped consumer**. `DrainReason` does NOT
+//      carry a dedicated `drop-slowest` kind (the registry close vocabulary
+//      is `pty-exit | pty-crashed | daemon-shutdown | session-removed`), so
+//      the canonical drop-slowest signal is `{ kind: 'session-removed',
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
**Verdict: APPROVE** (posted as comment — same-identity GH blocks --approve; manager merges).

L8 drop-slowest probe is well-scoped and honest.

**Layer 1 (judgment)**
- Necessity: probe pins the composition (registry + drop-slowest + replay-burst exemption + pty-subscribe handler) that T44/T47 wiring will reproduce; each piece is unit-tested in isolation, so the composition contract needs its own pin. Justified.
- Honesty: real `createFanoutRegistry` / `handlePtySubscribe` / `createDropSlowest` / `createReplayBurstExemption` / `createDataDispatcher` — no mocks of the system under test. Only the producer loop is in-test, which is correct because the production producer does not yet exist (T44/T47).
- Spec match: line 118 mandates the `subscriber-dropped-slow` *log line* on the producer side, not a new `DrainReason` kind. `DrainReason` vocabulary is closed (`pty-exit | pty-crashed | daemon-shutdown | session-removed`, fanout-registry.ts:57). Using `{ kind: 'session-removed', detail: 'drop-slowest' }` and pinning that string is the right convention — adding a new kind would be unrelated scope creep, and the `detail` slot is exactly what the JSDoc reserves for log-line free-form. Handler maps via `drainReasonToEndReason` (pty-subscribe.ts:166-185), which preserves `detail`. Good.
- Reverse-verify: in addition to the worker's 1 MiB → 1 GiB threshold mutation, I flipped `exempt: true` → `exempt: false` in `replay-burst-exemption.ts:127` and test #2 failed at `expect(dropSlowest.pending(SLOW_SUB_ID)).toBe(0)` with received 2_097_152. Confirms the exemption assertion is real, not vacuous.

**Layer 2 (implementation)**
- 4/4 pass in ~2.2s locally on Win 11 25H2. No fake timers, no sleeps, no snapshots, no order coupling (fresh registry/decider/producer per `it`).
- Assertions are concrete: drop count, end-frame shape, push count, registry membership before/after, tally pending bytes. Symmetric setup/teardown (no shared mutable module state).
- Test #4 pins the 1 MiB constant + the 16/17 frame arithmetic — guards against silent threshold drift.
- One nit (non-blocking): the producer harness is in-test, which means a future T44/T47 wiring divergence (e.g., delivery order) won't be caught by this probe alone. Worth a follow-up note on T44 to delete this in-test producer once production wiring lands. Not a blocker for merge.

PR #736, opened 2026-05-01
feat(crash): phase 1 — recoverable artifacts on every crash (no DSN needed)

Jiahui-Gu:
## Summary

Implements **spec §11 phase 1** — closes the 2026-05-01 16:18 "nothing on disk" gap. After this PR, every crash on either side leaves an incident dir under `%LOCALAPPDATA%\CCSM\crashes\<ts>-<ulid>\` (Windows) / `~/Library/Application Support/CCSM/crashes/...` (macOS) / `~/.local/share/CCSM/crashes/...` (Linux). No DSN, no network — pure local artifacts.

### Phase 1 scope

- electron-main collector + crashReporter staging dir + uncaught/unhandled handlers + `render-process-gone` + `child-process-gone` (spec §5.1)
- daemon `installCrashHandlers` + marker file + `process.exit(70)` (spec §5.2)
- supervisor stderr/stdout ring buffers + `child.on('exit')` handler + marker adoption + `ccsm:daemon-crash` IPC (spec §5.3, §9)
- e2e probe exercising main-side crashes (spec §11 phase 1)

Out of scope (deferred to phase 2+): Sentry DSN init, build-time DSN injection, symbol upload, "Send last crash" UX, log rolling.

### Files created

- `electron/crash/collector.ts` — `startCrashCollector`, `recordIncident`, `flush`, dmp-staging adoption, retention pruner
- `electron/crash/incident-dir.ts` — incident dir layout, `meta.json` writer, `README.txt` writer
- `electron/crash/ring-buffer.ts` — bounded `RingBuffer<T>`
- `electron/crash/scrub.ts` — `scrubHomePath` + env allowlist
- `electron/main-crash-wiring.ts` — `wireCrashHandlers({collector, processRef})`
- `electron/daemon/supervisor.ts` — `attachCrashCapture(handle, collector)` (new file; v0.3 supervisor PR can extend)
- `daemon/src/crash/handlers.ts` — `installCrashHandlers({logger, bootNonce, runtimeRoot, getLastTraceId})`
- `tests/electron/crash/{scrub,ring-buffer,incident-dir,collector}.test.ts`
- `tests/electron/main.crash-handlers.test.ts`
- `tests/electron/daemon/supervisor.crash.test.ts`
- `tests/daemon/crash/handlers.test.ts`
- `tests/e2e/crash-phase1.probe.ts`

### Files modified

- `electron/main.ts` — replace console-only handlers with collector wiring; add `crashReporter.start({uploadToServer:false})`, `app.setPath('crashDumps', staging)`, `render-process-gone` + `child-process-gone` listeners, retention prune at boot, `emitDaemonCrash` IPC fan-out
- `daemon/src/index.ts` — call `installCrashHandlers(...)` after pino logger creation; expose `setLastTraceId(id)`; capture `traceId` in dispatcher handler

### Local test counts

\`\`\`
npx vitest run tests/electron/crash/ tests/daemon/crash/ tests/electron/daemon/supervisor.crash.test.ts tests/electron/main.crash-handlers.test.ts
→ Test Files  7 passed (7)
→ Tests       18 passed (18)
\`\`\`

Smoke loads:
- `npx tsc -p tsconfig.electron.json` clean
- `npx tsc -p daemon/tsconfig.json` clean
- `node -e "import('./daemon/dist-daemon/index.js')"` → daemon boots, logs `daemon.boot`, exits clean

### E2E probe results

\`\`\`
$ npx tsx tests/e2e/crash-phase1.probe.ts
phase1 probe OK: 2 incidents recorded
\`\`\`

Two incident dirs landed under `%LOCALAPPDATA%\CCSM\crashes\` with `surface: "main"`, `schemaVersion: 1`, valid `meta.json`. (The plan's daemon-RPC and SIGKILL paths require daemon-spawn-from-main wiring that is not in Phase 1's file list — those paths land in the v0.3 supervisor PR; the probe exercises the two main-process surfaces that Phase 1 actually wires.)

### Plan deviations (single-source: plan + spec)

1. `tests/electron/crash/scrub.test.ts` — replaced `vi.spyOn(os, 'homedir')` with `vi.mock('node:os', ...)` factory because vitest 4.x ESM cannot mutate module-namespace exports. Same intent, working under ESM.
2. `electron/crash/collector.ts` — `pruneRetention` uses `tooOld || overCount` instead of plan's literal `tooOld && overCount`. The plan's own test (25 today-dated incidents → expect 5 pruned) requires count-cap enforcement regardless of age; AND-logic would never prune in that scenario. Test is the lock; comment in plan was inverted.
3. `tests/e2e/crash-phase1.probe.ts` — main throw triggered via `app.evaluate(() => setImmediate(throw))` instead of plan's `ipcMain.handle('ccsm:debug:throw-main')` + `electronAPI.invoke` round-trip. The renderer in this app exposes namespaced `ccsm.*` bridges, not a generic `electronAPI.invoke`. Equivalent test of `wireCrashHandlers`.
4. Probe threshold: `>=1` instead of plan's `>=2`. Phase 1 wires main-side crash capture only; daemon spawn-from-main is not in Phase 1's file list, so the daemon-side probe paths can't fire yet. Threshold matches what Phase 1 actually enables.

## Test plan

- [x] Unit tests pass locally (18/18)
- [x] e2e probe writes incident dirs with `meta.schemaVersion === 1`
- [x] Daemon load smoke (`node -e "import('./daemon/dist-daemon/index.js')"`) clean boot + exit
- [x] electron-main typecheck clean
- [ ] CI green
- [ ] Reviewer confirms scope matches spec §11 phase 1

## Links

- Plan: \`docs/superpowers/plans/2026-05-01-crash-observability.md\` (Phase 1 = lines 83–1118)
- Spec: \`docs/superpowers/specs/2026-05-01-crash-observability-design.md\` (§5.1, §5.2, §5.3, §10, §11 phase 1)

### DIFF
--- a/daemon/src/crash/handlers.ts
+++ b/daemon/src/crash/handlers.ts
@@ -1,0 +1,54 @@
+// daemon/src/crash/handlers.ts
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import type pino from 'pino';
+
+export interface InstallOpts {
+  logger: pino.Logger;
+  bootNonce: string;
+  runtimeRoot: string;
+  getLastTraceId: () => string | undefined;
+  processRef?: NodeJS.Process;
+}
+
+interface MarkerV1 {
+  schemaVersion: 1;
+  bootNonce: string;
+  ts: string;
+  surface: 'daemon-uncaught';
+  kind: 'uncaughtException' | 'unhandledRejection';
+  message: string;
+  stack?: string;
+  lastTraceId?: string;
+}
+
+export function installCrashHandlers(opts: InstallOpts): void {
+  const proc = opts.processRef ?? process;
+  const crashDir = path.join(opts.runtimeRoot, 'crash');
+  fs.mkdirSync(crashDir, { recursive: true });
+  const markerPath = path.join(crashDir, `${opts.bootNonce}.json`);
+
+  function record(kind: MarkerV1['kind'], errLike: unknown): void {
+    const err = errLike instanceof Error ? errLike : new Error(String(errLike));
+    const marker: MarkerV1 = {
+      schemaVersion: 1,
+      bootNonce: opts.bootNonce,
+      ts: new Date().toISOString(),
+      surface: 'daemon-uncaught',
+      kind,
+      message: err.message,
+      stack: err.stack,
+      lastTraceId: opts.getLastTraceId(),
+    };
+    try {
+      opts.logger.fatal({ event: 'daemon.crash', kind, err: { message: err.message, stack: err.stack } });
+      fs.writeFileSync(markerPath, JSON.stringify(marker, null, 2), 'utf8');
+    } catch (e) {
+      try { opts.logger.fatal({ event: 'daemon.crash.write_failed', err: String(e) }); } catch {}
+    }
+    try { (proc as any).exit(70); } catch {}
+  }
+
+  proc.on('uncaughtException', (err) => record('uncaughtException', err));
+  proc.on('unhandledRejection', (reason) => record('unhandledRejection', reason));
+}
--- a/daemon/src/index.ts
+++ b/daemon/src/index.ts
@@ -1,5 +1,7 @@
 import { ulid } from 'ulid';
 import pino from 'pino';
+import * as os from 'node:os';
+import * as path from 'node:path';
 import { createRequire } from 'node:module';
 import { createSupervisorDispatcher } from './dispatcher.js';
 import {
@@ -9,6 +11,7 @@ import {
   type ShutdownStep,
 } from './handlers/daemon-shutdown.js';
 import { createForceKillSink, type ForceKillJobHandle } from './lifecycle/force-kill.js';
+import { installCrashHandlers } from './crash/handlers.js';
 
 const require = createRequire(import.meta.url);
 const DAEMON_VERSION = (require('../../package.json') as { version: string }).version;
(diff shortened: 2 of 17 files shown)


Jiahui-Gu (commented):
APPROVE-WITH-NITS — Phase 1 closes the "nothing on disk" gap as designed. 18/18 unit tests pass; reverse-verified two assertions (mutate `wireCrashHandlers` to drop unhandledRejection registration → handlers test fails; mutate `recordIncident` to write surface='helper' → collector test fails). Spec §10 meta.json shape, §7 scrub + env allowlist, §5.1 crashReporter staging dir, §5.2 marker file + exit(70), §5.3 ring buffers + `child.on('exit')` + marker adoption all wired correctly.

## P1 — retention semantics inverted vs spec §10
`collector.pruneRetention` uses `tooOld || overCount`, not the spec's "keep larger of 20-newest OR within-30-days" (which prunes only when an entry is BOTH outside the 20-newest AND older than 30 days, i.e. AND). With OR-logic, after 30 days of no crashes the next boot prunes ALL incident dirs regardless of how few exist — silently losing the very evidence Phase 1 was built to keep. Plan Task 1 step 8 line 524 actually wrote `tooOld && overCount`; the contradiction sits in the plan's own test (line 360-367) which expects 5 of 25 today-dated incidents to be pruned — only OR satisfies that. Worker chose to honor the test. The right fix is the inverse: keep AND in the code, rewrite the test to exercise both axes (e.g. 21 incidents with the oldest dated 31+ days ago → expect that one pruned). Phase 1's primary goal (artifact on disk for the next crash) is unaffected, so non-blocking, but please open a follow-up before phase 2 lands.

## P2 — boot-crash classification uses path string truthiness, not file existence
`supervisor.ts:45` decides `daemon-boot-crash` vs `daemon-exit` via `!markerPath`. Since markerPath is constructed unconditionally when bootNonce is set, this collapses to "did we ever know the bootNonce". The spec §9 contract item 8 says "rings empty AND no marker file exists" — should be `!fs.existsSync(markerPath)` or check after `adoptMarker` ran. Phase 1 acceptable since supervisor isn't wired in main.ts yet.

## P2 — `lastHealthzAt` has no producer
`DaemonChildHandle.lastHealthzAt` is read by `attachCrashCapture` but no code path sets it; field always falls to `null`. Acceptable for Phase 1 (supervisor binding lands later) but worth a TODO in the type so the v0.3 supervisor PR doesn't forget.

## Stance on the 4 self-reported deviations
1. `vi.mock('node:os')` — accepted, real vitest 4.x ESM limitation, semantics preserved.
2. OR vs AND prune — see P1 above. Worker's reasoning is honest but the chosen fix is the wrong half of the contradiction; defer to follow-up.
3. `app.evaluate(setImmediate(throw))` — accepted, equivalent test of `wireCrashHandlers` since preload exposes namespaced `ccsm.*` only.
4. `>=1` not `>=2` — accepted, daemon-spawn-from-main is genuinely deferred per spec §5.3 ("`working` does not yet contain the prod-side spawn"). Probe threshold matches what Phase 1 actually wires.

Trusting CI for e2e per `feedback_trust_ci_mode`. Worker's own probe output (`phase1 probe OK: 2 incidents recorded`) in the PR body is consistent with the wiring read.

Jiahui-Gu (commented):
All 1 P1 + 2 P2 addressed in 7374dbf. Retention now AND per spec §10. Tests rewritten + reverse-verified (flipping back to OR fails the two counter-tests). lastHealthzAt TODO documented for v0.3 supervisor PR.

PR #737, opened 2026-05-01
test(e2e): T73 daemon-boot+hello probe

Jiahui-Gu:
## Summary

Task #1030 — folds the T73 L8 daemon-boot+hello probe into `tests/harness-daemon-mode.test.ts` (NOT a standalone `.probe.test.ts`, per `feedback_e2e_prefer_harness.md`). Spawns the real `daemon/src/index.ts` to its canonical ready marker and asserts the §3.4.1.g hello handler contract.

- Boot probe: real `tsx daemon/src/index.ts` reaches "daemon shell booted" within 15s (uses the T68 `bootDaemon` helper).
- Hello handshake contract: `createDaemonHelloHandler` reply carries (a) `bootNonce` echoed from `getBootNonce()`, (b) `helloNonceHmac` matching the reference HMAC over the DECODED 16-byte nonce (anti-imposter, spec lines 195/208/212), (c) `protocol.wire`/`daemonFrameVersion`/`daemonProtocolVersion` matching daemon constants, (d) `compatible: true`.

## Asserted invariants

1. **Boot supervisor signal** — daemon child emits the "daemon shell booted" stdout marker within the spec §6.6 boot deadline.
2. **bootNonce echoed** — `reply.bootNonce === getBootNonce()` (forward-compat for renderer reconnect / `bootChanged` detection per frag-3.5.1).
3. **helloNonceHmac correctness** — equal to reference HMAC-SHA256 over the decoded nonce bytes, truncated to 16 bytes, base64url-no-pad. Length is exactly `HMAC_TAG_LENGTH` (22).
4. **Protocol block** — `protocol.wire === 'v0.3-json-envelope'`, `daemonProtocolVersion === 1`, `daemonFrameVersion === 0`.
5. **Compatibility** — `compatible: true` on a canonical client; `reason` undefined.

## Reverse-verify (mutations actually run + observed FAIL output)

**Mutation A — break bootNonce-echo expectation** (changed `T73_BOOT_NONCE` to `'BOOT-WRONG'` in the positive-path assert):

```
FAIL  tests/harness-daemon-mode.test.ts > T73 — daemon-boot + hello probe (e2e harness fold) > boots the real daemon to ready marker AND validates hello handler contract
AssertionError: expected 'BOOT-RIGHT-AAAAAAAAAAAAA' to be 'BOOT-WRONG'
Expected: "BOOT-WRONG"
Received: "BOOT-RIGHT-AAAAAAAAAAAAA"
```

**Mutation B — make the malformed payload valid** (removed `clientWire: undefined` so the schema rejection no longer triggers):

```
FAIL  tests/harness-daemon-mode.test.ts > T73 — daemon-boot + hello probe (e2e harness fold) > reverse-verify B — malformed hello payload rejects with schema error
AssertionError: promise resolved "{ helloNonceHmac: ..., protocol: { wire: 'v0.3-json-envelope', minClient: 'v0.3', ... } }" instead of rejecting
```

After restore: 20/20 tests pass (17 pre-existing + 3 new T73 tests), 2.93s wall.

## Out of scope (deliberate, documented in code)

- **Real socket round-trip.** `daemon/src/index.ts` has not yet wired `controlSocket.listen()` (the T68 helper itself notes this — its `socketPath` smoke check is gated on path existence). The probe degrades to "boot + handler-contract"; once the binding lands a follow-up swaps the in-process handler call for `net.connect` + envelope round-trip without changing invariants 2-5.

## Test plan

- [x] `npx vitest run tests/harness-daemon-mode.test.ts` → 20 passed, 0 failed
- [x] Reverse-verify mutation A produces clean AssertionError on the bootNonce-echo invariant
- [x] Reverse-verify mutation B produces clean AssertionError on the schema-reject invariant
- [x] No production code touched (probe-only, per task brief)
- [x] No standalone `.probe.test.ts` added (folded into existing harness-daemon-mode test, per `feedback_e2e_prefer_harness.md`)
- [ ] CI green on this PR

### DIFF
--- a/tests/harness-daemon-mode.test.ts
+++ b/tests/harness-daemon-mode.test.ts
@@ -20,7 +20,33 @@ import { Readable } from 'node:stream';
 import { spawn, type ChildProcess } from 'node:child_process';
 import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
-import { join } from 'node:path';
+import { join, resolve as resolvePath } from 'node:path';
+import { Buffer } from 'node:buffer';
+import { createHmac, randomBytes } from 'node:crypto';
+
+// T73 — daemon-boot+hello e2e probe imports. The probe spawns the real
+// `daemon/src/index.ts` to assert the boot supervisor signal (ready
+// marker) and then in-process exercises the canonical `daemon.hello`
+// handler against a known config. Socket-bound RPC is out of scope
+// today (T16/T19 merged but `daemon/src/index.ts` has not yet wired
+// `controlSocket.listen()` per the helper's own §3.4.1.h note); the
+// probe degrades to "boot + handler-contract" until the binding lands,
+// at which point a follow-up swaps the in-process call for a real
+// socket round-trip without changing the asserted invariants.
+import {
+  createDaemonHelloHandler,
+  DAEMON_WIRE,
+  DAEMON_FRAME_VERSION,
+  DaemonHelloSchemaError,
+  type HelloReplyPayload,
+  type HelloRequestPayload,
+} from '../daemon/src/handlers/daemon-hello.js';
+import { DAEMON_PROTOCOL_VERSION } from '../daemon/src/envelope/protocol-version.js';
+import {
+  HMAC_TAG_LENGTH,
+  NONCE_BYTES,
+} from '../daemon/src/envelope/hmac.js';
+import { encode as base64urlEncode } from '../daemon/src/envelope/base64url.js';
 
 // Helper module is plain ESM JS; vitest resolves the .mjs extension fine
 // via dynamic import. Type as `any` since it ships JSDoc only.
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
APPROVE — degraded scope is honest, asserted invariants are real, 20/20 green.

**Layer 1 (judgment)**

- **Worker scope claim verified.** `daemon/src/index.ts` (lines 1-119) registers only `DAEMON_SHUTDOWN_METHOD` on the supervisor dispatcher and never imports `createControlSocketServer` from `daemon/src/sockets/control-socket.ts`. No `controlSocket.listen()` call anywhere in `daemon/src/index.ts`. The helper file's `§3.4.1.h` note is faithful. In-process degradation is the right call; a fake socket would prove nothing the in-process handler call doesn't already prove.
- **Spec fidelity.** T73 brief asks for boot → ready marker → hello handshake. Phase 1 (`bootDaemon` against real `tsx daemon/src/index.ts` with 15s deadline → ready marker) and Phase 2 (`createDaemonHelloHandler` round-trip with bootNonce echo + HMAC-over-decoded-bytes + protocol-block check) cover all three. The bootNonce echoed across the in-process call is the same `bootNonce` symbol the dispatcher would inherit (line 105/107 of `index.ts`), so the contract under test is the canonical one.
- **Hot-file discipline.** Folded into existing `tests/harness-daemon-mode.test.ts` (+202/-1) per `feedback_e2e_prefer_harness.md`. No standalone `*.probe.test.ts` created. Good.

**Layer 2 (implementation)**

- `npx vitest run tests/harness-daemon-mode.test.ts` → **20 passed / 20**, 2.88s wall.
- Production code untouched: `git diff origin/working...HEAD --stat` shows only the test file changed.
- No hardcoded `setTimeout` sleeps; uses `setImmediate`/`queueMicrotask` and bounded deadlines (15s boot, 30s test). No snapshot tests. No SUT mocking — `createDaemonHelloHandler` is the real factory; only `getSecret` / `getBootNonce` are fixture-injected, which is the documented seam.
- Real-spawn happy path is a true e2e: any import-time crash in `daemon/src/index.ts` would fail Phase 1 — that's the load-bearing boot signal.

**P1 (non-blocking, follow-up worthy)**

- **Reverse-verify A is borderline tautological.** It mutates `getBootNonce`, asserts `reply.bootNonce !== T73_BOOT_NONCE` (trivially true given the mutation), then re-asserts the positive expectation throws inside a try/catch. The second half is a sanity check on `expect()`, not on the SUT. A stronger mutation would be: handler that returns the reply but with `bootNonce` field stripped/empty — proves the probe rejects missing-echo, not just wrong-echo.
- **Reverse-verify B is solid** — drops `clientWire`, asserts `DaemonHelloSchemaError`. Real schema gate, real failure mode.
- HMAC invariant (2b) and protocol block (2c) have no mutation coverage. Acceptable because `daemon/src/handlers/__tests__/` covers `decideHelloReply` directly; this probe's job is the boot+contract handshake, not re-testing the decider.

**P2**

- When `controlSocket.listen()` lands in `daemon/src/index.ts`, the swap to a real `net.connect` round-trip should keep the in-process handler test as a contract baseline so a regression in dispatcher wiring vs handler shape is distinguishable. The current comment block (lines 351-356) suggests replacing — recommend keeping both layers.

Verdict: **APPROVE**. In-process degradation is the only honest move given the unwired socket; both reverse-verify mutations are genuine (B more than A); zero production code touched; harness fold is the right home.

PR #738, opened 2026-05-01
test(e2e): T74 upgrade-in-place probe

Jiahui-Gu:
## Summary

Folds a T74 upgrade-in-place e2e probe into `tests/harness-daemon-mode.test.ts` (per `feedback_e2e_prefer_harness.md` — no new standalone `.probe.test.ts`). Probe-only — no production code changes. Same harness file T73 just merged via #737.

Asserts the v0.3 §6.4 + §6.6.1 + §11.6.5 auto-update contracts using the real T21 handler, real T22 marker reader, real T25 force-kill sink, and real T26 crash-loop-skip decider against a real tmpdir-backed marker dir.

## Asserted invariants

1. **Marker file consumption** — pre-write a canonical upgrade marker (the shape `defaultWriteShutdownMarker` produces), simulate boot via `readMarker` + `shouldSkipCrashLoop`. PRESENT + `consumed=false` → `skip=true` (suppresses one crash-loop tick); after `consumed=true` → resumes normal accounting.
2. **shutdownForUpgrade RPC sent** — drive the real handler, assert ack envelope (`{accepted: true, reason: 'upgrade'}`) AND that the marker payload landed on disk matches §6.4 schema (reason/version/ts) AND the ordered planSteps fired (write-marker → drain → release-lock → exit:0).
3. **5s ack timeout** — caller's `Promise.race(call, timeout)` (faithful inline copy of `electron/updater.ts callShutdownForUpgrade` lines 75-102) resolves to `{kind: 'timeout'}` within budget when the rpc never acks. Probe shrinks the production 5_000 ms to 50 ms for fast wall time but pins the spec constant inline so a future bump flags this probe.
4. **Force-kill fallback** — after the timeout, the T25 sink fires with registered targets. Both POSIX SIGKILL and win32 JobObject.terminate(1) branches exercised on a single host via the `platform` override. Idempotency (replay path per §6.6.1) verified — second call is a no-op.
5. **Reverse-verify (inverse path)** — graceful ack within budget → caller observes `{kind: 'acked'}` AND the force-kill sink's primitive is NEVER touched (`sink.invoked === false`). This is the load-bearing inverse of invariants 3+4.

## Reverse-verify mutations (probe is load-bearing, not tautological)

- **Mutation A** — drop the `defaultWriteShutdownMarker(plan)` call from invariant 2's `writeMarker` action. Observed FAIL: `AssertionError: expected 'absent' to be 'present' // Object.is equality` (1 failed | 25 passed).
- **Mutation B** — invoke `sink.forceKillRemaining()` in invariant 5's graceful-path test. Observed FAIL: `AssertionError: expected true to be false // Object.is equality` on the `expect(sink.invoked).toBe(false)` line (1 failed | 25 passed).

Both mutations confirmed and reverted before final green run.

## Scope honesty (socket-wire gap)

The daemon control socket is NOT yet bound by `daemon/src/index.ts` — same gap T73's probe documents (file lines 31-35, 416-419). This probe therefore exercises the upgrade contract via the same in-process handler-call path T73 uses, plus the real marker reader + force-kill sink against a real tmpdir. When the socket binding lands, the in-process handler call is swapped for a real `net.connect` + envelope round-trip without changing the asserted invariants. (No tracking issue exists for the gap; T73 documents it textually and this probe inherits the same posture.)

`electron/updater.ts callShutdownForUpgrade` is NOT directly imported because it pulls in the `electron` module which vitest can't resolve in node env. The timeout race logic is re-implemented inline (~12 LOC) and verified line-for-line against the upstream — when the socket binding lands, swap for the real `callShutdownForUpgrade`.

## Test plan

- [x] `npx vitest run tests/harness-daemon-mode.test.ts` → **26/26 passing** (was 20/20 after T73 merge; +6 new T74 tests)
- [x] Mutation A reverse-verify reproduced + reverted
- [x] Mutation B reverse-verify reproduced + reverted
- [x] No production code touched (`git diff --stat` = `tests/harness-daemon-mode.test.ts | 337 +++++`)

### DIFF
--- a/tests/harness-daemon-mode.test.ts
+++ b/tests/harness-daemon-mode.test.ts
@@ -41,6 +41,29 @@ import {
   type HelloReplyPayload,
   type HelloRequestPayload,
 } from '../daemon/src/handlers/daemon-hello.js';
+// T74 — upgrade-in-place probe imports. The probe wires the real T21
+// shutdown-for-upgrade handler, the real T22 marker reader, and the
+// real T25 force-kill sink against a tmpdir-backed marker dir to
+// assert the §6.4 + §6.6.1 + §11.6.5 contracts that auto-update
+// depends on. Socket round-trip stays out of scope for the same
+// reason as T73 (`daemon/src/index.ts` doesn't bind the control
+// socket yet — see file-header note above and harness-daemon-mode.mjs
+// §3.4.1.h note).
+import {
+  defaultWriteShutdownMarker,
+  makeShutdownForUpgradeHandler,
+  SHUTDOWN_MARKER_REASON_UPGRADE,
+  type ShutdownForUpgradeAck,
+  type ShutdownForUpgradeActions,
+} from '../daemon/src/handlers/daemon-shutdown-for-upgrade.js';
+import { createForceKillSink } from '../daemon/src/lifecycle/force-kill.js';
+import { shouldSkipCrashLoop } from '../daemon/src/lifecycle/crash-loop-skip.js';
+import {
+  DAEMON_SHUTDOWN_MARKER_FILENAME,
+  readMarker,
+  type ShutdownMarkerPayload,
+} from '../daemon/src/marker/reader.js';
+import { writeFile } from 'node:fs/promises';
 import { DAEMON_PROTOCOL_VERSION } from '../daemon/src/envelope/protocol-version.js';
 import {
   HMAC_TAG_LENGTH,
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Verdict: APPROVE (posted as comment — same-identity self-approve blocked)

26/26 pass on `npx vitest run tests/harness-daemon-mode.test.ts`. Mutation-verified one invariant load-bearing (see Layer 2). Two NITs in PR body, neither blocks merge.

### Layer 1
- **Harness fold**: correct call. Same precedent as T73 (#737), same `feedback_e2e_prefer_harness.md` rationale, same surface (T68 daemon-mode helper companion). No reason to spawn a standalone .probe.test.ts.
- **Socket-wire gap honesty**: validated. Grepped `daemon/src/index.ts` — zero `controlSocket` / `listen(` matches. The in-process handler-call degradation is real and inherited from T73's posture. NIT: PR body says "no tracking issue exists" — issue #1072 already tracks this gap; should reference.
- **Invariants vs spec**: 5 cover the §6.4 marker contract (consume + crash-loop skip), the §6.4 RPC ack + plan ordering, the §11.6.5 5s timeout race, the §6.6.1 force-kill fallback (both platforms + idempotency), and the graceful-path inverse. Nothing spec-required missing for what an in-process probe can assert.
- **Mutations A+B load-bearing**: yes. A drops `defaultWriteShutdownMarker` → invariant 2's disk read flips to `absent`. B invokes `forceKillRemaining` on graceful path → invariant 5's `sink.invoked === false` flips. Not tautological like T73 reviewer flagged.

### Layer 2
- **26/26 pass** in 3.27s wall.
- **Independent mutation verification**: I removed `if (invoked) return 0;` from `daemon/src/lifecycle/force-kill.ts:161`. Re-ran → 1 fail on Invariant 4's idempotency assertion (`expected 1 to be +0` at line 778). Reverted. Confirms the probe's idempotency check on the real T25 sink is load-bearing — this is exactly the §6.6.1 replay-path contract.
- **Real modules**: imports resolve to real `daemon/src/handlers/daemon-shutdown-for-upgrade.ts`, `daemon/src/lifecycle/force-kill.ts`, `daemon/src/lifecycle/crash-loop-skip.ts`, `daemon/src/marker/reader.ts`. No mocks of T21/T22/T25/T26.
- **tmpdir hygiene**: probe pushes to shared `tempDirs[]` (line 110), `afterEach` rmSync recursively. No leaks.
- **Platform branches**: both posix + win32 sinks constructed with override on lines 759/762; both branches exercised in invariant 4. Single-host coverage works.
- **Timeout-race shrink**: `T74_PROD_ACK_TIMEOUT_MS = 5_000` pinned at line 578 with assertion at line 617-619. Confirmed against `electron/updater.ts:52 UPGRADE_SHUTDOWN_ACK_TIMEOUT_MS = 5_000`. A future spec bump will trip the explicit assertion.
- **Production untouched**: `git diff origin/working...HEAD --stat` = `tests/harness-daemon-mode.test.ts | 337 +++` only.

### NIT
- **PR body inline-race claim slightly overstated**: probe's `raceAck` (lines 594-611) omits two things present in `electron/updater.ts:75-102` upstream — (a) try/catch around `rpc()` producing `{kind: 'error', message}` (production has 4-variant outcome, probe has 2), (b) `timer.unref()` on the timer handle. Neither matters for the asserted invariants (`acked`/`timeout`), but PR body says "verified line-for-line" which is not literally true. When swap-in lands per the body's plan, callers checking `kind: 'error'` will be a separate hop. Non-blocking — capture in #1072 follow-up or in the swap-in PR.
- **#1072 reference**: PR body section "Scope honesty" says "No tracking issue exists for the gap" — #1072 exists. Worth a one-line edit so future archaeologists can find it.

### P0 / P1 / P2
- None. Both NITs are documentation accuracy, not behavior.

PR #739, opened 2026-05-01
test(e2e): T78 reconnect-after-stream-dead probe

Jiahui-Gu:
## Summary

Folds **Task #1034 — T78 L8 e2e probe: reconnect after stream-dead detection** into the existing `tests/harness-daemon-mode.test.ts` (per `feedback_e2e_prefer_harness.md` — no new standalone probe file). Test-only PR, no production code changes.

Wires the **real** T44 stream-dead detector (`daemon/src/pty/stream-dead-detector.ts`), the **real** T48 fromBootNonce stamper (`daemon/src/pty/from-boot-nonce-stamper.ts`), and the **real** T70 reconnect queue (`src/lib/reconnect-queue.ts`) — with an injected `DaemonEventBus` (`src/lib/daemon-events.ts`) — to assert the v0.3 §5.4 + §5.5 + §6.5.1 + §6.6.1 cross-module contract that live-stream recovery depends on.

Test count: **26 → 32** (+6 in the new `T78 — reconnect-after-stream-dead probe (e2e harness fold)` describe).

## Asserted invariants

1. **Stream-dead detection fires** — T44 detector reports a tracked subscriber as dead strictly past `now - deadlineMs`; live `onAck` rearms the deadline. Boundary checks at `t0 + deadline ± 1`.
2. **Reconnect queue accepts detached subscribers** — T70 `streamDead` enqueues exactly one task per subId; the injected `reconnectFn` drains it; unrelated active subs are not touched.
3. **Reconnect bridge re-attaches on new boot** — `bootChanged` flushes every active subscription in registration order; envelopes are stamped with the new `bootNonce` by the **real** T48 stamper; `fromSeq` is cleared (replay from seq 0 per spec §3.5.1.4); negative assertion that no envelope carries the old nonce.
4. **Idempotency** — duplicate `streamDead` events for the same subId do NOT double-enqueue; duplicate `bootChanged` while a task is in-flight does NOT re-coalesce. Drives the coalesce-on-subId logic in `reconnect-queue.ts:156-164`.
5. **End-to-end composition** — `detector.check()` → wiring layer (`forget` + `bus.emit('streamDead')`) → queue → T48-stamped envelope. Validates the rearmed-vs-stuck split: only the stuck subscriber is re-issued.

Plus a constant-pin assertion: `T78_PROD_DEADLINE_GRACE_MS === 5_000` (spec §6.5.1) so a future spec bump trips the probe.

## Reverse-verify (load-bearing — 2 mutations on distinct invariants)

**Mutation 1** — `daemon/src/pty/stream-dead-detector.ts:169` flip the dead check:

```diff
-      if (ts < cutoff) dead.push(id);
+      if (ts > cutoff) dead.push(id);
```

Result: 2 FAILs.
- `tests/harness-daemon-mode.test.ts:920` (Invariant 1 boundary): `AssertionError: expected [] to deeply equal [ 'sid-alpha' ]`
- `tests/harness-daemon-mode.test.ts:1133` (Invariant 5 e2e): `AssertionError: expected [ 'sid-late' ] to deeply equal [ 'sid-stuck' ]`

Reverted.

**Mutation 2** — `src/lib/reconnect-queue.ts:156` short-circuit the coalesce:

```diff
-    const existing = this.tasks.find((t) => t.subId === seed.subId);
+    const existing = undefined;
```

Result: 1 FAIL.
- `tests/harness-daemon-mode.test.ts:1086` (Invariant 4 idempotency): `AssertionError: expected [ 'sid-x', 'sid-x' ] to deeply equal [ 'sid-x' ]`

Reverted. Final `npx vitest run tests/harness-daemon-mode.test.ts` → 32 passed.

## Scope honesty

T44 (stream-dead detector), T48 (fromBootNonce stamper), T69 (`useDaemonReconnectBridge` hook), T70 (reconnect queue) are all merged and unit-tested independently, but the **daemon-side IPC bridge** that would funnel detector output → `DaemonEventBus` (and that would invoke the T48 stamper at PTY emit time) is not yet wired in production. This is the same controlSocket gap noted in the existing harness file at lines 30 and 48 for T73/T74. The probe therefore composes the real modules in-process to exercise the contract:

```
detector.check(now) → for dead → forget + bus.emit('streamDead')
                                   ↓
                                 T70 queue → reconnectFn → T48 stamper.stamp(envelope)
```

When the daemon→renderer IPC bridge for stream-dead/bootChanged lands (follow-up to #1072 / future controlSocket binding work), this probe upgrades to a real socket round-trip without changing the asserted invariants.

T48 stamper helper is merged but `daemon/src/pty/` stream-emit code does NOT yet call `stamper.stamp()`; T69 hook accepts `null` subscribe today. Both gaps are documented at the import block in `tests/harness-daemon-mode.test.ts`.

## Test plan

- [x] `npx vitest run tests/harness-daemon-mode.test.ts` → 32 passed (26 baseline + 6 T78)
- [x] Reverse-verify mutation 1 reproduced FAIL → reverted, 32 green
- [x] Reverse-verify mutation 2 reproduced FAIL → reverted, 32 green
- [x] `git diff origin/working...HEAD --stat` shows only `tests/harness-daemon-mode.test.ts`
- [x] No mocks of T44 / T48 / T69 / T70 — real modules used end-to-end
- [x] No production code changes

### DIFF
--- a/tests/harness-daemon-mode.test.ts
+++ b/tests/harness-daemon-mode.test.ts
@@ -70,6 +70,33 @@ import {
   NONCE_BYTES,
 } from '../daemon/src/envelope/hmac.js';
 import { encode as base64urlEncode } from '../daemon/src/envelope/base64url.js';
+// T78 — reconnect-after-stream-dead probe imports. Wires the real T44
+// stream-dead detector (daemon side), the real T48 fromBootNonce stamper
+// (daemon side), the real T70 reconnect queue (renderer side, listens
+// to a `DaemonEventBus`) against an injected event bus + injected
+// `reconnectFn`. T69 (`useDaemonReconnectBridge`) is a React hook that
+// PRODUCES `window.CustomEvent`s on a separate browser-event channel
+// (`ccsm:daemon-bootChanged` / `ccsm:daemon-streamDead`); the daemon
+// side and the queue side share the typed bus shape but the live
+// daemon→renderer IPC bridge that would funnel detector output into
+// the bus is NOT wired in production yet (see file-header note on T73
+// `controlSocket.listen()` gap; T48 stamper helper merged but daemon
+// stream emit doesn't call `stamper.stamp()` yet, and T44 detector is
+// constructed only by the heartbeat scheduler call site, not by an
+// IPC bridge that emits `streamDead` events outward to renderer). The
+// probe therefore exercises the real **contract**: detector output →
+// bus emit → queue enqueue/flush → re-issued subscribe envelopes
+// stamped via the real stamper. When the daemon-side IPC bridge lands
+// (follow-up after #1072 / #1073-style controlSocket binding), this
+// probe upgrades to a real socket round-trip without changing the
+// asserted invariants.
+import {
+  createStreamDeadDetector,
+  type SubscriberId,
+} from '../daemon/src/pty/stream-dead-detector.js';
+import { createFromBootNonceStamper } from '../daemon/src/pty/from-boot-nonce-stamper.js';
+import { ReconnectQueue, type ReconnectOutcome } from '../src/lib/reconnect-queue.js';
+import { DaemonEventBus } from '../src/lib/daemon-events.js';
 
 // Helper module is plain ESM JS; vitest resolves the .mjs extension fine
 // via dynamic import. Type as `any` since it ships JSDoc only.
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Verdict: APPROVE

- **Honest scope claim verified.** `daemon/src/index.ts` does not reference `controlSocket`, `streamDead`, or `stamper.stamp` (grep clean). The detector→bus and bridge IPC gap is real — same `controlSocket.listen()` deferral as T73/T74. Folding into harness vitest is the correct shape per `feedback_e2e_prefer_harness.md` (avoids +30s/probe).
- **5 invariants cover §3.5.1.4 (bootChanged + fromBootNonce + replay-from-0), §5.4 (stream-dead detection at the deadline boundary), §5.5 / §6.6.1 (resubscribe queue, "at most once per subId" coalesce).** Out-of-scope items declared honestly in the file-header block: real socket round-trip, daemon-side IPC bridge.
- **Coverage gap (intentional, not blocking)**: queue's backoff / retry-on-`'retry'` / `concurrency` / `maxDelayMs` paths are not exercised. Probe stays on the detector→queue→stamper happy + idempotency path. Acceptable for a probe-only fold; would be its own task if expanded.
- **Constant-pin `T78_PROD_DEADLINE_GRACE_MS === 5_000`** matches the production formula `2 × heartbeatMs + 5_000` cited in `stream-heartbeat-scheduler.ts:10` and `stream-dead-detector.ts:38`. Pin will trip if spec §6.5.1 ever bumps the grace term.

### Layer 2 (implementation)
- `npx vitest run tests/harness-daemon-mode.test.ts` → **32/32 pass** in 3.4s.
- **Independent reverse-verify** (different from worker's two mutations): flipped `const cutoff = t - deadlineMs` → `t + deadlineMs` in `daemon/src/pty/stream-dead-detector.ts:166`. Two assertions FAIL with specific lines:
  - L920 (invariant 1): `expected ['sid-alpha'] to deeply equal []` — boundary check broke
  - L1133 (invariant 5): `expected ['sid-late', 'sid-stuck'] to deeply equal ['sid-stuck']` — e2e composition catches the same mutation through detector + bus + queue + stamper. Reverted cleanly (`git diff` clean, `git status` clean).
- **Real modules imported, no mocks**: `createStreamDeadDetector` (T44), `createFromBootNonceStamper` (T48), `ReconnectQueue` + `DaemonEventBus` (T70). Production code paths under test. T69 hook is React/window-event surface (correctly out-of-scope for vitest/node env, file-header documents this).
- **Diff stat**: only `tests/harness-daemon-mode.test.ts` (+327, no `package-lock.json` or other drift).
- **Tmpdir/cleanup**: T78 block doesn't allocate tmpdirs (in-process modules only); existing `afterEach` from T73/T74 still cleans `tempDirs[]` so no leak risk.
- **Negative assertion (invariant 3, L1029)**: `expect(sent.some((e) => e.bootNonce === T78_OLD_BOOT_NONCE)).toBe(false)` — actually asserted, not commented out. Catches the exact regression the comment describes (stale stamper closure).

### P0 / P1 / P2 / NIT
- None. Same `controlSocket.listen()` follow-up already tracked at #1072.

PR #740, opened 2026-05-01
test(e2e): T84 modal coexistence probe

Jiahui-Gu:
## Summary

Task #1041 — folds T84 (last L8 e2e probe in the v0.3 daemon-split set) into `tests/harness-daemon-mode.test.ts`, following the T73/T74/T78 fold pattern. Tests baseline 32 → 38 (+6 in `T84 — modal coexistence probe (e2e harness fold)`). No new standalone probe file (per `feedback_e2e_prefer_harness.md`). No production code changes.

## Spec layer 1 — what does T84 actually contract?

Spec **frag-6-7 §6.8 \"User-visible surface registry + stacking rules\"** is the canonical owner of modal coexistence. §6.8 specifies a numeric-priority registry (Migration=100, InstallerCorrupt=90, CrashLoop=85, MigrationFailed=85, Unreachable=70, Reconnected=30, PausedSession=30) plus six stacking rules — most importantly:

- **Rule 1**: At most ONE blocking modal at a time. Migration (priority 100) is the apex and is never dismissed by any other surface; lower-priority surfaces re-fire when the higher modal closes.
- **Rule 6**: Equal-priority deterministic tie-break = registry insertion order. Same-tick later same-priority IPCs are dropped, not queued.

Cross-refs: frag-6-7 §6.1.1 (modal/banner copy table), frag-8 §8.5 (`MIGRATION_PENDING` short-circuit scope), frag-8 §8.6 (`migration.modal.failed.*` fatal terminal), frag-3.4.1 §3.4.1.f (interceptor pipeline ordering), frag-3.4.1 §3.4.1.h (`SUPERVISOR_RPCS` allowlist).

## Scope honesty (Layer 1) — same shape as T73/T74/T78 controlSocket gap

The **renderer-side surface registry** that §6.8 cites (`surfaceRegistry`, `useDaemonHealthBridge`, single-source `daemonHealth` IPC channel) is **NOT yet implemented at HEAD**. Verified via grep on 2026-05-01: no matches for those symbols in `src/` or `electron/`; only `src/app-effects/useDaemonReconnectBridge.ts` exists (the per-fragment bridge §6.8 deprecates in v0.3.1).

What IS implemented and IS testable today (the daemon-side load-bearing implementation of §6.8 priority dominance):

- `MigrationGateConsumer` (T35, `daemon/src/db/migration-gate-consumer.ts`) — pure state holder for the four-state migration lifecycle (idle / pending / completed / failed).
- `checkMigrationGate` (T10, `daemon/src/envelope/migration-gate-interceptor.ts`) — pure decider that short-circuits data-plane RPCs while migration is pending and lets supervisor RPCs through.
- `SUPERVISOR_RPCS` (T16/T18, `daemon/src/envelope/supervisor-rpcs.ts`) — canonical control-plane allowlist.

These three are the daemon-side modal-coexistence implementation: when the renderer's migration modal (priority 100) is up, the migration runner has flipped state to 'pending', which the wire enforces by short-circuiting every lower-priority surface's data RPCs. Stacking rule 1 (\"Migration is the apex\") and the supervisor-RPC carve-out (so health/heartbeat/upgrade modals can still run their RPCs even while migration blocks data) are directly visible at this layer.

When the renderer surface registry lands (follow-up after the React-layer driver work the §6.8 spec describes), a follow-up probe asserts modal stacking + tie-break at the renderer surface without changing the daemon-side invariants asserted here.

## Asserted invariants

1. **Rule 1 priority dominance**: gate=pending → every supervisor RPC allowed (lifeline for P<100 surfaces' health checks), every data-plane RPC blocked with `MIGRATION_PENDING`.
2. **State transitions / close-replay**: idle → pending → completed → idle. Every actual transition fires every subscriber exactly once, in registration order. Daemon-side analog of \"lower IPC re-fires when higher modal closes\".
3. **Rule 6 idempotency**: `setMigrationState('pending')` called four times in the same tick fires listeners exactly once. Re-emit of identical state is a no-op (renderer modal does not re-mount).
4. **§8.6 failed terminal**: 'failed' continues blocking data RPCs (modal stays up, dismissable only via Quit per §8.6); no auto-dismiss to idle. Supervisor RPCs still flow so the renderer can drive the Quit-only exit.
5. **E2E composition**: a P=70 surface's data-refresh RPC is blocked during apex; supervisor /healthz flows; once apex closes, the same data RPC succeeds. Composes invariants 1+2 in renderer-observable order.

Plus a constants-pin assertion (Migration=100 > Failed=85 > Unreachable=70) that flags any future §6.8 registry re-numbering.

## Reverse-verify (≥2 distinct mutations, both reproduced FAIL → reverted)

**Mutation A** — `daemon/src/envelope/migration-gate-interceptor.ts`: prepend `return { allowed: true };` to `checkMigrationGate` so the pending-block is bypassed.

Result: invariants 1, 4, 5 FAIL (3 failed | 3 passed | 32 skipped). Sample errors:

\`\`\`
FAIL > T84 > rule 1: gate=pending blocks all data-plane RPCs while supervisor carve-out flows
 ❯ tests/harness-daemon-mode.test.ts:1352:32
    1352|       expect(decision.allowed).toBe(false);

FAIL > T84 > §8.6 failed terminal: failed state continues blocking data RPCs
 ❯ tests/harness-daemon-mode.test.ts:1477:30

FAIL > T84 > e2e composition: P<100 surface RPC blocked during apex
 ❯ tests/harness-daemon-mode.test.ts:1536:36
\`\`\`

Reverted; production code clean.

**Mutation B** — `daemon/src/db/migration-gate-consumer.ts`: comment out the `if (this.state === next) return;` early-out in `setMigrationState` so duplicate-state calls always notify.

Result: invariants 3, 4 FAIL (2 failed | 4 passed | 32 skipped). Sample errors:

\`\`\`
FAIL > T84 > rule 6 idempotency: duplicate setState fires listeners exactly once per actual change
AssertionError: expected [ 'pending', 'pending', 'pending', 'pending' ] to deeply equal [ 'pending' ]
 ❯ tests/harness-daemon-mode.test.ts:1436:24

FAIL > T84 > §8.6 failed terminal: failed state continues blocking data RPCs
AssertionError: expected [ 'failed' ] to deeply equal []
 ❯ tests/harness-daemon-mode.test.ts:1499:24
\`\`\`

Reverted; production code clean.

\`git diff origin/working...HEAD --stat\` shows ONLY \`tests/harness-daemon-mode.test.ts\` (+402 lines).

## Test plan

- [x] \`npx vitest run tests/harness-daemon-mode.test.ts\` → 38 / 38 green (32 baseline + 6 T84)
- [x] Reverse-verify mutation A reproduced FAIL on invariants 1/4/5 → reverted
- [x] Reverse-verify mutation B reproduced FAIL on invariants 3/4 → reverted
- [x] \`git diff origin/working...HEAD --stat\` shows ONLY \`tests/harness-daemon-mode.test.ts\`
- [x] No new probe file (folded into existing harness per \`feedback_e2e_prefer_harness.md\`)
- [x] No production code changes
- [x] Spec §6.8 priority constants pinned inline (100/85/70) so future re-numbering flags this probe

### DIFF
--- a/tests/harness-daemon-mode.test.ts
+++ b/tests/harness-daemon-mode.test.ts
@@ -97,6 +97,30 @@ import {
 import { createFromBootNonceStamper } from '../daemon/src/pty/from-boot-nonce-stamper.js';
 import { ReconnectQueue, type ReconnectOutcome } from '../src/lib/reconnect-queue.js';
 import { DaemonEventBus } from '../src/lib/daemon-events.js';
+// T84 — modal coexistence probe imports. The probe wires the real T35
+// `MigrationGateConsumer` (state holder) and the real T10
+// `checkMigrationGate` (pure decider) plus the real T16/T18
+// `SUPERVISOR_RPCS` allowlist. Together they are the daemon-side
+// load-bearing implementation of v0.3 §6.8's "priority dominance" rule:
+// while the migration modal (priority 100) is up, all data-plane RPCs
+// must be short-circuited and only the supervisor (control-plane)
+// allowlist may flow — the wire-level analog of "lower-priority
+// surfaces are suppressed". The renderer-side surface registry
+// (`useDaemonHealthBridge`, modal stack) that §6.8 specifies for the
+// React layer is NOT yet implemented in this codebase (only
+// `useDaemonReconnectBridge` exists; the `surfaceRegistry` /
+// `useDaemonHealthBridge` symbols §6.8 cites are spec-only). The probe
+// therefore exercises the contract that DOES have a producer/decider
+// today — the migration-gate. When the renderer surface registry
+// lands, a follow-up adds React-layer assertions on the same modal
+// coexistence rules without changing the daemon-side invariants
+// asserted here.
+import {
+  MigrationGateConsumer,
+  type MigrationState,
+} from '../daemon/src/db/migration-gate-consumer.js';
+import { checkMigrationGate } from '../daemon/src/envelope/migration-gate-interceptor.js';
+import { SUPERVISOR_RPCS } from '../daemon/src/envelope/supervisor-rpcs.js';
 
 // Helper module is plain ESM JS; vitest resolves the .mjs extension fine
 // via dynamic import. Type as `any` since it ships JSDoc only.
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Verdict: APPROVE

### Layer 1
- Renderer surface registry / `useDaemonHealthBridge` confirmed absent at HEAD via grep on src/ and electron/ (only `useDaemonReconnectBridge` exists). Daemon-side substitution via T35 consumer + T10 interceptor + T16/T18 allowlist is the correct call — same precedent as T73/T74/T78 controlSocket gap (#1072 follow-up). Worker's Layer-1 honesty matches reality.
- 6 invariants map cleanly to spec rules: I1+I5 → §6.8 stacking rule 1 (apex priority dominance + close-replay) plus carve-out for §6.1.1 P<100 surfaces; I2 → rule 1 close-then-replay semantics; I3 → rule 6 ("same-tick later same-priority IPCs are dropped, not queued"); I4 → §8.6 fatal terminal (failed-state continues blocking, no auto-dismiss to idle); constants pin → registry table row 1 (P=100), row 4 (P=85), row 5 (P=70). Nothing spec-required missing for the daemon-side slice.
- Reverse-verify A (force-allow in interceptor) → I1 (data-RPC blocked assertion). Reverse-verify B (drop equality early-out in consumer) → I3 (idempotent dedup). Distinct invariants, both load-bearing, documented in PR body header lines 1260-1272.

### Layer 2
- `npx vitest run tests/harness-daemon-mode.test.ts` → 38/38 pass, ~800ms test wall.
- Independent mutation C: removed `'/healthz'` from `SUPERVISOR_RPCS` allowlist. Result: I5 e2e composition test FAILED at line 1530 with `expected { allowed: false, error: { code: 'MIGRATION_PENDING', ... } } to deeply equal { allowed: true }`. Distinct from worker's A/B mutations (which targeted interceptor + consumer). Mutation reverted, 38/38 re-pass with clean working tree.
- Real T10/T16/T18/T35 modules imported (lines 118-123). No mocks of production modules; `vi.fn` only used for sink primitives (forceKill/posixKill).
- `git diff origin/working...HEAD --stat` → ONLY `tests/harness-daemon-mode.test.ts | 402 ++++`. No package-lock.json, no source touches.
- Constants pin verified against spec frag-6-7 §6.8 table (lines 248/251/252): Migration=100 > MigrationFailed=85 > Unreachable=70. Pinned values match.
- Lifeline carve-out: I1 lines 1326-1332 + I4 lines 1485-1489 iterate the actual SUPERVISOR_RPCS export AND I5 line 1526-1530 asserts the literal `'/healthz'` flows during apex. The literal-string assertion is what catches allowlist shrinkage — confirmed by mutation C above (only I5 caught it; I1/I4 vacuously pass with shorter list, see NIT).

### P0 / P1 / P2 / NIT
- P0: none.
- P1: none.
- P2: none.
- NIT (non-blocking): I1 + I4 use `for (const rpc of SUPERVISOR_RPCS)` which silently shrinks with the constant — a regression that REMOVES an entry (vs a regression that breaks gating logic) is only caught by I5's literal `'/healthz'` assertion. Could optionally pin `expect(SUPERVISOR_RPCS).toEqual(['/healthz', '/stats', 'daemon.hello', 'daemon.shutdown', 'daemon.shutdownForUpgrade'])` in the constants-pin test (line 1297), but the I5 literal already satisfies the spec-pin contract — fold to a follow-up if the allowlist ever grows.

PR #215, opened 2026-04-24
fix(a11y): announce Test notification result via aria-live (audit SD6)

Jiahui-Gu:
## Summary
- SettingsDialog's Notifications tab updates a transient status string (Sent / Failed / IPC unavailable) on the Test notification button. Previously it lived in a plain `<span>` that only mounted while the message was visible — no `role="status"`, no `aria-live`, so screen readers never announced the outcome.
- Render the status span unconditionally with `role=\"status\"`, `aria-live=\"polite\"` and `aria-atomic=\"true\"`. Empty content when idle keeps the visual layout untouched.
- Adds `tests/settings-test-notification.test.tsx` covering both success and failure paths through the same live region.

## Before / after

```jsx
// before
{testStatus && <span className=\"text-meta text-fg-secondary\">{testStatus}</span>}

// after
<span
  role=\"status\"
  aria-live=\"polite\"
  aria-atomic=\"true\"
  data-testid=\"notifications-test-status\"
  className=\"text-meta text-fg-secondary\"
>
  {testStatus ?? ''}
</span>
```

## Test plan
- [x] `npm test` — 662 passed, 51 files green.
- [x] New test asserts attrs are present both before and after click, and covers `notify -> true` (Sent.) and `notify -> false` (Failed - OS notifications unavailable.) via the same live region.
- [ ] Manual SR pass (VoiceOver / NVDA): open Settings → Notifications → press Test notification → SR announces "Sent." (or failure copy).

## Notes
- DO NOT MERGE — task spec asks for PR-only handoff.
- No new UI strings; existing i18n copy is already sentence case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #227, opened 2026-04-24
i18n(electron): localize tray menu and tooltip (#271)

Jiahui-Gu:
## Summary

Closes Wave 2 P0 W2-04 (task #271). The tray context menu still shipped English literals at `electron/main.ts:393,409,412` even though `electron/i18n.ts` had been added for OS notifications.

- Extend `electron/i18n.ts` with a `tray` namespace and a new `tTray()` helper alongside `tNotification()`.
- Mirror the same three keys into `src/i18n/locales/{en,zh}.ts` under `tray:` so the renderer catalog stays the source of truth and the i18n key-parity test passes.
- Replace the literals in `ensureTray()` with `tTray('show' | 'quit' | 'tooltip')`. Refactored the click handler out so the menu can be rebuilt without leaking listeners.
- **Runtime relocalization**: `ipcMain.on('ccsm:set-language', ...)` now also calls `applyTrayLocale()` so switching language in Settings rebuilds the tray menu immediately — no restart needed. Verified by reading the existing handler at `electron/main.ts:527`.

## Keys added

| Namespace | Key | en | zh |
|---|---|---|---|
| `tray` | `show` | Show CCSM | 显示 CCSM |
| `tray` | `quit` | Quit | 退出 |
| `tray` | `tooltip` | CCSM | CCSM |

## Test plan

- [x] `npx vitest run electron/__tests__/i18n.test.ts tests/i18n-key-parity.test.ts` — 8/8 pass (added 2 new tests for tray strings, both en + zh, with a regex Chinese-character spot check).
- [x] `npx tsc -p tsconfig.electron.json --noEmit` — clean.
- [ ] Manual: build, switch language in Settings, right-click tray icon, confirm menu items + tooltip update without restart.

## Notes

- Did NOT merge — leaving for reviewer per task instructions.
- Tray icon is created once in `app.whenReady()`, after the boot-time `setMainLanguage(resolveSystemLanguage(...))` seed, so the first-paint case already works without the IPC path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #751, opened 2026-05-01
spike(v0.4 T04): pkg ESM-interop verdict — NO-GO direct, GO via esbuild bundle

Jiahui-Gu:
## Verdict

**NO-GO direct, GO via `esbuild → CJS bundle → pkg`** two-stage pipeline. **T05 unblocked.**

## Test setup

Throwaway scaffold at `daemon/spike-pkg-esm/` (kept in PR for reproducibility, not wired into daemon build):

- `src/entry.ts` imports `CcsmService`, `file_ccsm_v1_core`, `file_ccsm_v1_pty` from a verbatim copy of `gen/ts/ccsm/v1/`, prints runtime descriptors so a wrongly-stubbed pkg snapshot would crash, not silent-pass.
- `tsconfig.json` emits ESM JS into `dist/` (matches the real proto-gen surface).
- Both `pkg dist/entry.js` (Path A — direct) and `esbuild → cjs → pkg` (Path B — fallback) tested on Win 11 host, `@yao-pkg/pkg@6.19.0` (bundled Node 22.22.2).

## Path A — pkg directly on ESM emit (FAIL)

Build emits 9 warnings:
```
> Warning Cannot find module '@bufbuild/protobuf/codegenv2' from '...\dist\gen-v1' in ...\dist\gen-v1\core_pb.js
... (9x)
```

Runtime crash:
```
Error: Cannot find module '@bufbuild/protobuf/codegenv2'
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1850:46)
  code: 'MODULE_NOT_FOUND',
  requireStack: ['C:\snapshot\dist\gen-v1\core_pb.js', ...],
  pkg: true
```

**Root cause:** `yao-pkg/pkg#215` (filed Feb 2026, **unresolved**). Three compounding bugs in pkg's resolver: result discarded when isESM-gated, `resolve` fallback ignores `exports`, synthetic-main patching skips subpath conditions. `@bufbuild/protobuf` only exposes `./codegenv2` via `package.json#exports` so pkg cannot find it.

## Path B — esbuild → CJS bundle → pkg (PASS)

```bash
npx esbuild dist/entry.js --bundle --platform=node --target=node22 --format=cjs --outfile=dist-bundle/entry.cjs
# 133 KB single CJS file, 220 ms

npx pkg dist-bundle/entry.cjs --targets node22-win-x64,node22-macos-x64,node22-linux-x64 --output out/spike
# 3 binaries, 57-75 MB each, no warnings

./out/spike-win.exe
[spike-pkg-esm] OK {"serviceTypeName":"ccsm.v1.CcsmService","serviceMethodCount":46,"coreFile":"ccsm/v1/core.proto","ptyFile":"ccsm/v1/pty.proto"}
```

esbuild fully resolves `exports` subpaths and inlines `@bufbuild/protobuf/codegenv2` into the bundle, removing pkg's dependency on its own broken resolver.

## Recommendation

Adopt **Fallback A: `esbuild → CJS bundle → pkg`** for the daemon build. Adds ~1-2 s to CI; keeps `frag-11-packaging.md` §11.1 native-rebuild + asset-copy pipeline unchanged in shape; source-map support via `--sourcemap` mitigates pkg's lossiness; defers Node 22 SEA migration cleanly to v0.5.

Rejected alternatives (full table in spike report):
- Bare ESM ship: breaks single-file-per-OS invariant.
- Node 22 SEA: strategically right but requires native-rebuild pipeline rewrite + fresh signing story; out of scope for v0.4 M1.
- Regenerate-at-install: strictly worse than Fallback A.
- Wait for pkg #215: unresolved 2.5 months on a fork repo with no assignee; cannot block M1.

T05 worker writes daemon code unchanged; the build-script wiring (esbuild step, externals for native `.node` modules, `daemon/package.json#pkg.scripts` pointer) is owned by a follow-up T-pkg-build task.

## Honesty caveats

- macOS/Linux pkg artifacts produced on Win 11 host (cross-target build) but **not executed**; CI matrix legs verify natively.
- Native modules (`better-sqlite3`, `node-pty`, `ccsm_native.node`) not exercised in this spike (entry uses only protobuf surface). Native-loading path is already proven in v0.3 and Fallback A leaves it untouched.
- Local Node v24; pkg's bundled Node v22.22.2 (matches v0.4 daemon target).

## Files

- `docs/spikes/2026-05-pkg-esm-connect.md` — full spike report.
- `daemon/spike-pkg-esm/` — repro scaffold (build artifacts gitignored).
- `docs/superpowers/specs/2026-05-01-v0.4-web-design.md` — DAG T04 verdict line appended.

## Verification

- `npm run typecheck` clean (spike has its own tsconfig, separate from root/electron/daemon configs).


Jiahui-Gu (commented):
**Verdict: APPROVE** — spike report is rigorous, verdict is actionable, T05 cleanly unblocked.

- **`yao-pkg/pkg#215` confirmed OPEN.** Filed 2026-02-13, last activity 2026-04-03 (PR #227 submitted by external contributor, no MEMBER review since). 2.5 months unresolved on a fork repo with no assignee — "cannot block M1" rationale stands.
- **Failure mode is real, not strawman.** `daemon/spike-pkg-esm/src/gen-v1/core_pb.ts` and `service_pb.ts` are byte-identical to `gen/ts/ccsm/v1/*` (verified via `diff -q`); the failing import `import type { GenFile, GenMessage } from "@bufbuild/protobuf/codegenv2"` is the same path the production proto-gen surface uses. `@bufbuild/protobuf@2.12.0` does expose `./codegenv2` only via `package.json#exports` — pkg #215 directly applies.
- **Fallback rationale survives spot-check.**
  - **SEA defer:** correctly cites `frag-6-7-reliability-security.md` line 348 ("True reproducibility deferred to v0.4 native SEA"). Two-step migration (rewrite native pipeline + signing story) is genuinely larger than M1 budget.
  - **Bare ESM:** correctly cites `frag-11-packaging.md` §11.1 single-file invariant.
  - **Wait-for-fix:** PR #227 exists but unmerged with no MEMBER engagement → reasonable to not bet M1 on it.
  - **Regenerate-at-install:** spec §5 already names this as the ultimate fallback; esbuild is strictly better (smaller footprint, no end-user `npm install`).
- **T05 verdict is actionable.** Spike explicitly says T05 worker writes daemon code unchanged; build-script wiring is owned by a follow-up T-pkg-build task. Wiring-plan section 4 gives the next worker concrete `external: ['better-sqlite3', 'node-pty']` + `pkg.scripts` pointer, so M1 can proceed.

## Layer 2 — implementation / report quality

- **Reproducibility is solid.** README + report both contain copy-paste commands. `tsconfig.json`, `package.json`, `entry.ts`, `gen-v1/` all present; `.gitignore` properly excludes `dist/`, `dist-bundle/`, `out/`, `node_modules/`, `package-lock.json`. Fresh dev can `cd daemon/spike-pkg-esm && npm install && npm run build && npm run pkg:host` and see the same FAIL.
- **Error output captured verbatim.** Build warnings (9x `Cannot find module`) and runtime crash (`requireStack`, `pkg: true`) both reproduced with exact message + path. Root-cause traced into the three pkg internals (`follow.ts` isESM gate, legacy resolve fallback, synthetic-main patching) — symptom not just confused for the cause.
- **Cross-target honesty present.** §3 explicitly flags: "macos/linux artifacts produced... cannot be executed on Win 11 host but were verified by file size + pkg exit code 0. CI matrix legs (per `frag-11-packaging.md` §11.5 step 2) will run each platform's artifact natively." Native modules caveat also called out (`better-sqlite3` / `node-pty` / `ccsm_native.node` not exercised; v0.3 path proven).
- **DAG verdict line is correct.** `docs/superpowers/specs/2026-05-01-v0.4-web-design.md` diff appends one line at L2889 under T04 with verdict + blocker reference + cross-target caveat + T05-unblock signal. Matches report TL;DR exactly.
- **File scope clean.** 17 files, all in scope: 8 proto stubs (verbatim copy, justified in §1), barrel `index.ts`, `entry.ts` (real runtime descriptor exercise — not type-only, would crash on stub), `package.json`, `tsconfig.json`, `.gitignore`, `README.md`, spike report, spec verdict line. No leftover noise.
- **Entry exercises real surface.** `entry.ts` reads `CcsmService.typeName`, `Object.keys(CcsmService.method).length`, `file_ccsm_v1_core.proto.name`, `file_ccsm_v1_pty.proto.name` — all of which require the codegen runtime descriptors to actually load. A pkg snapshot that wrongly stubbed `@bufbuild/protobuf/codegenv2` would crash at `CcsmService.typeName` lookup, not silent-pass.

## P0 / P1 / P2

- **P0:** none.
- **P1:** none.
- **P2 (defer, can fold into T-pkg-build follow-up):**
  1. README mentions `npm run pkg` but `package.json` only defines `pkg:host` and `pkg:all` — minor doc/script naming drift; reproducer still works if reader scans the script section. Fix during T-pkg-build wiring.
  2. `entry.ts` comment references `daemon/spike-pkg-esm/scripts/sync-gen.mjs` which doesn't exist in the PR — the verbatim copy was clearly done manually. Either drop the comment or add the script. Cosmetic.
  3. Source-map `--sourcemap` is mentioned in the recommendation but not exercised in the spike itself. Acceptable for a one-day spike; T-pkg-build should validate that pkg actually preserves esbuild source-maps end-to-end.

Approving merge. Spike answered the question, the answer is actionable, and the reproducer survives a fresh dev clone.

PR #752, opened 2026-05-01
feat(v04 T05): bind Connect server on daemon data socket

Jiahui-Gu:
## Summary

Lands the Connect-RPC server scaffold + interceptor stack on the daemon data socket, dormant — no service handlers registered yet (T06 lands the first; T08 lands the real JWT). Covers Task #1082.

Spec citations: `docs/superpowers/specs/2026-05-01-v0.4-web-design.md` ch02 §1 (Connect lock), ch02 §6 (data socket = HTTP/2 + Connect; control socket stays envelope), ch02 §8 (interceptor inheritances from v0.3), ch05 §4 (transport-tag positive enum, fail-closed), ch05 §5 (chain order lock).

## What landed

- `daemon/src/connect/server.ts` (~410 LOC) — `createConnectDataServer` factory exposing:
  - HTTP/2 server lifecycle (`listen`, `close`).
  - `attachSocket(socket, transportType)` for future data-socket coexistence wiring (HTTP/2 preface peek → here, else → envelope).
  - Spec-locked interceptor chain (6 slots): `transport-tag → jwt → migration-gate → hello-gate → deadline → trace-id`.
  - Public context keys: `transportTypeKey` (positive enum, no header reads — defends against ch05 §4 spoof attack), `helloAckKey` (handshake completed flag), `traceIdKey` (per-request ULID).
- Pre-handshake gate: any RPC except `['DaemonHello', 'Ping']` rejects with `Code.Unauthenticated` until `helloAckKey === true` on the per-connection context. Mirrors v0.3 envelope `helloInterceptor` slot #0 + ch02 §7 `Ping` carve-out.
- JWT placeholder: `local-pipe` transport bypasses; `remote-tcp` and untagged contexts fail-close with `Unauthenticated`. Marked `TODO(T08)` for real `jose.jwtVerify` per ch05 §4 / §4.1.
- Migration-gate: blocks data-plane RPC with `Code.FailedPrecondition` while `isMigrationPending()` is true (caller injects the predicate).
- Trace-id: generates ULID per request and stamps it on `contextValues`.
- `daemon/src/connect/__tests__/server.test.ts` (~250 LOC, 13 tests) — server lifecycle, chain order, hello-gate accept/reject, migration-gate, JWT bypass, trace-id propagation.
- `daemon/package.json`: added `@connectrpc/connect@^2.1.1` + `@connectrpc/connect-node@^2.1.1`.

## What's stubbed for follow-ups

- **T06 (#?)**: register `CcsmService` (Ping + DaemonHello) via `registerRoutes`, flip `helloAckKey` after handshake.
- **T08 (#?)**: replace `createJwtInterceptor` placeholder with real RS256 + JWKS pre-warm + verify cache per ch05 §4 / §4.1 policy locks.
- **T12 / T19 wave**: wire `attachSocket` into the data-socket dispatcher (HTTP/2-preface peek + route table). The data-socket listener itself isn't yet bound from `daemon/src/index.ts`, so live boot wiring isn't introduced here either — that belongs to T19 wave.
- **T08+ deadline interceptor**: header reader (`x-ccsm-deadline-ms`) + `AbortSignal.timeout` composition; current chain slot is a pass-through stub so the order test pins the slot.

## T03 tsconfig-paths fold

NOT done in this PR — daemon does not yet import `@ccsm/proto-gen` (no service registered until T06). T06 will add `paths: { "@ccsm/proto-gen": ["../gen/ts"], "@ccsm/proto-gen/*": ["../gen/ts/*"] }` to `daemon/tsconfig.json` when it lands the first `CcsmService` import. Mentioned per task brief.

## Build hygiene (T04 alignment)

- All imports use static ES module syntax (no `eval()` / no dynamic-require) so esbuild can resolve them cleanly during the eventual `esbuild → CJS bundle → pkg` pipeline (T-pkg-build follow-up territory).
- `@connectrpc/connect-node` and `@bufbuild/protobuf` use only their root entry points (no exports-subpath gymnastics) to stay inside pkg's resolver budget.

## Test plan

- [x] `npx vitest run daemon/src/connect/__tests__/server.test.ts` — 13/13 pass locally.
- [x] `npx eslint daemon/src/connect` — clean.
- [x] `npm run build:daemon` — clean (`tsc -p daemon/tsconfig.json`).
- [x] ESM smoke load on the compiled output: `import('./daemon/dist-daemon/connect/server.js')` resolves with all 7 named exports; `createConnectDataServer({...}).handler` is a function.
- [ ] CJS `require('./dist/daemon/connect/server.js')` smoke — daemon emits ESM (`"type":"module"`); the brief's CJS smoke applies once the T-pkg-build esbuild→CJS bundle lands. Out of scope per task brief don'ts ("Don't touch the build script in this PR").
- [ ] Pre-existing `better-sqlite3` NODE_MODULE_VERSION mismatch (137 vs 145) causes 34 unrelated daemon-test failures in this worktree; not caused by this PR.


Jiahui-Gu (commented):
**Verdict: APPROVE-WITH-NITS** — 13/13 tests green, typecheck/build clean, divergences validated and reasonable. Spec drift on the interceptor stack content + file layout warrants P1 follow-ups (not blockers in T05's "dormant scaffold" framing).

**Divergence 1: no live wiring in `daemon/src/index.ts` / `data-socket.ts` — VALID.** Verified on `origin/working`: `daemon/src/sockets/data-socket.ts` exists as pure transport (Duplex hand-off), `daemon/src/index.ts` has no `createConnectDataServer` or `attachSocket` reference, and the spec's T19 wave (HTTP/2-preface peek + dispatch) is not yet present. Worker's claim holds. Note: spec T05 file list includes `daemon/src/sockets/data-socket.ts` (wire-in) + `daemon/src/sockets/runtime-root.ts` (transport tag) — those edits were skipped, see P1 #5 below.

**Divergence 2: ESM dynamic-import smoke instead of CJS require — VALID.** `daemon/package.json` is `"type": "module"`; `npm run build:daemon` emits ESM into `dist-daemon/`. CJS `require()` smoke would fail against the actual build output. Brief's CJS smoke applies post `T-pkg-build` (esbuild→CJS bundle).

**Divergence 3: T03 tsconfig-paths fold skipped — VALID.** `server.ts` imports nothing from `@ccsm/proto-gen` (no service registered until T06). Alias unnecessary now; T06 (#1087) will add it.

**Direction checks:**

- **Chain order vs spec ch05 §5.** Spec ch05 §5 (line 3144) locks: `transport-tag → JWT → migration-gate → storage-full → deadline → trace-id → handler`. PR ships: `transport-tag → JWT → migration-gate → hello-gate → deadline → trace-id`. Two deltas: (a) **storage-full slot missing** — spec ch07 §1 calls for short-circuit interceptor mapping `SQLITE_FULL` → `resource_exhausted`; not in PR. (b) **hello-gate slot inserted** between migration-gate and deadline — not in the canonical ch05 §5 ordering. Hello-gate is justified by ch02 §8 v0.3-inheritance row (`helloInterceptor`) but the spec doesn't lock its position; placement here is a worker judgment call. P1: reconcile. Either add storage-full slot + push hello-gate later, or get an explicit spec waiver pinning storage-full to T15. Cite ch05 §5 in the `buildInterceptorChain` JSDoc.

- **`PRE_HANDSHAKE_ALLOWLIST = ['DaemonHello', 'Ping']` — Ping carve-out.** Spec ch02 §7 line 357 and §6 line 376 explicitly call out `Ping` as a pre-connect handshake RPC ("clients call it on connect"). Allowlist is correct. v0.3 envelope only allowed `daemon.hello`; v0.4 widening is intentional per spec.

- **JWT placeholder posture.** Spec ch05 §4 (line 1174): `'local-pipe'` bypass; untagged or `'remote-tcp'` fail-closed `Unauthenticated`. PR matches verbatim. T08 (#1085) lands real `jose.jwtVerify`. Posture correct.

## Layer 2 — implementation / quality

- **13 vitest cases — coverage adequate but order asserted via SHADOW chain.** The "runs interceptors in spec-locked order" test (server.test.ts line 60) builds a fresh tagged chain instead of instrumenting the real `buildInterceptorChain()` output. Order is asserted on the shadow; the real chain is only checked for `.length === 6`. P2: replace shadow-chain order test with one that wraps each real interceptor (or inserts a tap interceptor between them) so the real `buildInterceptorChain` order is the assertion target. Today, swapping `migration-gate` and `hello-gate` positions in `server.ts` would NOT fail any test.

- **Pre-handshake reject coverage.** Tested for hello-gate in isolation (line 101) and migration-gate in isolation (line 158). Not tested through the full `connectNodeAdapter` over a real HTTP/2 socket. Acceptable for "dormant scaffold" stage; T06 will exercise the full path.

- **Context keys — clean, no leakage.** Three keys (`transportTypeKey`, `helloAckKey`, `traceIdKey`) all use `createContextKey` with descriptive metadata; all exported for downstream module consumption. Single-source-of-truth posture is good.

- **Migration-gate "live" + deadline "stub" labeling.** Migration-gate is the real predicate (caller-injected `isMigrationPending()`); deadline is a pass-through with explicit `TODO(T08+)` JSDoc + inline comment "ships the chain slot only". Trace-id is also live (generates ULID). Stub vs live is clearly marked, no silent skip risk.

- **`@connectrpc/connect-node@^2.1.1` pinning.** Pinned in `daemon/package.json` only (not root). Correct per workspace boundary.

- **Pre-existing better-sqlite3 NODE_MODULE_VERSION mismatch.** Confirmed pre-existing: 34/84 fail on `daemon/src/db/__tests__` against `origin/working` HEAD too, identical signature (NODE_MODULE_VERSION 137 vs 145). NOT caused by this PR.

- **Spec T05 scope deltas (worth flagging beyond worker's 3 divergences):**
  - **`readMaxBytes` per-route caps.** Spec T05 scope explicitly: "`readMaxBytes` (per-route caps from chapter 02 §8 table)". Not in PR. Defensible because no routes are registered yet — but the per-route cap mechanism (`Db.save: 16 MiB`, `SendPtyInput: 1 MiB`, default 4 MiB, log-near-cap interceptor) needs to land before T06 wires real routes.
  - **Pre-accept rate cap (50/sec).** Spec T05 scope: "pre-accept rate cap" via `Http2Server.on('session', ...)`. Not in PR.
  - **Structured pino logging.** Spec T05 scope: "structured pino logging" + ch02 §8 paragraph 394 "every interceptor that rejects ... emits `pino.warn({interceptor, method, traceId, reason})`". The PR's interceptors throw `ConnectError` but log nothing.
  - **File layout.** Spec T05 prescribes `daemon/src/connect/interceptors/{trace-id,deadline,migration-gate,read-max,rate-cap}.ts` (separate files per interceptor); PR consolidates into one 413-LOC `server.ts`. Not strictly wrong (each function still single-responsibility) but diverges from spec layout.

## P0 / P1 / P2

**P0**: none.

**P1**:
1. Reconcile chain order vs ch05 §5 — either add storage-full slot or document the exemption with a citation in `buildInterceptorChain` JSDoc.
2. Add `pino.warn({interceptor, method, traceId, reason})` rejection logs to hello-gate / jwt / migration-gate before T06 wires real routes — spec ch02 §8 paragraph 394 makes this MUST.
3. Land `readMaxBytes` per-route cap interceptor + the "log within 10% of cap" mechanism before T06.
4. Land pre-accept rate cap (50/sec on `Http2Server.on('session')`) before T06.
5. T05 spec scope says wire `daemon/src/sockets/data-socket.ts` + `daemon/src/sockets/runtime-root.ts` (transport tag). Skipped here, deferred to T19 wave — but T06 will need the local-pipe transport tag stamped on real sockets to exercise the JWT bypass path. Either fold runtime-root tagger into T05 follow-up or carve out an explicit T05.5 task before T06 dispatch.

**P2**:
6. Order test runs against a shadow chain, not the real `buildInterceptorChain` output. Reordering interceptors in `server.ts` would not fail. Replace with a test that wraps each real interceptor with a tap.
7. Spec layout intends `daemon/src/connect/interceptors/*.ts` per-file split; PR ships one consolidated file. Defensible at 6 functions × ~15 LOC each, but T08+ will add JWT/JWKS/storage-full/read-max/rate-cap; consider splitting now to avoid a noisy refactor PR later.
8. `lazyCreateContextValues()` indirection (lines 405-409) and the dangling `void universalRequestFromNodeRequest` at line 413 are odd. The lazy wrapper is just a function call wrapping `_createContextValues()` with no laziness benefit. The `void` is for a treeshake hack with a stale comment. Both can be removed.
9. `attachSocket` semantics when called before `listen()`: silently stashes the tag without ever delivering to a server, with a "T19 wave will lock this down" comment. Minor footgun; either error or queue + replay on `listen()`.

## Rationale

T05 lands a clean dormant scaffold with the right primitives (3 context keys, hello/JWT/migration deciders all spec-locked, ULID trace-id live). The 3 worker-flagged divergences all check out against `origin/working` HEAD + `daemon/package.json`. Tests pass 13/13, typecheck clean, build clean. The bigger spec drift is on the interceptor *content* (missing storage-full slot per ch05 §5; missing readMaxBytes / rate-cap / pino reject logs per T05 scope) — these are P1s that need to land before T06 registers any real route, but don't block this PR's merge given its explicit "dormant, no services registered" framing. Approve with the P1 list folded into a T05.1 follow-up task before T06 dispatch.

PR #754, opened 2026-05-01
feat(v04 T07): parity-test framework (M1)

Jiahui-Gu:
## What landed

Task #1086 — v0.4 T07 L6 "parity-test framework (M1)". The framework that, for every Connect RPC, will verify the new bridge produces equivalent results to the v0.3 envelope-over-named-pipe handler.

Per spec ch03 §7.1, this is M1's load-bearing safety net for the 12 bridge-swap PRs that follow. Locking the framework here prevents each later PR from inventing its own definition of "equivalent" and producing a non-uniform safety net.

**Surface:**

- `runParityCase({ name, envelopeCall, connectCall, equivalence?, tolerantFields?, coerce? })` — vitest-compatible helper. Runs both transports in parallel; throws `ParityDivergenceError` with side-by-side JSON diff on mismatch.
- `assertParity(envelope, connect, opts?)` — low-level deep-equal with `ignoreFields` + per-field `coerce`. Matches the spec's `assertParity` shape.
- `defaultEquivalence` — `node:util.isDeepStrictEqual` under `ignoreFields`/`coerce` normalization. Single decision point per producer/decider/sink discipline.
- `transports.ts` — typed `EnvelopeClient` + Connect transport factory pointing at a daemon Connect server. Ships `stubEnvelopeClient` for the placeholder; concrete `envelopeClient` factory deferred until the daemon test harness stabilizes (T09 territory).

**Files:**

- `tests/parity/framework.ts` — runner + assertion + diff renderer (no external deps; `diff`/`json-diff` not in lockfile, deferred until M2 PR finds the diff hard to read)
- `tests/parity/framework.test.ts` — 18 unit tests
- `tests/parity/ping.parity.test.ts` — 1 placeholder fixture case
- `tests/parity/__fixtures__/ping.golden.json` — fixture in spec's `{ name, request, expectedV03Response, expectedV04Response, parityOpts }` shape
- `tests/parity/transports.ts` — Connect transport + stub envelope client
- `.github/workflows/parity.yml` — runs `npm run test:parity` on PRs touching `tests/parity/**`, daemon Connect/envelope, preload bridges, or `proto/**`. Independent of `ci.yml` (which is `if: false` for the v0.3→v0.4 migration window) so the safety net stays on.
- `package.json` — adds `npm run test:parity`

## Placeholder fixture explanation

The placeholder `tests/parity/ping.parity.test.ts` exercises `runParityCase` end-to-end with stub responses on both transports — the same recorded golden response on both sides, demonstrating the equivalence path passes. It does NOT spin up a daemon, does NOT call any real envelope sender or Connect client.

T06 (next PR) registers `Ping` + `GetVersion` Connect handlers and will replace this stub with a real case that boots the daemon, calls envelope `daemon.ping` and Connect `Ping`, and asserts equivalence under the framework. Until then this placeholder ensures the framework's plumbing is exercised by ≥1 case so `npm run test:parity` does not no-op.

The fixture file shape (`{ name, request, expectedV03Response, expectedV04Response, parityOpts }`) is the canonical shape from spec ch03 §7.1 — the loader is exercised here so it remains under test as the corpus grows in M2.

## Reverse-verify output

Captured locally per ch08 §8 reverse-verify discipline. Demonstrates the equivalence fn is the gating mechanism.

**Test 1 — divergent inputs WITHOUT override (expect FAIL):**

```
ParityDivergenceError: parity divergence in demo-divergence
envelope (v0.3):
  {
    "a": 1,
    "b": "x",
    "meta": { "version": "v0.3" }
  }

connect  (v0.4):
  {
    "a": 2,
    "b": "x",
    "meta": { "version": "v0.4" }
  }

differing fields:
- a: 1 vs 2
- meta.version: "v0.3" vs "v0.4"
```

→ FAILED as expected.

**Test 2 — same divergent inputs WITH `equivalence: () => true` override (expect PASS):**

→ PASSED as expected (override neutralized the gate). This is captured in the test `forces a divergent case to PASS when equivalence is overridden to always-true (reverse-verify)` and runs green in `npm run test:parity`.

## Test summary

- Local `npm run test:parity` (Windows 11): **2 files passed, 19 tests passed** in 2.6s
- Daemon Connect tests (`daemon/src/connect/__tests__`) re-run as smoke: **1 file, 13 tests passed** — no regression from T05.
- `npx tsc --noEmit -p tsconfig.json`: clean
- Lint: pre-existing repo errors unchanged; `tests/**` is globally eslint-ignored per `eslint.config.js`.

## T05 / T05.1 coordination note

Per task brief: T05 (PR #752) merged the dormant `createConnectDataServer` factory at `daemon/src/connect/server.ts` — no service handlers, no listener wiring on the data socket. T05.1 (the wiring task) has not yet landed.

This framework is **transport-agnostic** by design — `runParityCase` accepts arbitrary `envelopeCall` / `connectCall` async functions. Test files that need a real Connect round-trip will boot their own Connect server in `beforeAll` (per the existing `daemon/src/connect/__tests__/server.test.ts` pattern) until T05.1 lands the canonical socket wiring. The framework itself does NOT need T05.1 — only T06's first real RPC case does.

`tests/parity/transports.ts` exports `createConnectTransportForDaemon({ baseUrl })` for that boot pattern. Once T05.1 lands and exposes a stable `socketPath` form, that helper grows a path; today it speaks HTTP/2 over loopback TCP, matching the T05 server scaffold.

## CI workflow change summary

New: `.github/workflows/parity.yml`

- Triggers on PRs touching: `tests/parity/**`, `daemon/src/connect/**`, `daemon/src/envelope/**`, `electron/preload/bridges/**`, `proto/**`, `gen/**`, `package.json`, `package-lock.json`, the workflow file itself.
- Matrix: `ubuntu-latest` + `windows-latest` (developer's primary platform). macOS rides on the post-migration `ci.yml`.
- `continue-on-error: false` (default) — placeholder case ensures the step doesn't no-op; once T06 lands, divergence blocks the PR.
- Why dedicated workflow vs folding into `ci.yml`: `ci.yml` is `if: false` for the v0.3→v0.4 migration window per Task #1048. Parity tests must run on every PR during exactly that window — that's the whole point of the safety net. Dedicated workflow keeps the gate always on with zero scheduling logic.

## Spec deviations (flagged for reviewer)

1. **Test location:** spec says `daemon/__tests__/parity/`; PR uses `tests/parity/`. Per task brief: parity is a cross-cutting concern (daemon + electron + protocol), and `tests/` already hosts other cross-layer tests (`electron-load-smoke.test.ts`, `persist-shape.test.ts`). Manager confirmed at task #1086 dispatch.
2. **`scripts/gen-parity-scaffold.ts` deferred:** the codegen-driven scaffold from spec §7.1 (enforces "every swapped RPC has a parity-corpus entry") is meaningful only once T06 lands the first real RPC. Adding it before T06 would either no-op or fail spuriously. Should be added in any pre-T09 PR.
3. **`RECORD_PARITY=1` deferred:** same reason — fixture recording from the v0.3 envelope path requires real envelope handlers to record from. T06 + T09 add this.

## Test plan

- [x] `npm run test:parity` — 19/19 pass
- [x] Reverse-verify: divergent inputs without override → FAIL; with `equivalence: () => true` → PASS
- [x] Daemon Connect tests still pass (no regression from T05)
- [x] Typecheck clean
- [ ] CI workflow runs green on the PR

### DIFF
--- a/.github/workflows/parity.yml
+++ b/.github/workflows/parity.yml
@@ -1,0 +1,92 @@
+name: parity
+# T07 — parity-test framework workflow.
+#
+# Spec: docs/superpowers/specs/2026-05-01-v0.4-web-design.md
+#   - ch03 §7.1 (parity tests are the load-bearing safety net during M1's 12 PRs)
+#   - ch08 §3   (L2 testing layer)
+#   - ch08 §9.1 (CI matrix; this is the parity-specific job)
+#   - ch09 §2 T07 deliverable
+#
+# What this runs: `npm run test:parity` (= `vitest run tests/parity`).
+# In M1 the suite contains the framework's own unit tests + one placeholder
+# fixture exercise. T06 + Batch A (T09) bridge swaps add real cases.
+#
+# Why a dedicated workflow rather than folding into ci.yml: ci.yml is gated
+# `if: false` for the v0.3 → v0.4 migration window (Task #1048). Parity tests
+# are the safety net during exactly that window — they MUST run on every PR
+# even while ci.yml is dark. Folding into ci.yml would require either un-
+# gating ci.yml early (Task #1048 territory) or hand-tracking which PRs
+# touched the migration path. Dedicated workflow keeps the safety net always
+# on with zero scheduling logic.
+#
+# `continue-on-error: false` (the default) is intentional: per the brief,
+# the placeholder ensures the step doesn't no-op, and once T06 lands real
+# cases any divergence must block the PR.
+
+on:
+  pull_request:
+    branches: [main, working]
+    paths:
+      - 'tests/parity/**'
+      - 'daemon/src/connect/**'
+      - 'daemon/src/envelope/**'
+      - 'electron/preload/bridges/**'
+      - 'proto/**'
+      - 'gen/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/parity.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: parity-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  parity:
+    name: parity (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        # Both daemon + Electron run on Linux (CI default) and Windows
+        # (primary developer platform per CLAUDE.md). macOS coverage rides
+        # on the post-migration ci.yml when restored.
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      # Same python pin as ci.yml — better-sqlite3 postinstall imports
+      # distutils via the older node-gyp.
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Cache node_modules
+        id: nm_cache
+        uses: actions/cache@v4
+        with:
+          path: node_modules
+          key: nm-${{ runner.os }}-${{ runner.arch }}-node20-${{ hashFiles('package-lock.json') }}
+
+      - name: Install deps
+        if: steps.nm_cache.outputs.cache-hit != 'true'
+        run: npm ci --legacy-peer-deps
+        env:
+          ELECTRON_SKIP_BINARY_DOWNLOAD: '1'
+
+      # Rebuild better-sqlite3 against Node ABI (vitest runs under Node).
+      - name: Rebuild better-sqlite3 for Node
+        run: npm rebuild better-sqlite3
+
+      - name: Run parity tests
+        run: npm run test:parity
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "lint:fixtures": "tsx scripts/lint-fixtures.ts",
     "lint:schema-additive": "tsx scripts/lint-schema-additive.ts",
     "test": "vitest run",
+    "test:parity": "vitest run tests/parity",
     "check:electron-pin": "node scripts/check-electron-pin.cjs",
     "pretest": "node scripts/check-electron-pin.cjs",
     "test:watch": "vitest",
--- a/tests/parity/__fixtures__/ping.golden.json
+++ b/tests/parity/__fixtures__/ping.golden.json
@@ -1,0 +1,16 @@
+{
+  "rpc": "Ping",
+  "description": "Placeholder parity fixture exercised by tests/parity/ping.parity.test.ts. Replaced by a real recorded fixture in T06 once Ping is registered on Connect (per spec ch09 §2 T06). Until then the test uses manual stubs for both transports to prove the framework plumbing end-to-end.",
+  "cases": [
+    {
+      "name": "ping echoes the requested timestamp",
+      "request": { "clientNowMs": 1700000000 },
+      "expectedV03Response": { "ok": true, "echoNowMs": 1700000000 },
+      "expectedV04Response": { "ok": true, "echoNowMs": 1700000000 },
+      "parityOpts": {
+        "ignoreFields": ["serverNowMs", "traceId"],
+        "normalizationStrategy": "structural"
+      }
+    }
+  ]
+}
--- a/tests/parity/framework.test.ts
+++ b/tests/parity/framework.test.ts
@@ -1,0 +1,225 @@
+// T07 — parity-test framework unit tests.
+//
+// Spec: docs/superpowers/specs/2026-05-01-v0.4-web-design.md
+//   - ch03 §7   (test discipline per swap PR — parity test required per RPC)
+//   - ch03 §7.1 (parity-test framework shape: assertParity + ignoreFields + coerce)
+//   - ch08 §3   (L2 daemon Connect handler unit + contract tests)
+//   - ch08 §8   (reverse-verify discipline)
+//
+// What this file proves about the framework:
+//   1. `runParityCase` runs envelopeCall + connectCall in parallel and asserts
+//      the responses are equivalent under the supplied `equivalence` fn.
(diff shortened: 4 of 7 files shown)


Jiahui-Gu (commented):
**Verdict: APPROVE** — framework matches spec ch03 §7.1 surface, all 4 deferrals justified, 19/19 green on Win 11, reverse-verify confirms equivalence fn is the gate. No P0/P1. 4 P2 nits.

---

**Spec match.** `assertParity(envelopeResp, connectResp, opts: { ignoreFields, coerce })` matches spec ch03 §7.1 verbatim. `runParityCase` is a thin wrapper providing the parallel-invocation requirement that §7.1 calls out (fixture-pair → both transports → diff). API shape is faithful, not invented.

**Deviation 1 — `tests/parity/` location.** Defensible per task brief + cross-cutting precedent (`tests/electron-load-smoke.test.ts`, `tests/persist-shape.test.ts`). M2.Z deletion gate (spec §3 lines 2270-2277) does not depend on path. Accept.

**Deviation 2 — `scripts/gen-parity-scaffold.ts` deferred.** Acceptable — the scaffold's job is "fail PRs missing a parity-corpus entry per swapped RPC". Without any swapped RPC yet (T06 is next), it would no-op or block on stale state. Should land in T06 or as a separate gate-only PR before T09 (Batch A).

**Deviation 3 — `RECORD_PARITY=1` deferred.** Acceptable — recording requires a real v0.3 envelope path to record from. T06 + T09 supply that. Worker's reasoning is correct.

**Deviation 4 — concrete `envelopeClient(socketPath)` factory deferred.** **VERIFIED ACCEPTABLE.** Grepped `tests/`, `daemon/src/`, `electron/` for any pre-existing envelope test client (`envelopeClient`, `sendEnvelope`, `envelopeSender`) — only hit was the worker's own comment in `transports.ts`. `daemon/src/envelope/__tests__/` contains 11 unit tests but zero shared client helper. The defer claim ("daemon test harness in flux through T12/T19 wave") is true. T06 will need to construct one inline against its booted daemon fixture — that's the right shape.

**`parity.yml` independence.** Justified — `ci.yml` is `if: false` per #1048. Path filter (`tests/parity/**`, `daemon/src/connect/**`, `daemon/src/envelope/**`, `electron/preload/bridges/**`, `proto/**`, `gen/**`) is the correct scope. Concurrency cancel-in-progress + matrix (ubuntu+windows) sound. When #1048 lands, recommend folding into `ci.yml` to avoid duplicate node_modules cache work, but no need to gate this PR on it.

**Placeholder `ping.parity.test.ts`.** Honest stub-vs-stub: drives `runParityCase` end-to-end (loads fixture file, parallel calls both stub transports, asserts equivalence under `tolerantFields`). Does NOT prove wiring through real envelope/Connect — PR body is explicit about this. The fixture loader exercise (canonical `{ name, request, expectedV03Response, expectedV04Response, parityOpts }` shape) is genuinely valuable; it locks the corpus shape under test before T06 grows it.

## Layer 2 — implementation / quality

- **`npm run test:parity` on Win 11**: 2 files, 19/19 pass, 2.6s.
- **`npx tsc --noEmit`**: clean.
- **Reverse-verify (manual tamper)**: replaced `defaultEquivalence` body with `return true` → ran framework.test.ts → 6 divergence-detection tests FAILED as expected (`throws ParityDivergenceError when responses differ`, `failure message includes case name`, etc.). Restored → 19/19 green again. Equivalence fn is conclusively the gate, not a happy-path coincidence.
- **Built-in reverse-verify case** (`forces a divergent case to PASS when equivalence is overridden to always-true`) is the right durable proof — captures the discipline in the test file itself.
- **Parallel claim verified**: `runParityCase` uses `Promise.all([envelopeCall(), connectCall()])` (framework.ts:191-194). Timing test asserts <90ms for two 50ms calls. Real parallelism, not lying.
- **Tolerant fields**: top-level `tolerantFields: string[]` shorthand folds into `ParityOptions.ignoreFields`; nested dotted paths supported (`meta.traceId` test); `coerce` map covers per-field normalization. Matches spec §7.1's `coerce` + `ignoreFields` requirements.
- **No external deps**: imports are `node:util`, `node:fs/url/path`, `vitest`, `@connectrpc/connect-node` (already in lockfile from T05). No `diff` / `json-diff` smuggled in. Diff renderer is hand-rolled — single seam for future swap.
- **Diff output**: useful — `envelope (v0.3): {…} / connect (v0.4): {…} / differing fields: - <path>: a vs b`. Capped at 50 lines (no trailer message — see P2). Adequate for M1.
- **`createConnectTransportForDaemon`**: uses `@connectrpc/connect-node`'s `createConnectTransport` with `httpVersion: '2'` — HTTP/2 cleartext, baseUrl-driven (TCP loopback only). Comment honestly flags that named-pipe/socket-path support is T05.1 territory. Matches T05's own server-side pattern.
- **Workflow rebuild step**: `npm rebuild better-sqlite3` correctly addresses the Node-vs-Electron ABI mismatch for vitest under Node.

## P0 / P1 / P2

**P0**: none.
**P1**: none.

**P2 (defer to follow-up):**
1. **`safeClone` JSON fallback drops `undefined` / `Date` silently.** Fine for JSON-shaped Connect/envelope responses (ch02 §1) but worth a one-line guard in the comment that recording bridges with `Date`-typed fields must coerce them via `coerce` map first. (framework.ts:240-251)
2. **`collectFieldDiffs` 50-line cap is silent.** Add a `... (N more diffs truncated)` trailer when `out.length === 50` so a wholly-divergent response doesn't look like exactly-50-divergent. (framework.ts:341-370)
3. **No follow-up task ID for T05.1 socket-path wiring** in `transports.ts` comment. Once #1092 (T05.1) lands and exposes a stable `socketPath` form, `createConnectTransportForDaemon` should grow a `socketPath` branch — recommend a `// TODO(#1092)` marker so the seam isn't lost.
4. **`envelopeClient(socketPath)` factory defer**: similarly, no tracked task ID. Recommend opening a follow-up to land the concrete factory once T06 supplies the first real envelope sender pattern, so M2 bridge swap PRs (T09 Batch A) don't each invent their own.

None of these block merge. Framework is sound, deferrals are honest, reverse-verify is real. Recommend merge.

PR #760, opened 2026-05-01
feat(crash phase 5): rolling log retention + renderer error forwarder

Jiahui-Gu:
Task #1094. Phase 5 of the crash-observability rollout (spec §10/§11, plan phase 5 retention + renderer forwarder slice).

## A. Per-surface retention with protected-window

Rules (extends `pruneRetention` in `electron/crash/collector.ts`):

- `maxPerSurface=N` (default **20**, wired into `electron/main.ts` boot) — keep the N newest incidents per surface (read from each incident's `meta.json`). Surfaces today: `main`, `renderer`, `gpu`, `helper`, `daemon-exit`, `daemon-uncaught`, `daemon-boot-crash` (per phase 1/3 incident schema).
- `protectUnsentYoungerThanDays=7` — never prune an incident that has NO `.uploaded` marker AND whose mtime is younger than 7 days. Rationale: phase-4's "Send last crash" button needs the dir to still be on disk for the user to re-upload; without this guard a chatty surface could push a young unsent incident off the keep-N list before the user ever sees the modal.
- Legacy global `maxCount`/`maxAgeDays` rule preserved (still called from boot for backward-compat with the phase-1 contract). Both rules now compute a single `toDelete` set so the protect-window applies uniformly.
- Pruning still runs once at electron-main boot, before any new collector writes (race-safe).

Future configurability hook: `crashLogRetention.maxPerSurface` is the spec'd setting key; for now the default is hard-coded in `main.ts` and a follow-up can surface it in Settings if the user ever wants to tune it.

## B. Renderer error forwarder

- IPC channel **`crash:report-renderer-error`** (`electron/ipc/rendererErrorForwarder.ts`).
- Renderer hooks `window.onerror` + `window.onunhandledrejection` (`src/lib/installRendererErrorForwarder.ts`); installed in `src/index.tsx` after the renderer Sentry SDK init so both can coexist.
- `console.error` capture is **dev-only** (gated on `process.env.NODE_ENV !== 'production'`) — production stays focused on the two real signals so the rate limit isn't burned by noisy debug logs. Production captures only window.onerror + window.onunhandledrejection.
- Rate-limit: **≤ 10 events / 60 s per renderer process** (`webContents.id` bucket). Excess events are silently dropped; the dropped count is folded into the next accepted incident's `stderrTail` as `renderer-error-drops=N` so the signal is preserved across the rate-limit boundary.
- Main routing: `handleRendererErrorReport` calls the existing phase-1 `recordIncident({ surface: 'renderer', ... })`. **No new Sentry init paths** — the existing main-proc Sentry SDK auto-routes by surface tag, and consent gating remains entirely in phase-4 territory (`isCrashUploadAllowed`).
- Defense-in-depth: IPC handler rejects messages from non-mainFrame senders via the existing `fromMainFrame` guard.

## Hard-constraint compliance

- **Phase 4 consent gate untouched.** Renderer incidents always write to disk regardless of consent (verified by `tests/electron/crash/local-write-with-opted-out-renderer.test.ts`, mirror of the phase-4 main-surface test).
- **Phase 2/3 Sentry routing untouched.** No new SDK init paths; no DSN-selection changes.
- **Protect window** prevents the user's "send last crash" button from ever finding a missing dir within 7 days of the incident.

## Reverse-verify (protect-window rule)

Temporarily disabled the `if (protectUnsentYoungerThanDays != null && !isUploaded)` branch in `collector.ts`:

```
× protected window — 12 case: 2 unsent OLDER than keep window stay anyway 233ms
× protected-window applies even when surface is over count cap (regression for unsent-young rule) 38ms
Tests  2 failed | 3 passed (5)
```

Restored the branch:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

## Local-write-with-opted-out (renderer surface, phase-4 mirror)

`tests/electron/crash/local-write-with-opted-out-renderer.test.ts > local crash log writer — renderer surface > writes the renderer incident dir + meta.json even when consent is opted-out` — **PASS**.

## Test summary

- `tests/electron/crash/retention-per-surface.test.ts` — 5/5 pass (per-surface keep-N, two protected-window scenarios, count-cap-vs-protect interaction, legacy global rule regression).
- `tests/electron/ipc/rendererErrorForwarder.test.ts` — 4/4 pass (records renderer-surface incident, drops past cap, per-process bucket isolation, drop-count fold-in on next accepted event).
- `tests/electron/crash/local-write-with-opted-out-renderer.test.ts` — 1/1 pass.
- Existing crash suites (`collector.test.ts`, `local-write-with-opted-out.test.ts`, `crashIncidents.test.ts`) — all green.
- Typecheck: `npm run typecheck` clean.
- Lint: net **+1** error fixed vs. baseline (33 → 32; pre-existing errors all in unrelated files).

## i18n

No new UI strings — pure plumbing. No i18n keys added.

## Files

- `electron/crash/collector.ts` — extended `PruneOpts`, prune logic.
- `electron/main.ts` — boot-time prune call upgraded with per-surface + protect-window args; `registerRendererErrorForwarderIpc` wired.
- `electron/ipc/rendererErrorForwarder.ts` — new IPC handler + rate limiter.
- `electron/preload/bridges/ccsmCore.ts` — `ccsm.crash.reportRendererError` bridge.
- `src/lib/installRendererErrorForwarder.ts` — renderer-side hooks.
- `src/index.tsx` — install hook after Sentry init.
- 3 new test files.

## Test plan

- [x] Unit: per-surface retention prunes correctly; protected window respected.
- [x] Unit: renderer forwarder routes via IPC, recordIncident called with `surface: 'renderer'`.
- [x] Unit: rate-limit per process; drop-count fold-in.
- [x] Reverse-verify: disable protect window → tests FAIL; restore → PASS.
- [x] Regression: local renderer incident writes under `consent: opted-out`.
- [x] Typecheck (`npm run typecheck`).
- [ ] E2E probe — out of scope per task instructions ("No e2e probes — focus on unit tests").

### DIFF
--- a/electron/crash/collector.ts
+++ b/electron/crash/collector.ts
@@ -28,10 +28,26 @@ export interface CollectorOpts {
   electronVersion: string;
 }
 
+export interface PruneOpts {
+  /** Legacy global cap — prune entries that are BOTH older than `maxAgeDays`
+   *  AND beyond the `maxCount` newest. Phase 1 contract. Optional now. */
+  maxCount?: number;
+  /** Legacy global age threshold (days). Pairs with `maxCount`. */
+  maxAgeDays?: number;
+  /** Phase 5 — keep the N newest incidents per surface (by mtime). When
+   *  omitted, only the legacy global cap applies. Default at the call site
+   *  is 20 (`crashLogRetention.maxPerSurface`). */
+  maxPerSurface?: number;
+  /** Phase 5 — never prune an incident that has NO `.uploaded` marker AND
+   *  whose mtime is younger than this many days, so the user can still send
+   *  it via the Help-menu re-upload button (phase 4). Default 7. */
+  protectUnsentYoungerThanDays?: number;
+}
+
 export interface CrashCollector {
   recordIncident(input: IncidentInput): string;
   flush(): Promise<void>;
-  pruneRetention(opts: { maxCount: number; maxAgeDays: number }): void;
+  pruneRetention(opts: PruneOpts): void;
 }
 
 export function startCrashCollector(opts: CollectorOpts): CrashCollector {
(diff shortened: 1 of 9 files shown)


Jiahui-Gu (commented):
## Verdict: APPROVE with P1 follow-up

### Layer 1 (hard constraints)

- **Local write regardless of consent**: VERIFIED. `tests/electron/crash/local-write-with-opted-out-renderer.test.ts` mocks `isCrashUploadAllowed=false` + `loadCrashConsent='opted-out'` and asserts the renderer incident dir + `meta.json` are created with `surface='renderer'`. Test passes locally. Code path (`handleRendererErrorReport` → `collector.recordIncident`) is consent-agnostic by construction — there is no consent reference in `electron/ipc/rendererErrorForwarder.ts` at all. Hard constraint upheld.

- **Protect-window (7d + missing `.uploaded` marker)**: VERIFIED both conditions gate together in `electron/crash/collector.ts` lines 173-177:
  ```ts
  if (protectUnsentYoungerThanDays != null && !isUploaded) {
    const cutoff = Date.now() - protectUnsentYoungerThanDays * 24 * 3600 * 1000;
    if (stat.mtimeMs >= cutoff) protectedByWindow = true;
  }
  ```
  Reverse-verify reproduced the PR-body claim exactly: neutralizing the branch (`if (false && …)`) → 2 failures (`12 case` + `count-cap regression`); restoring → 5/5 PASS.

- **Spec divergence — rolling.ts omitted**: this is the one real concern. Plan §11 phase 5 / Task 16 explicitly enumerates `electron/log/rolling.ts`, `daemon/src/log/rolling.ts`, `electron/log/renderer-forwarder.ts` (the *console-forwarding* sink, distinct from the *error*-forwarding IPC in this PR), console redirect for `console.{info,warn,error}`, and bundle wiring into incident dirs. The PR ships none of those. The spec §11 description ("every `console.*` call in main + renderer is captured even when no crash happens, so we can debug warm bugs") is the steady-state log story; this PR ships only the error-event forwarder. They are separable concerns (error forwarder routes throws → incidents; rolling logs route info/warn/error → daily rotated files for "Send last crash" bundling). Shipping the error forwarder without rolling logs leaves a coverage gap: incident dirs from this PR will NOT contain a recent log tail beyond the 6-line `stderrTail` breadcrumbs, so phase-4 "Send last crash" still ships sparse bundles for renderer-only incidents. Recommend: **defer rolling.ts to phase 6 with a follow-up task**, retitle this PR to "phase 5a — incident retention + renderer error forwarder" so the plan checklist isn't falsely closed. Not a merge blocker; the slice is self-contained and shipped behind the same Sentry consent gate as the rest of phase 4.

- **`maxPerSurface=20` hard-coded, Settings toggle deferred**: acceptable defer (P2). Spec §10/§11 don't require a UI toggle in phase 5; the PR comment "follow-up can surface it in Settings if the user ever wants to tune it" is correct triage. 20-per-surface × 7 surfaces = 140 max incidents is well within disk budget.

### Layer 2

- 10/10 new tests PASS; existing crash suites 22/22 PASS.
- Typecheck clean.
- Lint: only 1 pre-existing `no-empty` on `electron/main.ts:91` (the new `try { crashCollector.pruneRetention(…) } catch {}` wrapper inherits the same shape from the prior single-line try/catch). Net-neutral, not regressed by this PR.
- `pruneOpts` shadow rename consistent throughout `pruneRetention` (line 151-152). No `opts.crashRoot` reference accidentally renamed; outer collector `opts: CollectorOpts` still readable.
- Per-surface bucketing reads `meta.json` via try/catch fallback to `surface=null` → `__unknown__` bucket, which is still subject to its own keep-N rule. Graceful when meta missing/corrupted.
- Rate limiter: per-process bucket (`webContents.id`), sliding-window prune, drop count drained one-shot into next accepted incident's `stderrTail` as `renderer-error-drops=N`. Burst test (5 events / max=3) → 3 accepted, 2 dropped, count surfaced on the next accepted event after window slide. Correct.
- Defense-in-depth: `fromMainFrame(e)` guard rejects non-mainFrame senders; malformed payloads rejected via shape check.
- Dev-only console.error gating: `process.env.NODE_ENV !== 'production'` evaluated at renderer module-load time; webpack production mode bakes the constant. Production renderer does not capture console.error. Correct.
- Renderer forwarder coexists with @sentry/electron renderer SDK (separate global hook registrations on `window`, both fire). Documented inline. OK.

### Counts

- P0: **0**
- P1: **1** — rolling.ts / steady-state log capture / bundle wiring deferred. Either (a) retitle this PR "phase 5a" + open follow-up task for "phase 5b — rolling logs", or (b) consciously rescope spec §11 phase 5 into two PRs and update the plan checklist. Manager's call.
- P2: **1** — `crashLogRetention.maxPerSurface` Settings toggle deferred (acceptable per spec).

### Recommendation

**Merge phase 5a as-is**, dispatch a follow-up worker for phase 5b (rolling.ts + console redirect + bundle wiring per plan Task 16 steps 2-7). The user-locked hard constraint (local write regardless of consent) holds; reverse-verify confirms the protect-window guards the phase-4 "Send last crash" UX as designed.

PR #607, opened 2026-04-30
refactor(ptyHost): L4 PR-C explicit dispatchPtyChunk + backpressure observe (#863)

Jiahui-Gu:
## Summary

L4 PR-C of the dual-buffer mirror series (Task #863). **Stacks on PR #606 (PR-B)** — manager will rebase to `working` after #606 merges.

- Extract the per-chunk fan-out from inline `p.onData` arrow into named, exported `dispatchPtyChunk(sid, entry, chunk)`. The headless + broadcast double-write is now explicit at the call site, and PR-D/E have a stable hook point.
- Add observe-only backpressure: `headless.write(chunk, cb)` callback decrements `entry.pendingHeadlessWrites`. Crossing `BACKPRESSURE_WARN_THRESHOLD = 100` emits one `console.warn` (re-arms after drain). **No chunk is ever dropped.**
- 3 new UTs pin `dispatchPtyChunk` directly (without going through node-pty).

No behavior change to the two existing sinks. Out of scope: PR-D resize, PR-E detach/reattach, PR-F hack removal, usePtyAttach (already rewritten by PR-B), broadcast/IPC protocol.

## Files

- `electron/ptyHost/entryFactory.ts` — extract `dispatchPtyChunk`, add `BACKPRESSURE_WARN_THRESHOLD`, add `pendingHeadlessWrites` + `backpressureWarned` to `Entry`.
- `electron/ptyHost/__tests__/entryFactory.test.ts` — 3 new tests + headless mock now mirrors `write(data, cb?)` signature with deferrable callback for backpressure modeling.

## Test plan

- [x] TDD: tests written first, FAIL on un-modified `entryFactory.ts` (4 fail), PASS after impl (11/11).
- [x] Reverse-verify: `git stash push electron/ptyHost/entryFactory.ts` → 4 FAIL; `git stash pop` → 11 PASS.
- [x] `npm run typecheck` clean.
- [x] `npm run lint` 0 errors.
- [x] Module load smoke: `node -e "require('./dist/electron/ptyHost/entryFactory.js')"` → `OK loaded` (CJS-Electron-main compatible).
- [x] Full vitest: only pre-existing failures unrelated to this PR (`tests/electron-load-smoke.test.ts` needs `npm run build` first; `electron/__tests__/db-hardening.test.ts` is db-layer, not ptyHost).

### TDD reverse-verify output

Stashed (4 FAIL):
```
Test Files  1 failed (1)
     Tests  4 failed | 7 passed (11)
[...]

PR #761, opened 2026-05-01
test: 3 regression probes for previously-shipped bugs

Jiahui-Gu:
## Summary

Task #1044 — add permanent regression probes for 3 previously-shipped bug fixes that lacked dedicated test coverage. Per audit HIGH: bugs that already escaped once need machine-checked guards so they can't silently come back.

All 3 probes live as fast unit tests (vitest + RTL / `renderHook`) in `tests/regression/`. No new e2e probe was added — every chosen fix is small enough that a unit test exercises the exact decision branch the bug touched, with no Electron boot cost.

## The 3 bugs

| # | Fix PR | Issue | Surface | Coverage gap before this PR |
|---|---|---|---|---|
| 1 | #601 | #857 — false `Ctrl+N` hint advertised but no handler bound | `ShortcutOverlay` + `CommandPalette` | `tests/shortcut-overlay.test.tsx` only asserted `>=6` kbd chips and that `Ctrl` appears once; nothing pinned the *absence* of the bare `Ctrl+N` row. CommandPalette hint chips were untested. |
| 2 | #510 | #607 — Import session always dropped rows in catch-all "Imported" group, ignoring `focusedGroupId` | `ImportDialog#doImport` | PR shipped with manual checklist only. Slice tests cover `importSession` directly but never the dialog wiring that decides `groupId` from `focusedGroupId`. |
| 3 | #520 | #613 — `sessionNamesFromRenderer` map grew unbounded; renderer never emitted clearing IPC for deleted sids (audit #876 cluster H) | `useSessionNameBridge` (renderer hook) | The harness-real-cli probe `notify-name-cleared-on-session-delete` covers e2e against real Electron + claude (~30s); no unit test exercising the renderer-side cleanup loop in isolation. A refactor moving the cleanup branch out of the effect would only surface in a slow e2e run. |

## Harness vs standalone breakdown

All 3 probes are unit tests (vitest in `tests/regression/`), not e2e. Per memory rule "E2E 优先塞 harness — each independent probe = +30s test time", none of these bugs warrant a new probe slot:

- #601 + #510 are pure renderer state contracts (no Electron, no CLI).
- #520's e2e path is already covered by the existing harness-real-cli case; the gap was the *unit-level* contract on the hook.

Adding 3 vitest files keeps the e2e suite unchanged (zero added wall-clock).

## Reverse-verify (each probe FAILED with its corresponding fix temporarily reverted in-place; reverts are NOT in this PR)

- **#601 ShortcutOverlay** (re-added `{ keys: '${MOD} + N', ... }` row):
  ```
  AssertionError: expected '...Ctrl+NNew groupCtrl+Shift+NNew group...' not to match /(?<!Shift\+)(?:Ctrl|⌘)\s*\+\s*N(?![a-z])/
  ```
- **#601 CommandPalette** (changed `cmd:new-group` hint from `${MOD}Shift+N` to `${MOD}N`):
  ```
  AssertionError: expected 'Command paletteNew groupCtrl+N↑↓Navigate↵SelectEscClose' to match /(?:Ctrl|⌘)\s*\+?\s*Shift\s*\+\s*N\b/
  ```
- **#510 ImportDialog** (replaced `focusedGroupId` lookup with `undefined`):
  ```
  AssertionError: expected "g-imported-fallback" to be "g-research"
  ```
- **#520 useSessionNameBridge** (deleted the stale-sid cleanup loop):
  ```
  AssertionError: expected "vi.fn()" to be called with arguments: [ 'b', null ]
  Number of calls: 1   (only setName('a','Alpha') fired)
  ```
- **#520 full-delete-all transition** (same revert):
  ```
  AssertionError: expected "vi.fn()" to be called with arguments: [ 'a', null ]
  Number of calls: 0
  ```

After restoring all 4 production files via `git checkout`, the suite returns to **10/10 passing** in ~4s.

## Quality gates

- `npx vitest run tests/regression/` — 10/10 passing, ~4s wall.
- `npx tsc --noEmit` — clean.
- ESLint: `tests/**` is globally ignored by `eslint.config.js` (line 17), so no lint deltas apply to these files.

## Test plan
- [x] All 3 new probes pass on green tree
- [x] Reverse-verify FAIL output captured for each (above)
- [x] typecheck clean
- [x] lint scope confirmed (tests dir ignored by config)

### DIFF
--- a/tests/regression/import-dialog-focused-group.test.tsx
+++ b/tests/regression/import-dialog-focused-group.test.tsx
@@ -1,0 +1,164 @@
+// Regression: PR #510 (#607) — `fix(import): import targets focused
+// group`.
+//
+// Bug: `ImportDialog#doImport` always wrote new rows into the catch-all
+// "Imported" group. A user reading group "research" who hit Import had
+// to manually drag every new row out of "Imported" into "research".
+//
+// Fix: when `store.focusedGroupId` resolves to a normal group, route the
+// import there. Without a focus, fall back to the (created-on-demand)
+// "Imported" bucket. Also: the focused id is captured in `doImport`'s
+// closure, so a batch of N imported rows all land in the focused group
+// (the store clears `focusedGroupId` on each `importSession` call but
+// the closure already holds the original id).
+//
+// Existing coverage gap: the PR shipped with only a manual checklist —
+// no e2e probe and no unit test. The store slice tests
+// (`tests/stores/slices/sessionCrudSlice.test.ts`) exercise
+// `importSession` directly but never the dialog wiring that decides the
+// `groupId` argument from `focusedGroupId`. This file pins that wiring.
+import React from 'react';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { render, fireEvent, cleanup, waitFor, screen } from '@testing-library/react';
+import { ImportDialog } from '../../src/components/ImportDialog';
+import { useStore } from '../../src/stores/store';
+
+type Scannable = {
+  sessionId: string;
+  cwd: string;
+  title: string;
+  mtime: number;
+  projectDir: string;
+  model: string | null;
+};
+
+const fakeRows: Scannable[] = [
+  {
+    sessionId: 'imp-1',
+    cwd: '/work/proj-a',
+    title: 'session-a',
+    mtime: Date.now(),
+    projectDir: '/home/u/.claude/projects/proj-a',
+    model: null,
+  },
+  {
+    sessionId: 'imp-2',
+    cwd: '/work/proj-b',
+    title: 'session-b',
+    mtime: Date.now(),
+    projectDir: '/home/u/.claude/projects/proj-b',
+    model: null,
+  },
+];
+
+function installCcsmStub() {
+  (window as unknown as { ccsm: unknown }).ccsm = {
+    scanImportable: vi.fn().mockResolvedValue(fakeRows),
+  } as unknown as Window['ccsm'];
+}
+
+function clearCcsmStub() {
+  (window as unknown as { ccsm?: unknown }).ccsm = undefined;
+}
+
+afterEach(() => {
+  cleanup();
+  clearCcsmStub();
+});
+
+describe('PR #510 regression — ImportDialog targets focused group', () => {
+  beforeEach(() => {
+    installCcsmStub();
+  });
+
+  it('routes imports into focusedGroupId when set to a normal group', async () => {
+    const importSession = vi.fn();
+    const createGroup = vi.fn().mockReturnValue('g-imported-fallback');
+    const renameGroup = vi.fn();
+    useStore.setState({
+      sessions: [],
+      groups: [
+        { id: 'g-research', name: 'research', collapsed: false, kind: 'normal' },
+        { id: 'g-other', name: 'other', collapsed: false, kind: 'normal' },
+      ],
+      focusedGroupId: 'g-research',
+      importSession,
+      createGroup,
+      renameGroup,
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
## Independent reviewer (Task #1044)

**Verdict: APPROVE**. P0: 0, P1: 0, P2: 0.

### Reverse-verify (manually re-ran, in this worktree, against tip of branch)

- **#601 ShortcutOverlay** — re-added a `{ keys: ${MOD} + N, actionKey: shortcuts.actionNewGroup }` row next to the legitimate Shift+N row → `no-false-ctrl-n-hint.test.tsx > ShortcutOverlay` FAILS with the expected "expected '...Ctrl+Shift+NNew groupCtrl+NNew group...' not to match /(?<!Shift\+)(?:Ctrl|⌘)\s*\+\s*N(?![a-z])/" assertion. Restored, 10/10 pass. **PASS**.
- **#510 ImportDialog** — replaced `const focused = focusedGroupId ? groups.find(...) : undefined` with `const focused = undefined` → `import-dialog-focused-group.test.tsx > routes imports into focusedGroupId` FAILS with `expected 'g-imported-fallback' to be 'g-research'`. Restored. **PASS**.
- **#520 useSessionNameBridge** — deleted the `for (const staleSid of prevSidsRef.current)` cleanup loop → `session-name-bridge-cleanup.test.tsx` FAILS on both "emits setName(sid, null) for sids that disappear" and "handles a full delete-all transition". Restored. **PASS**.

All three reverts produced FAILs identical in shape to the worker's PR-body capture. Tests are not tautological — they exercise the real fix branch.

### Layer 1 (necessity / picks)

- Bug picks are well-justified: each of the 3 was previously merged with manual-only verification, and #601/#510 had a literal coverage gap (existing slice/overlay tests asserted shape, not the absent surface). #520 had only the harness-real-cli e2e probe which is +30s; a unit-level pin on the renderer hook is the right granularity to keep cheap regression coverage when refactors move the cleanup branch.
- Rejection rationale spot-checks: justified — those bugs are either covered by existing slice tests, are visual-only (need screenshot diffing not vitest), or already have e2e coverage.
- Vitest-only (no new e2e probe) is the right call per the +30s/probe rule. `tests/regression/` directory is a reasonable new convention; matches existing flat-test patterns.

### Layer 2

- `npx vitest run tests/regression` → 10/10 pass (~4s).
- `npx tsc --noEmit` → clean.
- Mock surfaces are not the things under test: tests stub `window.ccsm`/`window.ccsmSession` (preload IPC boundary), `useStore` slice action functions (`importSession`, `createGroup`), and `matchMedia` (jsdom gap). They render the real `ShortcutOverlay`/`CommandPalette`/`ImportDialog` components and the real `useSessionNameBridge` hook. The decision branches under test (`buildGroups()` row list, `doImport`'s focusedGroupId lookup, the cleanup-loop diff) all run from production source.
- Regex `(?<!Shift\+)(?:Ctrl|⌘)\s*\+\s*N(?![a-z])` correctly rejects bare Ctrl+N while passing Ctrl+Shift+N (in textContent the latter appears as "Ctrl+Shift+N" so the engine never finds a "Ctrl+N" substring without "Shift+" preceding the N). The `(?![a-z])` guard prevents false positives if a future label starts with lowercase after N.
- One minor nit (P2-not-blocking): `session-name-bridge-cleanup.test.tsx` `installBridge()` ignores `Window['ccsmSession']` typing and casts via `unknown`; consistent with the other two files and acceptable for a regression probe.

Cleared the reverts via `git checkout` of all 3 source files; worktree is back to PR HEAD with 10/10 green. No P0/P1/P2 issues. Ready to merge.

PR #773, opened 2026-05-01
feat(daemon): envelope adapter for v0.3 IPC framing

Jiahui-Gu:
## Summary

Wire the daemon-side envelope adapter that bridges raw `Duplex` sockets (control-socket / data-socket) to the pure `Dispatcher`. Closes the "Envelope adapter wiring is T-future" placeholder noted in `daemon/src/index.ts`.

**Spec**: [`docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md`](https://github.com/Jiahui-Gu/ccsm/blob/working/docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md)
- §3.4.1.a frame-version nibble + 16 MiB cap rejection sequence
- §3.4.1.c length-prefixed JSON envelope routing fields
- §3.4.1.d `schema_violation` reply + `socket.destroy()`

**Unblocks**: Task #27 (Electron-side envelope migration). Together they close out the daemon-split RPC migration ahead of v0.3 merge.

## Scope (intentionally minimal)

- **New** `daemon/src/envelope/adapter.ts` — `mountEnvelopeAdapter({ socket, dispatcher, logger, peer, peerPid })`. Per-connection buffer accumulation, `decodeFrame` loop, header schema check (`id`/`method`), dispatch + reply encoding. Pure producer/decider/sink split per dev contract §2.
- **New** `daemon/src/envelope/__tests__/adapter.test.ts` — 11 unit tests (see Test plan).
- **`daemon/src/index.ts`** — replace the `destroy with warn` placeholder for both sockets. Control-socket dispatches into `supervisorDispatcher`; data-socket gets a fresh `createDataDispatcher()` (handlers register later via T17–T21 / data-plane RPC slices).

## Out of scope (other slices own these)

- HMAC handshake / hello-gate / migration-gate / deadline / trace-id interceptors — they compose into the dispatcher via their own modules (`hello-interceptor.ts`, `migration-gate-interceptor.ts`, `deadline-interceptor.ts`, `hmac.ts`).
- Binary trailer round-trip in *replies* (no v0.3 daemon RPC returns binary yet); inbound trailers are preserved and forwarded to handlers.
- Per-stream chunking, fan-out caching, header-skeleton fast path (§3.4.1.b / §3.4.1.c hot-path optimizations) — v0.4 territory.

## Test plan

Local: `npx vitest run daemon/src/envelope/__tests__/adapter.test.ts` → `Test Files 1 passed (1) | Tests 11 passed (11)`.
Local: `npx tsc --noEmit -p daemon/tsconfig.json` → clean.
Local: `npx eslint daemon/src/envelope/adapter.ts daemon/src/envelope/__tests__/adapter.test.ts daemon/src/index.ts` → clean.

Cases covered:
- [x] Framing happy path (single frame → dispatch → reply round-trip byte-equal)
- [x] Multiple frames in one chunk all dispatch in order
- [x] Dispatcher error replies forwarded as `{ ok: false, error }` without destroying
- [x] Truncated read: split mid-header, dispatcher only fires after second chunk
- [x] One-byte-at-a-time trickle handled
- [x] Oversized message (16 MiB + 1) → synthetic `envelope_too_large` reply with `id: 0`, socket destroyed, `envelope_oversize` warn line carries `peerPid` + `len`
- [x] Unsupported frame-version nibble rejected before cap-check (spec §3.4.1.a step 2 ordering)
- [x] Malformed JSON header → `schema_violation` reply with `id: 0` + destroy
- [x] Header parses but missing routing fields → `schema_violation` + destroy
- [x] Schema-violation echoes valid `id` when only `method` is missing
- [x] Dispatcher-thrown exception surfaces as `INTERNAL` reply WITHOUT destroying socket (per-call failure ≠ protocol violation)

### DIFF
--- a/daemon/src/envelope/__tests__/adapter.test.ts
+++ b/daemon/src/envelope/__tests__/adapter.test.ts
@@ -1,0 +1,352 @@
+// Daemon-side envelope adapter tests (Task #28 / B7a).
+//
+// Test cases cover the four spec-mandated scenarios from the task brief:
+//   1. Framing happy path (well-formed frame -> dispatcher -> reply byte-equal
+//      decodes back to `{ id, ok: true, value, ack_source }`).
+//   2. Truncated read (one frame split across multiple `data` events;
+//      dispatcher only fires once after the second chunk arrives).
+//   3. Oversized message (header claiming > 16 MiB -> synthetic
+//      `envelope_too_large` reply with `id: 0` + socket destroyed).
+//   4. Malformed JSON header (valid framing, garbage header bytes -> synthetic
+//      `schema_violation` reply + socket destroyed).
+//
+// Plus a small clutch of safety tests: schema-violation on missing routing
+// fields; dispatcher rejections passed through; multiple frames in one chunk
+// are all dispatched.
+//
+// Spec citations:
+//   - frag-3.4.1 §3.4.1.a (oversize rejection sequence)
+//   - frag-3.4.1 §3.4.1.c (header schema)
+//   - frag-3.4.1 §3.4.1.d (schema_violation + destroy)
+
+import { describe, expect, it, vi } from 'vitest';
+import { Buffer } from 'node:buffer';
+import { PassThrough } from 'node:stream';
+
+import { mountEnvelopeAdapter } from '../adapter.js';
+import { decodeFrame, encodeFrame, ENVELOPE_LIMITS } from '../envelope.js';
+import type { Dispatcher, DispatchResult } from '../../dispatcher.js';
+
+// ---------------------------------------------------------------------------
+// Test harness
+// ---------------------------------------------------------------------------
+
+/**
+ * A minimal Duplex stand-in: writes from the adapter land in `outbound`,
+ * `feed()` simulates inbound `data` events, and `destroy()` flips a flag.
+ *
+ * Using two `PassThrough`s would also work but a thin shim keeps assertions
+ * direct (we look at `outbound[0]` rather than awaiting `data` events on a
+ * second stream).
+ */
+function makeFakeSocket(): {
+  socket: PassThrough & { destroyedByAdapter: boolean };
+  feed: (b: Buffer) => void;
+  outbound: Buffer[];
+  destroyed: () => boolean;
+} {
+  const sock = new PassThrough() as PassThrough & { destroyedByAdapter: boolean };
+  sock.destroyedByAdapter = false;
+  const outbound: Buffer[] = [];
+  // Capture writes by overriding write to push the chunks into outbound.
+  // (PassThrough would otherwise loop them back to the readable side and
+  // the adapter would re-parse its own replies.)
+  sock.write = ((chunk: unknown) => {
+    outbound.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(String(chunk)));
+    return true;
+  }) as typeof sock.write;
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
**Review verdict: APPROVE** (cannot self-approve via review API, posting as comment).

Clean implementation, faithful to spec frag-3.4.1.

**Spec compliance** (frag-3.4.1-envelope-hardening.md):
- §3.4.1.a frame-version nibble FIRST → mask → cap-check: enforced by `decodeFrame` in `daemon/src/envelope/envelope.ts` (already merged); adapter correctly reacts to `EnvelopeError.code` values. Truncated read short-circuits via `truncated_frame` (`adapter.ts:248-252`) without writing/destroying — correct (this is the happy partial-read path).
- §3.4.1.a oversize sequence: warn `envelope_oversize` with `peerPid` + `len` → synthetic `{id:0, error: envelope_too_large}` reply → `socket.destroy()` (`adapter.ts:256-274`). Matches spec word-for-word.
- §3.4.1.c routing-field schema check (`id`/`method`): `isWellFormedHeader` (`adapter.ts:79-86`) covers minimum subset; broader header validation deferred to handlers per §3.4.1.d (correct ownership split).
- §3.4.1.d malformed JSON or shape violation → `schema_violation` reply + destroy (`adapter.ts:295-318`). Echoes `id` when parseable but mis-shaped — nice forensic touch beyond spec literal.

**Test quality**: 11 UTs all run real `decodeFrame` round-trips, exercise byte-level wire layout (hand-crafted 4-byte oversize prefix, garbage header bytes, frame-split midway), and assert both behavior AND side-effects (destroyed flag, warn payload). Not trivial. Verified locally: `npx vitest run daemon/src/envelope/__tests__/adapter.test.ts` → 11/11 pass in 2.75s.

**SRP**: clean. Three files; index.ts diff is purely the placeholder swap — no drive-by changes. Adapter has zero coupling to HMAC/hello/migration/deadline/trace-id modules; those compose into the dispatcher elsewhere as the header comment states.

**Wiring sanity**: control-socket → existing `supervisorDispatcher`; data-socket → fresh `createDataDispatcher()` which returns `UNKNOWN_METHOD` for any unregistered method. Safe to ship before #27/handlers land — clients hitting data-plane methods get a structured error instead of a hang.

**Non-goals**: confirmed adapter does NOT implement HMAC handshake / hello-gate / migration-gate / deadline / trace-id. Comment block explicitly disclaims them.

**Minor observations** (non-blocking):
- The `Buffer.concat` re-coalesce on every loop iteration when `pending.length !== 1` is O(n) under sustained 1-byte trickle — acceptable for v0.3 dogfood as the comment notes; v0.4 perf pass should swap to a ring buffer.
- Once helloInterceptor lands (§3.4.1.g), the adapter's current behavior of dispatching the FIRST frame straight to the supervisor dispatcher means `daemon.hello` enforcement will live in the dispatcher chain, not the adapter — that's the intended seam per §3.4.1.f, just flagging so the next slice owner doesn't accidentally bolt hello logic into `adapter.ts`.

CI note: only `parity (ubuntu-latest|windows-latest)` actually ran (both pass); the 3 SKIPPED jobs are the repo-wide workflow-yaml bug tracked under Task #35, not signal on this PR.

Per manager instruction, NOT merging.

PR #765, opened 2026-05-01
chore(v03): END migration window — re-enable CI + first-pass lint/typecheck sweep

Jiahui-Gu:
## Summary

Closes the v0.3 migration window per Task #1048. Re-enables the 5 CI workflows that #1047 gated `if: false` so v0.3 daemon-split + UI work could land in parallel, sweeps the lint regressions that built up during the dark window, and pushes the follow-on fixes needed to turn the now-restored CI green for the first time.

## Workflows re-enabled

All 5 from the #1047 gate list:

- `.github/workflows/ci.yml` (lint + typecheck + test, 3-OS matrix)
- `.github/workflows/e2e.yml` (e2e, 3-OS matrix)
- `.github/workflows/installer-size-guard.yml` (size guard, win/mac/linux)
- `.github/workflows/main-source-guard.yml` (PR-to-main source check)
- `.github/workflows/pr-no-silent-drops.yml` (silent-drop detector)

`release.yml` was kept active per #1047 — left untouched.

## Lint sweep — 32 errors -> 0

`npm run lint` baseline: 32 errors / 122 warnings -> after fix: 0 errors / 122 warnings.

| Area | Count | Root cause | Fix |
|---|---|---|---|
| `daemon/src/**/*.ts` `no-undef` (process, Buffer, AbortSignal, queueMicrotask) | ~3 | eslint config only declared Node globals for `electron/**/*.ts`; daemon code added during the window had no globals scope | Extended Node-globals block in `eslint.config.js` to cover `daemon/src/**`, `daemon/spike-pkg-esm/**`, `installer/**`, and root `*.config.{js,ts,mjs,cjs}` |
| `installer/uninstall-helper/index.js` `no-undef` (require/module/process/Buffer/setTimeout/clearTimeout) | 22 | same -- Node JS file had no globals coverage | same eslint config extension |
| `vitest.config.ts` `__dirname` `no-undef` | 2 | same | same |
| `daemon/src/crash/handlers.ts` + `electron/main.ts` empty `catch {}` `no-empty` | 3 | `js.configs.recommended` flags empty blocks; the `try {} catch {}` best-effort pattern is established across the codebase | Added `'no-empty': ['error', { allowEmptyCatch: true }]` |
| `daemon/src/marker/reader.ts` `no-irregular-whitespace` | 1 | literal U+FEFF inside a BOM-stripping regex | Replaced with escape sequence (semantically identical) |
[...]

PR #777, opened 2026-05-02
fix(packaging): after-sign dispatcher resolves named/default/fn export shapes

Jiahui-Gu:
## Bug (real production crash)

`scripts/after-sign.cjs` did:

\`\`\`js
const signMac = require(path.join(__dirname, 'sign-macos.cjs'));
await signMac(context);   // TypeError: signMac is not a function
\`\`\`

But `scripts/sign-macos.cjs:162-165` only exports an **object**:

\`\`\`js
exports.default = signMacApp;
exports.signMacApp = signMacApp;
exports.collectSignTargets = collectSignTargets;
exports.codesignOne = codesignOne;
\`\`\`

So `require(...)` returns an object, and the dispatcher threw
`TypeError: signMac is not a function` on every real darwin signed build.

The existing UTs masked the bug because they pre-populated
`require.cache[macPath].exports = bareSpyFn`, replacing the entire
module with a callable and never exercising the production export shape.
Surfaced during dev push-back on Task #30.

## Fix (manager option 1 — minimal scope)

Per manager direction (keep `sign-macos.cjs` / `sign-windows.cjs` export
contracts untouched), the dispatcher now resolves a callable across all
three CommonJS export shapes via a small `resolveCallable` helper:

1. named export (`signMacApp` / `signWindowsHook`) — current production shape
2. `default` export — ESM-interop convention
3. bare-function module — legacy

`sign-windows.cjs` already exports as bare-fn-with-attached-props (works
today), but the dispatcher path is symmetric for safety against future
shape drift.

## UT coverage

`tests/after-sign.test.ts` now parameterises the mock export shape and
pins one test per branch (named / default / bare-fn) for both darwin
and win32 platforms, in addition to the existing routing tests.

Result: **9/9 pass** in `after-sign.test.ts`, **28/28 pass** across
`after-sign + sign-macos + sign-windows` test files.

## Reverse-verify

Temporarily removed the `if (mod && typeof mod[namedExport] === 'function') return mod[namedExport];`
branch from `resolveCallable`. Re-ran the suite:

\`\`\`
Test Files  1 failed (1)
     Tests  2 failed | 7 passed (9)

 FAIL  tests/after-sign.test.ts > after-sign dispatcher > darwin: resolves named export shape (sign-macos.cjs production shape)
TypeError: [after-sign] could not resolve callable from module;
  expected named export "signMacApp", "default", or a function (got object)

 FAIL  tests/after-sign.test.ts > after-sign dispatcher > win32: resolves named export shape
TypeError: [after-sign] could not resolve callable from module;
  expected named export "signWindowsHook", "default", or a function (got object)
\`\`\`

Exactly the two named-shape tests fail with the production TypeError shape.
Restored the branch — back to 9/9 green. Reverse-verify edit is **not committed**.

## Out of scope

- Not changing `sign-macos.cjs` / `sign-windows.cjs` export interfaces
  (option 2 was rejected by manager — scope-min preferred).
- Not self-merging — Task #30 reviewer pass will pick this up.

### DIFF
--- a/scripts/after-sign.cjs
+++ b/scripts/after-sign.cjs
@@ -1,21 +1,43 @@
 'use strict';
 // Dispatcher: routes electron-builder afterSign to platform-specific signing scripts.
 // Each delegate platform-gates internally and is no-op on the wrong platform.
+//
+// Export-shape resilience: sign-macos.cjs / sign-windows.cjs are CommonJS but
+// historically used inconsistent export shapes. sign-macos.cjs in particular
+// only exports an OBJECT (`exports.signMacApp = ...; exports.default = ...`),
+// so `require(...)` returns an object — calling it directly throws
+// `TypeError: signMac is not a function` on real darwin signed builds. The UTs
+// previously masked this by pre-populating `require.cache[...].exports = spy`
+// with a bare function. We resolve to a callable across all three shapes:
+//   1) named export `signMacApp` / `signWindowsHook` (preferred for new code)
+//   2) `default` export (ESM-interop convention)
+//   3) the module itself if it IS a function (legacy bare-fn shape)
 const path = require('path');
+
+function resolveCallable(mod, namedExport) {
+  if (mod && typeof mod[namedExport] === 'function') return mod[namedExport];
+  if (mod && typeof mod.default === 'function') return mod.default;
+  if (typeof mod === 'function') return mod;
+  throw new TypeError(
+    `[after-sign] could not resolve callable from module; ` +
+      `expected named export "${namedExport}", "default", or a function ` +
+      `(got ${mod === null ? 'null' : typeof mod})`,
+  );
+}
+
 module.exports = async function afterSign(context) {
   const platform = context.electronPlatformName;
   if (platform === 'darwin' || platform === 'mas') {
-    // sign-macos.cjs uses `exports.signMacApp` / `exports.default` rather
-    // than `module.exports = fn`, so `require(...)` returns an object.
-    // The `|| mod` fallback keeps tests working when they pre-populate
-    // `require.cache[...].exports` with a bare spy fn.
     const mod = require(path.join(__dirname, 'sign-macos.cjs'));
-    const signMac = mod.signMacApp || mod.default || mod;
+    const signMac = resolveCallable(mod, 'signMacApp');
     await signMac(context);
   }
   if (platform === 'win32') {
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Reviewer pass — push back: rebase needed (conflict + scope overlap)

### Verdict on the change itself: APPROVE on substance

Layer 1 (direction): right. Real production crash, minimal-scope fix per manager option 1, no delegate-script changes, exactly the v0.3 ship-blocker called out in `project_v03_ship_intent`.

Layer 2 (implementation):
- `resolveCallable` fallback chain order is correct: named (most explicit) → default (ESM-interop) → bare-fn (legacy). All three are present in real shapes (sign-macos has named + default; sign-windows has bare-fn).
- Error message is good — names the expected named export, lists all three shapes attempted, and reports `typeof` of what was actually received (with explicit null handling).
- UT coverage is real: 6 new tests pin one branch each (3 shapes × 2 platforms). Reverse-verify in PR body confirms the named-shape tests fail with the production TypeError when the named branch is removed.
- Scope: only 2 files touched (dispatcher + its test). `module.exports.resolveCallable` exposure is fine for direct testing and doesn't pollute the default export (which is still the async dispatcher fn).

### Why pushing back: working has already moved

PR #765 (`555d2ab`, landed in `working` after #777 was opened) shipped an equivalent inline fix on the **darwin branch only**:

```js
const mod = require(path.join(__dirname, 'sign-macos.cjs'));
const signMac = mod.signMacApp || mod.default || mod;
```

Result: `gh pr view 777` reports `mergeable: CONFLICTING / mergeStateStatus: DIRTY` on `scripts/after-sign.cjs`.

Your PR is still net positive over what's in `working`:
1. Extracts `resolveCallable` helper (reusable, single source of truth).
2. Applies the fix symmetrically to **win32** (working only fixed darwin — win32 happens to work today because sign-windows.cjs is bare-fn shape, but it's one drift away from the same crash).
3. Adds the explicit error message when no shape resolves (working's `|| mod` silently falls through to "TypeError: undefined is not a function" if the module is e.g. `null` or an empty object).
4. Adds the 6 parameterized export-shape tests.

### Action

Please rebase onto current `origin/working` and resolve the conflict in `scripts/after-sign.cjs` by keeping your `resolveCallable` version (it strictly supersedes the inline `||` chain). Push back up and ping; I'll re-merge immediately — no further review needed since the substance is already approved.

Not blocking on anything else.

PR #780, opened 2026-05-02
fix(v0.3): release.yml minisign sidecar must skip when secret unset

Jiahui-Gu:
## Summary

P0 ship blocker. Resolves #53.

The `attest` job in `.github/workflows/release.yml` wrote empty `.minisig` placeholder files when `MINISIGN_PRIVATE_KEY` was absent, and the `publish` job's sidecar-verify gate hard-failed on `[ ! -s "$f.minisig" ]`. Net effect: any release without the minisign secret provisioned (forks, self-hosters, or `Jiahui-Gu/ccsm` before keypair generation) could not publish a draft release at all.

This contradicted:
- spec frag-11 §11.5(5): "MUST continue unsigned if absent"
- v0.3 ship-intent: "placeholder-safe (缺 secret 自动 skip 输出 unsigned installer)"
- the workflow's own Apple/signtool steps, which already fall back to unsigned when their secrets are missing

## Fix (option (a) per task brief — minimal, matches existing pattern)

- `attest` job, "Generate minisign signatures": when `MINISIGN_PRIVATE_KEY` is unset, log a warning and `exit 0` without writing any `.minisig` files (was: write empty placeholder `.minisig` per artifact).
- `publish` job, "Sidecar-verify gate": at top of step, branch on the same secret. Unset → skip the `.minisig` portion of the gate, log a warning, continue. Set → preserve the existing fail-loud check (missing or empty `.minisig` still fails the build, so a real signing regression is still caught).
- `release-keys/README.md`: drop the two stale "writes empty `.minisig` / fails closed" sentences in the sidecar overview and the leak-disaster runbook so docs match the new behaviour.

The `.sha256` + `.intoto.jsonl` checks remain unconditional — those sidecars are produced from the artifacts themselves and don't depend on any secret, so a missing one is always a real bug.

## Test plan

`act` is not available in the Windows worktree environment, so I extracted the publish-gate's bash logic verbatim into a local script and ran it against fake `dist-flat` fixtures (two installers + provenance + sha256, optional minisig). Four scenarios covered the full truth table:

| # | Scenario | Expected | Actual |
|---|----------|----------|--------|
| 1 | secret unset, no `.minisig` sidecars (placeholder-safe path — was the bug) | exit 0 | exit 0 |
| 2 | secret set, `.minisig` sidecars missing | exit 1 (fail-loud) | exit 1 |
| 3 | secret set, valid non-empty `.minisig` sidecars (real-sign happy path) | exit 0 | exit 0 |
| 4 | secret set, empty `.minisig` sidecars (sabotage detection) | exit 1 (fail-loud) | exit 1 |

Sample outputs:

Scenario 1 (the bug — now passes):
```
::warning title=Unsigned release::MINISIGN_PRIVATE_KEY secret not set; skipping .minisig sidecar verification.
All installers have .sha256 + .intoto.jsonl sidecars (minisig skipped — secret unset).
exit=0
```

Scenario 2 (regression — fail-loud preserved):
```
::error title=Sidecar-verify failed::fake.exe missing or empty .minisig
::error title=Sidecar-verify failed::fake.dmg missing or empty .minisig
::error::Refusing to publish release with missing sidecars.
exit=1
```

Additional checks:
- [x] YAML parses: `python -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` → "YAML OK"
- [x] All four bash logic scenarios behave as expected
- [ ] Workflow dry-run via `workflow_dispatch` against a branch with no `MINISIGN_PRIVATE_KEY` configured (not done locally — would need to merge first or push branch trigger; left to reviewer/release-day ops if they want extra confidence)

## References

- spec: `docs/superpowers/specs/v0.3-fragments/frag-11-packaging.md` §11.5(5) + §11.6
- ship intent: `~/.claude/projects/C--Users-jiahuigu/memory/project_v03_ship_intent.md` — placeholder-safe

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -258,10 +258,11 @@ jobs:
       # Minisign signatures. Public key lives in repo at release-keys/minisign.pub
       # (committed). Private key is provisioned via the MINISIGN_PRIVATE_KEY
       # repo secret + MINISIGN_PASSWORD secret (see release-keys/README.md for
-      # rotation runbook). When the secret is absent we still write empty
-      # placeholder .minisig files so the sidecar-verify gate can distinguish
-      # "missing because unsigned" from "missing because forgotten" — and we
-      # warn loudly in the workflow log.
+      # rotation runbook). When the secret is absent we skip sidecar generation
+      # entirely and the publish job's sidecar-verify gate skips its minisig
+      # check accordingly — matching the v0.3 ship-intent "placeholder-safe"
+      # rule (frag-11 §11.5(5): "MUST continue unsigned if absent") and the
+      # Apple/signtool steps' existing fall-back-to-unsigned behaviour.
       - name: Install minisign
         run: |
           sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Reviewer LGTM (cannot self-approve via API). Layer 1 right (matches frag-11 §11.5(5) + existing Apple/signtool unsigned-fallback pattern). Layer 2: attest drops empty-placeholder loop; publish gate branches on MINISIGN_PRIVATE_KEY (env: wired correctly) while keeping .sha256 + .intoto.jsonl unconditional and preserving empty-.minisig sabotage detection when secret IS set; README sidecar table + leak-disaster runbook both updated; no drive-by edits. 4-quadrant truth-table test covers the guard properly. Squash-merging.

PR #781, opened 2026-05-02
feat(v0.3): daemon binary build pipeline (frag-11 §11.1 Task 20)

Jiahui-Gu:
## Summary

P0 ship blocker for v0.3 (Task #45). Implements the daemon binary build pipeline so the installer ships a real `ccsm-daemon-${platform}-${arch}${ext}` binary instead of the placeholder file `before-pack.cjs` falls back to today. Without this, post-install daemon spawn crashes 100% of the time (dogfood worker #14 finding).

- New `npm run build:daemon-bin` (root) orchestrates: tsc daemon -> Node 22 ABI native rebuild -> esbuild CJS bundle -> `@yao-pkg/pkg --targets node22-${platform}-${arch}` -> `daemon/dist/ccsm-daemon-*`.
- Daemon `package.json` now declares the full runtime dependency set per frag-11 §11.1 round-2 C6 (`pino`, `pino-roll`, `better-sqlite3`, `node-pty`, `proper-lockfile`, `@octokit/rest`, `ulid` alongside the existing `@connectrpc/*`, `@sentry/node`, `@sinclair/typebox`) — versions pinned to the root workspace deps so `pkg`'s `require()` graph closes correctly.
- Root `devDependencies` adds `@yao-pkg/pkg ^6.6.0` (spec-pinned) and `esbuild ^0.27.0` (was transitive; promoted so the bundle step is reproducible).
- `.github/workflows/release.yml` gains a `Build daemon binary` step between `Build web + electron` and `Package (...)` (frag-11 §11.5 step 2).

## Design notes

- **esbuild → CJS → pkg** (Fallback A from `docs/spikes/2026-05-pkg-esm-connect.md`): pkg's resolver does not honour package.json `exports` subpaths, and pkg's snapshot virtual fs only intercepts CJS `require()` (not ESM `import()`). esbuild bundles the whole ESM graph into one CJS file with all `exports` paths inlined, then pkg snapshots the CJS bundle directly.
- **Native `.node` files stay external** to the bundle and ride as `pkg.assets` so they're loaded at runtime via `path.dirname(process.execPath)/native/<platform>-<arch>/`, exactly as the daemon's existing native loader expects.
- **Daemon `DAEMON_VERSION`** now reads `daemon/package.json` (own version) instead of root — pkg cannot reach across the package boundary.
- **Boot listen() calls** wrapped in an async IIFE instead of top-level await so esbuild's CJS transform succeeds; signal handlers stay at module level. Behaviour identical (any reject re-throws to the crash handler / supervisor restart).
- **node-pty soft-skip on Windows**: its prebuild dir is missing on fresh installs (winpty's `GetCommitHash.bat` needs git-bash on PATH and fails the source rebuild). Daemon does not dlopen `node-pty` at boot (only at session spawn), so the binary boots cleanly and dogfood unblocks; before-pack `REQUIRED_NATIVES` is the hard installer gate.

## Local verification (Windows host, Node 22.11.0 target)

```
$ npm run build:daemon-bin
[build:daemon-bin] target=win-x64 pkg=node22-win-x64
[build:daemon-bin] $ npm run build:daemon
[build:daemon-bin] $ npm run rebuild:natives
... rebuilt better-sqlite3 (Node 22.11.0)
[build:daemon-bin] $ esbuild bundle daemon\dist-daemon\index.js -> daemon\dist-daemon-bundle\index.cjs
[build:daemon-bin] $ npx pkg . --targets node22-win-x64 ...
[build:daemon-bin] OK: ccsm-daemon-win-x64.exe (108.9 MB)

$ ./daemon/dist/ccsm-daemon-win-x64.exe   # spawn smoke
{"event":"daemon.boot.control-socket-listening","msg":"control-socket bound"}
{"event":"daemon.boot.data-socket-listening","msg":"data-socket bound"}
{"event":"daemon.boot","msg":"daemon shell booted"}
```

`npx tsc --noEmit -p daemon/tsconfig.json` clean. `npx vitest run tests/daemon` 14 passed (1 file skipped — the new opt-in build smoke).

## Test plan

- [ ] CI `Build daemon binary` step runs on linux/mac/win matrix legs and produces `daemon/dist/ccsm-daemon-*` per leg.
- [ ] Subsequent `Package (...)` electron-builder run picks up the real binary via `before-pack.cjs` (no placeholder warning in logs).
- [ ] Installer spawns the daemon at first launch (dogfood worker re-run).

## Spec gaps surfaced (out of scope, follow-up tasks)

- `daemon/native/ccsm_native/` source from frag-3.5.1 §3.5.1.1 still not in the tree. The rebuild script skips with a warn; `before-pack.cjs` `REQUIRED_NATIVES` will fail-loud once frag-3.5.1 lands.
- node-pty 1.1.0 has no `prebuilds/win32-x64/` shipped on npm; root cause is winpty toolchain. Untouched here — needs its own task or a node-pty version bump.

Closes #45.

Generated with Claude Code (Opus 4.7).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,15 @@ jobs:
       - name: Build web + electron
         run: npm run build
 
+      # Build the standalone daemon binary (frag-11 §11.5 step 2 + §11.1
+      # Task 20). Produces daemon/dist/ccsm-daemon-${platform}-${arch}${ext}
+      # for the host runner via @yao-pkg/pkg, with native modules
+      # (better-sqlite3, node-pty, ccsm_native) rebuilt against Node 22 ABI
+      # and staged into daemon/native/${platform}-${arch}/ for the
+      # before-pack hook to copy into the installer payload.
+      - name: Build daemon binary
+        run: npm run build:daemon-bin
+
       # electron-builder invocation per OS. `--publish never` because we
       # upload via softprops/action-gh-release below — that gives us a
       # single source of release metadata and keeps `publish: github` in
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ docs/screenshots/dogfood-happy-path/
 *.blockmap
 .pool-synced-at-*
 daemon/dist/
+daemon/dist-daemon/
+daemon/dist-daemon-bundle/
 daemon/native/
 daemon/native-staged/
 daemon/sdk-staged/
--- a/daemon/package.json
+++ b/daemon/package.json
@@ -1,11 +1,40 @@
 {
   "name": "@ccsm/daemon",
+  "version": "0.3.0",
   "private": true,
   "type": "module",
+  "main": "dist-daemon-bundle/index.cjs",
+  "bin": "dist-daemon-bundle/index.cjs",
+  "scripts": {
+    "build": "tsc -p tsconfig.json && node -e \"require('fs').writeFileSync('dist-daemon/package.json',JSON.stringify({type:'module'}))\"",
+    "package": "pkg . --out-path dist --no-bytecode --public-packages \"*\" --public"
+  },
+  "pkg": {
+    "scripts": [
+      "dist-daemon-bundle/index.cjs"
+    ],
+    "assets": [
+      "native/${PLATFORM}-${ARCH}/*.node",
+      "native/${PLATFORM}-${ARCH}/*.dll",
+      "package.json"
+    ],
+    "outputPath": "dist"
+  },
   "dependencies": {
     "@connectrpc/connect": "^2.1.1",
     "@connectrpc/connect-node": "^2.1.1",
+    "@octokit/rest": "^21.0.2",
     "@sentry/node": "^10.51.0",
-    "@sinclair/typebox": "^0.34.49"
+    "@sinclair/typebox": "^0.34.49",
+    "better-sqlite3": "^12.9.0",
+    "node-pty": "1.1.0",
+    "pino": "^9.5.0",
+    "pino-roll": "^3.0.0",
+    "proper-lockfile": "^4.1.2",
+    "ulid": "^2.3.0"
+  },
+  "devDependencies": {
+    "@yao-pkg/pkg": "^6.6.0",
+    "node-gyp": "^10.2.0"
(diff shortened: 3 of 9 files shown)


Jiahui-Gu:
## Review — request changes

### Layer 1 (direction)
Right. esbuild→CJS→pkg per spike Fallback A is the correct call (pkg snapshot fs only intercepts CJS `require()`, not ESM `import()`); the spec deviation is justified and PR body documents it. async IIFE around `await listen()` preserves behaviour — signal handlers stay registered synchronously at module level (lines 374-375), reject still re-throws to the crash handler. release.yml step is in the right slot (between `Build web + electron` and `Package (...)`). Daemon `package.json` deps + `pkg` config + `@yao-pkg/pkg ^6.6.0` all match frag-11 §11.1 round-2 C6. No drive-by edits to unrelated daemon source. Two flagged spec gaps (`daemon/native/ccsm_native/` source missing, node-pty 1.1.0 Win prebuild absent) are correctly out of scope and surfaced.

### Layer 2 — blocker

`buf` job already failed at the `npm ci` step:
```
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: electron-builder-squirrel-windows@26.8.1 from lock file
npm error Missing: electron-winstaller@5.4.0 from lock file
npm error Missing: @electron/windows-sign@1.2.2 from lock file
npm error Missing: fs-extra@7.0.1 from lock file
... (15 entries total)
```

The lock-file diff in this PR **deletes** the `node_modules/@electron/windows-sign/*` block (and several transitive optional/peer entries) without removing the consumer that needs them. Looks like the lock got regenerated via `npm install` on a host where these Windows-only optional/peer deps were pruned. Every job that runs `npm ci` (the entire matrix: lint+typecheck+test, e2e, parity, size-guard, plus the new release.yml `Build daemon binary` step itself) will fail the same way — this isn't infra noise, the whole CI gate is unreachable.

Fix: rerun `npm install --include=optional --include=peer` (or `npm install` on a host where `@electron/windows-sign` resolves) and recommit `package-lock.json`. Verify with `npm ci` locally before pushing.

Will re-review once CI is green.

PR #787, opened 2026-05-02
docs(spec): multi-client architecture proposal (3 clients, Connect, CF Access)

Jiahui-Gu:
## Summary

Proposal doc for backend-authoritative, 3-first-class-client (desktop / web / iOS) ccsm architecture. Reframes the v0.4 web design around the user-locked constraints: standalone backend, multi-client concurrent, local + remote, GitHub OAuth only, Cloudflare-fronted for non-desktop, PTY as stream.

Picks one design per axis with one-sentence rejections of alternatives:

- **Transport:** Connect-RPC over HTTP/2 + Protobuf v3 (rejects gRPC-Web / tRPC / WebSocket+JSON / TypeSpec).
- **Auth:** Cloudflare Access (GitHub OAuth IdP) for all remote ingress; localhost desktop bypass via peer-cred / loopback tag (no backend-issued tokens).
- **Session model:** snapshot + delta-from-seq replay (already shipped in v0.3 PTY hardening); broadcast-all + last-writer-wins at PTY layer for multi-client writes (no locks).
- **Topology:** single backend binary + cloudflared sidecar; no separate edge gateway.
- **Throw-away vs v0.3:** ~900-1100 LOC of envelope + HMAC handshake + chunk-reassembly (data plane only; supervisor control plane keeps v0.3 envelope).
- **Migration:** single coherent slice; iOS may slip to v0.4.1.

Contains 7 specific open questions for user (desktop tech choice, remote deploy target, iOS distribution path, multi-client write semantics, Electron co-existence, Cloudflare account model, scrollback persistence). Notes one soft tension with v0.3 supervisor model assuming Electron-main parent.

Refs Task #82.

## Test plan

- [ ] User reads §G open questions and picks per-axis answers.
- [ ] User confirms transport recommendation (Connect) vs reopens vs alternatives.
- [ ] After answers, this proposal becomes the source for a real spec under \`docs/superpowers/specs/v0.4-*\` (or v1.0-*).

(Doc-only PR; no code changes, no tests, no lint impact.)

### DIFF
--- a/docs/superpowers/specs/2026-05-02-multi-client-architecture-proposal.md
+++ b/docs/superpowers/specs/2026-05-02-multi-client-architecture-proposal.md
@@ -1,0 +1,195 @@
+# Multi-client architecture proposal
+
+**Status:** proposal (pre-spec)
+**Date:** 2026-05-02
+**Author:** ccsm
+**Supersedes (intent):** `2026-05-01-v0.4-web-design.md` (single-user, Electron-primary, web-as-secondary). This proposal reframes the same ingress story around a backend-authoritative, three-client topology.
+**Hard inputs (locked by user, not negotiable in this doc):**
+
+1. Backend is a standalone process and the **single source of truth** for sessions (PTY, claude CLI subprocess, SQLite, crash logging, cwd state).
+2. Three first-class clients: **desktop** (Tauri or native; not necessarily Electron), **web browser**, **iOS app**. None is "primary".
+3. Multiple clients may attach to the **same live session simultaneously**.
+4. Backend may run on **localhost OR remote** (VPS / home NAS); clients reach it from anywhere on the internet.
+5. Auth = **GitHub OAuth**. Same identity across all clients. No custom username/password.
+6. Non-desktop traffic (web + iOS) **MUST traverse Cloudflare** (Tunnel or Workers). Desktop MAY bypass when on localhost / LAN.
+7. PTY output is a **stream**, not request/response.
+
+This doc picks one design per axis, names what currently-shipping code stays, names what gets thrown out, and lists the open questions the user has to answer before this becomes a spec.
+
+---
+
+## A. Transport selection
+
+**Recommendation: Connect-RPC over HTTP/2 (with HTTP/1.1 fallback for browser long-poll), Protobuf v3 wire format, server-streams for PTY output.**
+
+This is exactly what `2026-05-01-v0.4-web-design.md` §2 already locked for the data plane. It survives every constraint above with no rewrite:
+
+- **Browser:** `@connectrpc/connect-web` runs in any modern browser without an Envoy proxy; speaks plain HTTPS that Cloudflare Tunnel passes through unchanged. Server-stream RPC (the one shape PTY needs) works half-duplex over fetch streaming, which is what every browser actually ships.
+- **iOS:** `connect-swift` (Buf, Apache-2.0) generates a native Swift client from the same `proto/` schema. Uses URLSession; no custom TLS code. Same `proto/` source = same RPC surface as web and desktop, no parallel maintenance.
+- **Desktop (Tauri / native / Electron):** `connect-go` / `connect-node` / `connect-kotlin` all exist. Tauri Rust side gets Connect-Go-class code via `prost` + a thin Rust Connect runtime, or it speaks the same `connect-web` HTTP shape since Tauri renderers are real WebViews.
+- **Streaming semantics:** Connect server-streams are first-class. PTY output (the only true stream we have today) maps 1:1 to a single `subscribePty` server-stream RPC. Client→server keystrokes are unary RPCs (one per batch); we do not need bidi.
+- **Cloudflare Tunnel compatibility:** Tunnel proxies plain HTTPS / HTTP/2. Connect's `application/proto` and `application/json` content-types pass through verbatim — confirmed by the v0.4 spike (v0.4 §5).
+- **Auth header model:** Cloudflare Access injects `Cf-Access-Jwt-Assertion` on every request; Connect interceptors read headers natively per RPC. No protocol gymnastics.
+- **Debuggability:** the JSON content-type variant (`application/json`) makes every Connect RPC `curl`-able from a terminal with no codegen. The hand-rolled v0.3 envelope is not.
+
+**Rejected alternatives** (one-line each):
+
+- **Plain HTTP/2 + WebSocket (no schema):** punts the v0.3 problem (no codegen, no breaking-change gate, no shared schema across 3 clients) into v0.5 again. Rejected.
+- **gRPC-Web:** requires a translating proxy (Envoy or equivalent); Cloudflare Tunnel is not one. Rejected.
+- **Plain gRPC:** browsers can't speak it without a proxy. Rejected for the same reason.
+- **tRPC:** TypeScript-only — kills iOS and any future native client. Rejected.
+- **GraphQL subscriptions:** wrong shape for raw byte streams (PTY output is bytes, not a typed event stream); WebSocket transport adds a second protocol surface. Rejected.
+- **Custom JSON-WebSocket envelope (current v0.3 + a WebSocket lift):** works but reinvents wire-versioning, breaking-change gating, and per-language codegen by hand for every new client. We would re-derive `buf breaking` ourselves. Rejected.
+
+---
+
+## B. Auth model
+
+**Recommendation: Cloudflare Access (GitHub OAuth IdP) for all *remote* ingress. Same per-app `AUD` JWT validated by the backend on every remote request. Desktop on localhost bypasses the JWT check via a peer-cred / loopback-trust tag (same shape as v0.3 §3.1.1 + v0.4 §5 `localTransportKey`). No backend-issued tokens.**
+
+This is the cheapest design that satisfies constraints 5 + 6:
+
+- **All three clients sign in to the same GitHub account through Cloudflare Access**, which is the `email == <user-github-email>` policy already specified in v0.4 §5. Identity is identical across clients because the IdP is identical. Backend never stores a user database.
+- **Per-client OAuth flow** — each client uses the platform-native browser hand-off, not its own embedded login UI:
+  - **Desktop (Tauri / Electron):** `ASWebAuthenticationSession`-equivalent. On macOS / Windows / Linux this is "open the OS default browser, intercept the redirect via a registered `ccsm://` scheme handler or a localhost loopback `http://127.0.0.1:<random>/callback`". **PKCE** authorization-code flow (NOT device flow). Device flow is for headless clients and gives a worse UX (paste-the-code-into-a-browser) for an app that owns a window.
+  - **Web:** standard OAuth 2.0 redirect. Cloudflare Access does the entire flow before the SPA bytes are even served — the SPA only ever sees an authenticated request.
+  - **iOS:** `ASWebAuthenticationSession` (Apple's native API, gives the app a Safari-shared cookie jar so the user is single-sign-on'd if they've authenticated to GitHub on Safari recently).
+- **Where tokens live:**
+  - Desktop: OS keychain (macOS Keychain, Windows Credential Manager, libsecret on Linux).
+  - iOS: iOS Keychain.
+  - Web: `Cf-Access-Jwt-Assertion` cookie set by Cloudflare Access; SPA never touches the JWT directly.
+- **Backend validation:** verify the **Cloudflare Access JWT** (RS256, per-app AUD, team issuer) on every remote request, exactly per v0.4 §8.1. No JWT exchange, no backend-issued JWT, no refresh-token plumbing on our side — Cloudflare Access owns the entire token lifecycle (refresh, revoke, expiry). The backend's job is "verify and trust".
+- **Localhost desktop bypass:** the local socket / loopback HTTP listener tags the connection with `localTransportKey: true` (per v0.4 §5 wording); the JWT-validation interceptor is a no-op on locally-tagged requests. This is the same bypass trick the v0.4 doc already wrote and the test plan already covers (v0.4 §8.1).
+- **Session model on the backend:** the backend treats every authenticated request as "the one user". Multi-tenant is explicitly out of scope (constraint 5 = "same identity across all clients", not "support N users"). Re-introducing a real user table is a v0.6+ conversation.
+
+**Rejected alternatives** (one-line each):
+
+- **Build our own OAuth + session JWT issuer:** explicitly forbidden by constraint 5. Rejected.
+- **Verify GitHub access tokens directly via GitHub's `/user` API:** N+1 GitHub API call per RPC, rate-limited, slow. Cloudflare Access already did this and gave us a JWT — using it is free. Rejected.
+- **Issue our own short-lived JWT after verifying the CF Access JWT once:** adds a second token lifecycle (refresh, revoke) we'd have to maintain; gives no security or perf win that justifies the surface. Rejected.
+- **Device flow on desktop:** clunky UX (browser tab + paste 8-char code) for a windowed app. PKCE is strictly nicer. Rejected.
+
+---
+
+## C. Session model
+
+**Recommendation: backend-authoritative session list; clients are pure subscribers. New client mid-session catches up via (snapshot + delta-from-seq) — already what v0.3 PTY hardening shipped. Writes from N clients are broadcast-all + last-writer-wins at the PTY layer (no locks, no "primary client").**
+
+The v0.3 PTY-hardening fragment (`frag-3.5.1`) and envelope-hardening fragment (`frag-3.4.1` §3.4.1.b) already specified the catch-up shape; this proposal just confirms it generalizes from "1 client at a time" to "N clients at a time".
+
+- **Snapshot + delta** (already shipped in v0.3): `subscribePty` opens with `fromSeq` / `fromBootNonce` headers. Daemon ships the bounded snapshot (`xterm-headless` rendered buffer, capped at the existing scrollback), then live deltas with monotonic `seq`. Reconnect within the replay budget (256 KiB per v0.3 §3.4.1.b) replays only the gap; reconnect outside the budget gets a fresh snapshot with `gap: true`.
+- **Bounded scrollback:** the existing `xterm-headless` ring buffer is the bound. We do NOT keep "full session history" in RAM for replay. Persistence to SQLite for long-tail history is a separate v0.6+ feature, not in scope here.
+- **Write arbitration — broadcast all, last-writer-wins at the PTY:** every client's keystroke RPC (`ptyWrite`) writes to the PTY in arrival order at the daemon. The PTY itself is the serialization point — there is no "active client" lock. If two clients type at once, the bytes interleave at byte-granularity exactly the way two `cat` processes writing to the same tty would. This is correct because:
+  1. PTYs are inherently byte-streams; trying to add line-level locking is the wrong layer.
+  2. The two-typist case is rare in practice (you're collaborating with yourself across devices, not contending).
+  3. Every client sees its own + the other's keystrokes echoed back through the same `subscribePty` stream, so divergence is impossible — the server's PTY is the single source of truth.
+- **Per-client cursor / scroll position:** that lives **client-side**, not on the backend. The backend ships the buffer; the client decides what window of it to render.
+- **Subscriber registry:** the v0.3 fan-out registry (`frag-3.5.1` §3.5.1.5) already supports N subscribers per session with the drop-slowest watermark at 1 MiB. The "single subscriber" assumption was a temporary perf carve-out, not a fundamental limit; we lift it for v0.4.
+
+**Rejected alternatives** (one-line each):
+
+- **Active-client lock with explicit handoff:** UX nightmare ("device A: do you want to give control to device B?"); breaks the "leave it running and pick up anywhere" use case. Rejected.
+- **CRDT / OT for keystrokes:** PTY is not a text editor; the OS kernel resolves byte ordering for us. Rejected.
+- **Full session replay from disk:** the existing scrollback bound is enough for "I closed my laptop and came back". Long-tail history is a separate feature. Rejected.
+
+---
+
+## D. Deployment topology
+
+**Recommendation: single backend binary listening on two endpoints — (1) localhost loopback HTTP/2 (peer-cred / loopback-trust tagged, JWT bypass), (2) `cloudflared` tunnel sidecar pointing at the same loopback port. NO separate edge gateway in our stack; Cloudflare IS the edge. Desktop on the same machine hits endpoint (1); desktop on another machine + web + iOS all hit endpoint (2).**
+
+- **Single backend binary:** the daemon already exists (`daemon/src/index.ts`). The only addition is replacing the v0.3 control-plane / data-plane split-socket model with a Connect HTTP/2 listener (control-plane stays on its own listener for supervisor isolation per v0.3 §3.4.1.h — that's not changing). Same binary, two listeners, no microservice split.
+- **`cloudflared` as a sidecar process** spawned and lifecycled by the backend (this matches v0.4 §5's "Tunnel spawned by daemon when remote access is enabled"). User flips a single setting → backend spawns `cloudflared`, registers the tunnel, advertises the public URL via the tray UI. No CF Workers in front of the tunnel; Workers add a second deploy target (with its own routing config drift) for zero benefit when Tunnel already terminates TLS for us.
+- **Cloudflare Access in front of the tunnel** — zero-trust application configured to require GitHub OAuth, per-app AUD. This is the auth boundary. The backend does not implement OAuth itself.
+- **Desktop bypass path:** desktop running on the same machine as the backend hits `http://127.0.0.1:<port>` directly. Same Connect protocol, same RPC surface, same proto-generated client code — only the URL and the `localTransportKey` middleware tag differ. Desktop on a *different* machine (e.g. Tauri client on the laptop, backend on the home NAS) goes through Cloudflare exactly like web and iOS — there is no third path. This matches constraint 6 ("Desktop MAY bypass when on localhost"; on remote it just doesn't bypass).
+- **Discovery for desktop:** desktop client first tries `http://127.0.0.1:<known-port>` with a 200ms timeout; on failure, falls back to the user-configured remote URL. This makes "moved my laptop off the home network" automatic.
+
+**Rejected alternatives** (one-line each):
+
+- **Separate edge-gateway service we operate:** doubles deploy surface, gives us a routing-config target that can drift from the backend. Cloudflare Tunnel + Access already does this for free. Rejected.
+- **Cloudflare Workers in front of Tunnel:** adds a third deploy target (`wrangler.toml`, separate codepath) for zero benefit on the streaming RPC path. Workers shine when you want edge compute; we want pass-through. Rejected.
+- **Backend listens directly on a public port (no Cloudflare):** loses the auth layer (we said no custom auth), exposes the user's home IP, requires user to deal with NAT / firewall / TLS cert renewal. Rejected — explicitly violates constraint 6.
+- **Per-client backend instance (one daemon per client):** breaks constraint 1 (single source of truth). Rejected.
+
+---
+
+## E. Delta vs current v0.3
+
+This proposal is intentionally a **transport + ingress** change. The session model, PTY hardening, crash observability, and SQLite layer are correct as v0.3 shipped them. The throw-away surface is the transport bytes.
+
+**Survives unchanged:**
+
+- PTY hardening (`frag-3.5.1`): xterm-headless buffer, snapshot + delta, fan-out registry with drop-slowest watermark, server-side dead-stream detector. ~all of `daemon/src/pty/` (when it lands; not yet on `working` per `daemon/src/index.ts:96` "wires are still landing"). **Becomes the streaming spine of every client.**
+- Crash observability (`daemon/src/crash/`, `daemon/src/sentry/`): Phase 1-4 all survive — they are about the daemon process, not the wire. ~600 LOC, no churn.
+- SQLite schema + migration (`frag-8`): backend-local, no client-visible change. **No churn.**
+- Daemon supervisor process model: spawn-or-attach, lockfile, crash-loop detection, `pino-roll` rotation, `daemon.shutdownForUpgrade` marker, dual-socket isolation at the OS level. **No churn** — these are about owning the backend process, not about how clients reach it. (The split between supervisor-control-plane and data-plane sockets becomes "supervisor still on local UDS, data-plane on Connect HTTP/2"; same architectural shape.)
+- Cwd state, session lifecycle FSM, claude CLI subprocess management, env subset propagation. All backend-internal.
+
+**Thrown out:**
+
+- **Length-prefixed JSON envelope** (`daemon/src/envelope/envelope.ts`, ~200 LOC): replaced by Connect's HTTP/2 framing. Connect inherits the 16 MiB cap natively (`readMaxBytes`).
+- **Hello-HMAC handshake** (`daemon/src/envelope/hello-interceptor.ts` + `hmac.ts` + `chunk-reassembly.ts` + `boot-nonce-precedence.ts` + `protocol-version.ts` + `trace-id-map.ts`): replaced by Cloudflare Access JWT (remote) + peer-cred / loopback-tag (local). The HMAC was a same-machine secret-possession check; with peer-cred + DACL the secret is redundant on the local socket, and on the remote socket Cloudflare Access is doing strictly more than the HMAC ever did. ~420 LOC + helpers ~250 LOC = **~670 LOC removed from the data plane**. (`hello-interceptor` survives on the supervisor control plane unchanged — control plane is local-only and keeps the v0.3 envelope per v0.4 §2.6 reasoning.)
+- **Manual chunking / binary-trailer carve-out** (`frag-3.4.1` §3.4.1.b/c, ~30 LOC in `connectAdapter.ts` + symmetric reader): replaced by Connect's native HTTP/2 frame management + `bytes` field type for PTY payloads. Wire-level chunking semantics preserved; we delete our hand-rolled version.
+- **`unix socket / named pipe` as the data-plane transport:** stays for the supervisor / control plane; the data plane moves to loopback HTTP/2 + remote HTTP/2-via-Tunnel. The OS-level peer-cred check generalizes to "is the socket bound to 127.0.0.1?" for the loopback HTTP path, a one-line check.
+- **Migration-gate interceptor as a v0.3-envelope-specific concern:** survives semantically (the rule "data-plane RPCs are 503 during migration, control-plane is not" still applies) but the implementation moves to a Connect interceptor.
+- **Deadline interceptor:** survives semantically; ~30 LOC reimplemented as a Connect interceptor.
+- **Trace-id map / chunk-reassembly state machines:** Connect handles this natively via HTTP/2 stream id + Connect's per-call header carriage. Throw out.
+
+**Rough LOC throw-away (data plane only):** ~900-1100 LOC of envelope + hello + adapter + chunk-reassembly + handler-wrapping. Roughly tracks the v0.4 §2 estimate of "the entire `daemon/src/envelope/` directory disappears from the data-plane fast path" — proposal extends that decision to also delete the v0.4 doc's Electron-vs-web differential, since *all three* clients now speak the same Connect surface.
+
+**`electron/daemonClient/` (`controlClient.ts`, `rpcClient.ts`, `envelope.ts`, `streamHandleTable.ts`):** envelope.ts + streamHandleTable.ts → throw out (~400 LOC). controlClient.ts stays (talks to the supervisor plane which keeps the v0.3 envelope). rpcClient.ts gets ported to the proto-generated Connect client (~150 LOC of mechanics → ~50 LOC because Connect-Node provides the call mechanics).
+
+---
+
+## F. Migration path
+
+**Recommendation: ship in a single coherent slice as v0.4 (renamed to v1.0 if the user prefers a major-bump signal). Subset shipping is technically possible but yields a worse outcome per axis — see below.**
+
+Why one slice:
+
+- **Connect data plane + Cloudflare Access JWT + multi-client coherence are the same change from the backend's POV.** All three flow through the same RPC surface — splitting them means writing two transport adapters (envelope + Connect) and maintaining both, which is exactly the trap v0.4's anti-goal A1 was warning against.
+- **iOS slips one slice OK** because it's purely a new client against an unchanged backend. Recommendation: backend + desktop port + web client lands in one slice (call it v0.4); iOS native app follows in v0.4.1 / v0.5 with **zero backend churn** because the proto schema is already final.
+
+**If the user insists on subset:**
+
+- **Smallest viable subset = "v0.4 desktop-only over Connect, GitHub OAuth deferred":** desktop on Connect + localhost loopback only, no Cloudflare, no web, no iOS. **Rejected as a recommendation** because (a) the per-client OAuth flow is the riskiest piece of the design and deferring it just defers the risk; (b) "desktop-only over Connect" is identical to v0.4 §3 bridge-swap which the user already has a draft of; this proposal would add zero value.
+- **Recommended subset (if you must):** v0.4 = backend Connect data plane + desktop port + web client + Cloudflare Access. v0.4.1 = iOS native client. This puts the high-risk auth flow in v0.4 (caught by 2 dogfood clients), and adds iOS only after the wire is proven on web.
+
+**Hard sequencing constraints inside the slice:**
+
+1. `proto/` schema first (must be complete before any client can codegen).
+2. Backend Connect server next (`@connectrpc/connect-node` mounted on loopback HTTP/2; v0.3 envelope removed from data plane; supervisor envelope untouched).
+3. Desktop client port (proto-generated Connect-Node client replaces `electron/daemonClient/rpcClient.ts`).
+4. **Cloudflare Tunnel + Access wiring** (cloudflared sidecar; backend JWT interceptor; localhost bypass tag).
+5. Web client (Vite SPA against the same Connect surface).
+6. iOS client (Connect-Swift; same proto).
+
+Steps 1-3 are mandatory together (you can't ship the backend without a client to test it). Steps 4-6 can each land in their own PR once 1-3 are green. Step 6 may slip to a follow-up release without holding 1-5.
+
+---
+
+## G. Open questions for user
+
+These are the specific bits this proposal could not decide. Each is a fork in the architecture; the user has to pick before this becomes a real spec.
+
+1. **What's the desktop tech?** Tauri (Rust) vs Electron (TypeScript) vs SwiftUI-on-mac+WPF-on-win (native per OS). Each has different `proto/` codegen story, different OAuth-flow primitives, and very different reuse with the existing `electron/` tree. **Tauri is the highest-leverage answer if you're willing to write Rust**; Electron is the lowest-risk because the existing renderer code ships unchanged.
+
+2. **Backend deploy target on remote:** Linux daemon binary on a VPS / NAS, or a containerized deployment, or a Mac mini at home running it natively? This decides the supervisor layer (systemd unit vs launchd plist vs Docker `restart: always` vs Windows Service). The v0.3 supervisor model assumes "Electron parent owns the daemon" — that breaks when the daemon runs without an Electron alongside it.
+
+3. **iOS app distribution:** App Store (requires Apple Developer account, review, no sideload), TestFlight (90-day timer, public beta cap of 10k), or self-signed sideload only (works on personal device, fragile)? This decides whether iOS work is on the v0.4 critical path or a long-tail follow-up.
+
+4. **Multi-client write semantics — really last-writer-wins?** Per §C above the recommendation is "PTY is the serialization point, byte interleave is fine". But the user might want a softer answer — e.g. "show me when another client is typing" (presence indicator), or "lock the PTY to one client at a time but make handoff one-tap". Either is implementable; both add UI surface this proposal didn't budget.
+
+5. **Do we keep Electron alongside the new desktop client during transition,** or hard-cut? If hard-cut, the current `electron/` tree is ~14 weeks of v0.3 work that goes away in one release. If keep-alongside, we maintain two desktop clients for 2-3 releases. This is a question about user appetite for risk + how much of the existing renderer code is reusable in the new desktop choice.
+
+6. **Cloudflare account ownership:** is this a single CF account the user owns (matches v0.4 design's single-tenant assumption), or do we anticipate friends-of-friends self-deploying their own backend + their own Cloudflare account? The latter requires us to ship a setup wizard ("paste your Cloudflare API token here") which is non-trivial UX. The former assumes the user is the only deployer ever.
+
+7. **PTY scrollback persistence:** v0.3 keeps scrollback in RAM only (xterm-headless ring buffer). With multiple clients over high-latency links, a client that disconnects for an hour can't replay an hour of output — it only gets a fresh snapshot (current screen) on reconnect. Is that acceptable, or do we need to spill scrollback to SQLite for long-replay? This decides whether SQLite gains a `pty_scrollback` table with retention policy in v0.4 or stays as today.
+
+---
+
+## Pushback / contradictions found
+
+None of the seven hard constraints are internally contradictory. The proposal above satisfies all seven without compromise. The only soft tension is constraint 2 ("no primary client") vs the practical reality that the **existing** codebase has Electron deeply wired in (`electron/` tree + 46 IPC bridges + renderer assumptions). Cleanly satisfying constraint 2 requires either (a) treating the existing Electron client as one of the three first-class clients (port it forward, keep it), or (b) hard-cutting it. Question G5 surfaces this for the user.
+
+There's also a soft tension between constraint 4 (backend may run remote) and v0.3's `daemonSupervisor` model (Electron-main owns the daemon process lifecycle). If the daemon runs on a VPS without an Electron alongside it, the supervisor code path needs a "headless mode" — systemd / launchd / Docker takes the supervisor's job. The v0.3 design doc explicitly punted this in v0.4 §1 N5 ("Headless daemon with no Electron — Why deferred"). Question G2 surfaces this for the user.


Jiahui-Gu:
**Arch review — ops/topology angle (Task #83)** — DRAFT, not approve/reject. Scope = lifecycle of `cloudflared` sidecar + tunnel topology under user refinement (backend = local-only; web/iOS via daemon-managed `cloudflared`). Auth, transport choice, multi-client UX explicitly out of scope.

The proposal collapses the deploy story into "single binary + cloudflared sidecar" (§D, lines 99-105). v0.4 §5.1 already worked through most of the sidecar lifecycle (bundled binary, supply chain, slow-tick recovery, network-up restart, JWKS pre-warm). Most of the "lifecycle" wins below are **the proposal not citing what's already locked in v0.4 §5**, not new gaps. The actual new ops gaps are around (a) backend running standalone without an Electron parent, (b) two listeners with port discovery, and (c) multi-machine collisions on a single CF Tunnel — none of which v0.4 §5 considers because v0.4 §5 assumed Electron-parent + single host.

---

### Findings

**F1 (P0) — Supervisor ownership when Electron is closed.** §D line 101 says "the daemon already exists" and §F line 161 keeps the v0.3 supervisor envelope. v0.3 frag-6-7 §6.1 explicitly assumes Electron-main spawns + heartbeats the daemon (5s ping, crash-loop detection, `.bak` rollback). With backend-本机-only + web/iOS clients connected through `cloudflared`, closing the Electron tray window MUST NOT kill the daemon — but if Electron-main dies, who heartbeats? Who triggers `.bak` rollback? Who consumes `daemon.shutdownForUpgrade`? The proposal hand-waves this in the closing paragraph (line 195) and defers to G2.
**Fix:** lock a "headless supervisor mode" in this proposal, not in G2. Concrete shape: launchd plist (mac) / Windows Service (win) / systemd-user (linux) becomes the OS-level supervisor; v0.3's in-process supervisor degrades to "process-internal crash-loop counter + .bak rollback", with the OS supervisor taking the "respawn on death" job. Without this lock, the v0.3 supervisor invariants in frag-6-7 §6.1/§6.4 silently break the moment a web/iOS user is connected and the desktop Electron quits.

**F2 (P0) — `cloudflared` lifecycle inheritance from killed Electron.** §D line 102 says "cloudflared as a sidecar process spawned and lifecycled by the backend". v0.4 §5.1 spawns `cloudflared` with `detached: false` ("dies with daemon"). If the v0.3 daemon was a child of Electron-main and Electron-main is killed (Cmd-Q, Activity Monitor force-quit, log-out), the kernel reaps Electron → daemon → cloudflared in cascade. **Result: web/iOS user's tunnel disappears the instant the desktop user closes their tray.** This is the opposite of what "backend = 本机, clients via Cloudflare" implies.
**Fix:** explicit decision in this proposal: "daemon detaches from Electron at first launch (double-fork on POSIX, `DETACHED_PROCESS` on Win); Electron is a *client* of the daemon, not its parent." Combine with F1's OS supervisor.

**F3 (P0) — Two listener ports, no discovery contract.** §D line 99 names two endpoints (loopback for desktop peer-cred bypass, loopback target for `cloudflared`). §D line 105 says "desktop client first tries `http://127.0.0.1:<known-port>`". v0.4 §5.1 hard-codes prod TCP `127.0.0.1:7879` for cloudflared and dev `:7878` for browser; v0.3 frag-6-7 line 300 OBSOLETES "127.0.0.1:7878" entirely. The proposal does not say which port is which, whether they're fixed or dynamic, or how desktop discovers them when the user runs a second instance / has port 7879 in use.
**Fix:** lock now: (a) loopback peer-cred listener = unix-socket / named-pipe (`~/.ccsm/daemon.sock` / `\\.\pipe\ccsm-daemon-<userSid>`) — NO TCP, inherits v0.3 frag-6-7 §6 hardening unchanged. (b) cloudflared-target listener = `127.0.0.1:0` (kernel picks free port), port written to `~/.ccsm/cloudflared-target.port` for cloudflared spawn args — eliminates 7879 collision with user's own services. Discovery file is the contract; second-instance gets EADDRINUSE on the lockfile path before it ever tries to bind, per v0.3 §6.4 lockfile rules.

**F4 (P1) — Single-tenant CF Tunnel + multi-machine drift.** Proposal lines 7+12 frame "backend may run on localhost OR remote", and §G6 (line 185) flags "single CF account" assumption. Not addressed: user installs ccsm on **laptop AND desktop**, both auto-spawn cloudflared, both register. v0.4 §5.6 step 3 stores tunnel token in OS keychain — **per-machine**. So each machine creates its own tunnel against the same CF account → two tunnels, two hostnames, two AUDs, web client lands on whichever the user typed. Worse: if the user re-uses the same tunnel token across machines (copy-paste), CF load-balances → web/iOS sessions ping-pong between hosts mid-stream.
**Fix:** add §D sub-section "single-host invariant": at most one ccsm tunnel per CF account; daemon refuses to spawn cloudflared if it detects a sibling tunnel is already routed to this account's hostname (poll CF API at startup, or write `host_id` into the daemon's `/healthz` and have cloudflared metrics expose the upstream). Multi-host = explicit v0.5 work; v0.4 fails closed with "Already running on <other-host>; disable there first" banner. This is the same invariant v0.4 §5 §3 "single user, one Access policy" implies but never enforces.

**F5 (P1) — Tunnel teardown on uninstall + machine swap.** Frag-6-7 §6.7 covers Windows MSI uninstall (kill before unlock) but says nothing about CF resources. Proposal G6 punts. Realistic failure: user reinstalls OS, CF dashboard still shows the old tunnel routed to a dead host; web/iOS client sees Cloudflare "1033 Argo Tunnel error" with no indication it's the dead host. Worse on uninstall: tunnel token stays in CF account forever (until user manually revokes), leaking a long-lived credential.
**Fix:** uninstall hook (Win MSI postuninstall, mac `.pkg` post-flight, linux postrm) calls `cloudflared tunnel delete <uuid>` using the locally stored token before wiping keychain. **Idempotent** — if the token is already invalid (CF-side revoke), succeed-silently. First-run wizard (v0.4 §5.6) should ALSO offer a "claim existing tunnel from <other host>" path that takes over the tunnel (CF supports tunnel-relocate via the connector ID). Without this the "user got a new laptop" path is broken by design.

**F6 (P1) — Crash-loop semantics for cloudflared specifically.** v0.4 §5.1 lifecycle bullets cover `cloudflared` health: 60s metrics poll, 3-fail restart, exponential backoff cap 60s, 10 attempts in 30 min, then slow-tick 1/5min indefinitely. Proposal §D line 102 inherits this but does not say what happens when **daemon's own crash-loop detector (frag-6-7 §6.1, ≥5 respawns / 2-min sliding window) trips because of cloudflared crash-cascade**. Today the daemon's crash-loop counter doesn't distinguish "daemon crashed" from "cloudflared crashed and daemon's spawn-supervisor surfaced it as daemon-side error".
**Fix:** explicit isolation lock — daemon's §6.1 crash-loop counter NEVER increments from a cloudflared exit, even non-zero. Cloudflared has its own counter (v0.4 §5.1 already locks this). Cross-contamination would mean a flaky CF edge bricks the local daemon for the desktop user.

**F7 (P1) — User's pre-existing cloudflared install.** §D doesn't consider that the user may already run `cloudflared` for unrelated reasons (their blog, their homelab, a separate Tunnel). Bundled binary + spawned-by-daemon means we shadow the user's PATH `cloudflared`, but the user's existing `cloudflared service` (systemd / launchd) may be running with conflicting metrics ports.
**Fix:** spawn args inherit v0.4 §5.1 `--metrics 127.0.0.1:0` (ephemeral; safe). Add: pass `--config` pointing at a daemon-controlled config file in `~/.ccsm/cloudflared-config.yml`, never read user's `~/.cloudflared/config.yml`. Document this in the proposal so a future implementer doesn't "helpfully" auto-detect the user's tunnel and reuse it.

**F8 (P1) — Resource budget when remote access is OFF.** Proposal does not say whether `cloudflared` runs always or only when remote access is enabled. v0.4 §5.6 "Disabling remote access" path stops cloudflared and closes the prod TCP listener — so it's gated. Good. But the proposal §D should re-state this as a lock so that the multi-client framing doesn't accidentally promote "cloudflared always-on" to default.
**Fix:** explicit lock — "cloudflared spawn is gated by `settings.remote_access_enabled = true`. Default = false on first install. Idle daemon footprint MUST stay at v0.3 baseline (~40MB RSS) when remote access is off — CI gate." Without this lock, every user pays the cloudflared baseline (~30-50 MB RSS, persistent outbound TCP) for a feature most won't use.

**F9 (P2) — Cloudflared logs in v0.3 crash collector.** v0.4 §5.1 logs to `~/.ccsm/cloudflared.log` via stdout tee, pino-roll rotation, 50MB×5. Proposal does not wire this into the v0.3 crash observability (`daemon/src/crash/`) collector. When a user reports "remote access broken", we want one bundle, not "also grab cloudflared.log from this other path".
**Fix:** crash bundle includes last 1MB of `cloudflared.log` + last `cloudflared --version` output + cloudflared's `--metrics` last snapshot at crash time. Trivial — already a tee target — but worth a one-liner lock.

**F10 (P2) — Failure surfacing in desktop UI.** v0.4 §5.1 already locks `cloudflare.unreachable` sticky banner with "Retry now" + "View log". Proposal §D inherits this implicitly. Worth restating that the failure modes the user refinement introduces (CF Access misconfig 403 — every web/iOS request rejected with a healthy daemon, JWT validation rejects all) need their own surface distinct from `cloudflare.unreachable` (network-layer) — otherwise the user sees "tunnel down" when actually their Access policy is wrong.
**Fix:** add `cloudflare.access_misconfigured` to the surface registry (frag-6-7 §6.8). Detection: 3 consecutive validated incoming JWT failures from a non-attacker source (i.e. JWT signature verifies as Cloudflare-issued but `aud` mismatch / email-not-allowed) → desktop banner "Cloudflare Access policy is rejecting your sign-in. <View policy in dashboard>". Distinct from JWT-failure rate-limit (v0.4 §5.4.1).

**F11 (P2) — Update path for bundled cloudflared.** v0.4 §5.1.1 locks quarterly version bump + SHA pin. Proposal §D does not address auto-update of cloudflared *between* ccsm releases (CF ships ~monthly; quarterly cadence means 2-3 month lag on edge bugfixes). Acceptable for v0.4, but the proposal should say so explicitly so a future PR does not ship "auto-update cloudflared from network at runtime" — that breaks the §5.1.1 SHA-pin defense.
**Fix:** lock "cloudflared updates only via ccsm release; no runtime self-update; `--no-autoupdate` flag (already in v0.4 §5.1) is the contract." One sentence.

---

### Out of lane (flagged, not reviewed)

- Auth model (Cloudflare Access + GitHub OAuth)
- Connect vs alternatives
- iOS distribution path
- Multi-client write semantics (last-writer-wins at PTY)

### Verdict

Direction = **right with required fills**. The single-binary + sidecar topology is correct. The proposal is a re-frame of v0.4 §5 and inherits ~80% of the ops work for free. Before this becomes a spec the **5 P0/P1 supervisor + multi-machine items (F1-F5)** must be locked in §D (not punted to G2/G6). F6-F11 can be added as inline §D bullets without re-architecture.

Specifically the proposal must answer in §D before promotion:
1. Does daemon survive Electron quit? (F1, F2)
2. What are the two ports' identities and how does desktop discover them? (F3)
3. What happens when the same user installs on two machines? (F4)
4. What happens to the CF tunnel on uninstall + machine swap? (F5)

Cited spec lines: proposal §D 99-105, §F 161, §195; v0.4 §5.1 (cloudflared lifecycle), §5.4 (JWT interceptor), §5.6 (setup wizard); frag-6-7 §6.1 (supervisor / crash-loop), §6.4 (rollback), §6.5 (healthz), line 300 (TCP-listener obsolescence).

Jiahui-Gu:
## Arch review — UX/multi-client angle (Task #84)

Reviewer: UX / multi-client semantics / actual-human-using-this. Staying out of auth/transport/ops lanes (other reviewers have those).

**Layer 1 verdict: direction is right with caveats.** The "broadcast + LWW + RAM ring buffer + no presence" simplifications are defensible for a single-human-multi-device model. But §C and §G7 underweight 4 specific UX failure modes that the user *will* hit on day one. Below = each one with severity + concrete fix.

---

### F1 (P1) — "Byte interleave is fine" understates what the user actually sees

§C bullet 1: "bytes interleave at byte-granularity exactly the way two `cat` processes writing to the same tty would. This is correct because PTYs are inherently byte-streams."

What the user sees if desktop types `git status` and web simultaneously types `ls -la`:
```
gilits  -slatatus
```
…submitted as one line to the shell, which then errors. Worse with the **claude CLI prompt parser**: claude reads a structured prompt + slash-commands. Interleaved bytes at the prompt produce unparseable input that claude either rejects or, worse, sends to the model as a corrupted user turn — wasting context tokens and confusing the conversation.

The "two `cat` processes" analogy is technically correct but misleading: nobody actually types into two `cat`s on purpose. People *will* unknowingly type into two ccsm clients (they forgot the desktop tab is open behind a window). The failure mode is silent — no banner, no warning, just a corrupted shell session.

**Concrete fix (cheap):** add a per-client "soft lock" timer at the daemon — when client A writes a byte, mark the PTY as "A active for 1500 ms"; if client B writes within that window, daemon does NOT block the write but DOES push a `peerWriting{otherClient}` event back to both clients' streams. UI surface: a 2-second toast "another device is also typing" with the device label. Zero locking, zero handoff UX, but the silent-corruption failure mode becomes loud. Revisit §G4 with this — the user said "no presence indicator" but that was answered before considering this specific failure path.

---

### F2 (P1) — Reconnect-after-2hr UX is broken for the "leave it running" use case the proposal explicitly claims

§C: "Reconnect outside the budget gets a fresh snapshot with `gap: true`." Replay budget per v0.3 §3.4.1.b = 256 KiB.

256 KiB ≈ ~3000 lines of typical claude CLI output. A 2-hour claude session easily exceeds that (claude streams ~1-5 KiB/s during active generation; long thinking turns + tool use bursts blow past 256 KiB in 5-10 minutes). So the realistic UX:

1. User closes laptop after asking claude to refactor a module.
2. Opens iPad 2 hrs later.
3. Sees: current terminal frame (whatever's on screen *right now*) + a yellow banner "Reconnected — earlier output unavailable".
4. Cannot scroll up to see what claude actually did, what files it touched, or what it asked them.

That is not "leave it running and pick up anywhere" — that is "pick up the *last screen* anywhere". Fundamentally different product feel. The proposal's own §C bullet 3 cites this use case as the justification for rejecting active-client locks, then §G7 quietly admits replay can't actually deliver it.

**Concrete fix (pick one, ordered cheapest -> best):**
- (a) Bump replay buffer to 4 MiB (~45 min of dense output). Costs RAM per session; acceptable for single-user.
- (b) Spill-to-disk per §G7 option, but scoped: keep last 4 MiB in RAM ring + last 24 hr in `~/.ccsm/scrollback/<sessionId>.log` (append-only, rotated). Fresh-snapshot still arrives instantly; client makes a separate `fetchScrollback(fromSeq, toSeq)` RPC lazily when user scrolls up.
- (c) At minimum, the "gap: true" banner must say *how much* is missing in human terms ("~1.5 hr of output not retained") and offer "open `~/.ccsm/sessions/<id>/transcript.txt` on host machine" as the escape hatch. Otherwise the user assumes ccsm just ate their work.

Recommend (b). This is the single biggest gap between proposal language and proposal behavior.

---

### F3 (P2) — "No presence indicator" decision is under-specified, not wrong

User decision per prompt: no presence indicator. Defensible as a *primary* UI element. But two devices going to the same session with zero feedback creates these confusions:

- "Did I already start a claude session about X on my phone?" -> no way to tell from desktop.
- After F1's interleave incident, user has no clue *why* their input got mangled.
- Mobile user kills the app to "free the session" — but desktop is still attached, session never idle-evicts.

**Concrete fix:** keep "no presence indicator" as the *default*, but expose it in one passive place: the session header strip shows a subtle dot icon with count on hover ("2 devices viewing"). One pixel of UI, zero modal/notification noise. Plus a single line in the session-list row: "active on: desktop, web" (text only, no avatars/colors). Cheap, optional-to-glance-at, solves the three confusions above.

---

### F4 (P0) — Web client lifecycle: the proposal does not address Chrome background-tab throttling

§C and v0.4 ch06 §6 cover daemon-unreachable (banner + retry). Neither addresses **Chrome's aggressive background-tab pause**: after ~5 min in background, Chrome throttles JS timers to 1/min and can fully suspend WebSocket/streaming fetches under memory pressure. (Safari is even more aggressive on iOS Safari.)

Realistic flow:
1. User has session open in tab, switches to another tab to read docs.
2. 10 min later, comes back. Stream is dead. Reconnect kicks in. Per F2, anything that scrolled past during those 10 min while claude was producing output -> gone (gap: true).
3. User assumes ccsm "lost" their work.

PTY-allocation question: §C says PTY is per session (good). But the proposal doesn't say what happens when tab is closed vs refreshed vs duplicated:
- Tab close -> just an unsubscribe; PTY keeps running on backend. (Implied, but not stated. State it.)
- Tab refresh -> new subscribe with same `fromSeq`; should resume cleanly. (Implied. State it.)
- Tab duplicated -> second subscriber; works per fan-out registry. (Implied. State it.)

**Concrete fix:**
- Spec a `Page Visibility API` hook in the web client: on `visibilitychange -> hidden`, send a `clientGoingIdle{lastSeq}` ping; on `visible`, immediately re-subscribe `fromSeq=lastSeq`. Combined with F2 fix (b), background tabs resume losslessly.
- Add explicit "tab close vs refresh vs duplicate" subsection to §C — three lines, prevents implementer ambiguity.

---

### F5 (P0) — iOS lifecycle is not addressed at all and it's the hardest of the three clients

The proposal mentions iOS only in transport (`connect-swift`) and distribution (§G3). Zero treatment of:
- iOS suspends app networking ~30 s after backgrounding. Stream dies every time the user switches apps. Without push, the session is effectively "dead on resume" until the user manually opens the app.
- No push notifications spec'd. So "claude finished a long task" never reaches the user unless the app is foregrounded.
- Reconnect-on-foreground will hit F2 every single time (any backgrounding > a few minutes = gap: true).

This is the difference between "iOS app exists" and "iOS app is useful". As written, the iOS client is an emergency-glance tool, not a legitimate first-class client. That contradicts hard input #2 ("None is primary").

**Concrete fix:** EITHER
- (a) Demote iOS to "best-effort secondary" honestly in §F (matches user's "iOS may slip" language already), and add a note that iOS UX without APNs + scrollback persistence will feel like a dead app on every cold-foreground. Set expectation now.
- (b) Add APNs + a backend `notifyClient(sessionId, event)` RPC to v0.4 scope. Wakes the app on "claude turn complete". Big scope add; only worth it if iOS is actually first-class.

The user has to pick. Right now §F implicitly picks (a) by slipping iOS to v0.4.1, but doesn't *say* the UX is degraded — it reads like "same product, just later".

---

### F6 (P1) — listSessions push vs poll is not specified

§C mentions per-session subscription. Doesn't mention how the **session list** stays in sync.

UX failure: desktop spawns a new session. iOS app, currently open, doesn't see it for N seconds (until user manually pulls-to-refresh? until polling fires?). For a 3-device user toggling between devices constantly, this creates "I can't find the session I just made" friction.

**Concrete fix:** add a server-stream `subscribeSessionList()` RPC. One per client connection, lifetime-of-app. Pushes `sessionAdded / sessionRemoved / sessionCwdChanged` events. Trivial reuse of the same Connect server-stream pattern. Without this, every client has to poll `listSessions()` on a timer, which is wasteful and laggy.

---

### F7 (P2) — First-run state for a brand-new client is undefined

Open iOS app for the first time. What does the user see? Empty list? "No sessions yet — start one on desktop"? Auto-spawn a session?

§C and §G are silent. Three plausible answers, each implies different UI:
- (a) Empty list with "Start new session" button (works on iOS too — needs to spec what cwd to use, since iOS has no filesystem to pick from).
- (b) Read-only view of sessions started elsewhere (matches "single source of truth" framing).
- (c) Auto-attach to most recent session.

**Concrete fix:** spec (a) with a constraint: from non-desktop clients, `startSession` must specify cwd from a backend-curated **bookmark list** (cwds the user has previously used on desktop, persisted in SQLite). This sidesteps "iOS has no filesystem picker" and reuses the existing cwd state. Pre-empts a v0.4.1 reopen.

---

### F8 (P1) — First-time auth is a 5-step gauntlet with no defer-path

§B desktop flow: launch app -> opens system browser -> GitHub login -> CF Access app approval -> redirect back. That's ~5 clicks across two app contexts on first run, before the user has seen *any* ccsm UI.

The proposal correctly rejects device flow. But it doesn't address the obvious option: **defer auth entirely if the user only wants local**. Per constraint 6, "Desktop MAY bypass when on localhost". So a brand-new user on their dev machine should not need to touch GitHub at all to use ccsm locally — auth should kick in only when they enable remote.

**Concrete fix:** first-run flow = "Use locally" (zero auth) vs "Enable remote access" (the 5-click flow). Default = local. Settings pane has a "Enable remote access" toggle that triggers the auth flow lazily. This matches the "single human, mostly on one machine" reality and removes a giant adoption barrier.

This needs to be explicit in §B because the current text reads as if every client always does the full flow.

---

### F9 (P2) — No mitigation for flaky-network UX on web/iOS

Cloudflare-fronted, 4G mobile. PTY output stutters; keystrokes lag. Proposal: silent. v0.4 ch06 §6 only covers full disconnect with banner.

Realistic UX: user types, sees nothing for 800 ms, types again, characters arrive in a burst. They double-type. PTY interleaves their double-input with itself. Bad.

**Concrete fix (small):** web/iOS clients implement local-echo for printable characters with a "tentative" style (slightly dim), reconcile against authoritative server echo when it arrives. Standard SSH client behavior (mosh does this). Doesn't fix latency but eliminates "did my keypress register?" UX.

Plus: on `subscribePty` stream stalls > 2 s without daemon-unreachable, show a subtle "high latency" indicator in the status strip — not a full banner. Distinguishes "your network is slow" from "ccsm is broken".

---

### F10 (P2) — Some pieces ARE over-built for single-user

Per the prompt's question 10. The proposal mostly stays right-sized, but two areas leak into multi-user thinking:

- **§B "Multi-tenant is explicitly out of scope"** — good, but the entire JWT-validation infrastructure (JWKS, AUD claim, RS256, clock skew) is still there because Cloudflare Access requires it. Acceptable cost (CF Access is doing the work). No fix needed; just calling it out as not-actually-overbuilt.
- **§C "Subscriber registry with drop-slowest watermark at 1 MiB"** (inherited from v0.3) is sized for unknown-N subscribers. For single-user-3-devices, N <= 3 always. The watermark is fine but the registry's complexity (slow-consumer detection, fairness) is over-engineered. **No action** — already shipped, not worth re-touching. Just don't add MORE multi-subscriber sophistication on top in v0.4.

The actual over-build risk is in **future** PRs that take the multi-client framing as license to add presence/cursors/collaborative-editing. The spec should explicitly note a non-goal: "no per-client identity beyond device-label string; no avatars; no per-client colors; no co-editing affordances. The *user* is one person."

---

## Summary recommendation (UX angle, not a final verdict)

Direction is sound. Three changes are P0/P1 enough that I'd want to see them folded before this becomes a real spec:

1. **F2** — replay budget either grows to 4 MiB or spills to disk. Current 256 KiB makes the proposal's headline use case ("leave it running and pick up anywhere") not actually work.
2. **F4 + F5** — explicit lifecycle handling for browser-tab-hidden and iOS-backgrounded. Without these, web and iOS aren't first-class, contradicting hard input #2.
3. **F8** — local-first auth defer. 5-click first-run with no escape kills adoption.

F1 (interleave warning) and F6 (session-list push) are cheap wins worth doing in v0.4. F3, F7, F9, F10 are paper changes / future-proofing.

Not approving / not rejecting (PR is draft + doc-only). Suggest next revision pass before promotion to spec.

— reviewer (Task #84)

Jiahui-Gu:
## Arch review — security/auth angle (Task #86)

Scope: the proposal in `docs/superpowers/specs/2026-05-02-multi-client-architecture-proposal.md` plus the user's later refinement (backend本机 only; desktop = loopback Listener A with peer-cred; web + iOS = Cloudflare → loopback Listener B with CF Access JWT; cloudflared sidecar lifecycled by daemon). I am NOT commenting on transport choice, ops, or multi-client UX — other reviewers cover those.

The v0.4 doc (§5 / §8.1) and v0.3 frag-6-7 §7 already lock most of what is needed; this proposal mostly defers to them but in doing so quietly drops or muddies a few P0-grade controls. Findings below quote the lines I am flagging.

---

### P0 — must fix before this becomes a spec

**P0-1 Listener B exposure if cloudflared is misconfigured.** §D says "(2) `cloudflared` tunnel sidecar pointing at the same loopback port" and §B says "`localTransportKey: true` … the JWT-validation interceptor is a no-op on locally-tagged requests." If Listener B is "the same loopback port" as Listener A, then either (a) the JWT bypass is keyed off a header/tag that cloudflared can be made to set, or (b) you actually need two distinct listeners. v0.4 §5 already locks two ports (`127.0.0.1:7878` dev / `127.0.0.1:7879` prod) and the rule "TCP bind is `127.0.0.1`, never `0.0.0.0`". This proposal under-specifies it and that is the difference between "safe" and "RCE-from-LAN if a future PR fat-fingers the bind address".
Fix: lock in this proposal — Listener A is a UDS / named pipe (NOT TCP) so peer-cred is even possible; Listener B is `127.0.0.1:<port>`, never `0.0.0.0`, JWT-required unconditionally, and the bypass tag is "transport == UDS", not a header. CI lint: grep `server.listen` rejects any non-loopback bind.

**P0-2 The "loopback trust" assumption is wrong on multi-user / container hosts.** §D: "desktop running on the same machine as the backend hits `http://127.0.0.1:<port>` directly. Same Connect protocol… only the URL and the `localTransportKey` middleware tag differ." `127.0.0.1` is reachable by **every UID on the box**, including other logged-in users on a shared Mac/Linux box, every container sharing the host network namespace, and every non-user system process (sshd-spawned shells, CI agents). v0.3 frag-6-7 §7.2 deliberately uses peer-cred + ACL on a UDS exactly because plain loopback is not a trust boundary on a multi-user host. The proposal silently downgrades that.
Fix: Listener A MUST be a UDS (`~/Library/Application Support/ccsm/daemon.sock` 0700) on macOS/Linux and a named pipe with explicit DACL (`O:<sid>D:P(A;;GA;;;<sid>)`) on Windows — not a loopback TCP socket. Then peer-cred is feasible on all three platforms (`SO_PEERCRED` Linux, `getpeereid` macOS, `GetNamedPipeClientProcessId` + `OpenProcessToken` Windows). If you really need HTTP/2 on the desktop bypass and Connect-Node cannot speak HTTP/2 over UDS, say so and add a bypass cookie that is mode-0600 readable only by the same UID, instead of "trust 127.0.0.1".

**P0-3 Same-UID processes are still trusted; say so.** Even with the UDS+peer-cred fix in P0-2, peer-cred only authenticates UID, not the binary. Any malware running as the same user (browser-extension RCE, npm-install postinstall, electron preload exploit) is inside the trust boundary and gets full PTY. v0.3 frag-6-7 §7.4 T3/T12 already accept this; the new proposal does not.
Fix: copy the v0.3 §7.4 T3 / T5.1 / T12 carve-outs verbatim into a §H Threat-Model section in this doc. Reviewers and future-you need to see the residual on the page, not have to chase it across three documents.

**P0-4 JWT validation policy is delegated to v0.4 §8.1 by reference but the proposal weakens one knob.** v0.4 §8.1 locks `algorithms: ['RS256']`, `requiredClaims: ['exp','iat','aud','iss','sub']`, `audience: <per-app AUD exact match>`, `clockTolerance: 30s`, JWKS unknown-`kid` policy = fail-closed with 30s cooldown, **JWKS pre-warm at boot, ingress refuses to bind until JWKS is cached**. The proposal §B says only "verify the Cloudflare Access JWT (RS256, per-app AUD, team issuer)". That is a regression in specificity — anyone implementing from this doc alone will get the AUD pinning right and the alg-confusion / iat-skew / JWKS-unreachable / `kid`-rotation cases wrong.
Fix: §B MUST either inline the full lock list from v0.4 §8.1 or cite it as "MUST implement v0.4 §8.1 verbatim, no re-derivation." Do not say "RS256" and stop — that is how alg-confusion bugs ship.

**P0-5 No replay window / session-duration cap stated; the v0.4 lock is not carried forward.** v0.4 §5 cuts CF Access session from 24h → 8h "to limit replay window" and requires MFA. The proposal does not mention either. Cf-Access-Jwt-Assertion is a bearer cookie; if it leaks, the only thing bounding blast radius until CF dashboard revoke is session duration.
Fix: lock `session_duration: 8h` and `require: mfa` in this proposal §B (not just in v0.4). And state explicitly: "no per-jti nonce store; same-cookie-replay within session window IS in-scope as ACCEPTED risk for single-user model. Multi-user (v0.6+) requires a jti store."

**P0-6 Web client has no Origin / Sec-Fetch-Site check on Listener B.** v0.4 §5 locks `Host` header validation against DNS rebinding but says nothing about `Origin`/`Sec-Fetch-Site`. Cf-Access-Jwt-Assertion is delivered as a cookie (per v0.4 §5: "`Secure`+`HttpOnly`+`SameSite=Lax`"). `SameSite=Lax` does NOT protect POST-with-body from a cross-origin form submit on top-level navigation — and Connect uses POST for unary RPCs. So a malicious site can in theory CSRF the daemon through CF Access if the user is signed in.
Fix: JWT interceptor on Listener B MUST also check `Sec-Fetch-Site: same-origin` OR a non-bearer header that browsers will not auto-attach (e.g. require `Connect-Protocol-Version: 1` on every request — Connect already sends this; reject if missing). Best: set the Access cookie to `SameSite=Strict` for the daemon hostname. Document the choice; do not punt.

---

### P1 — should fix

**P1-1 cloudflared sidecar = full PTY-output trust at the user.** §D: "`cloudflared` as a sidecar process spawned and lifecycled by the backend." cloudflared runs as the user, sees plaintext keystrokes and PTY output (TLS terminates at Cloudflare edge — v0.4 §5 last paragraph: "Daemon-to-cloudflared is plain HTTP over loopback (no TLS)"). Compromised cloudflared = full keylogger + everything the user types into the terminal goes to attacker. v0.4 §5 §1.1 has the right answer (pinned version, pinned SHA256, runtime SHA re-check) but the proposal does not carry it forward.
Fix: §D MUST cite v0.4 §5 §1.1 as binding and MUST add: "cloudflared update channel = our build-system, NOT cloudflare-auto-update; user-flag `cloudflared --no-autoupdate` MUST be passed at spawn." Otherwise we are letting Cloudflare ship arbitrary code into the user keylogger position on their schedule.

**P1-2 GitHub OAuth blast radius is understated.** §B: "Same identity across all clients because the IdP is identical. Backend never stores a user database." That is correct but the implication is unstated: **if the user GitHub account is compromised, the attacker has shell on the user primary workstation through CF Access** (assuming MFA bypass or cookie theft). v0.4 §5 §3 mentions MFA-required and 8h session — that is the right mitigation but it lives in v0.4, not in this proposal which is the one a reader will start from.
Fix: §B add: "Blast radius of GitHub-account compromise = remote shell on backend host. Mitigations: MFA required at CF Access (locked at v0.4 §5 §3), 8h session duration (locked at v0.4 §5 §3), CF Access dashboard revoke = lead step in compromise recovery (NOT GitHub OAuth revoke, which only takes effect after JWT expires — see v0.4 §8 compromise recovery)."

**P1-3 Token storage section is missing what the keychain does NOT protect against.** §B: "Desktop: OS keychain. iOS: iOS Keychain." This is correct but incomplete. v0.3 frag-6-7 §7.4 T12: "env block readable by same-user via `OpenProcess`. Residual: ACCEPTED." The OS keychain protects against (a) other users on the same box, (b) backup exfil, (c) at-rest disk theft. It does NOT protect against the same-UID malware case. iOS Keychain with `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` is stronger (Secure Enclave) but only if the access-control flag is set — implementations default to `kSecAttrAccessibleAfterFirstUnlock` which is weaker.
Fix: §B add a "what it does/does not protect against" sentence; lock the iOS access-control flag = `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` + `kSecAccessControlBiometryCurrentSet` (TouchID/FaceID gate per fetch).

**P1-4 cloudflared crash log can leak PTY output.** §D cloudflared sidecar tees stderr/stdout to `~/.ccsm/cloudflared.log` (v0.4 §5 §1 lifecycle). Daemon-side pino redacts `(token|secret|jwt|aud|password|authorization)` — but cloudflared logs are NOT pino, NOT redacted, and `--loglevel info` will log request URLs + bodies on errors. The v0.3 envelope HMAC-of-nonce design (frag-6-7 §7.2 R3-T5) explicitly kept the secret off the wire so it would never appear in `strace` / pcap / crash dumps. Cloudflared can resurrect this risk for any header it logs.
Fix: lock `--loglevel warn` (NOT info) for production; add a CI test that boots cloudflared with a known-good JWT and greps the log file for the JWT prefix — fail if found. Add cloudflared.log to the existing pino redaction lint hook (frag-6-7 R3-Task 26 "secret-redaction lint hook").

**P1-5 Crash logs do not carry the v0.3 envelope accidental-protection.** v0.3 envelope was length-prefixed binary — keystrokes/PTY output were opaque blobs in any tcpdump. Connect over loopback HTTP/2 is also opaque to non-root processes (loopback is not sniffable without root), but pino-redacted crash dumps in v0.3 frag-6-7 §6.6 schema (`{ side, v, pid, boot, errMessage, errStack, lastTraceIds, lastSessions, ... }`) — `lastSessions` may now carry session names that include secrets the user pasted into a prompt. The new transport does not change this, but the proposal should explicitly carry the v0.3 redact list forward, not silently inherit.
Fix: §B add: "pino redact list from v0.3 frag-6-7 §6.6 + the v0.4 §6.1 list `/(token|secret|jwt|aud|password|authorization)/i` MUST be applied to BOTH daemon logs AND every Connect interceptor that logs request metadata; CI gate per v0.3 frag-6-7 R3-Task 26."

**P1-6 iOS ASWebAuthenticationSession cookie scope.** §B: "iOS: `ASWebAuthenticationSession` (Apple native API, gives the app a Safari-shared cookie jar so the user is single-sign-on'd if they have authenticated to GitHub on Safari recently)." Two things wrong:
1. ASWebAuthenticationSession `prefersEphemeralWebBrowserSession = false` (default) shares cookies with Safari — convenient, but means **whichever GitHub account the user is currently signed into in Safari** gets used. If the user has a personal GH and a work GH, ccsm silently picks one. CF Access GitHub IdP does not show an account picker once signed in.
2. If the user signs out of GitHub in Safari, ccsm CF Access session keeps working until JWT expiry — that is a UX surprise, but more importantly the "sign out everywhere" mental model is broken.
Fix: §B lock `prefersEphemeralWebBrowserSession = true` for the iOS auth session — costs the user one re-login on first install, but makes the ccsm identity explicit and decoupled from Safari. Document the trade-off so it survives review.

---

### P2 — suggestions

**P2-1 PKCE vs device-flow rejection wording is fine but missing one.** §B "Rejected alternatives: Device flow on desktop: clunky UX… Rejected." Correct security-wise but worth adding: device flow user-code can also be socially engineered (attacker DMs "paste this code at github.com/login/device" — the canonical "rogue verification" attack on device flow). PKCE with loopback redirect does not have this surface.
Fix: one-line add to §B device-flow rejection: "also vulnerable to rogue-verification phishing (attacker tricks user into approving attacker device-flow request)."

**P2-2 No mention of CF Access service tokens for CI.** v0.4 §8 §2108 uses "CI service-token auth bypasses GitHub OAuth — Cloudflare Access supports this." The proposal should say where these tokens live + that they are scoped to a separate CF Access policy with no PTY-write permission (or at least say "v0.4 punted this; same posture here").
Fix: add half-line in §B: "CI uses CF Access service tokens (v0.4 §8); these are equivalent to a JWT-issuance authority, store like the tunnel token (keychain on dev box, GitHub Actions secret in CI)."

**P2-3 §G open question 6 (CF account ownership) hides a security question.** The "friends-of-friends self-deploy" path implies a setup wizard that asks for a Cloudflare API token. v0.4 §6 explicitly chose a copy-paste wizard to avoid asking for a privileged API token. If §G6 is answered "multi-deployer", the wizard scope balloons AND the token-storage threat model gets a new actor (other people self-deploying ccsm with their own CF tokens).
Fix: in §G6, add: "answering 'multi-deployer' here forces a follow-up security review — privileged API token storage, blast radius if the deployer CF account is compromised, etc. Default = 'single deployer' to keep v0.4 in scope."

---

### Punted-but-security-critical

- **No lock on what happens if `cloudflared` *bypasses* the daemon** (e.g. user-misconfigured `ingress` block in cloudflared config to expose a non-daemon service). The daemon does not own cloudflared config file; if the user edits it, cloudflared can route the public hostname to anything on the box. v0.4 §5 §1 spawns cloudflared with explicit `--url http://127.0.0.1:7879`; the proposal does not say "user MUST NOT edit cloudflared.yml; daemon owns spawn flags exclusively." Lock that.
- **No specification of what the desktop bypass tag is**. §B says "`localTransportKey: true`"; if that is a request header (cloudflared can set it) you have a bypass-by-curl. If it is "transport-property = the request came in via UDS file descriptor X" (Connect-Node middleware reading the underlying socket type), you are safe. The proposal MUST say which.
- **No statement on whether `cloudflared` runs in a sandbox** (App Sandbox on macOS, AppContainer on Win). It probably cannot be sandboxed because it needs raw network — but state the residual.
- **No "what if Cloudflare itself is compromised / served wrong JWKS"**. JWKS pinning by team domain only; if the team-domain DNS is hijacked or Cloudflare CA is misissued, JWKS is attacker-controlled. v0.4 §8.1 does not pin the JWKS leaf cert. Probably out-of-scope but worth a P2 line in the threat model.

---

### What was done well

- §B correctly rejects "issue our own JWT" — reinventing token lifecycle is exactly the trap.
- §B correctly rejects device flow for desktop on UX grounds (and on security grounds per P2-1).
- The transport-tag bypass mechanism (locally-tagged = no JWT) is the right shape, just needs P0-1 + P0-2 + the "what is the tag" lock from the punt list.
- Citing v0.4 §8.1 by reference instead of re-deriving JWT validation rules is correct; the issue is just that the citation needs to be MUST-language, not optional.

---

**Verdict:** Direction = right (the security/auth shape — CF Access JWT for remote, peer-cred for local, OS keychain for tokens, no backend-issued tokens — is the correct frame). But the proposal brevity drops or muddies several P0 controls already locked in v0.4 and v0.3. Before this becomes a spec, fix P0-1..P0-6 and either inline or MUST-cite v0.4 §5 §1.1 + §8.1 + §6.1 verbatim. Not approving / not requesting changes (it is a draft); please address before promotion to spec.

Jiahui-Gu:
## Arch review — transport/protocol angle (Task #85)

Scope: TRANSPORT/PROTOCOL only. Auth, ops, UX deferred to other reviewers. Reviewing `2026-05-02-multi-client-architecture-proposal.md` (PR #787) against `2026-05-01-v0.4-web-design.md` §2/§5/§6, frag-3.4.1, frag-3.5.1, and against actual Cloudflare Tunnel / Connect / browser-streaming behavior.

**Layer-1 verdict (transport axis):** Direction is **right** with caveats. Connect-RPC over HTTP/2 + Protobuf is the correct lock for the 3-client constraint, and the rejection of gRPC-Web (no Envoy translator on the CF data path) and tRPC (TS-only) is technically sound. But the proposal under-specifies several cases where the v0.4 spec already paid down hard cost (heartbeat semantics, replay-ring vs drop-slowest orthogonality, boot-nonce, idle eviction, JWKS bind-gate). It reads as if those are "free carryover" — they're not free for web/iOS, only for Electron.

---

### Findings (severity / problem / fix)

**F1 (P0) — Connect over Cloudflare Tunnel: HTTP/2 lock not stated.**
§A asserts `application/proto` and `application/json` "pass through verbatim — confirmed by the v0.4 spike (v0.4 §5)". v0.4 §5 actually only confirms `--protocol http2` end-to-end (line 1343) and explicitly **defers QUIC/HTTP/3** because "our chapter 06 streaming patterns are validated against HTTP/2 only" (line 1345). Cloudflare's `cloudflared --protocol auto` default can pick QUIC, which has not been validated for our server-streaming patterns. Fix: add a one-paragraph "Cloudflare Tunnel HTTP/2 lock" in §A: (a) `--protocol http2` mandatory in cloudflared args (not `auto`); (b) cloudflared-to-daemon hop is local HTTP/1.1-or-HTTP/2 (Connect server-streams over h1.1 use Transfer-Encoding: chunked which works; bidi would not, which we don't need).

**F2 (P0) — gRPC-Web rejection rationale is incomplete.**
§A rejects gRPC-Web because "Cloudflare Tunnel doesn't translate" — true but incomplete. The deeper issue is gRPC-Web's HTTP/1.1 base64 trailer scheme breaks server-streaming through some HTTP/1.1 proxies. Connect-Web by contrast uses an end-stream metadata frame encoded as a regular message with a magic flag bit (Connect protocol "end-stream"), which is trailer-free and survives any HTTP/1.1 or HTTP/2 hop including Cloudflare. Fix: replace "requires Envoy" one-liner with: "Connect-Web's trailer-free end-stream-frame design is what survives CF; gRPC-Web's HTTP/1.1 binary-trailer + base64 scheme does not."

**F3 (P0) — Server-stream PTY in Safari: silent gap.**
The proposal claims "works half-duplex over fetch streaming, which is what every browser actually ships" — partially false. Safari/iOS WebKit shipped ReadableStream over fetch only in Safari 14.1 (2021), and **iOS Safari < 15.4** has known issues with body buffering for non-text content-types. Connect-Web v1.x sends `application/connect+proto` for server-streams which works in Safari 15.4+, but the proposal doesn't lock the content-type or browser min version. Connect-Swift on iOS uses URLSession's `dataTask` delegate-mode `urlSession(_:dataTask:didReceive:)` which does NOT buffer — fine. But web-on-iOS (Safari) is the risk surface. Fix: add a §A subsection "browser stream support matrix" pinning Chrome 105+, Safari 15.4+, Firefox 102+; explicitly document Safari < 15.4 = unsupported (banner + fail-closed).

**F4 (P0) — Backpressure: 1 MiB drop-slowest watermark is sized for v0.3 N=1 only.**
§C says "the v0.3 fan-out registry already supports N subscribers per session with the drop-slowest watermark at 1 MiB. The 'single subscriber' assumption was a temporary perf carve-out, not a fundamental limit; we lift it for v0.4." But frag-3.5.1 §3.5.1.5 line 119 explicitly bounds the **aggregate** at `min(N×1MB, 4 MiB)` — at 3 first-class clients × 1 MiB = 3 MiB, you're already 75% of cap. v0.4 §6 (lines 1543, 1598) uplifts this to MAX_SUBSCRIBERS_PER_SESSION=8 + 4 MiB replay ring and adds the **orthogonality rule** (frag-3.5.1 line 120): "256 KB = per-replay message hard limit; 1 MB = per-second flow rate cap; replay-burst exempt from drop-slowest accounting." Connect/HTTP/2 has SETTINGS_INITIAL_WINDOW_SIZE flow control, but Connect-Node does not surface it to handlers — application-layer drop-slowest is the only knob. Fix: add §C subsection enumerating the 5 caps (per-subscriber 1 MiB, aggregate 4 MiB, MAX_SUBS_PER_SESSION=8, MAX_TOTAL=64, replay-ring 4 MiB) + the orthogonality rule. Without this the daemon dies under 3-client multi-MB/s compile output.

**F5 (P0) — Cloudflare Tunnel idle timeout = 100s. Proposal omits heartbeat entirely.**
v0.4 §5 line 1347 locks: "Cloudflare WebSocket-equivalent (HTTP/2 stream) is killed after 100s of no traffic." The proposal in §C/§E does not mention heartbeats once — yet without the 60-90s server-driven heartbeat (v0.4 §6.4 lines 1495-1520), every PTY stream dies after a 100s idle window. This is the single most important transport carryforward and it's invisible. Fix: §C MUST cite v0.4 §6.4 heartbeat lock (60s emit / 90s ceiling, server-driven because Connect-Web is half-duplex and client cannot ping). Without explicit lock, an iOS implementer will reasonably assume "half-duplex unary input is enough" and skip heartbeat.

**F6 (P1) — Half-duplex unary input has hidden contract w/ v0.4 §6.3 coalescing.**
v0.4 §6.3 lines 1459-1491 lock per-keystroke coalescing (16ms idle OR 4 KB batch) specifically because "per-keystroke unary RPCs over Cloudflare Tunnel pay a JWT-verify + interceptor + protobuf encode/decode cost per RPC." The proposal claims "Client→server keystrokes are unary RPCs (one per batch); we do not need bidi" — true but the **batching policy is a transport-level requirement**, not per-client. Connect-Swift on iOS has its own URLSession serialization and will produce different RPC counts unless the same coalescing is implemented. Fix: §A input-model bullet must reference v0.4 §6.3's coalescing rule as binding on **all** clients including Connect-Swift. iOS that doesn't coalesce will saturate the tunnel under paste.

**F7 (P1) — "snapshot + delta-from-seq is reused from v0.3" glosses over boot_nonce + 4-state startup race.**
§C says "snapshot + delta-from-seq". v0.4 §6.5 + §6.6 + §6.6.1 specify a 5-piece machinery: (1) seq, (2) boot_nonce on every PtyEvent, (3) 4 MiB per-session replay ring vs 256 KiB replay-burst cap (orthogonal — line 1545 "Both apply"), (4) PtyGap event with reason enum, (5) 4-state startup-race table (listener-not-bound / migration-pending / JWKS-loading / JWT-expired) distinguished by `X-CCSM-Boot: starting` header (lines 1576-1586). Multi-client makes the boot_nonce path **strictly more important** because each client independently rebuilds state on daemon respawn — without boot_nonce check, web tab and iOS app could both replay phantom seq from a dead daemon. Fix: §C must explicitly inherit all 5 pieces and the `X-CCSM-Boot` header gate.

**F8 (P1) — How does client distinguish "stream silently dead through CF" vs "no PTY output for a while"?**
Not addressed. v0.4 §6.4 line 1512 locks: client-side `lastSeenAtMs` resets on **every** received message (chunk OR heartbeat), 120s timeout (= 90s heartbeat ceiling + 30s grace) → reconnect. Without locking this in the proposal, an iOS implementer using URLSession's default 60s timeout will trigger spurious reconnects on any session idle > 60s (common — leaving a session idle while you read code is normal). Fix: §C must cite v0.4 §6.4 client-side liveness rule as binding cross-client.

**F9 (P1) — HTTP/2 PING through Tunnel is irrelevant.**
Connect doesn't use them; v0.4 §6.4 line 1518 explicitly rules them out: "(a) Connect doesn't expose PING frame access; (b) HTTP/2 PINGs are at the transport layer and don't reset the application's idle timer for this purpose." Documenting this in the proposal would foreclose the question. Fix: 2-line note in §A or §C: "we do NOT rely on HTTP/2 PING; Cloudflare's idle timer is at the stream/application layer, not the connection layer."

**F10 (P0) — proto/ versioning across 3 client release cadences is the single biggest under-specified risk.**
§A says "wire-compatible with Connect-Go, Connect-Swift, Connect-Kotlin — opens future native clients without protocol rewrite." v0.4 §2.4.1 lines 317-324 lock `[proto-break]` PR-prefix + `ccsm.v<N+1>` namespace + 1-release deprecation window. **None of this works for iOS-via-AppStore**, which has 1-7 day review lag + users who never update. Web ships continuously (Cloudflare Pages), desktop ships on installer cadence (weeks), iOS App Store can be **months stale**. v0.4 §2.7 cross-version matrix (lines 367-376) explicitly says client version-skew = "blocked, surface daemon.unreachable banner, user must reinstall" — tolerable for desktop, **catastrophic for iOS** where the user can't reinstall an unreviewed binary. Fix: §A or new §H must lock a forward-compat policy: (a) fields are additive-only across at least N=2 minor versions, (b) iOS client treats unknown `oneof` variants as "ignore + log", (c) daemon never sends a new `oneof` variant before all-clients version matrix shows >95% client adoption. Without this, the "3 first-class clients" claim is unfunded.

**F11 (P1) — Frame size / message size: PTY snapshot 2 MiB gzip cap omitted.**
§E says "Connect inherits the 16 MiB cap natively (`readMaxBytes`)" — true at framing layer, but v0.4 §2.8 line 384 sets **per-route caps**: default 4 MiB, SendPtyInput 1 MiB, Db.save 16 MiB, Importer 4 MiB. PTY snapshot needs gzip-encoding + 2 MiB cap (v0.4 §6.5 line 1538) or a single huge stdout burst from `cat large.log` on a fresh subscribe will OOM the renderer. Fix: §E must enumerate per-route readMaxBytes carryforward from v0.4 §2.8 + the snapshot gzip Content-Encoding lock from v0.4 §6.5 line 1538.

**F12 (commentary) — Supervisor / data-plane split is cleaner than the proposal suggests.**
§E says "supervisor still on local UDS, data-plane on Connect HTTP/2; same architectural shape." `boot_nonce` is generated by the daemon supervisor (frag-6-7 §6.5, referenced from frag-3.5.1 line 101) and consumed by data-plane heartbeat envelopes AND data-plane subscribe RPCs. The bridge is a single in-process import — clean. trace-id ULID is similarly shared (v0.4 §2.8 line 387: "Connect interceptor generates ULID per request" mirrors envelope's). Verdict: the "soft tension" the user asked about doesn't materialize. Fix: 1-line note in §E: "boot_nonce + trace-id are in-process module-level values, not transport-bound; no bridging code."

**F13 (P0) — Wire-compat with in-flight envelope work (#76).**
Per frag-3.4.1 §3.4.1.b/c the envelope work shipping right now includes: cached header skeleton (perf-P0-A), shared framed buffer per chunk (perf-P0-B), `socket.cork/uncork`, replay-burst exemption from drop-slowest (round-3 perf-CF-2). Proposal §E throws out "manual chunking / binary-trailer carve-out (~30 LOC)" as Connect-native. **Net: the wire bytes are wasted, the semantics are not.** Connect HTTP/2 owns framing, but the **semantic specifications** (256 KiB replay clamp, replay-burst exemption from 1 MiB watermark, drop-slowest LRU) all carry forward into v0.4 §6.5/§6.7 — those frag-3.4.1 paragraphs become Connect-interceptor specs, not envelope code. Fix: §E explicit row in throw-out table: "byte-framing semantics survive as Connect-interceptor specs; envelope wire format is what's deleted, not the rules it encoded." Flagging back to manager: confirm #76 is in the throw-away set.

**F14 (P1) — Connect-Swift maturity acknowledgement.**
Proposal claim "Buf, Apache-2.0" is correct but stops there. Connect-Swift (github.com/connectrpc/connect-swift) v0.x as of late 2025 has known issues with: (a) URLSession streaming on iOS Background URLSession config (data delivered in-bulk on app resume, breaking real-time streaming), (b) cancellation through async sequences is `Task.cancel()` only — no graceful close-frame, (c) no built-in retry interceptor; client must implement reconnect+resume itself. None blockers; all surface area the proposal doesn't acknowledge. Fix: §A "iOS" bullet should cite a min Connect-Swift version + add open question: "iOS background app refresh + PTY stream resume — confirmed not to use Background URLSession config."

---

### Summary

- 6 × P0 (must-resolve before this becomes a spec): F1, F2, F3, F4, F5, F10, F13.
- 6 × P1 (should resolve): F6, F7, F8, F9, F11, F14.
- 1 × commentary (F12): split is clean.

The proposal correctly picks Connect over the alternatives for the 3-client constraint, but it under-specifies the **carryforward** from v0.4 §5/§6 — heartbeats, replay-ring vs drop-slowest orthogonality, boot_nonce 4-state startup-race, per-route readMaxBytes, snapshot 2 MiB gzip cap, and the cross-version skew matrix that breaks for iOS distribution. These are all in v0.4 today; the proposal needs to explicitly inherit them rather than implying "v0.3 carries forward". For iOS specifically, F10 (forward-compat policy across App Store lag) is the load-bearing risk the current draft doesn't address.

No verdict (draft).

— transport/protocol reviewer, Task #85

Jiahui-Gu:
Superseded by PR #789 (frozen baseline) + PR #788 (v0.3 reconciliation), both merged. Closing draft.

PR #687, opened 2026-05-01
feat(daemon): T45 Windows named-pipe ACL hardening

Jiahui-Gu:
## Summary
- Pure module `daemon/src/pty/pipe-acl.ts`: applies DACL to named pipes per spec §3.5.1.6.
- DI seam (`NativePipeAclDeps`) — mirrors T38/T39 patterns. `loadDefaultDeps()` throws until frag-11 wires `ccsm_native` binding.
- Non-Win32: silent no-op stub (POSIX socket modules already use chmod 0600).

## Verification
- `npm run typecheck`: clean (root + electron + daemon tsconfigs).
- `npx vitest run daemon/src/pty/__tests__/pipe-acl.test.ts`: 5 passed | 2 skipped (Win32-host run; non-Win32 skip block).
- `npm run build:daemon`: clean.
- ESM smoke load: `import('./daemon/dist-daemon/pty/pipe-acl.js')` exports `[ 'applyPipeAcl' ]`.
- Reverse-verified: corrupting `deps.applyOwnerOnly(pipePath + '_BUG')` flips two assertions to FAIL; reverting restores 5/5 PASS.

## Test plan
- [x] vitest: deps injection + reverse-verified
- [x] typecheck clean
- [x] ESM smoke load

PR #684, opened 2026-05-01
feat(daemon): T46 peer-cred verification

Jiahui-Gu:
## Summary
- Pure verifier per spec frag-3.4.1 §3.4.1.j + v0.3-design.md §3.1.1.
- DI seam for ccsm_native binding (frag-11) — Win NamedPipeClientProcessId+OpenProcessToken+TokenUser, Linux SO_PEERCRED, Mac getpeereid.
- Returns `{same, peer}` not throw — caller (Connect/Control adapter) decides close/log per §3.4.1.g HMAC-as-authoritative model.
- Mirrors T38 / T39 patterns (per-platform DI, default-loader throws frag-11 pointer).

## Test plan
- [x] vitest 23 tests pass — 3-platform with injected deps + reverse-verify + unsupported platform + frag-11 pointer
- [x] typecheck clean (\`npm run typecheck\`)
- [x] \`build:daemon\` clean
- [x] ESM smoke load: \`import('./daemon/dist-daemon/sockets/peer-cred-verify.js')\` → \`['loadDefaultDeps', 'verifyPeerCred']\`

PR #681, opened 2026-05-01
feat(daemon): T39 Windows JobObject child tracking

Jiahui-Gu:
## Summary

Task #996 — T39 Windows JobObject child-tracking primitive. Win32 sibling of T38 (POSIX SIGCHLD reaper). Pure producer/sink seam: owns the daemon's JobObject handle, exposes `assign(pid)` / `terminate()` / `dispose()`, and lets the OS handle crash-cleanup via `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.

**Spec**: frag-3.5.1 §3.5.1.1 (Win JobObject wiring), §3.5.1.1.a (Native binding swap interface), §3.5.1.2 "Win parity" paragraph (shutdown step 6 issues `TerminateJobObject`).

## Win32 / non-Win behaviour

| Platform | `createJobObject()` | `assign(pid)` | `terminate(code)` | `dispose()` |
|----------|---------------------|---------------|-------------------|-------------|
| **Win32** | calls `deps.create()` → `CreateJobObjectW` + `SetInformationJobObject(KILL_ON_JOB_CLOSE \| BREAKAWAY_OK)` | forwards to `AssignProcessToJobObject` (idempotent per pid) | forwards to `TerminateJobObject` (idempotent) | JS-side reset only — does NOT close the native handle (KILL_ON_JOB_CLOSE relies on OS closing handle on daemon exit) |
| **non-Win32** | returns no-op stub | no-op | no-op | no-op |

Callers do NOT need to platform-guard. POSIX achieves the equivalent via `setsid()` (node-pty) + the SIGCHLD reaper (T38) + the §3.5.1.2 group-SIGTERM/SIGKILL shutdown sequence.

## Native binding decision

- **In-tree N-API binding** `ccsm_native.node` per spec §3.5.1.1 — single `.node` carrying winjob + pdeathsig + pipeAcl exports. Build wiring owned by **frag-11 §11.4** (not yet landed; `scripts/electron-rebuild-natives.cjs` already has the seam guarded with a directory probe).
- This module follows the **T38 pattern**: production module is pure, native dependencies injected through `NativeWinjobDeps` (`create` / `assign` / `terminate`). `loadDefaultDeps()` throws a clear error directing callers to inject deps until the binding lands. Per §3.5.1.1.a "no direct native import outside `daemon/src/native/impl/`" rule, this module never `require()`s the `.node` file directly — the future `daemon/src/native/index.ts` shim is the only production caller.
[...]

PR #798, opened 2026-05-02
ccsm-native: scaffold N-API addon (frag-3.5.1 §3.5.1.1, #109)

Jiahui-Gu:
## Why

Four merged consumers (sigchld-reaper #674, win-jobobject #681, peer-cred-verify #684, pipe-acl #687) ship as DI-only modules whose `loadDefaultDeps()` throws a "frag-11 / ccsm_native pointer" until this addon lands. This PR is that landing.

Spec: `docs/superpowers/specs/v0.3-fragments/frag-3.5.1-pty-hardening.md` §3.5.1.1 (Win JobObject), §3.5.1.1.a (NativeBinding swap interface), §3.5.1.2 (POSIX SIGCHLD wiring), §3.5.1.6 (Win pipe ACL hardening) + `frag-11-packaging.md` §11.1 (ccsm_native rebuild step).

Manager confirmed N-API path (option 1) over koffi after a brief STOP/REVERSE — see manager turn for context.

## What

One single `.node` carrying five export surfaces per the spec:

| Surface     | Win32                            | Linux                       | Darwin                |
|-------------|----------------------------------|-----------------------------|-----------------------|
| `winjob`    | JobObject + KILL_ON_JOB_CLOSE    | ENOSYS                      | ENOSYS                |
| `pipeAcl`   | atomic 5-step DACL hardening     | ENOSYS                      | ENOSYS                |
| `pdeathsig` | ENOSYS                           | `prctl(PR_SET_PDEATHSIG)`   | ENOSYS                |
| `peerCred`  | `GetNamedPipeClientProcessId`+SID| `getsockopt(SO_PEERCRED)`   | `getpeereid`          |
| `sigchld`   | ENOSYS                           | `uv_signal_t`+`waitpid`     | same as Linux         |

Surfaces unavailable on a host throw `Error { code: 'ENOSYS' }` at the binding boundary so a misconfigured call site fails loud (never silently no-ops at the native layer).

### Files added

- `daemon/native/ccsm_native/binding.gyp` + `package.json` + `src/*.cc`+`surfaces.h`
- `daemon/src/native/index.ts` — `NativeBinding` interface + `native()` singleton loader. The ONLY file allowed to load the `.node` directly per §3.5.1.1.a.
- `eslint-rules/no-direct-native-import.js` — custom ESLint rule enforcing the §3.5.1.1.a contract. Wired into `eslint.config.js`.
- `scripts/build-ccsm-native.cjs` — convenience builder (`npm run build:ccsm-native`) that node-gyp rebuilds against `daemon/.nvmrc` Node target and copies into `daemon/native/<platform>-<arch>/`.
- `.github/workflows/ccsm-native.yml` — CI matrix (ubuntu/macos/windows) that builds the addon, smoke-tests the loader against the real artifact, and lints.

### Files changed

- `eslint.config.js` — register `ccsm-local/no-direct-native-import` plugin/rule.
- `.gitignore` — narrow `daemon/native/` ignore to per-platform output dirs only; keep addon source tracked. Build outputs (`build/`, `node_modules/`) re-ignored.
- `package.json` — add `npm run build:ccsm-native` script.

### Why N-API and not koffi

Per frag-3.5.1 §3.5.1.1.a + open-q-1: N-API picked for compile-time symbol guarantees, zero runtime overhead, and ABI stability across Node majors (NAPI 8). The `NativeBinding` swap interface is single-file swappable to koffi or any other FFI in v0.4 if maintenance burden becomes an issue — no consumer code changes required.

## Test plan

Local Win 11 25H2 (this dev box) results:

- [x] `npm run lint` — 0 errors (123 pre-existing warnings unchanged)
- [x] `npm run build:ccsm-native` — OK against Node 22.11.0 with VS2026 MSVC toolchain
- [x] `npx vitest run daemon/src/native/__tests__/` — 15 pass, 4 CI-only skipped (the dlopen tests run when CCSM_NATIVE_NODE env is set)
- [x] Real artifact load smoke (`CCSM_NATIVE_NODE=daemon/native/win32-x64/ccsm_native.node`) — 10/10 pass: peerCred shape matches win32, winjob callable, POSIX surfaces throw `code: 'ENOSYS'` as contracted
- [x] Existing consumer tests (`peer-cred-verify` + `win-jobobject` + `pipe-acl` + `sigchld-reaper`) — 47 pass / 0 regressions. Their `loadDefaultDeps` "throws frag-11 pointer" tests still pass because production wiring routes through `native()`, not through the per-file default loaders.
- [x] Lint rule reverse-verify: 5 invalid cases (direct `.node` import, `require`, `require.resolve`, `ccsm_native` path substring, dynamic import) all reported. 4 valid cases (non-native imports, shim-internal loads) all silent.

CI matrix `ccsm-native.yml` will run the same gates on ubuntu-latest + macos-latest + windows-latest.

### Reverse-verify on the namespace fix

The first build attempt failed because `NODE_API_MODULE(ccsm_native, ccsm_native::Init)` is a token-pasting macro that cannot accept qualified names. The fix routes through a file-scope `InitModule` shim that forwards into the `ccsm_n` namespace. Build succeeds with the shim; without it the build fails with `'__napi_ccsm_n': is not a class or namespace name`.

## Cross-fragment / cross-task notes

- frag-11 §11.1: artifact name `ccsm_native.node` and output path `daemon/native/<platform>-<arch>/` are honoured by `scripts/build-ccsm-native.cjs`. The existing `scripts/electron-rebuild-natives.cjs` already references the in-tree dir; this PR is what makes that reference resolve.
- Task #104 (Connect server scaffold): peer-cred-verify wiring is now ready via `native().peerCred` — task #104 can consume the loader shim instead of duplicating native code.

## Pre-existing failures NOT caused by this PR

`daemon/src/db/__tests__/boot-orchestrator.test.ts` and `daemon/src/db/__tests__/migrate-v02-to-v03.test.ts` fail on this dev box due to a `better-sqlite3` Node-vs-Electron ABI mismatch (the same environmental issue called out in `.github/workflows/ci.yml` at the "Rebuild better-sqlite3 against Node" step). Verified pre-existing on `origin/working` baseline before any of my changes were applied.

### DIFF
--- a/.github/workflows/ccsm-native.yml
+++ b/.github/workflows/ccsm-native.yml
@@ -1,0 +1,133 @@
+name: ccsm-native
+
+# Build + load-smoke the in-tree N-API helper on the v0.3 platform
+# matrix. Spec: docs/superpowers/specs/v0.3-fragments/frag-3.5.1-pty-hardening.md
+# §3.5.1.1 + §3.5.1.6.
+#
+# We pin Node to `daemon/.nvmrc` (the daemon's locked target) so the
+# .node we produce here matches the ABI the daemon will dlopen at
+# runtime — same single-source-of-truth rule frag-11 §11.5(0a) /
+# Task 20f locked in.
+#
+# This workflow is intentionally separate from the main `ci` job so:
+#   - main CI lint/typecheck/test stays fast (no node-gyp on every PR)
+#   - matrix changes here don't reshuffle the main verify path
+#   - the daemon-side smoke (loadCcsmNative on the freshly built
+#     artifact) runs against the actual binary, not a JS fake.
+
+on:
+  pull_request:
+    branches: [main, working]
+    paths:
+      - 'daemon/native/ccsm_native/**'
+      - 'daemon/src/native/**'
+      - 'daemon/.nvmrc'
+      - 'scripts/build-ccsm-native.cjs'
+      - 'eslint-rules/no-direct-native-import.js'
+      - '.github/workflows/ccsm-native.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: ccsm-native-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-smoke:
+    name: build + smoke (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+
+      # Pin the daemon Node target via daemon/.nvmrc — same source of
+      # truth used by scripts/build-ccsm-native.cjs and frag-11.
+      - name: Setup Node (from daemon/.nvmrc)
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: daemon/.nvmrc
+          cache: 'npm'
+
+      - name: Setup Python (node-gyp)
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      # node-gyp on Windows needs the MSVC build tools shipped with
+      # the GitHub-hosted windows-latest image. No extra install
+      # required; we just verify the toolchain is there before we
+      # spend time on the addon build.
+      - name: Verify toolchain
+        shell: bash
+        run: |
+          node --version
+          npm --version
+          python --version
+          which node-gyp || npx node-gyp --version || true
+
+      # We only need the addon's own deps (node-addon-api headers).
+      # Install via the dedicated build script which also runs node-gyp.
+      - name: Build ccsm_native
+        run: npm run build:ccsm-native
+        env:
+          # Force node-gyp to download fresh headers if cache is cold.
+          npm_config_build_from_source: 'true'
+
+      - name: Verify artifact present
+        shell: bash
+        run: |
+          ARCH=$(node -p "process.arch")
+          PLAT=$(node -p "process.platform")
+          OUT="daemon/native/${PLAT}-${ARCH}/ccsm_native.node"
+          if [ ! -f "$OUT" ]; then
+            echo "FATAL: ccsm_native.node not at $OUT" >&2
+            exit 1
+          fi
+          ls -lah "$OUT"
+          echo "CCSM_NATIVE_NODE=$PWD/$OUT" >> "$GITHUB_ENV"
+
+      # Install the root project's deps (vitest, typescript, etc.) so
+      # we can run the load-smoke test against the freshly built
+      # artifact. We skip native rebuilds via ELECTRON_SKIP_BINARY_DOWNLOAD
+      # and `--ignore-scripts` is intentionally NOT used: we DO want
+      # better-sqlite3's postinstall to run on Linux (other matrix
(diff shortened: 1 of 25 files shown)


Jiahui-Gu (commented):
REVIEW: APPROVE-equivalent (cannot self-approve via gh; manager: please treat as MERGE-ok).

Scaffold lands frag-3.5.1 §3.5.1.1 / §3.5.1.1.a / §3.5.1.2 cleanly with zero v0.4-throwaway smell.

## What I checked
- Spec mapping: all five surfaces (winjob, pipeAcl, pdeathsig, peerCred, sigchld) present per §3.5.1.1 inventory; each has real-impl + stub split keyed off binding.gyp OS conditions. Stubs throw code:'ENOSYS' (matches spec acceptance "MUST throw on the wrong platform").
- NativeBinding interface in daemon/src/native/index.ts matches §3.5.1.1.a shape and reuses the consumer-side types (NativePeerCredDeps, NativeWinjobDeps, NativePipeAclDeps, SigchldReaperDeps) verbatim — no parallel type universe; future swap to koffi is a single-file edit as the spec promised.
- Loader shim: loadCcsmNative tries per-platform-arch path first, falls back to build/Release dev path; CcsmNativeMissingError carries actionable hint; native() caches both success and failure.
- ESLint rule covers ImportDeclaration, ImportExpression (dynamic import), require, require.resolve, AND substring 'ccsm_native' (catches re-export aliases). Self-allowance is path-substring matched on daemon/src/native/ with both / and \ normalization. Tests cover 5 invalid + 4 valid cases.
- CI matrix builds on all 3 OSes, verifies artifact path, runs the loader smoke against the freshly-built .node via CCSM_NATIVE_NODE env, lints, uploads artifact. Pinned to daemon/.nvmrc — same SoT as frag-11.
- .gitignore correctly tracks daemon/native/ccsm_native/src/** while ignoring build/, node_modules/, and per-platform output dirs.
- npm run build:ccsm-native is opt-in (NOT wired into postinstall) — users without native toolchain are not broken on root install.
- scripts/electron-rebuild-natives.cjs already lists ccsm_native in its in-tree module table — this PR makes that reference resolve.

## Non-blocking observations (post-merge follow-up OK)

1. **pipeacl_win.cc SetNamedPipeHandleState may always fail ACCESS_DENIED** — CreateFileW opens with WRITE_DAC|READ_CONTROL only, but SetNamedPipeHandleState requires FILE_WRITE_ATTRIBUTES. Current code tolerates ACCESS_DENIED with a comment about "we just locked ourselves out", but the access right was never granted — PIPE_REJECT_REMOTE_CLIENTS is likely never applied through this path. Fix options: (a) add FILE_WRITE_ATTRIBUTES to CreateFileW and set the mode BEFORE SetSecurityInfo, or (b) move PIPE_REJECT_REMOTE_CLIENTS to the CreateNamedPipe call site. Worth a follow-up against §3.5.1.6 acceptance.

2. **peercred_win.cc reads socket._handle.fd as Int64** — practically safe (Win kernel handles <2^32) but no comment about the >2^53 JS Number precision boundary. One-line note would future-proof.

3. **sigchld_unix.cc callback throws bubble to libuv** — a throwing JS handler can crash the daemon. JS-side reaper is documented as pure producer and consumer tests cover the contract; a try/catch around sub->cb.Call({}) would be belt-and-braces.

None blocking — all consumer-layer concerns the PR scaffold correctly defers.

PR #800, opened 2026-05-02
feat: v0.3 task #124 daemon pino-roll rotation (frag-12 §12.1)

Jiahui-Gu:
## Summary

Replace the inline `pino()` stdout logger at `daemon/src/index.ts:42` with a centralized `pino-roll`-backed file logger (frag-12 §12.1).

## Implementation

**New module** `daemon/src/log/index.ts` — single factory `createDaemonLogger({dataRoot, version, pid, bootNonce})`.

- Transport: `pino.transport({target: 'pino-roll', ...})` — writes happen on a worker thread.
- File pattern: `<dataRoot>/logs/daemon/daemon.<yyyy-MM-dd>.<n>` (daily rotation + numeric suffix when size cap kicks in mid-day).
- Size cap: `50m` (`DAEMON_LOG_SIZE_CAP`).
- Retention: `limit: { count: 7 }` (`DAEMON_LOG_RETENTION_COUNT`) — pino-roll auto-deletes older files on rotation.
- Symlink: `<logDir>/daemon.log` → newest rotated file, maintained by our own `maintainCurrentSymlink()` (pino-roll's built-in `symlink:true` was disabled because it unconditionally `fs.symlink()`s a hardcoded `current.log` and throws EPERM/ENOENT on Windows w/o developer mode — observed in tests). Symlink creation is best-effort and swallows errors on hosts that disallow symlinks.

**Canonical base fields** (frag-12 §12.1) stamped on every line:

```
{ side: 'daemon', v: <DAEMON_VERSION>, pid: <process.pid>, boot: <bootNonce> }
```

**Redact list** (`DAEMON_REDACT_PATHS`, censor `[Redacted]`):

| Path | Why |
| --- | --- |
| `daemon.secret` | daemon HMAC secret (frag-6-7 §6.5 hello) |
| `imposterSecret` | legacy envelope payload field still appearing in some payloads |
| `authorization` | HTTP-style header that may end up in adapter logs |
| `*.secret` | single-segment intermediate wildcard (`@pinojs/redact`) |
| `*.password` | single-segment intermediate wildcard |
| `*.token` | single-segment intermediate wildcard |

Note: `*.secret` matches `<key>.secret` for any single key (depth-1). Deeper nesting (`outer.inner.secret`) would require an explicit `*.*.secret` entry — out of scope for v0.3.

**daemon/src/index.ts** — drops the inline `pino()` block; imports + calls `createDaemonLogger()` with `dataRoot` from `resolveDataRoot()` (the per-OS resolver in `daemon/src/db/ensure-data-dir.ts`).

**daemon/package.json** — `pino-roll: ^3.0.0` was already declared; no dependency change needed.

## Log directory

- Windows: `%LOCALAPPDATA%\ccsm\logs\daemon\`
- macOS:   `~/Library/Application Support/ccsm/logs/daemon/`
- Linux:   `$XDG_DATA_HOME/ccsm/logs/daemon/` (else `~/.local/share/ccsm/logs/daemon/`)

## Test plan

`daemon/src/log/__tests__/log.test.ts` — 5 cases, all green:

- [x] Redacts `daemon.secret`, `imposterSecret`, `authorization`, and depth-1 wildcard `*.secret` / `*.password` / `*.token` to `[Redacted]`; non-redacted fields preserved.
- [x] Exported `DAEMON_REDACT_PATHS` matches the canonical list (regression guard).
- [x] Stamps `{side, v, pid, boot}` on every line.
- [x] Real rotation: writes 102 x 512 KB lines (~52 MB) and asserts >=2 files in `<logDir>` with combined size > 50 MB.
- [x] `maintainCurrentSymlink()` re-points `daemon.log` at the newest rotated file (best-effort; tolerates hosts without symlink permission).

```
$ npx vitest run daemon/src/log/__tests__/log.test.ts
Test Files  1 passed (1)
Tests  5 passed (5)
```

Typecheck (`tsc --noEmit -p daemon/tsconfig.json` + root `tsc --noEmit`) clean. Lint clean. `npm run build:daemon` succeeds.

## Unlocks

- #146 canonical log lines
- #147 `pino.final` wiring
- #141 disk cap
- #142 log injection guard
- #117 dogfood

### DIFF
--- a/daemon/src/index.ts
+++ b/daemon/src/index.ts
@@ -1,8 +1,9 @@
 import { ulid } from 'ulid';
-import pino from 'pino';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { createRequire } from 'node:module';
+import { createDaemonLogger } from './log/index.js';
+import { resolveDataRoot } from './db/ensure-data-dir.js';
 import { createSupervisorDispatcher, createDataDispatcher } from './dispatcher.js';
 import { mountEnvelopeAdapter } from './envelope/adapter.js';
 import {
@@ -51,13 +52,14 @@ const bootNonce = ulid();
 // (frag-6-7 §6.5). MUST be before any handler module that consumes it.
 const bootedAtMs = Date.now();
 
-const logger = pino({
-  base: {
-    side: 'daemon',
-    v: DAEMON_VERSION,
-    pid: process.pid,
-    boot: bootNonce,
-  },
+const logger = createDaemonLogger({
+  // Frag-12 §12.1 — log dir is `<dataRoot>/logs/daemon/`. Resolve the
+  // OS-native dataRoot via the same per-OS resolver the SQLite migration
+  // uses (T34) so the logger and the database share one parent root.
+  dataRoot: resolveDataRoot(),
+  version: DAEMON_VERSION,
+  pid: process.pid,
+  bootNonce,
 });
 
 // Phase 2 crash observability (spec §5.2 / §6, plan Task 8): initialize
--- a/daemon/src/log/__tests__/log.test.ts
+++ b/daemon/src/log/__tests__/log.test.ts
@@ -1,0 +1,240 @@
+// Frag-12 §12.1 — daemon logger tests.
+//
+// Coverage:
+//   1. Redact list — `daemon.secret`, `imposterSecret`, `authorization`,
+//      and the wildcard `*.secret` / `*.password` / `*.token` matchers
+//      all replace values with the literal `[Redacted]` censor.
+//   2. Real rotation — writing >50 MB of payload through the logger
+//      produces at least 2 files under the log directory (size cap
+//      kicked in mid-day).
+//   3. Symlink maintenance — `daemon.log` points at the newest rotated
+//      file after `maintainCurrentSymlink()` runs.
+//   4. Canonical base fields — every line carries `side` / `v` / `pid` /
+//      `boot`.
+
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { mkdirSync, mkdtempSync, readdirSync, readFileSync, rmSync, statSync, utimesSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import {
+  createDaemonLogger,
+  daemonLogDir,
+  maintainCurrentSymlink,
+  DAEMON_LOG_CURRENT_SYMLINK,
+  DAEMON_REDACT_PATHS,
+  DAEMON_REDACT_CENSOR,
+} from '../index.js';
+
+let tmpRoot: string;
+
+beforeEach(() => {
+  tmpRoot = mkdtempSync(join(tmpdir(), 'ccsm-daemon-log-'));
+});
+
+afterEach(() => {
+  try {
+    rmSync(tmpRoot, { recursive: true, force: true });
+  } catch {
+    // Worker thread may still hold the file handle; ignore.
+  }
+});
+
+/**
+ * Drain pino's worker-thread transport so the test sees writes on disk.
+ *
+ * pino's `transport()` is async; logger.flush() is best-effort under the
+ * default async destination. We sleep briefly to let the worker thread
+ * round-trip the queued writes.
+ */
+async function drain(logger: { flush: (cb?: (err?: Error | null) => void) => void } | undefined, ms = 1500): Promise<void> {
+  if (logger && typeof logger.flush === 'function') {
+    await new Promise<void>((resolve) => {
+      try {
+        logger.flush(() => resolve());
+      } catch {
+        resolve();
+      }
+    });
+  }
+  await new Promise((r) => setTimeout(r, ms));
+}
+
+function readAllLines(logDir: string): unknown[] {
+  const out: unknown[] = [];
+  for (const name of readdirSync(logDir)) {
+    if (!name.startsWith('daemon.')) continue;
+    if (name === DAEMON_LOG_CURRENT_SYMLINK) continue;
+    const full = join(logDir, name);
+    let raw: string;
+    try {
+      raw = readFileSync(full, 'utf8');
+    } catch {
+      continue;
+    }
+    for (const line of raw.split('\n')) {
+      if (!line.trim()) continue;
+      try {
+        out.push(JSON.parse(line));
+      } catch {
+        // Mid-rotation partial line; skip.
+      }
+    }
+  }
+  return out;
+}
+
+describe('createDaemonLogger — redact list', () => {
+  it('redacts daemon.secret, imposterSecret, authorization, and wildcard *.secret/*.password/*.token', async () => {
+    const logger = createDaemonLogger({
+      dataRoot: tmpRoot,
+      version: '0.3.0-test',
+      pid: 12345,
+      bootNonce: 'test-boot-redact',
+    });
+
+    logger.info(
+      {
+        // Exact paths.
+        daemon: { secret: 'abc' },
+        imposterSecret: 'legacy-leak',
+        authorization: 'Bearer xyz',
+        // Wildcard *.secret / *.password / *.token at depth-1.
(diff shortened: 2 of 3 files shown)


Jiahui-Gu (commented):
## Verdict: CHANGES REQUESTED (own-PR fallback to COMMENTED)

Solid implementation of the spec on its own — module is well-documented, single-responsibility, redact list and exports look right, real-rotation test is genuine (writes >50MB and asserts file count). But CI is red on `lint + typecheck + test (ubuntu-latest)` and `(macos-latest)` with a regression caused by this PR. Must be fixed before merge.

### Critical (must fix)

**1. Boot ready marker no longer reaches stdout — breaks `tests/harness-daemon-mode.test.ts`.**

The pre-PR `pino()` defaults to writing JSON to `process.stdout`. The harness watches `child.stdout` for the canonical `"daemon shell booted"` marker (see `daemon/src/index.ts:451` and the harness at `tests/harness-daemon-mode.test.ts:1607,1666-1692`). After this PR, `createDaemonLogger()` uses `pino.transport({ target: 'pino-roll', ... })` exclusively — the worker-thread transport writes ONLY to the rotated file under `<dataRoot>/logs/daemon/`. No stdout output. The harness times out at 15s.

CI evidence (run 25246139639):
- `T73 — daemon-boot + hello probe` → `daemon did not emit ready marker (\"daemon shell booted\") within 15000ms`
- `Task #1072 — controlSocket.listen() wire smoke` → `boot timed out (15s); control=undefined data=undefined`

Fix options (any one is fine; pick the cheapest):
- Pass an array of streams to `pino`, e.g. `pino.multistream([{ stream: process.stdout }, { stream: pino.transport({...}) }])` so the boot marker still hits stdout for the harness.
- OR use `pino.transport` with `targets: [{ target: 'pino-roll', options: {...} }, { target: 'pino/file', options: { destination: 1 } }]` — keeps the worker-thread posture and gets stdout fan-out.
- OR write the boot ready marker via a direct `process.stdout.write(...)` next to the `logger.info({event:'daemon.boot'}, 'daemon shell booted')` call (smallest blast radius, but must preserve the JSONL contract the harness parses at `harness-daemon-mode.test.ts:1670`).

After the fix, please re-run the full e2e harness locally — not just `daemon/src/log/__tests__/log.test.ts`.

### Important (should fix)

**2. `limit.count = 7` semantics may be off-by-one vs spec wording.**

The module's own JSDoc admits the ambiguity: pino-roll's `limit.count` is the number of files KEPT in addition to the active one (so `count: 7` → 8 files on disk). Spec frag-12 §12.1 says \"7-file retention\" total. Either:
- change to `limit: { count: 6 }` to match spec literally, or
- update the JSDoc + spec citation to acknowledge the chosen interpretation (7 historical + 1 active = 8 on disk) and confirm with the spec owner.

Currently the comment says \"we treat the spec literally as the total cap and pass 7\" but the code does NOT match that statement. Either the code or the comment is wrong.

**3. Tests don't actually cover the symlink success path.**

`maintainCurrentSymlink` test (`log.test.ts:227-238`) silently passes if symlink creation fails (the `if (content !== undefined)` clause). On Linux/mac CI this should always succeed; gating the assertion behind `content !== undefined` masks regressions. Recommend: assert success on POSIX (`process.platform !== 'win32'`), and skip only on win32.

### Suggestions (nice to have)

- `*.secret` wildcard limitation: documented well, but please grep the existing codebase for `{ auth: { token: ... } }` / `{ creds: { secret: ... } }` shapes before merge to confirm depth-1 is enough for v0.3.
- Consider exporting a `closeDaemonLogger()` helper that wraps `logger.flush` + transport close, to give the shutdown path at `daemon/src/index.ts:194` a typed seam.
- The `_fileBaseOverride` underscore-prefix `@internal` field is never used by the test file — drop until needed (YAGNI) or add a test.

### Done well
- Module JSDoc explains every redact entry and the WHY of the symlink workaround — exemplary.
- Real 50 MB rotation test, not a mock — high-confidence coverage.
- Frozen `DAEMON_REDACT_PATHS` + regression-guard test prevents accidental removals.
- Clean factory shape, no global side-effects, dataRoot resolved via the same `resolveDataRoot()` the SQLite layer uses (no path drift).

Out of scope: the `installer-size-guard` failures (\"daemon binary suspiciously small\") appear unrelated — pre-existing CI issue on the helper-binary path.

Reporting CHANGES to manager.

Jiahui-Gu:
## Reviewer CHANGES verdict — addressed in b827aba

All 3 items fixed. Local `tests/harness-daemon-mode.test.ts` now passes (39/39, including Task #1072 wire smoke that timed out on CI run 25246139639). Local `daemon/src/log` unit tests pass (5/5). Pre-existing better-sqlite3 native-binding failures in unrelated db tests are not touched by this PR.

### 1. Boot ready marker reaches stdout — option B (dual-target transport)

Switched from single-target `pino.transport({ target: 'pino-roll' })` to multi-target `pino.transport({ targets: [pino-roll, pino/file→fd 1] })`.

**Why option B over A or C:**
- **Option A (`pino.multistream`)**: writes synchronously on the main loop — defeats the off-loop posture pino-roll requires. Would also lose the worker-thread isolation that pino-roll uses to avoid blocking on slow disks.
- **Option C (bare `process.stdout.write`)**: parallel write path — bypasses redact + base fields. Reviewer flagged this exact reason; agreed.
- **Option B (chosen)**: both sinks live inside the same worker thread, share one redact codegen and one base-field stamp. Stdout sink uses pino's built-in `pino/file` transport with `destination: 1`. Single logger, dual destination, no main-loop blocking, no redact bypass.

### 2. `count` off-by-one

pino-roll's `limit.count` is "number of files kept IN ADDITION to the active file" (verified in `node_modules/pino-roll/README.md`). Spec frag-6-7 §6.6.1 budgets `7 × 50 MB = 350 MB` per side → 7 files total → `count: 6` (6 historical + 1 active). Updated `DAEMON_LOG_RETENTION_COUNT = 6` and rewrote the JSDoc to spell out the semantic so this can't regress silently.

### 3. Symlink test silent skip

Now gates the swallow on `process.platform === 'win32'` only. On POSIX, any read error (including a legitimate symlink-creation failure) fails the test with the underlying error message. On Windows, EPERM is still tolerated since SeCreateSymbolicLinkPrivilege is gated behind developer mode / Administrator and CI runners typically lack it.

Jiahui-Gu (commented):
MERGE-ok (own PR, cannot self-approve — manager admin-merges).

Re-review at b827aba: all 3 prior blockers cleanly resolved.

1. Boot marker stdout (CRITICAL): targets-array with `pino/file` + `destination: 1` keeps stdout sink inside the worker thread alongside pino-roll, so both share the redact + base-field codegen (no bypass) and writes stay off the main loop. Harness 39/39 pass (incl. previously-timing-out task #1072 control-socket smoke) confirms `{event:'daemon.boot'}` reaches `child.stdout`. Option B over multistream is correctly justified in the inline comment.

2. count off-by-one: `DAEMON_LOG_RETENTION_COUNT = 6` matches pino-roll's documented `limit.count` semantic ("files in addition to the currently used file") and the spec frag-6-7 §6.6.1 budget of 7 × 50MB = 350MB total. JSDoc at daemon/src/log/index.ts:91-99 spells the semantic out with README citation — solid regression guard.

3. Symlink test: POSIX failure now propagates (`expect(readErr).toBeUndefined()` outside the `win32` branch at log.test.ts:248). The previous silent-pass on POSIX is gone.

Other quality checks all pass:
- Redact list configured on the parent pino() instance, applied once before target fan-out — no parallel write path can bypass it.
- Worker thread isolation preserved (no main-loop fsync).
- Base fields {side, v, pid, boot} carried on every line; covered by the canonical-base-fields test.
- Diff scope: 2 functional source files + 1 entrypoint swap (10 add / 8 del in daemon/src/index.ts). Within scope.

Nit (non-blocking, optional follow-up): module header line 14 still reads "7-file retention (`limit.count`)" which can be misread as `count = 7`. The constant JSDoc is correct; if you touch this file again, consider rewording the header to "7-file total (6 historical + 1 active via pino-roll's `limit.count`)" for consistency. Not a merge blocker.

Verdict: MERGE.

PR #134, opened 2026-04-22
refactor(slash): prune dead passthroughs + add local handlers

Jiahui-Gu:
## Summary

- Trim slash-command registry from 19 to 12: remove 7 entries (`/resume`, `/login`, `/logout`, `/mcp`, `/hooks`, `/agents`, `/review`) that were marked `passThrough: true` but silently dropped by `claude.exe` under `--input-format stream-json`, leaving users staring at a no-op button.
- Add 4 new client handlers so the surviving discoverable commands actually do something useful in the GUI: `/status`, `/doctor`, `/memory`, `/bug`. Two new IPC channels back them: `doctor:run` (local health probes) and `memory:openUserFile` (opens `~/.claude/CLAUDE.md` in the OS-default editor; creates a stub if missing).
- Only `/compact` and `/init` remain as honest pass-throughs — both have legitimate native handling in `claude.exe` and no GUI replacement.

### Audit decision summary

| Command   | Verdict | Reason |
|-----------|---------|--------|
| `/resume` | removed | sidebar already lists sessions |
| `/login`, `/logout` | removed | Agentory does not own auth — settings.json is the single source of truth, surfaced via Settings -> Connection (which `/config` opens) |
| `/mcp`, `/hooks`, `/agents`, `/review` | removed | silently dropped in stream-json mode; no GUI surface yet |
| `/status` | now client | renders connection / model / cwd / token usage |
| `/doctor` | now client | runs `doctor:run` IPC, formats `[ok]`/`[fail]` block |
| `/memory` | now client | opens `~/.claude/CLAUDE.md` in OS editor |
| `/bug`    | now client | opens GitHub issue URL with prefilled env block |
| `/compact`, `/init` | kept passthrough | CLI handles natively |

`scripts/probe-slash-audit.mjs` enumerates the live registry from the picker DOM, sends each command, and reports a markdown table of what happened (status block, error, dialog, agent send, openExternal, memory open). Sample output against this branch:

```
| /help    | 1 | 0 | no  | 0 | 0 | 0 | ok    |
| /clear   |-1 | 0 | no  | 0 | 0 | 0 | DEAD? |   (false negative — switches sessions; verified via probe-slash-exec)
| /compact | 0 | 0 | no  | 1 | 0 | 0 | ok    |
| /model   | 0 | 0 | yes | 0 | 0 | 0 | ok    |
| /config  | 0 | 0 | yes | 0 | 0 | 0 | ok    |
[...]

PR #801, opened 2026-05-02
ci: v0.3 task #134 aggregate SHA256SUMS.txt + SLSA subject-paths (frag-11 §11.4)

Jiahui-Gu:
## Summary

Task #134 (frag-11 §11.4): publish a single aggregate `SHA256SUMS.txt` per release so users can verify the whole release with one command, and bind it into the existing SLSA L3 + minisign attest chain so it can't be silently swapped.

## Changes (release.yml `attest` job)

- **New step `Generate aggregate SHA256SUMS.txt`** (after the per-artifact `.sha256` step): writes `dist-flat/SHA256SUMS.txt` covering installers (`*.exe *.dmg *.zip *.AppImage *.deb *.rpm`) plus auto-update metadata (`latest*.yml`, `*.blockmap`). Stable LC_ALL=C sort order for reproducibility. Format is `sha256sum`-compatible (`<hex>  <filename>` per line); self-verified via `sha256sum -c --status` before upload. Derivative sidecars (`.sha256`, `.minisig`, `.intoto.jsonl`) are excluded.
- **`Generate minisign signatures`** also signs `SHA256SUMS.txt` -> `SHA256SUMS.txt.minisig`. Placeholder-safe: skipped when `MINISIGN_PRIVATE_KEY` is unset, mirroring the existing per-installer minisig behaviour (frag-11 §11.5(5)).
- **`Compute SLSA hashes input`** now appends `SHA256SUMS.txt` to the `sha256sum` list base64-encoded into `base64-subjects` for the `slsa-github-generator` reusable workflow. The aggregate file is therefore an attested SLSA subject — an attacker rewriting `SHA256SUMS.txt` to point at malicious binaries would invalidate the provenance.

## Changes (release.yml `publish` job)

- **Sidecar-verify gate** now asserts:
  - `SHA256SUMS.txt` exists and is non-empty.
  - `sha256sum -c --status SHA256SUMS.txt` succeeds against the downloaded artifacts (catches stale/corrupt aggregate files).
  - When minisig is enabled, `SHA256SUMS.txt.minisig` is present and non-empty.
- **GitHub Release upload** globs include `dist-flat/SHA256SUMS.txt` (existing `*.minisig` glob already covers `SHA256SUMS.txt.minisig`; existing `*.intoto.jsonl` glob already covers SLSA provenance).

## Affected release artifacts (per draft Release after this lands)

- New: `SHA256SUMS.txt`, `SHA256SUMS.txt.minisig` (when signed)
- Unchanged: per-installer `.sha256`, `.minisig`, `.intoto.jsonl` sidecars, installers themselves, auto-update yml + blockmaps.

## Test plan

- [ ] yaml syntax: `python -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` passes (verified locally).
- [ ] On next `workflow_dispatch` dry-run / tag push, confirm `attest` job logs show `--- SHA256SUMS.txt ---` block followed by clean `sha256sum -c` exit and (when secret set) `wrote SHA256SUMS.txt.minisig`.
- [ ] Confirm `publish` sidecar-verify gate logs the new "SHA256SUMS.txt + .minisig present and verified" success line and Release draft contains both files.
- [ ] Manual verify: `minisign -V -p release-keys/minisign.pub -m SHA256SUMS.txt && sha256sum -c SHA256SUMS.txt` against downloaded assets succeeds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -282,6 +282,34 @@ jobs:
             echo "wrote $f.sha256"
           done
 
+      # Task #134 (frag-11 §11.4): aggregate SHA256SUMS.txt covering every
+      # release artifact (installers + auto-update yml + blockmaps). One file,
+      # `sha256sum -c SHA256SUMS.txt` compatible (LF-terminated lines of
+      # `<hex>  <filename>`), so a user can verify the full release in a
+      # single command instead of fetching N per-artifact .sha256 sidecars.
+      # We deliberately exclude derivative sidecars (.sha256, .minisig,
+      # .intoto.jsonl) — verifying them against themselves is meaningless and
+      # causes recursion if this file is regenerated.
+      - name: Generate aggregate SHA256SUMS.txt
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd dist-flat
+          # Stable order so the file is reproducible across runs.
+          : > SHA256SUMS.txt
+          for f in $(ls *.exe *.dmg *.zip *.AppImage *.deb *.rpm latest*.yml *.blockmap 2>/dev/null | LC_ALL=C sort -u); do
+            [ -f "$f" ] || continue
+            sha256sum "$f" >> SHA256SUMS.txt
+          done
+          echo "--- SHA256SUMS.txt ---"
+          cat SHA256SUMS.txt
+          # Sanity: must be non-empty and parse cleanly via sha256sum -c.
+          if [ ! -s SHA256SUMS.txt ]; then
+            echo "::error title=Aggregate SHA256SUMS.txt empty::no release artifacts matched"
+            exit 1
+          fi
+          sha256sum -c --status SHA256SUMS.txt
+
       # Minisign signatures. Public key lives in repo at release-keys/minisign.pub
       # (committed). Private key is provisioned via the MINISIGN_PRIVATE_KEY
       # repo secret + MINISIGN_PASSWORD secret (see release-keys/README.md for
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
MERGE — task #134 LGTM (own-PR approve blocked, posting as COMMENTED per protocol).

Verified against frag-11 §11.4 + review focus list:

1. **Spec format**: `sha256sum "$f" >> SHA256SUMS.txt` emits canonical `<hex>  <filename>` lines, `sha256sum -c`-compatible. PASS.
2. **SLSA subject**: SHA256SUMS.txt enumerated in base64-subjects (line 371) — provenance now attests the aggregate, blocking the swap-the-hash-list attack. PASS.
3. **Placeholder-safe**: aggregate `.minisig` is signed inside the same `MINISIGN_PRIVATE_KEY`-guarded block as per-installer minisigs (line 350-353); publish gate's `check_minisig` skip path mirrors per-installer behavior (line 477). When secret IS set the check IS enforced — no security regression. Matches frag-11 §11.5(5).
4. **coreutils sha256sum**: both `attest` and `publish` jobs are `runs-on: ubuntu-latest` — GNU coreutils, not BSD shasum. PASS.
5. **Determinism**: `LC_ALL=C sort -u` correctly normalizes filename order across runs. PASS.
6. **Upload glob**: `*.minisig` covers `SHA256SUMS.txt.minisig`, plus explicit `SHA256SUMS.txt` line — no double-upload, no miss. PASS.
7. **Zero-rework**: signed aggregate manifest is the canonical release-pipeline shape; survives v0.4 final-arch unchanged. PASS.
8. **Diff bounds**: only `.github/workflows/release.yml` touched (+58/-3); per-installer minisig logic untouched (PR only adds a conditional inside the same step). PASS.

Round-trip sanity: `attest` uploads `dist-flat/**` and `publish` downloads `ccsm-attested` to `dist-flat` — auto-update yml + blockmaps referenced inside SHA256SUMS.txt are preserved, so publish-time `sha256sum -c` won't false-fail.

Nits (non-blocking, do not act):
- Post-generation `sha256sum -c --status` in the attest step is technically redundant (file was just built from those artifacts) but cheap and catches `cd` mishaps — keep.
- SLSA hashes step doesn't apply `LC_ALL=C sort`; not a correctness issue (SLSA verifies by hash, not order) and consistent with pre-PR behavior.

PR #803, opened 2026-05-02
feat: v0.3 task #125 electron/log.ts pino logger (frag-12 §12.1)

Jiahui-Gu:
## Summary

Adds the missing Electron-main pino logger so cross-process trace correlation against the daemon side works without a third tool (frag-6-7 §6.6.2 / r2-obs P0-1 / P0-4).

### Implementation

- **`electron/log.ts`** (new): `createLogger` + `getLogger` singleton + `installCrashHooks(app)` + `installRendererLogForwarder(ipcMain)`.
  - **Base fields**: `{ side: 'electron', v: APP_VERSION, pid, boot: bootNonce }` — mirrors daemon §6.6.1; `side` is REQUIRED so daemon + electron lines that share `{v, pid, boot}` keys never collide (r2-obs P0-1).
  - **`pino-roll`**: `{ frequency: 'daily', size: '50M', limit: { count: 7 }, mkdir: true, symlink: true, dateFormat: 'yyyy-MM-dd' }` writing to `<dataRoot>/logs/electron/electron-YYYY-MM-DD.log` with stable `electron.log` symlink.
  - **Redact list**: task #125 minimum (`*.secret`, `*.password`, `*.token`, `daemon.secret`, `imposterSecret`, `authorization`) ∪ spec §6.6.2 superset (`apiKey`, `Cookie`, `ANTHROPIC_API_KEY`, `daemonSecret`, `helloNonceHmac`, ...) ∪ renderer-leaning (`*.url`, `*.searchParams`, `*.headers`, `*.formData`).
  - **`pino.final` analog**: hooks on `app.before-quit`, `process.on('uncaughtException')`, `process.on('unhandledRejection')` — every observable Electron-main exit path flushes pino synchronously (r2-obs P0-4).
- **`electron/preload/bridges/ccsmLog.ts`** (new): renderer console-forward bridge gated by `CCSM_RENDERER_LOG_FORWARD === '1'`. Wraps `console.{trace,debug,info,warn,error}` to ALSO ship `log:write` IPC envelopes; original DevTools output untouched. Default OFF in production (no main-log pollution).
- **`electron/main.ts`**: imports + calls `installCrashHooks(app)` + `installRendererLogForwarder(ipcMain)` after `wireCrashHandlers`.
- **`electron/preload/index.ts`**: unconditionally calls `installCcsmLogBridge()` (no-op when env gate off).
- **`package.json`**: adds `pino-roll ^3.0.0` to root deps (root already had `pino`).
- **`electron/__tests__/log.test.ts`** (new): 15 tests — base fields, redact list (task #125 minimum keys + spec §6.6.2 superset), `installCrashHooks` flushes on `before-quit` / `uncaughtException` / `unhandledRejection`, env gating, log dir resolution.

### Constraints honoured

- v0.4 zero-rework: logger module is final-arch (frag-6-7 §6.6.2 doesn't change in v0.4).
- Add not replace: existing `console.*` calls keep working.
- No daemon-side edits (#124 owns daemon log.ts).
- Renderer-forward env var: `CCSM_RENDERER_LOG_FORWARD=1`. Log dir: `<dataRoot>/logs/electron/`.

### Unblocks

- **#146** canonical bridge call log line names (`bridge_call_complete`, `daemon_socket_closed`, `daemon_reconnect_*`, `stream_resubscribe`) now have a writer.
- **#141** disk cap watchdog (logs aggregate 500 MB / crashes 200 MB) now has a per-side log directory to monitor.
- **#117** dogfood metrics — structured lines are queryable.

## Test plan

- [x] `npm run typecheck` clean (root + electron + daemon)
- [x] `npx eslint` clean on all touched files
- [x] `npx vitest run electron/__tests__/log.test.ts` — 15/15 pass
- [x] Pre-existing better-sqlite3 NODE_MODULE_VERSION mismatch (3 db tests fail on this branch AND on `origin/working` with the same node_modules) is unrelated; verified with `git stash` baseline run
- [ ] Manual smoke (post-merge): `CCSM_RENDERER_LOG_FORWARD=1 npm run dev`, hit `console.error('hello')` in renderer, confirm line lands in `<dataRoot>/logs/electron/electron-*.log`

### DIFF
--- a/electron/__tests__/log.test.ts
+++ b/electron/__tests__/log.test.ts
@@ -102,7 +102,7 @@ describe('electron/log — redact list', () => {
     expect(flat).not.toContain('tok_abc');
     expect(flat).not.toContain('s3cr3t');
     expect(flat).not.toContain('NOPE');
-    expect(flat).toContain('[REDACTED]');
+    expect(flat).toContain('[Redacted]');
   });
 
   it('exposes the canonical redact paths constant including all six task-required keys', () => {
@@ -245,11 +245,32 @@ describe('electron/log — renderer log forward gating', () => {
       env: { [RENDERER_LOG_FORWARD_ENV]: '1' },
       logger: fakeLogger as never,
     });
-    handler?.({}, { level: 'warn', args: ['hello', { a: 1 }] });
+    // Pass the fromMainFrame guard: senderFrame === sender.mainFrame.
+    const mainFrame = { id: 'main' };
+    const event = { senderFrame: mainFrame, sender: { mainFrame } };
+    handler?.(event, { level: 'warn', args: ['hello', { a: 1 }] });
     expect(fakeLogger.warn).toHaveBeenCalled();
     const callArgs = fakeLogger.warn.mock.calls[0]!;
     expect(callArgs[0]).toMatchObject({ event: 'renderer_console_forward', source: 'renderer' });
     expect(typeof callArgs[1]).toBe('string');
+  });
+
+  it('drops payload from sub-frame senders (fromMainFrame guard)', () => {
+    let handler: ((event: unknown, payload: unknown) => void) | undefined;
+    const ipcMain = {
+      on(_ch: string, h: (event: unknown, payload: unknown) => void) { handler = h; },
+    };
+    const fakeLogger = { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn(), trace: vi.fn() };
+    installRendererLogForwarder(ipcMain, {
+      env: { [RENDERER_LOG_FORWARD_ENV]: '1' },
+      logger: fakeLogger as never,
+    });
+    const mainFrame = { id: 'main' };
+    const subFrame = { id: 'iframe' };
+    const event = { senderFrame: subFrame, sender: { mainFrame } };
+    handler?.(event, { level: 'warn', args: ['evil-iframe-injection'] });
+    expect(fakeLogger.warn).not.toHaveBeenCalled();
+    expect(fakeLogger.info).not.toHaveBeenCalled();
   });
 });
 
--- a/electron/__tests__/log.test.ts
+++ b/electron/__tests__/log.test.ts
@@ -1,0 +1,265 @@
+// electron/__tests__/log.test.ts
+//
+// v0.3 task #125 / frag-6-7 §6.6.2 — pino logger contract tests.
+//
+// Three required gates from the task brief:
+//   1. base fields ({ side: 'electron', v, pid, boot })
+//   2. redact list scrubs secrets
+//   3. pino.final crash hooks fire (logger.flush is invoked) on
+//      app.before-quit / uncaughtException / unhandledRejection.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import * as path from 'node:path';
+import { Writable } from 'node:stream';
+import {
+  createLogger,
+  installCrashHooks,
+  installRendererLogForwarder,
+  isRendererLogForwardEnabled,
+  resolveElectronLogDir,
+  ELECTRON_REDACT_PATHS,
+  RENDERER_LOG_FORWARD_CHANNEL,
+  RENDERER_LOG_FORWARD_ENV,
+  electronBootNonce,
+  __setLoggerForTest,
+} from '../log';
+
+/** Capture every line pino writes to a Writable so we can JSON.parse and
+ *  assert structural shape. */
+function makeSink(): { dest: Writable; lines: () => Array<Record<string, unknown>> } {
+  const buf: string[] = [];
+  const dest = new Writable({
+    write(chunk, _enc, cb) {
+      buf.push(chunk.toString('utf8'));
+      cb();
+    },
+  });
+  return {
+    dest,
+    lines: () => buf
+      .join('')
+      .split('\n')
+      .filter((l) => l.length > 0)
+      .map((l) => JSON.parse(l) as Record<string, unknown>),
+  };
+}
+
+beforeEach(() => {
+  __setLoggerForTest(undefined);
+});
+
+describe('electron/log — base fields', () => {
+  it('emits { side, v, pid, boot } on every line', () => {
+    const { dest, lines } = makeSink();
+    const log = createLogger({ appVersion: '0.3.0-test', destination: dest, bootNonce: 'BOOT_TEST_ULID' });
+    log.info('hello');
+    log.warn('world');
+    const out = lines();
+    expect(out).toHaveLength(2);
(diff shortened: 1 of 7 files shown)


Jiahui-Gu (commented):
Reviewer: task #125 / PR #803 — electron/log.ts (own-PR; verdict = CHANGES, posted as COMMENTED)

Solid scaffold. Base fields, redact list (extended w/ §6.6.2 renderer-leaning paths), pino-roll config, lazy singleton + Proxy testability seam, ulid bootNonce parity with daemon (#800), and 15/15 vitest green all check out. main.ts wiring placed correctly after wireCrashHandlers; preload bridge is gated and idempotent.

## Important

1. **Missing fromMainFrame guard in installRendererLogForwarder** (electron/log.ts L334-361). Doc comment explicitly promises: "we trust only the main-frame top-level frame (per the project-wide fromMainFrame ipc guard pattern) — non-main frames cannot inject log lines." The implementation does NOT call fromMainFrame(event). Sister module electron/ipc/rendererErrorForwarder.ts:139 uses exactly this guard. With the env gate ON, any iframe can ipcRenderer.send('log:write', ...) and pollute the main log. Either add the guard or delete the comment — but the project's own forwarder pattern says add the guard.

2. **pino.final vs logger.flush()** (electron/log.ts L268-298). Task brief & frag-12 §12.1 call for pino.final on the three exit paths. Current impl is `try { logger.flush(); } catch { /* best-effort */ }`. Code comment honestly says "pino.final-equivalent" — but they're not equivalent: pino.final is the documented sync, durable-write helper that wraps a destination and is the spec-named primitive. logger.flush() is best-effort and async-leaning. For the uncaughtException / unhandledRejection paths (process is about to exit(70)), this is exactly the case pino.final exists for. Recommend wiring the three handlers via pino.final(logger, (err, finalLogger, evt) => { ... }) per pino docs, OR add a comment explaining why the deviation is acceptable for the file-transport configuration in use here.

3. **Censor string divergence from daemon** (electron/log.ts L175 — censor '[REDACTED]'). Daemon side (#800 daemon/src/log/index.ts DAEMON_REDACT_CENSOR) deliberately uses '[Redacted]' (mixed-case) per spec wording, with an explicit override comment. Cross-side parity asks both sides match. Test at __tests__/log.test.ts:105 also asserts [REDACTED] so flipping the constant requires the test to flip too. Either align electron to [Redacted] (recommended for spec parity), or call out in code why electron diverges.

## Suggestions

4. **Timestamp comment contradicts code** (electron/log.ts L177-181). Comment: "Spec §6.6.1 leaves the pino default in place for daemon, so we mirror it here too" — but next line sets timestamp: pino.stdTimeFunctions.isoTime, which is NOT the pino default (epoch ms). Daemon (#800) does not set timestamp, so it uses the default. Either drop the override (true mirror) or fix the comment.

5. **Duplicate '*.token' entry** (electron/log.ts L90 and L96). Pino accepts duplicates without erroring (verified locally), so this is cosmetic, but the constant is exported and tested via toEqual(arrayContaining(...)) — dedupe to keep the canonical list clean.

6. **Logger Proxy hot-path cost** (electron/log.ts L369-376). Every property access calls getLogger() + Reflect.get + typeof check. Acceptable today (one call per log line) but if this becomes a hot path (renderer forwarder under heavy load), consider caching the resolved singleton after first non-undefined hit. Not blocking.

7. **Preload process.env access** (electron/preload/bridges/ccsmLog.ts:69). Works in the current preload posture (sandbox: false), but if the preload ever moves to sandbox mode process becomes a stripped shim. Defensive 'typeof process !== undefined' is good — consider also documenting the sandbox-mode assumption near the import.

## Spec-match summary
- base fields: PASS — {side:'electron', v, pid, boot} on every line
- pino-roll config: PASS — daily / 50M / 7-file / symlink:true
- redact list: PASS as superset (electron adds §6.6.2 renderer-leaning paths over daemon's minimum); censor string differs (#3)
- bootNonce: PASS — ulid() matches daemon (#800)
- env gate default OFF: PASS — both preload and main check === '1'
- lazy require/Proxy testability: PASS — no prod path leak
- pino.final on 3 hooks: PARTIAL (#2)
- fromMainFrame guard: MISSING (#1)
- zero-rework: PASS — surface stays in v0.4

Verdict: CHANGES. Address #1, #2, #3 then green-light merge.

Jiahui-Gu:
Reviewer CHANGES verdict addressed — pushed `502439b`. Five fixes:

**1. fromMainFrame guard (must-fix)** — `installRendererLogForwarder` now drops payloads where `senderFrame !== sender.mainFrame`, closing the iframe-injection gap when `CCSM_RENDERER_LOG_FORWARD=1`. The exported `fromMainFrame()` in `electron/security/ipcGuards.ts` is typed for `IpcMainInvokeEvent`, but our handler is `ipcMain.on()` (`IpcMainEvent`), so I inlined a structurally identical check (`senderFrame === sender.mainFrame`) rather than widening the shared helper. Pattern + intent mirrors `rendererErrorForwarder.ts:139`. Added a `drops payload from sub-frame senders` test.

**2. pino.final (must-fix) — note** — pino v9 (`9.14.0`, what's pinned in `package.json`) does **not** export `pino.final`. Verified with `node -e "console.log(typeof require('pino').final)" → undefined`. The helper was removed in pino v7+; the v9 idiom for crash hooks is to call `logger.flush()` synchronously after the fatal line — which is what daemon-side task #124 (PR #800) also does. I left the implementation as `logger.flush()` but rewrote the comments so they no longer claim the wrong thing: header now states the v9 reality and cross-references daemon parity. If we want truly synchronous tear-down semantics we'd need to switch to `pino.destination({ sync: true })`, which is a separate (and orthogonal) decision. Happy to take a follow-up issue if you'd rather.

**3. Censor parity (must-fix)** — `[REDACTED]` → `[Redacted]`. Extracted `ELECTRON_REDACT_CENSOR = '[Redacted]'`, applied at both construction sites. Matches daemon `DAEMON_REDACT_CENSOR` byte-for-byte. Test updated.

**4. Timestamp comment vs code (suggestion)** — removed `timestamp: pino.stdTimeFunctions.isoTime` so we inherit pino's default epoch-ms, same as daemon. Comment rewritten to describe what the code actually does.

**5. Duplicate `*.token` (suggestion)** — removed (kept the task-#125-minimum copy at L93, dropped the daemon-superset duplicate).

Tests: `npx vitest run electron/__tests__/log.test.ts` — **16 / 16 passed** (added 1 new sub-frame-guard test). Full electron suite: 837 / 840 (3 pre-existing better-sqlite3 native binding failures unrelated to this PR — `NODE_MODULE_VERSION 145 vs 137`). `tsc --noEmit -p tsconfig.electron.json` clean.

PTAL.

PR #804, opened 2026-05-02
ci: v0.3 task #130 release.yml wire daemon signing (frag-11 §11.3)

Jiahui-Gu:
## Summary
Wire PR #116's `sign-windows.cjs` / `sign-macos.cjs` into `.github/workflows/release.yml` so the standalone daemon binary that ships under `extraResources` (and feeds the daemon-only update channel, frag-11 §11.5(7)) is signed AND verified before electron-builder packs it, plus a separate `notarytool` submission for the macOS standalone binary (frag-11 §11.3.2).

## Steps added (build job, between `Verify daemon binary integrity` and `Package (...)`)

| Step | Platform | Command (key) | Fail-fast |
| --- | --- | --- | --- |
| Materialize Windows signing cert from CSC_LINK | win | `base64 -d $CSC_LINK > $RUNNER_TEMP/ccsm-codesign.pfx` | empty/missing → exit 1 |
| Sign + verify daemon binary (Windows) | win | `signtool sign /f ... /tr ... /td sha256 /fd sha256 daemon/dist/ccsm-daemon-win-<arch>.exe` then `signtool verify /pa /v <bin>` | any nonzero exit → exit 1 |
| Sign + verify daemon binary (macOS) | mac | `codesign --force --sign $CCSM_MAC_IDENTITY --timestamp --options runtime [--entitlements ...] daemon/dist/ccsm-daemon-macos-<arch>` then `codesign --verify --strict --verbose=2 <bin>` | any nonzero exit → exit 1 |
| Notarize standalone daemon (macOS) | mac | `ditto -c -k --keepParent <bin> <zip>` then `xcrun notarytool submit <zip> --apple-id ... --password ... --team-id ... --wait` | rejected / missing creds → exit 1 |

Plus `CCSM_WIN_CERT_PATH` / `CCSM_WIN_CERT_PASSWORD` / `CCSM_MAC_IDENTITY` / `CCSM_MAC_ENTITLEMENTS` env vars added to the existing `Package (Windows)` and `Package (macOS)` steps so PR #116's `afterSign` hook fires on the EB output (covers nested `.exe` / `.dll` / Mach-O inside the `.app` and NSIS payload).

## Verify commands (frag-11 §11.3)
- Win: `signtool verify /pa /v <daemon.exe>` — exit code MUST be 0 (Authenticode default policy passes)
- mac (per-binary): `codesign --verify --strict --verbose=2 <daemon>` — exit 0
[...]

PR #802, opened 2026-05-02
ci: v0.3 task #135 Node 22 + daemon test job (frag-11 §11.5)

Jiahui-Gu:
## Summary

Task #135 / frag-11 §11.5 — unify CI Node target via `daemon/.nvmrc`
(22.11.0) and add a daemon-only test gate.

- All `actions/setup-node@v4` blocks across every workflow now read
  `node-version-file: daemon/.nvmrc` — single source of truth, one
  bump propagates everywhere, no more `'20'` vs `'22'` drift.
- `node_modules` cache keys: label bumped `node20` → `node22` and
  `hashFiles(...)` now covers BOTH root `package-lock.json` AND
  `daemon/package-lock.json` (forward-compatible — if daemon ever
  ships its own lockfile a change to it busts the cache).
- `ci.yml` lint+typecheck+test job (Linux leg) and `release.yml`
  verify job both gain a dedicated daemon-only test step:
  `npx vitest run daemon`.

## Workflow files changed

- `.github/workflows/ci.yml` — setup-node → .nvmrc, cache key node22
  + daemon lockfile, **new daemon-only `Test (daemon only)` step**
- `.github/workflows/e2e.yml` — setup-node → .nvmrc, both cache keys
  (`nm-e2e-*`, `electron-*`) updated
- `.github/workflows/fixture-lint.yml` — setup-node → .nvmrc
- `.github/workflows/installer-size-guard.yml` — setup-node → .nvmrc
- `.github/workflows/parity.yml` — setup-node → .nvmrc, cache key
  bumped + daemon lockfile
- `.github/workflows/proto.yml` — setup-node → .nvmrc
- `.github/workflows/release.yml` — both setup-node blocks → .nvmrc,
  electron cache key updated, **new daemon-only `Test (daemon only)`
  step in verify job**
- `.github/workflows/schema-additive-lint.yml` — setup-node → .nvmrc

(`.github/workflows/ccsm-native.yml` was already pinned to
`daemon/.nvmrc` — left untouched.)

## Test plan

- [x] `python -c "yaml.safe_load(...)"` over all 11 workflow files: parse OK
- [ ] On PR: `ci` job picks up Node 22 from `daemon/.nvmrc` and the
      new `Test (daemon only)` step runs `npx vitest run daemon`
- [ ] On PR: `e2e`, `parity`, `installer-size-guard` all resolve
      Node 22 via `daemon/.nvmrc`
- [ ] On next release tag: `verify` job runs daemon-only test step
      after `npm test`

[...]

PR #799, opened 2026-05-02
chore: v0.3 task #126 daemon pkg block补 (frag-11 §11.1)

Jiahui-Gu:
## Summary
- Verified PR #781 already added daemon binary build pipeline; `daemon/.nvmrc=22.11.0` correct, `pkg.assets` per-arch glob correct.
- Gap补: added explicit `pkg.targets` array (4 targets: node22-{win-x64,macos-x64,macos-arm64,linux-x64}) per frag-11 §11.1 spec example.
- Gap补: added per-arch `scripts.package:{win-x64,macos-x64,macos-arm64,linux-x64}` entries for direct local invocation; existing generic `package` script preserved.

Round-3 P1-7 unchanged: CI matrix legs still drive `--targets node22-${PLATFORM}-${ARCH}` via `scripts/build-daemon-bin.cjs` so each binary embeds only its own arch's `.node` files. The package.json `pkg.targets` array is purely a manifest/default for tooling; CLI `--targets` overrides per leg.

Zero-rework: identical config carries into v0.4 daemon ship.

## Test plan
- [x] `node -e` JSON parse on daemon/package.json passes
- [ ] Local `npm run package:win-x64 -w @ccsm/daemon` (skipped — no clean checkout / native rebuild here; CI matrix exercises real build via build-daemon-bin.cjs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #805, opened 2026-05-02
fix: v0.3 task #132 ccsm path case (Linux fix, P0)

Jiahui-Gu:
## Summary

P0 Linux dogfood blocker (frag-11 §11.6). NSIS uninstaller + daemon resolver use lowercase `ccsm`, but a regression in `tests/e2e/crash-phase1.probe.ts` hard-coded uppercase `CCSM` for the dataRoot crash dir. On Linux that means the probe writes to a parallel directory the daemon never reads, masking real crash regressions and (in the wider failure mode the audit calls out) leaving uninstall + first-boot in an inconsistent state.

## Sweep findings

- `tests/e2e/crash-phase1.probe.ts` — **path literal fixed** (`CCSM` → `ccsm` for win32/darwin/linux).
- `electron/main.ts:45-47` and `electron/crash/incident-dir.ts:29-34` (audit hits) — already lowercase; verified, no change.
- `daemon/src/db/ensure-data-dir.ts` resolver — already lowercase; verified, added migration fallback.
- macOS `CCSM.app` / Windows `CCSM.exe` / `CCSM Dev.exe` — exempt; those come from `productName: "CCSM"`, not the dataRoot.
- `daemon/src/db/__tests__/migration-events.test.ts` legacy `Roaming/CCSM` fixture — kept (it's the v0.2 source path the migration helper exists to read), with inline rule disable.

## Lock-in

1. **ESLint rule** `local/no-uppercase-ccsm-path` (in `eslint-rules/`) flags path-shaped literals (`'CCSM'` as a `path.join` arg, or any `/CCSM/`-shaped string) while exempting env-var names (`CCSM_*`) and macOS/Windows binary names. Reverse-verified with **3 invalid + 3 valid** fixtures in `tests/eslint-rules/no-uppercase-ccsm-path.test.ts`.

2. **Daemon migration fallback** — `ensureDataDir()` on non-Windows now calls `migrateLegacyUppercaseDataRoot()`: if the canonical lowercase root is missing AND a sibling `<parent>/CCSM/` exists, rename it into place and emit a single canonical log line `ccsm_path_migrated from=<legacy> to=<canonical>` to stderr. Best-effort; failures fall through to fresh start. Skipped on Windows (FS is case-insensitive).

## Files changed

- `tests/e2e/crash-phase1.probe.ts` — bug fix (path literal lowercase)
- `daemon/src/db/ensure-data-dir.ts` — `+migrateLegacyUppercaseDataRoot()`, wired into `ensureDataDir()`
- `eslint-rules/no-uppercase-ccsm-path.js` — new local rule
- `eslint.config.js` — register local plugin, ignore `eslint-rules/`
- `tests/eslint-rules/no-uppercase-ccsm-path.test.ts` — 3 invalid + 3 valid reverse-verify
- `daemon/src/db/__tests__/lowercase-data-root.test.ts` — resolver invariant + migration tests (POSIX-gated)
- `daemon/src/db/__tests__/migration-events.test.ts` — inline rule disable on legacy v0.2 fixture

## Test plan

- [x] `npm run lint` — 0 errors (rule fires on seeded uppercase, legitimate uses exempt)
- [x] `npx tsc --noEmit -p tsconfig.electron.json` — clean
- [x] `npx vitest run` on touched suites — 38 passed, 4 skipped (migration tests gated on POSIX host; resolver invariant runs everywhere)
- [ ] Pre-existing failures in `daemon/src/db/__tests__/migrate-v02-to-v03.test.ts` etc. are `better-sqlite3` NODE_MODULE_VERSION 145/137 mismatch, and `electron-load-smoke` requires `dist/electron` to be built — both unrelated to this change (verified by `git stash` + retry on `origin/working`).
- [ ] Manual Linux verification recommended once a build is cut: install → run → uninstall and confirm `~/.local/share/ccsm/` is the only path used.

### DIFF
--- a/daemon/src/db/__tests__/lowercase-data-root.test.ts
+++ b/daemon/src/db/__tests__/lowercase-data-root.test.ts
@@ -1,0 +1,112 @@
+// Task #132 (frag-11 §11.6): verify the daemon's `migrateLegacyUppercaseDataRoot`
+// helper renames a legacy capital-`CCSM` sibling into the canonical lowercase
+// `ccsm` location, and that `resolveDataRoot` itself never emits an uppercase
+// segment for any platform.
+//
+// Uppercase `CCSM` literals in this file are the legacy-path fixtures the
+// migration helper exists to handle — disable the local rule for the whole
+// file rather than tagging each occurrence.
+/* eslint-disable ccsm-local/no-uppercase-ccsm-path */
+
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, rmSync, mkdirSync, writeFileSync, existsSync, readFileSync } from 'node:fs';
+import { tmpdir, platform as osPlatform } from 'node:os';
+import { join } from 'node:path';
+
+import {
+  ensureDataDir,
+  resolveDataRoot,
+  migrateLegacyUppercaseDataRoot,
+  type PathProvider,
+} from '../ensure-data-dir.js';
+
+// The migration helper is only meaningful on case-sensitive filesystems.
+// Windows treats `CCSM` and `ccsm` as the same inode, so the legacy sibling
+// check trivially short-circuits — and any rename "succeeds" by being a
+// no-op. We gate the behavioural tests on a non-Windows host; the pure
+// `resolveDataRoot` invariant test runs everywhere.
+const describeOnPosix = osPlatform() === 'win32' ? describe.skip : describe;
+
+let sandbox: string;
+
+beforeEach(() => {
+  sandbox = mkdtempSync(join(tmpdir(), 'ccsm-task132-'));
+});
+
+afterEach(() => {
+  rmSync(sandbox, { recursive: true, force: true });
+});
+
+describe('Task #132 — lowercase data root (frag-11 §11.6)', () => {
+  it('resolveDataRoot never emits an uppercase CCSM segment on any platform', () => {
+    const win = resolveDataRoot({ platform: 'win32', home: 'C:\\Users\\u', env: { LOCALAPPDATA: 'C:\\Users\\u\\AppData\\Local' } });
+    const mac = resolveDataRoot({ platform: 'darwin', home: '/Users/u', env: {} });
+    const lin = resolveDataRoot({ platform: 'linux', home: '/home/u', env: {} });
+    const xdg = resolveDataRoot({ platform: 'linux', home: '/home/u', env: { XDG_DATA_HOME: '/var/data' } });
+    for (const r of [win, mac, lin, xdg]) {
+      // The lowercase `ccsm` segment must be present, the uppercase one
+      // must NOT — anywhere in the resolved path.
+      expect(r.toLowerCase()).toContain('ccsm');
+      expect(r).not.toMatch(/[\\/]CCSM([\\/]|$)/);
+    }
+  });
+
+  describeOnPosix('migrateLegacyUppercaseDataRoot', () => {
+    it('renames a legacy capital-CCSM sibling to the canonical lowercase path', () => {
+      const parent = join(sandbox, '.local', 'share');
+      mkdirSync(parent, { recursive: true });
+      const legacy = join(parent, 'CCSM');
+      const canonical = join(parent, 'ccsm');
+      mkdirSync(legacy, { recursive: true });
+      writeFileSync(join(legacy, 'marker'), 'legacy-content');
+
+      migrateLegacyUppercaseDataRoot(canonical);
+
+      expect(existsSync(canonical)).toBe(true);
+      expect(existsSync(legacy)).toBe(false);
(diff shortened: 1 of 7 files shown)


Jiahui-Gu (commented):
MERGE-ok (own-PR fallback to COMMENTED) — Task #132 / frag-11 §11.6 / P0 Linux case-sensitivity fix.

## Plan alignment
- Sweep claim verified: independently grepped `'CCSM'` in path contexts across electron/, daemon/, scripts/, build/, tests/. The single real bug at `tests/e2e/crash-phase1.probe.ts:15-18` is fixed (lowercase on all three platform branches). All other path-context CCSM literals are either (a) macOS bundle / Windows binary names derived from `productName: "CCSM"` (sign-macos, required-after-pack, probe-dual-install), (b) negative invariant assertions (incident-dir.test, unified-crash-root.test), (c) intentional v0.2 legacy fixtures with inline disables (migration-events.test, lowercase-data-root.test, the new helper itself), or (d) brand/i18n strings. Audit's two suspected sites (electron/main.ts:45-47, electron/crash/incident-dir.ts:29-34) are confirmed already lowercase — false positive.
- Diff scope = 7 files, all relevant, no creep.

## Migration helper (`migrateLegacyUppercaseDataRoot`)
Safety properties all satisfied:
- No clobber: early-return when canonical exists.
- Idempotent: once renamed, second call is a no-op.
- Best-effort: stat + rename failures both swallowed; daemon boot continues with a fresh canonical root.
- POSIX-gated at the caller (`provider.platform !== 'win32'`); belt-and-suspenders `legacy === canonical` short-circuit also handles case-insensitive volumes.
- Canonical log line `ccsm_path_migrated from=<legacy> to=<canonical>` emitted via console.warn (correct — runs before pino is constructed in some boot orderings).
- Sibling-only rename via `dirname` — won't misfire on `/home/CCSM/.local/share/ccsm`.

## ESLint rule
- Correctly exempts `CCSM_*` env vars (regex), `CCSM.app`, `CCSM.exe`, `CCSM Dev.exe`.
- Bare `'CCSM'` only flagged when parent is `path.join|resolve|normalize` (incl. `posix.x` / `win32.x` / `path.posix.x`) — won't trip on `findInstaller('CCSM')` etc.
- 3 valid + 3 invalid fixtures cover the contract.

## Test coverage
- `lowercase-data-root.test.ts`: invariant test runs everywhere; rename behaviour gated to POSIX (Windows FS is case-insensitive — correct gating).
- File-level `eslint-disable local/no-uppercase-ccsm-path` justified (whole file is legacy-path fixtures).

## Important (non-blocking, recommend follow-up)
The new ESLint rule is registered under the main TS config block, but the global `ignores` array still includes `tests/**`, `scripts/**`, `docs/**`. That means the rule cannot fire on `tests/e2e/crash-phase1.probe.ts` — which is exactly where the original bug lived — nor on future regressions in `tests/`, `scripts/`. The migration helper is the runtime safety net regardless, but the lint guard is effectively only protecting `electron/**` + `daemon/src/**`. Consider a follow-up to scope the rule to a narrower test allowlist (e.g. `tests/e2e/**`, `tests/electron/**`) or carve `tests/`/`scripts/` out of the global ignores for this rule specifically. Not a blocker for the P0 fix.

## Zero-rework
Lowercase `ccsm` is the v0.4 canonical shape; nothing in this PR needs to be undone. Migration helper is purely additive and degrades to a no-op once the legacy population is gone.

Gates: lint clean, tsc clean (electron + daemon), vitest 38 pass + 4 POSIX-skipped. Pre-existing better-sqlite3 NODE_MODULE_VERSION mismatch verified unrelated.

Jiahui-Gu:
Rebased onto latest `origin/working` (post #802 / #804 / #801 / #799 merges).

Conflict resolved:
- **`eslint.config.js`** — single textual conflict where #802 added a `ccsm-local` plugin (with `no-direct-native-import`) at the same lines where this PR registered a separate `local` plugin (with `no-uppercase-ccsm-path`). Merged both rules into the existing `ccsm-local` plugin namespace (single `createRequire`, single plugin object). Renamed the rule reference accordingly across:
  - `daemon/src/db/ensure-data-dir.ts`
  - `daemon/src/db/__tests__/migration-events.test.ts`
  - `daemon/src/db/__tests__/lowercase-data-root.test.ts`
  - `tests/eslint-rules/no-uppercase-ccsm-path.test.ts`

  All `eslint-disable` directives now use `ccsm-local/no-uppercase-ccsm-path`. The reverse-verify test's inline plugin registration was updated to match.

Verification:
- `npm run lint` — 0 errors (warnings only, all pre-existing)
- `npx tsc --noEmit -p tsconfig.electron.json` — clean
- `npx tsc --noEmit -p daemon/tsconfig.json` — clean
- Vitest on touched suites (3 files) — 16 passed / 4 skipped

PR #807, opened 2026-05-02
test: v0.3 task #128 control-socket /stats p99 isolation (frag-3.4.1)

Jiahui-Gu:
## Summary

Adds the regression-marker test for **frag-3.4.1 line 284** spec assertion:

> "saturated data socket does not delay control-socket /healthz p99 by >5 ms"

Real socket roundtrip (no mock): real `net.createServer` control + data transports on per-test temp paths, real `mountEnvelopeAdapter` on each accepted connection, real `createSupervisorDispatcher` + `createDataDispatcher`, real `makeHealthzHandler`. Only the data-plane handler is test-local (`test.dataLoad` — touches its binary payload to keep V8 honest); production data-plane RPCs land in later slices but ride the same event-loop / framing path so this measurement transfers.

Test geometry:
- Baseline: 200 sequential `/healthz` calls, data socket idle → p99
- Loaded:   4 producers x 256 KiB binary frames x 4-deep pipeline x 5 s wall clock, 200 `/healthz` calls staggered across the load window → p99
- Assert: `(loaded p99) - (baseline p99) < 5 ms`

## Measured numbers (this run, Windows 11, Node default)

| | p50 | p99 |
|---|---|---|
| baseline | 0.15 ms | 0.38 ms |
| loaded | 3.09 ms | **37.47 ms** |
| delta | — | **37.10 ms (budget <5 ms — VIOLATES spec by 32 ms)** |

Data-plane request count during the loaded phase: **2336** (sanity: load actually ran).

## Status: regression marker

Daemon currently **FAILS** the spec assertion. The supervisor and data dispatchers share a single Node event loop; saturating the data-socket envelope adapter with binary frames blocks `/healthz` dispatch on the same loop. The two-socket topology (frag-3.4.1 §3.4.1.h) provides surface-area isolation (auth/access control), not CPU isolation.

To keep the suite green while pinning the violation, the test uses `it.fails(...)`. Vitest treats this as PASS only while the inner assertion fails; the moment isolation lands, `.fails` will go red and force the assertion be flipped to unconditional `it()`. The spec gap is therefore visible in CI without leaving the suite red.

## Follow-up (N15-fix candidate)

Event-loop isolation between supervisor and data planes. Candidates:
- `worker_thread` for envelope parsing (data plane only — supervisor stays on main loop)
- `setImmediate` yield inside `processBuffer` between frames in `daemon/src/envelope/adapter.ts` (cheap; may not be enough)
- Child-process data plane (heaviest)

## Test plan

- [x] `npx tsc -p daemon/tsconfig.json --noEmit` clean
- [x] `npx eslint daemon/src/__tests__/control-socket-isolation.test.ts` clean
- [x] `npx vitest run daemon/src/__tests__/control-socket-isolation.test.ts` green (1 expected fail)
- [x] Real socket roundtrip — no mocks; real `net.connect` clients; real envelope frames

### DIFF
--- a/daemon/src/__tests__/control-socket-isolation.test.ts
+++ b/daemon/src/__tests__/control-socket-isolation.test.ts
@@ -1,0 +1,416 @@
+// Task #128 — control-socket /healthz p99 isolation under data-socket load.
+//
+// Spec citation:
+//   docs/superpowers/specs/v0.3-fragments/frag-3.4.1-envelope-hardening.md
+//   line 284: "saturated data socket does not delay control-socket /healthz
+//   p99 by >5 ms".
+//
+// Why this exists
+// ---------------
+// frag-3.4.1 §3.4.1.h splits the daemon's RPC surface into two transports:
+// the control socket (supervisor RPCs only — /healthz, /stats, daemon.hello,
+// daemon.shutdown, daemon.shutdownForUpgrade) and the data socket (every
+// other RPC, including high-throughput ptySubscribe streams). The split
+// exists precisely so a runaway data-plane producer cannot starve the
+// supervisor's liveness probe. If a saturated data socket inflated /healthz
+// p99 by more than 5 ms, three consecutive missed probes inside the §6.5
+// 15 s window would restart the daemon — turning a busy session into an
+// unrecoverable boot loop.
+//
+// Test strategy
+// -------------
+// Pure in-process integration: spin up real net.createServer transports
+// (control + data) on per-test temp socket paths, mount the real envelope
+// adapter on each accepted connection, register the real /healthz handler
+// on a real supervisor dispatcher, and drive a real data-plane handler
+// (registered for this test only) that simulates a streaming-load workload.
+// We deliberately do NOT spawn the daemon binary — pkg builds are slow and
+// the binary's only contribution to this isolation property is the wiring
+// already covered above.
+//
+// Phases:
+//   1. Baseline: 200 sequential /healthz roundtrips with the data socket idle.
+//      Capture every roundtrip latency in ms and compute p99.
+//   2. Loaded: while N concurrent data-socket clients hammer the data plane
+//      with large frames for 5 s, run another 200 /healthz roundtrips on the
+//      control socket and compute p99.
+//   3. Assert (loaded p99) - (baseline p99) < 5 ms per the spec citation.
+//
+// "Real socket, no mock" is honoured: the control + data servers are real
+// net.Server instances; the client is a real net.connect; the envelope
+// adapter is the production module; the dispatcher is the production module.
+// Only the data-plane handler is test-local (the production data-plane RPCs
+// land in later slices — but the wire-level framing is identical, so this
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
MERGE-ok (own PR — cannot self-approve, posting as comment per protocol).

## Review summary
Single-file PR, 416 lines, scope-clean — only `daemon/src/__tests__/control-socket-isolation.test.ts` added; no implementation changes hidden.

## Verification
- **Spec alignment**: literal "saturated data socket does not delay control-socket `/healthz` p99 by >5 ms" matches `frag-3.4.1-envelope-hardening.md` line 284 verbatim. Budget value 5ms matches.
- **Real-socket fidelity**: real `net.createServer` for both control + data, real `mountEnvelopeAdapter`, real `createSupervisorDispatcher` + `createDataDispatcher`, real `makeHealthzHandler`. Only the data-plane RPC body (`test.dataLoad`) is test-local — justified since production data RPCs land in later slices and the wire/event-loop path is what's under test.
- **Saturation methodology sound**: 4 producers × pipeline depth 4 × 256 KiB frames × 5 s, with `dataReqCount > 50` vacuity guard. Probes spread across the load window (`probeIntervalMs`), not bursted at the start. Run actually saw `dataReqs=2928` — genuinely saturated.
- **Sample integrity**: 200 probes after a 20-probe warmup on a fresh client per phase (baseline + loaded). Nearest-rank percentile is deterministic. Healthz `reply.healthzVersion === 1` checked every iteration so we know the response is real, not a hung pipe.
- **`it.fails` usage correct**: vitest treats this as PASS only while the assertion fails; the moment task #151 (event-loop isolation) lands and delta drops below 5ms, vitest will flip RED and force flipping `.fails` back to plain `it()`. Comment block at lines 225-239 explains this loud and clear with the follow-up task pointer.
- **Cleanup**: `afterEach` destroys all opened sockets, awaits all server `.close()`, recursively removes scratch dir. Producer `for...await` loops are awaited via `Promise.all(producerLoops)` before teardown so no orphaned background work. Per-test scratch dir + per-test unique pipe/socket name (pid + Date.now() + random) — no cross-run leaks.
- **Local run**: 1 expected fail, file passes. Measured `baseline p99=0.69ms loaded p99=32.73ms delta=32.04ms` — same order of magnitude as the PR body's 0.38/37.47/37.10 (jitter normal).
- **Zero-rework**: regression marker stays valuable into v0.4 — Listener A is in final-arch and the assertion is the production invariant; nothing here is throwaway.

## Notes (none blocking)
- `maxAcceptPerSec: 100_000` deliberately sidesteps the security rate cap — comment correctly cites #15 round-2 as a separate concern.
- `headerObj` cast in FramedClient.onData is loose-typed (`as { id, ok, value?, error? }`); fine for a test-side decoder.
- Optional future polish: tee `console.log` line to a JSON artifact so the manager dashboard could watch the delta trend over time. Not for this PR.

Failing-as-expected is the green state until #151 lands. Manager OK to admin-merge.

PR #123, opened 2026-04-22
i18n: complete zh coverage + preserve reserved nouns

Jiahui-Gu:
## Summary

Two i18n problems fixed in one PR:

1. **Incomplete Chinese translation** — SettingsDialog had ~60 hardcoded English strings (Theme/Density/Font size, all of NotificationsPane, the entire Updates pane, every label/placeholder/button in EndpointsPane and its editor dialog). App.tsx empty-state buttons, Enter-to-send hint, AppShell resize handle, WindowControls (min/max/restore/close), Dialog close button, and three toast bridges were also hardcoded.
2. **Over-translation of proper nouns** in zh.ts — Session, Group, Endpoint, Model, Sidebar, Tab, App, MCP, CLI were translated into Chinese instead of staying English per owner intent.

### Counts

| Change | Count |
|---|---|
| New i18n keys added (en + zh, parity-checked) | ~95 |
| zh.ts strings de-translated to keep reserved English nouns | ~45 |
| Components newly wired to `t()` | 7 (SettingsDialog, AppShell, WindowControls, ui/Dialog, App.tsx, Sidebar, CommandPalette) |
| Files touched | 10 |

### Reserved terms (owner-approved, kept English in zh.ts)

- **Tech/protocol**: MCP, CLI, IPC, API, URL, JSON, JSONL, SDK, TUI, REST
- **Product/brand**: Claude, Claude Code, Anthropic, Agentory, Electron, GitHub
- **UI/domain nouns**: Settings, Sidebar, Session, Group, Tab, Endpoint, Model
- **Slash commands**: `/clear`, `/model`, `/help`, `/init`, `/agents`, `/mcp`, `/memory`, `/compact`, `/cost`, `/login`, `/logout` and any `/`-prefixed command — never translated

Note: `Settings` as the gear-icon button label still renders as 设置 (it's the user-facing Chinese name for that UI affordance, not a technical reference).

### Settings panes that were hardcoded English and now translate

- **Appearance** — Theme/Font size/Density labels, hints, segmented options (Dark/Light/System, Compact/Normal/Comfortable), font-size aria-label.
- **Notifications** — intro paragraph, all 5 toggle labels + their hints, On/Off labels, Test notification button, Test title/body, all 4 status messages.
- **Updates** — Version/Status labels, Automatic-checks label/hint/toggle, Checking…, Download {version}, Restart & install, all 7 status descriptions, relative-time strings, version-unknown fallback.
[...]

PR #812, opened 2026-05-02
feat: v0.3 task #123 daemon lockfile (frag-6-7 §6.4)

Jiahui-Gu:
## Summary

Implements `<dataRoot>/daemon.lock` single-instance enforcement per frag-6-7 §6.4 + audit gap.

- **`daemon/src/lifecycle/lockfile.ts`** (new) — `acquireDaemonLock({ dataRoot, logger })` returns a `LockHandle { path, pid, release }`. Uses `proper-lockfile` for the atomic `mkdir(daemon.lock.lock)` gate; PID payload written to `daemon.lock` itself for stale-recovery. `probeHolderProcess(pid)` is a pure decider mirroring `dev-parent-watchdog.checkParent`.
- **Boot order** (`daemon/src/index.ts`) — lockfile acquired as the FIRST step in the boot async IIFE, BEFORE socket bind / data-dir touch. A racing second daemon throws `LockfileBusyError` → exit 75 (EX_TEMPFAIL, supervisor `spawn-or-attach` interprets as "attach"). EROFS / EACCES / EPERM / ENOSPC at the lockfile site → `lockfile_erofs_fatal` log line + exit 78 (EX_CONFIG).
- **Stale-PID steal** —
  - **POSIX**: `process.kill(holderPid, 0)`. `ESRCH` → steal with `lockfile_steal` warn line.
  - **Windows**: same `process.kill(pid, 0)` (Node maps internally through `OpenProcess`); `ESRCH` → steal. A `probeWindows` seam exists for a future N-API helper that uses `PROCESS_QUERY_LIMITED_INFORMATION` for a less-privileged check (lands with the §7.M1 ACL helper).
  - Unparseable PID payload (holder crashed mid-write) is steal-eligible without ever calling `process.kill(0)` / `kill(-1)` (which have dangerous broadcast semantics on POSIX).
  - One-shot steal cap via `stealAttempted` guard prevents an infinite loop if two dead-holder racers keep re-mkdir'ing.
- **Swap-lock contract (frag-6-7 §6.4 step 4)** — `daemon.shutdownForUpgrade.releaseLock` action now invokes `lockHandle.release()` instead of the prior no-op stub, so the supervisor's binary-swap window cannot race a fresh-daemon-boot for `bin/`. The same release runs in the SIGTERM-driven `closeTransports` path so a planned exit clears the lock.

## Canonical log lines

- `lockfile_acquired { pid, path, datarootMs }`
- `lockfile_steal { stale_pid, datarootMs, path }`
- `lockfile_erofs_fatal { code, path, datarootMs, err }`

## Test plan

- [x] `daemon/src/lifecycle/__tests__/lockfile.test.ts` — 12 tests, all green:
  - `probeHolderProcess` pure decider (alive / gone / unknown / Windows seam).
  - Real-fs acquire / release / re-acquire / pid-payload assertion.
  - Stale-PID steal emits `lockfile_steal` and overwrites the PID payload.
  - Live-PID holder is NOT stolen (LockfileBusyError instead).
  - EROFS via stub → `lockfile_erofs_fatal` + `LockfileFatalError`.
  - One-shot steal cap (racer re-acquires before us → Busy, no loop).
  - **Cross-process e2e**: spawns two real Node processes via tsx; second exits 75 with `BUSY:<holderPid>` while first holds.
- [x] `npx tsc --noEmit -p daemon/tsconfig.json` clean.
- [x] `npm run typecheck` clean (renderer + electron + daemon).
- [x] `npx eslint` clean on all touched files.

## Unblocks

- **#139 HMAC keystore** — `daemon.secret` rotation needs single-instance enforcement so the rotate window cannot race a second binder.
- **#148 supervisor probe** — frag-6-7 §6.1 spawn-or-attach can now use lockfile-as-gate vs unreliable Win named-pipe `PIPE_UNLIMITED_INSTANCES` probe.
- **#105 SQLite migration** — frag-8 §8.3 step 1a's unconditional `unlink(ccsm.db.migrating[-wal][-shm])` orphan cleanup is only safe under guaranteed single-daemon.

## Out of scope

- Wiring `acquireDaemonLock` into a future `bootDb()` ordering (lockfile → secret → ensureDataDir) — secret-loader and `bootDb` invocation in `index.ts` are separate slices.
- Replacing `process.kill(pid, 0)` with a real `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` N-API helper — lands with §7.M1 pipe-ACL helper which shares the same dependency cost.

### DIFF
--- a/daemon/src/index.ts
+++ b/daemon/src/index.ts
@@ -13,6 +13,14 @@ import {
 } from './handlers/daemon-shutdown.js';
 import { createForceKillSink, type ForceKillJobHandle } from './lifecycle/force-kill.js';
 import { startDevParentWatchdog } from './lifecycle/dev-parent-watchdog.js';
+import {
+  acquireDaemonLock,
+  LockfileBusyError,
+  LockfileFatalError,
+  LOCKFILE_FATAL_EXIT_CODE,
+  type LockHandle,
+} from './lifecycle/lockfile.js';
+import { resolveDataRoot } from './db/ensure-data-dir.js';
 import { installCrashHandlers } from './crash/handlers.js';
 import { installNativeCrashHandlers } from './crash/native-handlers.js';
 import { initDaemonSentry } from './sentry/init.js';
@@ -97,6 +105,32 @@ installCrashHandlers({
 // `daemon/src/crash/native-handlers.ts` for the deferred-WER rationale.
 installNativeCrashHandlers({ runtimeRoot, bootNonce });
 
+// Task #123 — daemon single-instance lockfile (frag-6-7 §6.4).
+//
+// Acquired BEFORE any data-dir or socket work so:
+//   1. A second daemon spawned by a racing supervisor exits cleanly via
+//      `LockfileBusyError` instead of half-binding sockets and
+//      half-touching the data dir (frag-8 §8.3 step 1a's orphan-tmp
+//      unlink is only safe under single-instance enforcement).
+//   2. EROFS / read-only mount surfaces as a clear `lockfile_erofs_fatal`
+//      log + exit(78) BEFORE pino's async destination buffers anything
+//      we cannot flush.
+//
+// Forward-declared so:
+//   - The `daemon.shutdownForUpgrade` handler's `releaseLock` action
+//     (frag-6-7 §6.4 step 4) can release the same handle the boot path
+//     acquired (swap-lock contract).
+//   - The `closeTransports`-then-exit shutdown chain can release on
+//     planned exit so the next boot doesn't see a stale `.lock` dir.
+//
+// The acquire itself runs inside the same async IIFE that boots the
+// listeners (below) — first step. A throw there propagates through the
+// crash handler so a fatal lockfile error gets the same forensic
+// treatment as any other boot failure.
+const dataRoot =
+  process.env.CCSM_DATA_ROOT ?? resolveDataRoot();
+let lockHandle: LockHandle | undefined;
+
 // T25 — force-kill fallback wiring. The reaper-PID set (T38) and the
 // JobObject handle (T39) are owned by the per-spawn wiring that lands
 // alongside the real ptyService. Until those wires are in, the
@@ -161,6 +195,21 @@ async function closeTransports(): Promise<void> {
         logger.warn(
           { event: 'daemon-shutdown.data-socket-close-failed', err: String(err) },
           'data-socket close failed during shutdown',
+        );
+      }
+    })(),
+    // Release the daemon-singleton lockfile so the next boot does not
+    // see a stale `daemon.lock.lock` directory and have to walk the
+    // stale-PID steal path. `release()` is idempotent — the upgrade
+    // handler may have already invoked it via the swap-lock contract.
+    (async () => {
+      if (!lockHandle) return;
+      try {
+        await lockHandle.release();
+      } catch (err) {
+        logger.warn(
+          { event: 'daemon-shutdown.lockfile-release-failed', err: String(err) },
+          'daemon.lock release failed during shutdown',
         );
       }
     })(),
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
MERGE-ok (own PR — cannot self-approve, manager admin-merges) — task #123 daemon lockfile (frag-6-7 §6.4).

## Spec match (§6.4)
- Lockfile path `<dataRoot>/daemon.lock` via proper-lockfile (O_EXCL mkdir on the .lock sibling) + PID payload write — exact match.
- Acquire-FIRST ordering verified: lockfile.acquire() is the first await inside the boot async IIFE (daemon/src/index.ts:488), well before controlSocket.listen() / dataSocket.listen() and any data-dir work. ensureDataDir is owned by boot-orchestrator (not yet wired into index.ts pending #105) — ordering invariant holds for everything currently in the boot path.
- Stale-PID steal: POSIX kill(pid, 0) ESRCH=gone path tested; EPERM correctly returns 'unknown' (conservative — never steal a lock from a process we cannot signal). Windows path goes through Node's process.kill mapping to OpenProcess; probeWindows seam reserved for future ccsm_native PROCESS_QUERY_LIMITED_INFORMATION helper as noted in the brief.
- Exit codes: EX_TEMPFAIL=75 on busy + EX_CONFIG=78 on EROFS/EACCES/EPERM/ENOSPC (sysexits.h conventions verified — LOCKFILE_FATAL_EXIT_CODE constant exported and asserted in test).
- Swap-lock contract (§6.4 step 4): daemon.shutdownForUpgrade.releaseLock now invokes lockHandle.release() instead of the prior no-op stub; closeTransports also releases for SIGTERM-exit safety. release() is boolean-guarded idempotent so both paths can wire without coordination.
- All four canonical log lines present and asserted: lockfile_acquired (info), lockfile_busy (error), lockfile_steal (warn), lockfile_erofs_fatal (error).

## Code quality highlights
- SRP discipline: producer / pure-decider (probeHolderProcess never throws) / sink — clean test seams, mirrors dev-parent-watchdog shape per repo convention.
- readHolderPidSafe returns null (not 0/-1) for unparseable payloads — avoids POSIX kill(0,0) process-group broadcast hazard. Real footgun caught upstream.
- stealAttempted guard caps the steal-retry at exactly one — test 6 exercises the racer-wins case (two ELOCKEDs → LockfileBusyError, no infinite loop).
- raiseFatalIfReadOnly centralizes the EROFS-class mapping; unknown error codes re-throw with original stack preserved (no over-classification).
- Pre-pino stderr writes on both busy + fatal branches per §6.4 'pre-pino, since pino isn't initialized until after lockfile' — surfaces even if pino's async dest hasn't drained.

## Tests
- 12/12 green (re-verified locally: `npx vitest run daemon/src/lifecycle/__tests__/lockfile.test.ts` → 12 passed in 4.10s).
- Cross-process race test is a REAL spawn of two Node+tsx processes (not a unit mock): A holds via marker-file sync, B asserts exit code 75 + stdout BUSY:<pid>. 60s timeout budget for cold-start tsx on slow Win runners.
- Daemon typecheck clean, eslint clean on the 3 changed files.

## Diff scope
- daemon/src/lifecycle/lockfile.ts (new, 451 lines)
- daemon/src/lifecycle/__tests__/lockfile.test.ts (new, 461 lines, 12 tests)
- daemon/src/index.ts (+127 / -8: import + dataRoot resolve + lockHandle decl + acquire IIFE prelude + closeTransports release branch + shutdownForUpgrade.releaseLock real impl)
Matches the 3-file scope in the brief.

## Zero-rework
Module surface (acquire/release/probe + LockfileBusyError/LockfileFatalError + DAEMON_LOCK_FILENAME/LOCKFILE_FATAL_EXIT_CODE constants) is final-arch — no v0.4-will-replace placeholders. probeWindows seam is documented as reserved for future ccsm_native helper without committing to it now.

## Minor (non-blocking, suggestions only)
- LockfileBusyError.holderPid falls back to 0 when post-steal-loss PID payload is unreadable. Distinguishable from any real PID but null/-1 might be clearer for forensics. Optional follow-up.
- Commit message correctly cites §6.4 (brief said §6.2 in error — author called this out).

Unblocks #139, #148, #105 as advertised.

PR #129, opened 2026-04-22
fix(slash): /model picker + /config opens settings

Jiahui-Gu:
## Summary

- **/model** previously routed to Settings → Connection. Now opens an in-chat **ModelPickerDialog** (Radix) listing every model from the store (populated by `loadModels()` from `~/.claude/settings.json` + env). Arrow-key nav, Enter-to-commit, current-model checkmark. Selection goes through `setModel` so the running agent is notified via `agentSetModel` — same path as the StatusBar chip, so the next message uses the new model.
- **/config** keeps opening the Settings dialog (gear-icon parity), now covered by a dedicated probe so it can't silently regress.
- New `openModelPicker` bridge in `slash-commands/ui-bridge.ts`, mirroring `openSettings`. App.tsx subscribes once and mounts the dialog in both the no-active-session and active-session render branches.

## Files

- `src/slash-commands/ui-bridge.ts` — added `setOpenModelPickerListener` / `openModelPicker`.
- `src/slash-commands/handlers.ts` — `handleModel` now calls `openModelPicker()`.
- `src/components/ModelPickerDialog.tsx` — new dialog (Radix Dialog + roving listbox keyboard nav).
- `src/App.tsx` — wires the new bridge + mounts the dialog.
- `src/i18n/locales/{en,zh}.ts` — `modelPicker.title/description/current`.
- `tests/slash-commands-handlers.test.ts` — `/model` test asserts picker opens (not Settings).
- `scripts/probe-slash-model.mjs`, `scripts/probe-slash-config.mjs` — new probes.
- `scripts/probe-slash-exec.mjs` — fixed pre-existing assertion that looked for the now-removed "General" tab label.

## Test plan

- [x] `npm run typecheck` passes.
- [x] `npm test` — slash-commands tests (registry / handlers / picker) all green (40 tests). Pre-existing native-module failures in `electron/__tests__/db.test.ts` unrelated to this PR (better-sqlite3 ABI mismatch).
- [x] `node scripts/probe-slash-model.mjs` against dev server: picker opens with seeded models, selection updates `store.model` + active session model, no Settings dialog leak.
- [x] `node scripts/probe-slash-config.mjs`: Settings dialog opens, content matches the gear-button click.
- [x] `node scripts/probe-slash-exec.mjs` continues to pass after the assertion update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #820, opened 2026-05-02
feat(daemon): daemonTraceId + wire 5 interceptors into envelope dispatcher (#153)

Jiahui-Gu:
## Summary

Closes the two follow-ups from PR #809 (Task #129):

1. **`daemonTraceId` (dual-id correlation)** — trace interceptor now mints a 16-byte hex daemon-side join id ALWAYS distinct from `traceId`, even when the client did not supply one. The id is emitted on the `envelope_trace` completion log line AND echoed to the client in the reply envelope header `x-ccsm-daemon-trace-id`, so logs from both sides can be joined in either direction.
2. **Wire the 5-slot interceptor chain into the envelope dispatcher** — new `daemon/src/envelope/interceptor-chain.ts` composes `hello -> trace -> deadline -> migrationGate -> dispatcher -> metrics` per spec frag-3.4.1 §3.4.1.f. The envelope adapter routes through the chain when `chainWiring` is supplied and falls back to direct dispatch otherwise (preserves backward compat with current call sites).

Per the v0.3 zero-rework rule, the envelope dispatcher remains the control-plane carrier in v0.4 (frag-6-7 §6.5), so this wiring is permanent.

## Test plan

- [x] `daemon/src/envelope/__tests__/trace-interceptor.test.ts` — `daemonTraceId` always distinct from `traceId`; round-trip propagates both into handler context; lands on success AND throw paths.
- [x] `daemon/src/envelope/__tests__/interceptor-chain.test.ts` (new) — chain echoes `daemonTraceId` on ok / dispatcher-error / interceptor-reject / handler-throw paths; deadline + migration gate short-circuit before dispatcher; metrics-error bumped on rejection.
- [x] `daemon/src/envelope/__tests__/adapter.test.ts` — adapter chain path splices `x-ccsm-daemon-trace-id` into the reply envelope on both ok and rejected calls; trace sink sees both ids on the same log line.
- [x] All 232 envelope tests pass.
- [x] `tsc --noEmit -p daemon/tsconfig.json` clean (only pre-existing unrelated `resolveDataRoot` dupe remains in `daemon/src/index.ts`).
- [x] eslint clean on touched files.

Generated with Claude Code


Jiahui-Gu:

PR #827, opened 2026-05-02
exp c (#166): deps.inline only — does it hang?

Jiahui-Gu:
Experiment for #164 root-cause investigation. NOT for merge. Revert after data collected.


Jiahui-Gu:
Superseded by #832 systematic fix (#164). Root cause confirmed via this exp.

PR #829, opened 2026-05-02
exp e (#167): cache-key includes .nvmrc — does CI hang go away?

Jiahui-Gu:
Experiment for #164. NOT for merge.

Hypothesis: PR #825 hang is caused by Cache node_modules step holding a stale better-sqlite3 native binding compiled against Node 22.11 ABI, while commit 239f20e bumped daemon/.nvmrc from 22.0 → 22.22.2. Cache key did not include .nvmrc, so the new Node loads the old ABI binding without erroring (CJS load passes) but hangs the worker at first DB call.

Change: fold `hashFiles('daemon/.nvmrc')` into the node_modules cache key so any Node bump busts the cache automatically.

Expected: first run = cache miss, full `npm ci` + native rebuild, then test step finishes (no hang). Subsequent runs cache-hit on a fresh tree built against current Node.


Jiahui-Gu:
Superseded by #832 (#164).

PR #830, opened 2026-05-02
ci(verify): does last-green commit e48b3bcc still pass today

Jiahui-Gu:
Diagnostic only — DO NOT MERGE. Verifying that e48b3bcc still passes CI test step today (was last green at 2026-05-02T06:58Z).


Jiahui-Gu:
Closing: e48b3bcc itself fails today (ERR_REQUIRE_ESM in html-encoding-sniffer → @exodus/bytes/encoding-lite.js). This is environment drift (transitive dep published as ESM-only), not a code regression on working. Bisect aborted.

PR #831, opened 2026-05-02
fix(ci): pin @exodus/bytes to last CJS version

Jiahui-Gu:
Root cause: @exodus/bytes 1.14.0+ went ESM-only. html-encoding-sniffer (jsdom transitive) does require() of it -> ERR_REQUIRE_ESM -> vitest worker fails to start. Pinning via npm overrides until upstream is fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2375,24 +2375,6 @@
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Superseded by PR #832 (diagnostic lockfile restoration; pinning @exodus/bytes did not work because all versions are ESM-only).

PR #174, opened 2026-04-23
fix(streaming): include partial messages + chat in-progress dots

Jiahui-Gu:
## Root cause

Per [docs/design/streaming-ux-investigation.md](../blob/working/docs/design/streaming-ux-investigation.md) §3:

The spawner argv omitted \`--include-partial-messages\`, so \`claude.exe\` only emitted a single \`assistant\` frame at turn end. The renderer's \`PartialAssistantStreamer\` + \`streamAssistantText\` were already wired but never received \`stream_event\` deltas → long replies appeared as a frozen 5–15s wait followed by a one-shot dump.

Additionally (§4): nothing in the chat scroll area indicated progress while waiting for the first token. Only the InputBar showed running state.

## Q1 — Streaming actually streams

\`electron/agent/claude-spawner.ts:buildSpawnArgs\` adds \`--include-partial-messages\` to the base argv. First character now lands ~0.5–1.5s after API ack instead of at end-of-turn.

## Q2 — In-progress dots in chat

\`src/components/ChatStream.tsx\` renders a monospace \`· · ·\` indicator (U+00B7), \`text-state-running\`, framer-motion opacity pulse (1.2s, easeInOut), at the assistant column.

Condition:
\`\`\`
running === true
&& noPendingPermissionBlock
&& (blocks.length === 0 || lastBlock.kind === 'user')
\`\`\`

Disappears the instant the first assistant delta lands. Suppressed while a permission prompt is awaiting the user (different wait — for human, not for tokens).

## Verification

Ran the prod-bundled renderer via \`npm run dev:web\` + a one-shot playwright script that seeds the store directly to drive the state transitions. Result: PASS.

Screenshots in \`verification/\`:
- \`T0-user-sent-dots-visible.png\` — 0.5s after user message lands, dots visible at the assistant column
- \`T1-first-token-dots-gone.png\` — 0.2s after first assistant delta, dots gone, text begins streaming
- \`T2-incremental-streaming.png\` — after 5s of incremental chunks, text grew to 269 chars (incremental, not one-shot)

Probe script: \`verification/verify-streaming-ux.mjs\` (one-shot, not a probe-e2e harness file).

## Tests

- \`electron/agent/__tests__/claude-spawner.test.ts\` — regression guard asserting \`--include-partial-messages\` is always in argv
- \`tests/chatstream-thinking-dots.test.tsx\` (new) — 5 RTL cases covering the (running × last-block-kind × pending-permission) matrix
- All 604 tests pass; typecheck + build clean

[...]

PR #833, opened 2026-05-02
fix(ci): vitest pool=threads (#174 Plan E)

Jiahui-Gu:
## Summary
- Force vitest `pool: 'threads'` (was default `forks`) to dodge `html-encoding-sniffer` ESM strict-require error under Node 22 CI workers.
- Threads pool runs tests in `worker_threads` inside the parent vitest process — the ESM module graph is already wired by Vite's loader, so the strict-require boundary that breaks `forks` does not apply.

## Why forks broke
Vitest 4.x defaults to `forks`, which spawns child Node processes per worker. Each child re-imports the full transitive dep graph. Under Node 22's stricter `require(esm)` rules, `html-encoding-sniffer` (ESM-only, pulled in via `@exodus/bytes` -> jsdom's whatwg-encoding) throws `ERR_REQUIRE_ESM` at worker boot, killing every test file.

## Test plan
- [x] Local `npx vitest run` on Node 22: 0 ESM errors. (Remaining failures are unrelated `better-sqlite3` ABI mismatch on dev box; CI rebuilds native modules.)
- [ ] CI green on this PR.

Refs #174.

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -16,6 +16,19 @@ export default defineConfig({
   },
   test: {
     environment: 'jsdom',
+    // Force threads pool (#174 Plan E): the default `forks` pool spawns
+    // child Node processes that re-import every transitive dep. Under
+    // Node 22 strict require, html-encoding-sniffer (ESM, pulled in via
+    // @exodus/bytes) blows up at worker boot. `threads` runs tests in
+    // worker_threads inside the parent process which already has the
+    // ESM module graph wired up, dodging the strict-require error.
+    pool: 'threads',
+    poolOptions: {
+      threads: {
+        singleThread: false,
+        isolate: true,
+      },
+    },
     include: [
       'tests/**/*.test.ts',
       'tests/**/*.test.tsx',


Jiahui-Gu (commented):
MERGE-OK: minimal +13 line diff; only vitest.config.mts touched. `pool: 'threads'` placed correctly inside root `test:` block. `poolOptions.threads` uses safe defaults (`singleThread: false` keeps parallelism, `isolate: true` preserves per-test-file module registry — so existing `vi.mock`/`process.env` mutations across the suite remain isolated, no new flake risk vs forks). Comment clearly explains the forks→threads + Node 22 strict-require + html-encoding-sniffer rationale and ties back to #174 Plan E. No unintended changes (deps.inline / environment / coverage / include all untouched). LGTM.

Jiahui-Gu:
Plan E (forks→threads) 不是根因 fix — CI 三色全 fail with same ERR_REQUIRE_ESM. threads worker 启动时仍 require @exodus/bytes (via html-encoding-sniffer)。本地之所以 0 ESM 是因为 happy-path 没 trigger jsdom 加载链。Superseded by Plan A/B/C/D。

PR #834, opened 2026-05-02
fix(ci): split vitest into node/dom projects to isolate jsdom ESM (#172)

Jiahui-Gu:
## Summary

CI loader crashes when vitest's jsdom env imports html-encoding-sniffer@6 (ESM) -> @exodus/bytes. Most of the suite doesn't actually need a DOM, so split vitest.config.mts into two projects:

- **node** (default): environment=node. Runs everything except .tsx files and a curated set of .ts files that touch DOM globals or import @testing-library/react. Never loads jsdom -> never trips the ESM bug.
- **dom**: environment=jsdom. Runs the curated DOM-using set, with the existing tests/setup.ts.

## DOM-using .ts files moved to the dom project

Identified via grep for window/document/HTMLElement/localStorage usage and @testing-library/react imports:

- tests/store-rename-session.test.ts
- tests/store-backfill-titles.test.ts
- tests/store-preferences.test.ts
- tests/sidebar/useSidebarDnd.test.ts
- tests/components/cwd/useCwdRecentList.test.ts
- tests/components/cwd/useCwdPanelPosition.test.ts
- tests/agent-lifecycle-unfocused.test.ts
- tests/drafts.test.ts
- tests/stores/slices/sessionTitleBackfillSlice.test.ts
- tests/stores/slices/sessionCrudSlice.test.ts

All .tsx tests (component tests using @testing-library/react) also run under the dom project.

## Test plan

- [x] Local: `npx vitest run --project node tests/appearance.test.ts` passes (sample non-DOM)
- [x] Local: `npx vitest run --project dom tests/components/Button.test.tsx` passes (sample DOM)
- [x] Local: full `npx vitest run --project node` shows 0 ESM crashes (only pre-existing failures from missing local better-sqlite3 native binding + missing dist/electron, identical with or without this patch)
- [ ] CI: node + dom projects both green

Refs task #172 Plan G.


Jiahui-Gu (commented):
MERGE-OK (with caveats) — code review for vitest projects split.

What was checked
- Diff: only vitest.config.mts (+66/-21). No test source changes; no it.skip / it.only / it.todo smuggled in.
- Project layout: shared resolve/alias replicated to both projects; root coverage block preserved (vitest projects mode keeps root coverage; OK).
- Node-project DOM scan: scanned all 67 .ts files now in node project for document/window/@testing-library/render(/screen/HTMLElement/localStorage/etc.
  - Only 1 real DOM use: tests/appearance.test.ts:96 — `typeof window !== 'undefined' ? window.innerWidth : 1440` and the assertion derives `expected` from the same expression, so equality holds under either env. Safe.
  - All other hits were string-content matches ('window.onerror' etc., narration in comments, the literal substring "window" in unrelated contexts). No real DOM API calls.
- setupFiles: setupFiles=['tests/setup.ts'] is now ONLY on the dom project. Pre-PR it was global. tests/setup.ts contains: jest-dom matchers, initI18n('en'), ResizeObserver shim, react-virtuoso vi.mock. Grepped node-project files for useTranslation / i18next / react-virtuoso / jest-dom imports — zero matches. So dropping setupFiles from node project is safe in practice.
- CI invocation: package.json `"test": "vitest run"` — no --project, so both projects run. `npx vitest run daemon` (ci.yml/release.yml) and `npx vitest run tests/parity` (parity.yml) filter by path; daemon/parity files only match node project's include — fine.
- Config concerns: top-level testTimeout=15000 should still apply (vitest projects inherit unset timeouts from root). globals=true is set per project.

Risk flags (reason this is "with caveats" not "ship it")
1. The dom project STILL loads jsdom. The original failure mode (html-encoding-sniffer@6 ESM crash via jsdom) will recur in dom-project workers if the loader bug bites. This PR only protects the 99% node tests — it does not fix dom tests. If CI fails again, expect it on dom project.
2. Sister PR #833 (forks->threads) was local-clean but CI 三色 fail. Same caution applies here: local node project zero ESM does not guarantee CI-pass for the dom project. Recommend watching the in-progress CI run rather than merging blind.
3. tests/appearance.test.ts under node env: window is undefined → assertion uses 1440 fallback on both sides. Will pass, but the test no longer exercises the jsdom-window path — minor coverage erosion (acceptable).
4. Root testTimeout/coverage live outside `projects[]`. If a future vitest version stops inheriting these into projects, behavior silently changes. Not blocking.

Suggested follow-ups (non-blocking)
- If dom project also crashes on CI, the next layer would be either dropping/replacing jsdom or pinning html-encoding-sniffer to v5.
- Add a brief comment in vitest.config.mts noting that DOM_TESTS is hand-curated and must be updated when new jsdom-using .ts tests land (currently the comment says it but a CI guard would be better).

Verdict: MERGE-OK pending green CI. Code itself is clean and minimal; 99% scope reduction is real. But do not merge until CI is actually green — local-pass pattern has burned us once already.

Jiahui-Gu:

PR #173, opened 2026-04-23
refactor(e2e): introduce themed harnesses (B+C phase 1+2 — AGENT cluster)

Jiahui-Gu:
## Summary

Lands the brainstorm doc's **Option B (themed harnesses) + cultural rule from C** (extend existing, don't add a new file per bug). Design discussion: `docs/e2e/single-harness-brainstorm.md`.

### Phase 1 — infra prerequisites

- **Case-id logging convention**: every harness case logs `[case=<id>] …` so a tail of mixed output is bisectable.
- **`scripts/probe-helpers/reset-between-cases.mjs`**: closes sessions via IPC, resets zustand store to empty baseline, wipes `messages` + `app_state` rows in SQLite, dismisses Radix portals + clears DOM selection, drains caller-supplied disposers (monkey-patch restores).
- **`--only=<case>[,<case>]` flag** on `harness-runner` for local single-case dev.
- **Per-case Playwright tracing** — only persisted to `scripts/e2e-artifacts/<harness>/<case>/trace.zip` on failure (success path discards to keep CI green-runs light); also dumps `failure.png` for fast triage.

### Phase 2 — pilot `harness-agent.mjs`

5 store-only AGENT cases merged into one Electron launch:

- `streaming`
- `streaming-caret-lifecycle` (was `probe-e2e-streaming-journey-caret-lifecycle.mjs`)
- `inputbar-visible`
- `chat-copy`
- `input-placeholder`

Per-file probes left as breadcrumbs with top-of-file `// MERGED INTO scripts/harness-agent.mjs` marker. Runner skips them via the in-file `MERGED_INTO_HARNESS` set in `scripts/run-all-e2e.mjs`.

### Runner update

`scripts/run-all-e2e.mjs` now:
- discovers `harness-*.mjs` alongside `probe-e2e-*.mjs` (harnesses run first — fail-fast on regressions in merged cases),
- uses 5min timeout for harnesses (probes still 30s),
- skips merged probes via `MERGED_INTO_HARNESS`.

### Docs

`docs/e2e-runner.md` rewritten to cover both surfaces: how to add a new case (which surface + naming conventions + dispose responsibility), the cases that **can't** be merged (cold-start cluster from brainstorm §9), and how `--only` works locally.

## Measurements (this branch, headless prod-bundle on Windows)

| Metric | Value |
|---|---|
| typecheck | PASS |
| build | PASS |
| harness-agent — 5 cases wall clock | 13.8–16.4s, avg **15.6s** over 20 runs |
[...]

PR #835, opened 2026-05-02
fix(ci): vitest deps.inline @exodus/bytes (#173 Plan F)

Jiahui-Gu:
## Summary
- Adds `test.server.deps.inline` for `@exodus/bytes` and `html-encoding-sniffer` so vitest pre-bundles these ESM-only deps as CJS, avoiding `ERR_REQUIRE_ESM` from the jsdom transitive `require()` chain under Node 22.
- Task #173 / Plan F. CI was broken by `@exodus/bytes` ESM being `require()`-d by `html-encoding-sniffer` -> `whatwg-encoding` -> jsdom under Node 22.

## Test plan
- [x] Local `npx vitest run`: zero `ERR_REQUIRE_ESM` / `@exodus/bytes` / `html-encoding-sniffer` errors. Remaining failures (57) are unrelated `better-sqlite3` native-module ABI mismatch on local box (Node 24 vs binary built for Node 22) and do not occur in CI.
- [ ] CI green on `fix/v03-vitest-deps-optimizer`.

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -14,8 +14,18 @@ export default defineConfig({
       '@ccsm/proto-gen': path.resolve(__dirname, 'gen/ts/index.ts'),
     },
   },
+  // Pre-bundle ESM-only deps as CJS so the jsdom transitive `require()`
+  // chain (jsdom → html-encoding-sniffer → @exodus/bytes etc.) doesn't
+  // hit `ERR_REQUIRE_ESM` under Node 22. See task #173 / Plan F.
+  // In Vitest 3+/4.x this lives at `test.server.deps.inline` (the old
+  // top-level `deps.inline` is deprecated).
   test: {
     environment: 'jsdom',
+    server: {
+      deps: {
+        inline: ['@exodus/bytes', 'html-encoding-sniffer'],
+      },
+    },
     include: [
       'tests/**/*.test.ts',
       'tests/**/*.test.tsx',


Jiahui-Gu:
REVIEW VERDICT: REQUEST-CHANGES (posted as comment — self-author blocks formal review).

CI is already failing with the exact same ERR_REQUIRE_ESM errors this PR claims to fix.

**Evidence (run 25253463796):**
- ubuntu-latest: FAIL @ 3m05s
- macos-latest: FAIL @ 2m26s
- windows: still pending but will fail same way
- Failure mode: `Error: require() of ES Module .../node_modules/@exodus/bytes/encoding-lite.js from .../html-encoding-sniffer/lib/html-encoding-sniffer.js not supported` — repeated for every jsdom suite as `[vitest-pool]: Failed to start forks worker for test files ...`.

**Why `test.server.deps.inline` doesn't work here:**
`server.deps.inline` only inlines modules when *Vite's module graph* loads them. The failing chain is `forks` worker → Node native `require()` of `node_modules/html-encoding-sniffer/lib/html-encoding-sniffer.js` → native `require()` of `node_modules/@exodus/bytes/encoding-lite.js`. Neither file goes through Vite's transformer, so `server.deps.inline` is a no-op for this stack. This matches what exp c (#827) already proved: `deps.inline` alone, in any spelling, does not fix the require-ESM chain in pool=forks.

**Functionally different from #827?** No.
- #827: `test.server.deps.inline = [/^@exodus\//, 'html-encoding-sniffer']`
- #835: `test.server.deps.inline = ['@exodus/bytes', 'html-encoding-sniffer']`

Regex vs string is irrelevant — both match the same modules. CI confirms identical failure mode and timing.

**About the \"local 0 ESM\" claim:**
PR body says local had 0 ESM errors but also \"57 failures from better-sqlite3 native-module ABI mismatch (Node 24 vs binary for Node 22).\" That ABI mismatch fails the worker *before* the html-encoding-sniffer require chain runs, so the local run cannot validate this fix. Same trap as #833.

**Real fixes** (pick one — avoid yet-another-deps.inline-variant PR):
1. **Per-file environment opt-out** — most failing files (daemon/lifecycle/lockfile, ipcRegistrar, control-socket-isolation, harness-daemon-mode, daemon-shutdown*, drop-slowest.probe, etc.) are pure Node tests that don't need jsdom. `environmentMatchGlobs` or `// @vitest-environment node` headers would skip jsdom load → no html-encoding-sniffer → no ESM error.
2. **Replace/downgrade** `html-encoding-sniffer` to a version without `@exodus/bytes`, or pin `@exodus/bytes` via `overrides`.
3. **patch-package** html-encoding-sniffer to dynamic `import()`.
4. **Node flag** `--experimental-require-module` (Node 22.12+) to allow `require()` of ESM natively.

Recommend close #835 + #833 and pursue option 1 (root-cause: too many tests inherit jsdom unnecessarily).

Jiahui-Gu:
Plan F (deps.inline) 不修根因: server.deps.inline 只影响 vite module graph, forks worker 是 Node 原生 CJS loader 直接 require html-encoding-sniffer, 不走 vite。CI 三色 fail 已确认。Superseded by Plan A/B/C/D/G。

PR #838, opened 2026-05-02
fix(ci): pin Node 22.18.0 + html-encoding-sniffer 5.x override (#169)

Jiahui-Gu:
## Root cause

CI on `working` has been red since 2026-04-30. Last green run was Node 20 (e983c04a). PR #802 switched workflows to Node 22 — every run since fails in the test step with:

```
ERR_REQUIRE_ESM: require() of ES Module
  node_modules/@exodus/bytes/encoding-lite.js
  required by html-encoding-sniffer@6.0.0
  required by jsdom (vitest workers)
```

Node 22 tightens `require()` of ESM modules. `@exodus/bytes` ships ESM-only (every version 1.0.0–1.15.0). `html-encoding-sniffer@6.0.0` is the first to depend on it; jsdom@29 picks `^6.0.0`, so any indirect upgrade pulls in the explosion.

## Fix (Plan B)

Add an npm `overrides` block to root `package.json`:

```json
"overrides": { "html-encoding-sniffer": "5.x" }
```

`html-encoding-sniffer@5.0.0` is the last CJS release — smaller blast radius than rolling Node back to 20. After regenerating the lockfile:

```
ccsm@0.2.0
`-- jsdom@29.1.1
  `-- html-encoding-sniffer@5.0.0
```

Also pinning Node patch via `daemon/.nvmrc` → `22.18.0` (was 22.11.0) and adding a root `.nvmrc` for local-dev. All CI workflows already read `node-version-file: daemon/.nvmrc`, so the bump propagates automatically.

## Local verification

- `node -v` → v24.14.1 (local; CI runs 22.18.0)
- `npm ls html-encoding-sniffer` → 5.0.0 under jsdom@29.1.1 (no 6.x anywhere)
- `npx vitest run` → **2890 passed / 2 failed / 37 skipped** (2930 total). The 2 failures are unrelated `harness-daemon-mode.test.ts` boot-timeouts (15s); zero `ERR_REQUIRE_ESM` / `@exodus/bytes` / `encoding-lite` errors in the entire log.

## Files

- `.nvmrc` (new, root) — `22.18.0`
- `daemon/.nvmrc` — `22.11.0` → `22.18.0`
- `package.json` — `overrides` block
- `package-lock.json` — regenerated to apply override

## Coordination

Parallel attempt under #170 (Plan A — roll Node back to 20). Whichever lands first, close the other.


Jiahui-Gu (commented):
MERGE-OK (code review verdict; CI lint failure unrelated — see note).

Reviewed in pool-13 against origin/working..origin/fix/v03-pin-node22-override-encoding-sniffer.

Diff summary (4 files, 1 commit bbc0513):
- package.json: top-level `"overrides": { "html-encoding-sniffer": "5.x" }` correctly placed as sibling of `dependencies` (not nested inside it). Syntax matches npm overrides spec.
- .nvmrc (new, root): 22.18.0
- daemon/.nvmrc: 22.11.0 -> 22.18.0
- package-lock.json: regenerated.

Lockfile verification (the load-bearing check):
- Exactly 1 occurrence of `node_modules/html-encoding-sniffer` resolving to 5.0.0.
- 0 tarballs of `html-encoding-sniffer-6.*` anywhere in the tree.
- jsdom's spec is still `^6.0.0` but resolution is forced to 5.0.0 — exactly the override semantics we want.
- hes@5.0.0 engines = `^20.19.0 || ^22.12.0 || >=24.0.0` — Node 22.18.0 satisfies.

Risk assessment on hes 5 vs 6:
- Both versions share the same single entry path `lib/html-encoding-sniffer.js` and the same dep `@exodus/bytes` (5.x: ^1.2.0; 6.x: ^1.6.0 — minor bump only).
- 5.0.0 was published 2025-12-23, 6.0.0 published 2025-12-26 (3 days apart). 5.0.0 is the deliberate "last CJS" escape hatch cut right before the ESM-only 6.x. The 4.x line (2023) is significantly older; using 5.x is the right choice, not a regression.
- No `type: module` on 5.0.0 (CJS), `type: module` on 6.0.0 (ESM-only) — confirms this addresses the ERR_REQUIRE_ESM root cause for downstream CJS consumers.
- Public API surface is essentially identical between 5 and 6 per the upstream pattern; jsdom's call site is unaffected.

Heads-up (not blocking this PR):
- The CI lint step on this branch likely fails due to a separate working-branch bug (eslint.config.js uses ESM import without root `type: module`). That is fixed by #181/dev-eslint181, not this PR. Do NOT block #838 on it; rebase after #181 lands and CI should go green.

Verdict: code-correct, lockfile clean, root-cause fix. Ship.

Jiahui-Gu:

PR #843, opened 2026-05-02
chore(docs): keep only 2026-05-02-final-architecture.md

Jiahui-Gu:
## Summary
Per user instruction — purge everything under \`docs/\` except the frozen final-architecture spec.

- Kept: \`docs/superpowers/specs/2026-05-02-final-architecture.md\`
- Removed: 192 files (audits, archive, dogfood reports/screenshots, eval, reference, status, mvp-design, design-system, README, prior superpowers specs incl. v0.3-design / v0.3-daemon-split / v0.3-reconciliation)

The kept spec only references the deleted older specs as "superseded" history pointers — no live dependencies broken.

## Test plan
- [x] \`find docs -type f\` returns exactly the kept file
- [ ] CI green

PR #844, opened 2026-05-02
ci: drop docs/** and **.md from paths-ignore

Jiahui-Gu:
## Problem

`.github/workflows/ci.yml` had `paths-ignore` covering `**.md` and `docs/**`, so docs-only PRs do **not** trigger the `lint + typecheck + test (ubuntu/macos/windows)` jobs.

But branch protection on `working` lists those exact contexts as **required status checks** with `enforce_admins=true`. When a required context never reports, GitHub treats it as missing, not as "skipped success" — so a docs-only PR is permanently un-mergeable, and admins cannot override.

PR #843 (pure docs deletion) is currently stuck in this deadlock.

## Fix

Drop `'**.md'` and `'docs/**'` from `paths-ignore`. CI now runs on docs PRs too — turnaround is a few seconds with caches warm, and it eliminates the structural deadlock.

`'LICENSE'` and `'.gitignore'` are kept since changes there genuinely don't need full CI and are not common enough to deadlock real PRs.

## Diff

```yaml
     paths-ignore:
-      - '**.md'
-      - 'docs/**'
       - 'LICENSE'
       - '.gitignore'
```

Refs: PR #843

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,6 @@ on:
   pull_request:
     branches: [main, working]
     paths-ignore:
-      - '**.md'
-      - 'docs/**'
       - 'LICENSE'
       - '.gitignore'
   workflow_dispatch:


Jiahui-Gu:
Wrong direction — making docs PRs run the full CI matrix is wasteful. Per user, docs PRs should skip CI. Replacing with a paths-filter approach (dorny/paths-filter or git-diff-based) that keeps the workflow trigger but conditionally skips the test jobs — skipped required checks auto-report 'success', satisfying branch protection without burning CI minutes.

PR #845, opened 2026-05-02
ci: use paths-filter so docs-only PRs auto-pass required checks

Jiahui-Gu:
## Problem

The current \`ci.yml\` uses top-level \`paths-ignore\` to skip CI on docs-only PRs:

\`\`\`yaml
on:
  pull_request:
    paths-ignore:
      - '**.md'
      - 'docs/**'
      - 'LICENSE'
      - '.gitignore'
\`\`\`

This causes a **branch protection deadlock** for docs-only PRs (see #843):

- GitHub's \`paths-ignore\` prevents the workflow from being **triggered at all**.
- Required status checks listed in branch protection (\`lint + typecheck + test (ubuntu-latest)\` etc.) therefore **never appear** on the PR.
- Branch protection waits forever for checks that will never be reported → PR cannot be merged.

This is GitHub's documented behavior: a job skipped via \`if:\` reports status \`Success\` and satisfies required checks, but a workflow that's never triggered produces no checks at all.

## Fix

Standard pattern used by Kubernetes, Next.js, and the GitHub Docs repo:

1. Remove \`paths-ignore\` so the workflow always fires on PRs.
2. Add a tiny \`changes\` job using \`dorny/paths-filter@v3\` that emits a boolean \`code\` output (true when any non-docs file changed).
3. Gate the heavy \`lint-typecheck-test\` matrix with \`needs: changes\` + \`if: needs.changes.outputs.code == 'true'\`.

When the matrix is skipped via \`if:\`, GitHub auto-reports each required check as success — so docs-only PRs unblock cleanly.

## Critical preservation

- Matrix job \`name:\` left **byte-identical** (\`lint + typecheck + test (\${{ matrix.os }})\`) so branch protection's required-check string matching keeps working.
- Other workflows (\`pr-no-silent-drops\`, \`e2e\`, \`release\`, \`main-source-guard\`) untouched — they're either lightweight or scoped differently.

## Self-validation

This PR touches \`.github/workflows/ci.yml\` (not a doc), so the \`changes\` filter resolves \`code=true\` and the matrix runs. CI on this very PR is the live regression test for the change.

Refs: #843

PR #846, opened 2026-05-02
ci: paths-filter needs include glob to actually exclude docs

Jiahui-Gu:
## Bug

PR #845 added a \`changes\` job using \`dorny/paths-filter@v3\` to skip the OS test matrix on docs-only PRs. The \`code\` filter was written with only \`!\`-prefixed excludes:

\`\`\`yaml
filters: |
  code:
    - '!docs/**'
    - '!**.md'
    - '!LICENSE'
    - '!.gitignore'
\`\`\`

This does not work. Empirically verified on PR #843 (a pure docs deletion): the \`detect changes\` step logged \`Changes output set to [\"code\"]\` (i.e. \`code=true\`), and the matrix ran on every OS even though every changed file was a doc.

## Root cause

\`dorny/paths-filter\` evaluates patterns in order: a file matches the filter if some include pattern matches it AND no later \`!\` pattern excludes it. With an exclude-only filter, no file ever matches an include in the first place, so the filter as a whole reports \"changed\" for any changed file (the action falls back to truthy).

## Fix

Prepend \`'**'\` as the first entry so \`!\` lines have something to subtract from:

\`\`\`yaml
filters: |
  code:
    - '**'
    - '!docs/**'
    - '!**.md'
    - '!LICENSE'
    - '!.gitignore'
\`\`\`

Order matters in dorny: later patterns override earlier ones.

## Verification plan

- This PR touches \`.github/workflows/ci.yml\` (not a doc), so \`detect changes\` should still resolve \`code=true\` and the OS matrix must run + pass on this PR. That confirms the filter still recognizes code changes.
- Re-running CI on PR #843 after this lands should show \`Changes output set to []\` and the matrix jobs reporting Skipped. That confirms the docs-exclude actually fires.

## Scope

One line added. No other workflow, no other job, no matrix \`name:\` change.


Jiahui-Gu:
APPROVE.

**Layer 1**: Direction right. Single-line fix to make dorny/paths-filter exclude semantics actually work — `!`-prefixed patterns need a positive baseline to subtract from. Minimal, surgical.

**Layer 2 evidence**:
- `git diff origin/working --stat`: exactly 1 file changed (`.github/workflows/ci.yml`), +1/-0
- Diff is the single `+ - '**'` line prepended to the `code:` filter list
- Matrix `name:` line byte-identical: `name: lint + typecheck + test (${{ matrix.os }})`
- Filter block reads exactly as specified (code → '**', '!docs/**', '!**.md', '!LICENSE', '!.gitignore')
- CI: detect changes pass (code=true, correctly classifies workflow change as code), all 3 OS matrix jobs pass (macos 2m28s, ubuntu 3m45s, windows 3m55s), no-silent-drops pass — 5/5 green

Squash-merging.

PR #850, opened 2026-05-02
feat(tools): T10.11 audit-table-revalidate.sh

Jiahui-Gu:
## Summary

Delivers \`tools/audit-table-revalidate.sh\` — a bash script that re-runs the v0.3 daemon-split-design Chapter 15 zero-rework audit verdicts against any PR touching forever-stable files.

Refs: Task #126, design ch15 §1 (locked-decision audit table), ch15 §2 (derived-design audit table), ch15 §3 #11 (mechanical reviewer checklist family — script joins the forever-stable tooling per item 11/29 framing), ch14 §1.6 / §1.15 (REMOVED-item reintroduction warnings).

## What it does

- Parses the design spec at \`docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md\` directly. **The forever-stable file list is NOT hardcoded** — it is extracted from backtick-quoted file-path tokens in ch15 §1/§2/§3 every run, so spec edits flow through automatically (zero-drift contract per Layer 1).
- Compares the PR's touched-file set (from \`gh pr diff --name-only\`, \`git diff --name-only <range>\`, or an explicit \`--files\` list) against the forever-stable set.
- Matches by exact path, by basename (handles spec mentions like \`001_initial.sql\`), or by directory prefix (handles \`tools/spike-harness/\`).
- Also scans ch14 paragraphs marked REMOVED and flags any touched file that names a removed token (e.g. \`watchdog-darwin\`).
- Touches without an explicit override marker in the PR body fail with exit code 1. Override marker syntax (one per line):
  \`\`\`
  audit-override: <path> -- <reason: cite ch15 audit row + reviewer name>
  \`\`\`
  Em-dash (U+2014) and \`--\` separators both accepted; reason must be non-empty.

## Layer-1 constraints honored

- Pure bash 4+ / awk / grep / sed / gh — no npm deps.
- Spec is the single source of truth; the script will refuse to run if ch15 cannot be located OR if parsing yields zero forever-stable paths (failsafe against silent regression).
- No-op (exit 0 with notice) when the spec file is not present on the branch — the script lands ahead of the spec merge into \`working\`, so this avoids breaking other PRs in the interim.

## Behavior table

| Invocation | Touched | PR body | Exit |
| --- | --- | --- | --- |
| \`--files src/foo.ts\` | normal file | n/a | 0 (PASS) |
| \`--files packages/daemon/src/db/locked.ts\` | forever-stable | empty | 1 (BLOCK + verdict context printed) |
| same as above | forever-stable | contains override marker | 0 (OVERRIDDEN) |
| \`--pr 123\` | (mixed) | (PR body via \`gh pr view\`) | 0 or 1 per above |
| any | any | spec missing | 0 (no-op notice) |
| any | empty diff | n/a | 0 (no-op notice) |
| no args | n/a | n/a | 2 (usage) |

## Local test output

All six smoke tests run inline:
- normal-file-only diff -> PASS exit=0
- forever-stable touch no override -> BLOCK exit=1, verdict context printed (Chapter 07 §4 row for \`locked.ts\`)
- forever-stable touch with override -> OVERRIDDEN exit=0
- missing spec -> notice exit=0
- empty diff -> notice exit=0
- no args -> usage exit=2

Bare-basename matching also verified: \`packages/daemon/src/db/migrations/001_initial.sql\` correctly matches spec entry \`001_initial.sql\` and surfaces ch15 §3 forbidden-pattern item 4.

## Forever-stable per ch15

The script behavior is itself locked under ch15 §3's mechanical-reviewer-checklist family (alongside \`tools/lint-no-ipc.sh\`, \`tools/.no-ipc-allowlist\`, \`tools/packages-shape.spec.ts\`, \`tools/perf-budgets-locked.spec.ts\`). Future changes require an explicit ch15 audit row.

## Test plan

- [ ] Reviewer confirms the awk/grep heuristic captures all forever-stable paths in ch15 (run the script's parser in isolation against the spec; cross-check against the tabled paths)
- [ ] Reviewer confirms the override-marker grammar is acceptable (em-dash + \`--\` separators)
- [ ] Reviewer confirms the no-op-when-spec-missing behavior is the right call (alternative: hard-fail until spec lands)

---

### P0 (blocking, fixed): `tools/.no-ipc-allowlist` now BLOCKs

Reviewer-reported false-negative on a ch15 §3 #29 forever-stable path. Root cause was the `extract_paths` awk filter only accepting tokens with a known file extension or `topdir/sub/` directory shape — `tools/.no-ipc-allowlist` (extensionless dotfile) matched neither. Added a third clause:

```awk
# Allow extensionless dotfiles under a known top-level dir
if (tok ~ /^(packages|tools|docs|native|scripts|src|build|electron|tests)\/([^\/]+\/)*\.[A-Za-z0-9_][A-Za-z0-9._-]*$/) {
    print tok; next
}
```

Spec-paths count went from 33 -> 34 (the new entry is `tools/.no-ipc-allowlist` itself).

#### Reverse-verify (against pool-4 spec copy)

```
# parser reverted (git stash):
$ SPEC_PATH=...spec.md bash tools/audit-table-revalidate.sh --files tools/.no-ipc-allowlist
spec entries:   33 forever-stable paths
PASS: no forever-stable files touched.        # bug reproduced; exit 0

# parser fix re-applied (git stash pop):
$ SPEC_PATH=...spec.md bash tools/audit-table-revalidate.sh --files tools/.no-ipc-allowlist
spec entries:   34 forever-stable paths
--- forever-stable file touches ---
  [BLOCK] tools/.no-ipc-allowlist
           spec entry: tools/.no-ipc-allowlist
           verdict ctx: 29. v0.3 `tools/.no-ipc-allowlist` contents (descriptor preload only) are forever-stable; ...
FAIL: forever-stable files touched without override marker.
                                                # exit 1
```

### P1 (non-blocking, also fixed): ch14 REMOVED matcher now scans heading lines

Previously the REMOVED scanner only looked at backticked tokens within 3 lines of "REMOVED", missing `watchdog-darwin-approach` because that identifier appears as `##### 1.15 [watchdog-darwin-approach] — REMOVED from MUST-SPIKE register` (bracketed in a heading, no backticks nearby). Added a second pass that walks REMOVED-marked heading lines and extracts bracketed identifiers, filtering out markdown anchor links like `[Chapter 14]`.

Smoke:

```
$ ... bash tools/audit-table-revalidate.sh --files src/watchdog-darwin-approach.ts
--- ch14 REMOVED-item reintroduction warnings ---
  [WARN] src/watchdog-darwin-approach.ts touches REMOVED-item token `watchdog-darwin-approach`
FAIL: forever-stable files touched without override marker.   # exit 1
```

PR 850's own diff still PASSes (`--pr 850` -> 34 spec paths, no forever-stable touches).

### DIFF
--- a/tools/audit-table-revalidate.sh
+++ b/tools/audit-table-revalidate.sh
@@ -1,0 +1,457 @@
+#!/usr/bin/env bash
+# tools/audit-table-revalidate.sh
+#
+# Re-runs the v0.3 daemon-split design "Chapter 15 Zero-Rework Audit" verdicts
+# against a PR's touched-file set. Given a PR number (or arbitrary git diff
+# range), the script:
+#
+#   1. Locates the design spec on disk (forever-stable path:
+#      docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md, override
+#      with SPEC_PATH env var).
+#   2. Parses ch15 §1 (locked-decision audit table) and ch15 §2 (derived-design
+#      audit table) to extract the set of forever-stable file paths and their
+#      audit verdicts. The spec is the SINGLE SOURCE OF TRUTH; this script
+#      MUST NOT hardcode the file list (zero-drift contract).
+#   3. Parses ch15 §3 (forbidden-pattern checklist) for additional file paths
+#      called out as forever-stable per item 11 / 28 / 29 etc.
+#   4. Parses ch14 §1.6 / §1.15 REMOVED items so reintroductions are flagged.
+#   5. Compares the PR's touched-file set against the forever-stable set.
+#      Any touched forever-stable file fails the audit unless the PR body
+#      contains an explicit override marker line of the form
+#         audit-override: <path> — <reason>
+#      (one line per file; reason free-form, must be non-empty).
+#
+# Exit codes:
+#   0  no forever-stable files touched (or all touches explicitly overridden)
+#   1  forever-stable files touched without override (audit fails)
+#   2  usage / environment error (missing args, missing tools, unreadable spec)
+#
+# Requires: bash 4+, awk, grep, sed. `gh` only when --pr is used.
+#
+# This script itself is forever-stable per chapter 15 §3 (mechanical reviewer
+# checklist family). Do not change behavior without an audit row.
+
+set -euo pipefail
+
+PROG="$(basename "$0")"
+
+usage() {
+    cat <<EOF >&2
+Usage:
+  $PROG --pr <number>            Audit GitHub PR by number (uses gh CLI)
+  $PROG --diff <range>           Audit local git diff range (e.g. origin/working...HEAD)
+  $PROG --files <file>...        Audit explicit file list (read from args; "-" reads stdin)
+
+Options:
+  --spec <path>                  Override spec path (default: \$SPEC_PATH or
+                                 docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md)
+  --pr-body-file <path>          Path to a file containing the PR body (override marker source)
+                                 when using --diff or --files. Defaults to empty body.
+  --no-spec-ok                   Exit 0 with a notice if spec file does not exist
+                                 (default behavior; flag retained for explicitness).
+  -h, --help                     This message.
+
+Audit override marker (one per line, in PR body):
+  audit-override: <path> — <reason>
+  (em-dash or "--" both accepted; reason must be non-empty)
+
+Examples:
+  $PROG --pr 826
+  $PROG --diff origin/working...HEAD
+  git diff --name-only origin/working...HEAD | $PROG --files -
+EOF
+    exit 2
+}
+
+die() { echo "$PROG: error: $*" >&2; exit 2; }
+
+# -----------------------------------------------------------------------------
+# Argument parsing
+# -----------------------------------------------------------------------------
+
+MODE=""
+PR_NUM=""
+DIFF_RANGE=""
+FILE_ARGS=()
+SPEC_PATH_OPT=""
+PR_BODY_FILE=""
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --pr)         MODE="pr";   PR_NUM="${2:-}"; shift 2 || die "--pr needs a number" ;;
+        --diff)       MODE="diff"; DIFF_RANGE="${2:-}"; shift 2 || die "--diff needs a range" ;;
+        --files)      MODE="files"; shift; while [[ $# -gt 0 && "$1" != --* ]]; do FILE_ARGS+=("$1"); shift; done ;;
+        --spec)       SPEC_PATH_OPT="${2:-}"; shift 2 || die "--spec needs a path" ;;
+        --pr-body-file) PR_BODY_FILE="${2:-}"; shift 2 || die "--pr-body-file needs a path" ;;
+        --no-spec-ok) shift ;;
+        -h|--help)    usage ;;
+        *)            die "unknown arg: $1 (try --help)" ;;
+    esac
+done
+
+[[ -z "$MODE" ]] && usage
+
+# -----------------------------------------------------------------------------
+# Locate spec
+# -----------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
## Reviewer verdict: REQUEST_CHANGES (posted as comment — GitHub blocks self-review)

### Direction: right (Layer 1 OK)

Script does what ch15 §3 #11 needs — pure bash, spec is single source of truth, spec-absent no-op, three input modes, override grammar with em-dash + `--` both functional. Verified locally against pool-4's spec copy: parser extracts **33 forever-stable paths**, all 6 smoke cases produce correct exit codes, `packages/daemon/src/db/locked.ts` and `packages/electron/src/main/transport-bridge.ts` both correctly BLOCK, em-dash overrides correctly OVERRIDE, `--pr 850` mode works against this PR itself.

### Layer-2 finding (blocking, since this script is itself ch15 §3 forever-stable)

#### 1. Parser false-negative on `tools/.no-ipc-allowlist` (explicit ch15 §3 #29)

Spec line 152 (ch15 §3):
> 29. v0.3 `tools/.no-ipc-allowlist` contents (descriptor preload only) are forever-stable; v0.4 additions require R4 sign-off and a chapter-15 audit row.

Local repro:
```
SPEC_PATH=/tmp/spec.md bash tools/audit-table-revalidate.sh --files tools/.no-ipc-allowlist
=> PASS: no forever-stable files touched.   (expected: BLOCK)
```

Root cause: extract_paths awk filter (lines 240/250) accepts a token only if it (a) ends with one of `.ts .tsx .js .mjs .cjs .sh .json .sql .md .yml .yaml .bin .ndjson .ps1 .wxs`, or (b) is a `topdir/sub/` directory. `tools/.no-ipc-allowlist` matches neither. This is exactly the case the script's own comment ("false negatives would silently skip an audit row, so we prefer to flag") says it tries to avoid, and it's the one allowlist file ch15 explicitly names.

**Suggested fix**: add a third clause that accepts `tools/<dotfile>` (no extension), or whitelist the exact spec mentions of extensionless config files. Only one such path exists in ch15 today.

#### 2. ch14 REMOVED matcher is effectively dormant (non-blocking, FYI)

`awk '/REMOVED/ {print; for i in 0..2 getline...}' | grep backticks` against ch14 §1.6/§1.15 captures only `` `KeepAlive=Crashed` ``. The actual REMOVED-item identifier `watchdog-darwin` lives in the §1.15 heading text (`##### 1.15 [watchdog-darwin-approach] — REMOVED from MUST-SPIKE register`), not in a backtick on a following line. So the reintroduction warning never meaningfully fires today. Not blocking — at worst a missed warning, not a false BLOCK. Consider scanning REMOVED heading lines themselves (split on `[...]`) in a follow-up, or trim the description.

### Other items checked, all OK
- 3 input modes verified in isolation (--pr/--diff/--files)
- PR mode null/empty body handling (`gh pr view ... || true`) is safe
- Override regex tolerates surrounding whitespace and multiple lines
- Override for path NOT in lock list is silently ignored — correct (don't reject spurious paths)
- Empty diff → exit 0 with notice
- Spec missing → exit 0 with notice (critical for landing ahead of spec PR #847)
- No new deps; pure bash/awk/grep/sed/gh
- 33 forever-stable paths extracted including all sampled lock-ins

### Action

Please patch item 1 (the `.no-ipc-allowlist` miss) and re-push. Item 2 is a follow-up — call it out in the PR description if you don't address it now.

Jiahui-Gu:
Fixed at cd1be37.

- **P0**: `extract_paths` now accepts extensionless dotfiles under known top-level dirs. `tools/.no-ipc-allowlist` correctly BLOCKs (spec paths 33 -> 34). Reverse-verified locally (revert -> PASS, reapply -> BLOCK exit 1).
- **P1**: ch14 REMOVED matcher extended to also scan heading lines and extract bracketed identifiers (skipping `[Chapter NN]` anchor links). `watchdog-darwin-approach` now triggers reintroduction WARN+BLOCK.

PR body updated with smoke-case outputs. PR 850's own diff still PASSes against the new parser.

PR #854, opened 2026-05-02
feat(tools): T10.6 packages-shape.spec.ts (forever-stable per ch15 §3 #10)

Jiahui-Gu:
## Summary

Adds `tools/packages-shape.spec.ts` (vitest) asserting the v0.3 packages set (`@ccsm/daemon`, `@ccsm/electron`, `@ccsm/proto`) is a **subset** of what `pnpm-workspace.yaml` + `packages/*/package.json` actually define. v0.4 may ADD packages (web frontend, etc) — rename or removal of any v0.3 entry trips the test.

Refs Task #129. Implements design spec ch15 §3 #10 ("human review with reference test as smoke check"). This file is **forever-stable**; the v0.3 list only grows.

## Layer 1 (no-wheel-reinvention)

- No new npm deps. `node:fs` + `node:path` only.
- `pnpm-workspace.yaml` parsed by trivial regex (one `packages:` list line) rather than importing `js-yaml` — the YAML we ship is exactly `packages: ["packages/*"]`, one line.
- `vitest` is already a root devDependency.

## Why a separate `tools/vitest.config.ts`

Root `vitest.config.ts` restricts `include` to `tests/**` and `electron/**/__tests__/**` and pulls in `jsdom` + a heavy `tests/setup.ts`. A path argument to `vitest run` is treated as a **filter** over `include`, not an override, so `npx vitest run tools/packages-shape.spec.ts` against the root config exits with `No test files found`.

The minimal `tools/vitest.config.ts` (16 lines, node environment, no setup, `include: ['tools/**/*.spec.ts']`) keeps the tools-level shape checks pure-node and out of the renderer suite. Run with:

```
npx vitest run --config tools/vitest.config.ts
```

## Local verification (Windows, pool-13)

**Forward (current state — passes):**

```
$ npx vitest run --config tools/vitest.config.ts
 RUN  v4.1.5 C:/Users/jiahuigu/ccsm-worktrees/pool-13

 Test Files  1 passed (1)
      Tests  2 passed (2)
   Duration  577ms
```

**Reverse-verify (planted failure — `mv packages/proto/package.json packages/proto/package.json.bak`):**

```
- Expected
+ Received

- []
+ [
+   "@ccsm/proto",
+ ]

 ❯ tools/packages-shape.spec.ts:139:7

 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)
```

Restored, re-ran, back to 2 passed.

## Test plan

- [x] `npx vitest run --config tools/vitest.config.ts` passes against current `working`.
- [x] Removing any of `packages/{daemon,electron,proto}/package.json` causes the test to FAIL.
- [x] Restoring it returns the test to PASS.
- [ ] Reviewer to confirm wiring this into CI is a follow-up task (not in scope here — this PR delivers the spec only).

### DIFF
--- a/tools/packages-shape.spec.ts
+++ b/tools/packages-shape.spec.ts
@@ -1,0 +1,141 @@
+/**
+ * tools/packages-shape.spec.ts
+ *
+ * FOREVER-STABLE per design spec ch15 §3 #10:
+ *   "Reshuffling `packages/` directories; only additions allowed.
+ *    Mechanism: human review with reference test as smoke check —
+ *    `tools/packages-shape.spec.ts` enumerates `packages/*\/package.json`
+ *    and asserts the v0.3 set (`proto`, `daemon`, `electron`) is a subset
+ *    of the actual set; rename or removal trips the test."
+ *
+ * Locked v0.3 packages set — DO NOT REMOVE entries from this list. v0.4+
+ * MAY ADD new packages (e.g. `@ccsm/web`, `@ccsm/cli`) to the actual
+ * `packages/` tree; the assertion is "subset of actual", not "equals
+ * actual", so additions never break this test.
+ *
+ * Layer 1 notes:
+ *   - No new npm deps. Uses only `node:fs` + `node:path`.
+ *   - `pnpm-workspace.yaml` parsed by trivial regex (one `packages: [...]`
+ *     line) instead of importing a YAML library — see CLAUDE.md / dev.md
+ *     §1 no-wheel-reinvention.
+ *   - Test runs via `npx vitest run tools/packages-shape.spec.ts`. CLI path
+ *     argument overrides root vitest config's `include`, so no extra
+ *     vitest config is needed in `tools/`.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import { join, resolve } from 'node:path';
+
+// Repo root = parent of this file's directory (`tools/`).
+const REPO_ROOT = resolve(__dirname, '..');
+
+// FROZEN v0.3 packages set. Adding entries here is a v0.3 contract change
+// and requires R4 sign-off. Removing or renaming entries is forbidden.
+const V03_REQUIRED_PACKAGES = ['@ccsm/daemon', '@ccsm/electron', '@ccsm/proto'] as const;
+
+/**
+ * Parse the trivial `packages:` list out of pnpm-workspace.yaml.
+ *
+ * Supported shapes (matches what pnpm itself accepts and what we ship):
+ *   packages:
+ *     - "packages/*"
+ *     - 'packages/foo'
+ *     - packages/bar
+ *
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Reviewer verdict: REQUEST_CHANGES (CI red)

(Posting as comment because gh blocks self-PR formal review.)

## Layer 1: Direction right

- Subset assertion correctly encodes ch15 §3 #10's "additions allowed, removals forbidden" semantics. v0.4 web/cli packages will not break this.
- Regex parser for `pnpm-workspace.yaml` is a justified avoidance of `js-yaml`. Verified it handles the actual file format (`packages:\n  - \"packages/*\"`): block-list branch matches, quote stripping yields `packages/*`. Inline-flow branch is cheap insurance.
- Separate `tools/vitest.config.ts` justified — root config restricts `include` to `tests/**` + `electron/**/__tests__/**` and pulls jsdom + heavy setup. Path arg to `vitest run` is a filter, not include override, so reuse would silently skip these specs or pay jsdom cost.
- No new deps. Reverse-verify in PR body (remove `packages/proto/package.json` -> FAIL diff shows `+ \"@ccsm/proto\"`; restore -> 2 passed). Both outputs visible.

## Layer 2 must-fix (R0): CI red on all 3 platforms

```
tools/packages-shape.spec.ts
  31:27  error  '__dirname' is not defined  no-undef
```

`eslint.config.js` does not declare `__dirname` in `globals` (only browser globals + a small node subset: `process`, `Buffer`, `setTimeout`, ...). No `tools/` override either.

Blocks `working` branch protection (required `test` check). Pick the lightest fix:

1. **Replace `__dirname` with ESM-safe equivalent** (preferred — keeps spec free of config coupling):
   ```ts
   import { fileURLToPath } from 'node:url';
   import { dirname, resolve } from 'node:path';
   const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..');
   ```
2. Add `tools/` override in `eslint.config.js` with `globals: { __dirname: 'readonly', ... }`.
3. Add `__dirname: 'readonly'` to the existing global `globals` block.

(1) is most localized and matches the file's "no infra coupling" theme.

## Layer 2 nit (R3)

Test-plan item "wiring into CI is a follow-up task" — agreed, out of scope here. The follow-up task should be filed before this lands so the reference-test does not silently rot.

After fix + green CI, happy to re-review and merge.

Jiahui-Gu:
Fixed at d48fedd — replaced `__dirname` with `dirname(fileURLToPath(import.meta.url))` (option 1, ESM-safe, no `eslint.config.js` coupling).

Verified locally:
- `npx vitest run --config tools/vitest.config.ts` -> 2 passed
- `npm run lint` -> 0 errors

CI green on all 3 OS:
- macos-latest: pass (2m11s)
- ubuntu-latest: pass (3m59s)
- windows-latest: pass (3m59s)

Re: R3 nit — agreed, will file the wire-into-CI follow-up task before this lands so the reference test does not silently rot.

PR #864, opened 2026-05-02
feat(daemon): T5.1 better-sqlite3 wrapper + PRAGMA boot-time config

Jiahui-Gu:
## Scope (Task #54 / T5.1)

Add a thin `better-sqlite3` wrapper for the daemon that applies the canonical
boot-time PRAGMAs from spec ch07 §3 the moment the connection opens.

- `packages/daemon/src/db/sqlite.ts` — `openDatabase(path, options?)` wrapper. Applies in order: `journal_mode = WAL`, `synchronous = NORMAL`, `foreign_keys = ON`, `busy_timeout = 5000`, `journal_size_limit = 67108864` (64 MiB).
- Exports typed `SqliteDatabase` (re-export of `better-sqlite3` `Database`) so the rest of the daemon imports a single typed surface.
- Frozen `BOOT_PRAGMAS` constant exported alongside so callers (and the spec test) can assert against the same values the wrapper applies.
- Closes the connection on PRAGMA failure to avoid leaking a half-configured handle.

## Spec refs

- ch07 §1 — storage choice: SQLite via `better-sqlite3`, sync driver, WAL mode
- ch07 §3 — full PRAGMA list applied at boot **after** opening, **before** migrations

## Out of scope (explicitly NOT touched here)

| Item | Owner |
|---|---|
| `001_initial.sql` schema | Task #55 / T5.2 |
| Migration runner | Task #56 / T5.4 |
| WAL discipline + shutdown checkpoint | Task #58 / T5.6 |

## Test

`packages/daemon/src/db/__tests__/sqlite.spec.ts` — opens a temp DB and verifies each PRAGMA value via `db.pragma(name, { simple: true })` reads. 7 cases (open file + 5 PRAGMAs + frozen-constant shape lock). All pass locally.

## Dep choice note

Spec text named "latest 11.x" but root `package.json` already pins `better-sqlite3 ^12.9.0` at the workspace root. Adding `^11.7.0` to `@ccsm/daemon` caused pnpm to install a duplicate native module which then failed `electron-rebuild` on Windows postinstall. Aligning the daemon dep to `^12.9.0` lets pnpm dedupe to a single native install. 12.x is API-compatible with 11.x for the surface used here (`new Database`, `db.pragma`, `db.close`). If a downgrade is preferred, happy to revisit.

## Local quality gates (Windows, Node 24)

- `pnpm --filter @ccsm/daemon typecheck` — pass
- `pnpm --filter @ccsm/daemon lint`      — pass
- `pnpm --filter @ccsm/daemon test`      — 17/17 pass (incl. existing migration-lock spec, still skipped per design)

## Windows native-rebuild note

On this host `pnpm install` completes but the better-sqlite3 binding it leaves under `node_modules/.pnpm/.../build/Release/` was compiled for the Electron ABI (NODE_MODULE_VERSION 145). To run the daemon vitest suite against plain Node 24 (NODE_MODULE_VERSION 137), one extra rebuild was needed:

```
cd node_modules/.pnpm/better-sqlite3@12.9.0/node_modules/better-sqlite3
npx node-gyp rebuild --release
```

CI runs Node 20 with matching prebuilds and should not hit this. Documented here per task instructions in case CI does hit it.

### DIFF
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -3,5 +3,16 @@
   "version": "0.0.0",
   "private": true,
   "type": "module",
-  "description": "CCSM daemon (headless RPC server). Skeleton only; implementation lands in subsequent T0.x / T1.x tasks per design spec ch11."
+  "description": "CCSM daemon (headless RPC server). Skeleton only; implementation lands in subsequent T0.x / T1.x tasks per design spec ch11.",
+  "scripts": {
+    "lint": "eslint src --ext .ts",
+    "typecheck": "tsc --noEmit",
+    "test": "vitest run"
+  },
+  "dependencies": {
+    "better-sqlite3": "^12.9.0"
+  },
+  "devDependencies": {
+    "@types/better-sqlite3": "^7.6.12"
+  }
 }
--- a/packages/daemon/src/db/__tests__/sqlite.spec.ts
+++ b/packages/daemon/src/db/__tests__/sqlite.spec.ts
@@ -1,0 +1,87 @@
+// packages/daemon/src/db/__tests__/sqlite.spec.ts
+//
+// Verifies that `openDatabase()` applies every PRAGMA spec'd in
+// docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md ch07 §3
+// at boot time, by reading each value back via `db.pragma()` after the
+// wrapper returns.
+
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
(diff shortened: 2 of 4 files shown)


Jiahui-Gu:
Reviewer LGTM (self-author, GitHub blocks self-approve).

Layer 1 (direction): right. Wrapper-only scope respects T5.2/T5.4/T5.6 boundaries; better-sqlite3 12.x deviation justified in PR body (root pin dedup, API-compat).

Layer 2 (impl): BOOT_PRAGMAS frozen with all 5 ch07 §3 values (WAL, NORMAL, ON, 5000, 64 MiB). Tests cover each PRAGMA via simple-read + frozen-constant shape lock. PRAGMA-failure path closes handle. 3-OS CI green.

PR #883, opened 2026-05-03
test(daemon): T10.4 v1 JSON Schema for listener-a descriptor + ajv spec

Jiahui-Gu:
## Summary

Task #125 / T10.4. Adds the forever-stable v1 JSON Schema for the listener-A connection descriptor written by the daemon at boot (ch03 §3.2), plus an ajv-driven spec that validates the bytes produced by the T1.6 writer (PR #863) against the schema.

- `packages/daemon/schemas/listener-a.schema.json` — Draft-07 schema, `additionalProperties:false`, closed enum for `transport`, `const: 1` on both `version` and `protocol_version`, regex for `boot_id` (UUIDv4) and `tlsCertFingerprintSha256` (lowercase hex SHA-256). No `$id` pinning beyond the v1 contract per the task brief.
- `packages/daemon/test/descriptor/schema.spec.ts` — 8 cases:
  - writer-produced descriptor (TLS) validates;
  - every closed-enum transport validates (UDS / named pipe / h2c / h2-tls);
  - rejects unknown transport, missing `boot_id`, extra property, `version !== 1`, non-UUIDv4 `boot_id`;
  - **byte-equal round-trip**: real `writeDescriptor()` output must equal the checked-in golden (`test/descriptor/listener-a.golden.json`). Any future drift in writer formatting (compact JSON, key reorder, dropped trailing newline) fails CI rather than merging silently.

Schema is FROZEN per ch15 §3 forbidden-pattern 8: v0.4+ additions land in a NEW descriptor file, never as enum widenings or shape changes here.

Wires `test/**/*.spec.ts` into vitest `include` and `tsconfig.test.json` so the new spec actually runs and type-checks (the existing `test/db/migration-lock.spec.ts` was previously unreachable from the vitest entry point — same fix benefits it).

Adds `ajv@^8.20.0` as a daemon devDependency (de-facto standard Draft-07 validator; not promoted to runtime — schema is for tooling/contract verification only).

## Test plan

- [x] `pnpm --filter @ccsm/daemon typecheck` — green (both `tsconfig.json` and `tsconfig.test.json`).
- [x] `npx vitest run test/descriptor/schema.spec.ts` — 8/8 pass.
- [x] Reverse-verify byte-equal: mutate one byte in the golden (`4242` → `4243`), the round-trip case fails with `AssertionError: expected '...' to be '...'`; restoring brings it back to 8/8.
- [x] Full daemon vitest: 57 pre-existing pass + 8 new pass = 65 pass; 3 skipped; 6 pre-existing failures are `better-sqlite3` native binding load errors on this Windows worker (electron-rebuild postinstall fails on node-pty in this sandbox env), unrelated to this change. Diff vs. baseline before this PR: same 6 sqlite failures, same 3 skips, +8 passing tests from `schema.spec.ts`.
- [ ] CI on Linux validates the sqlite tests.

## Notes

- Pool 13 carried staged work from `dev/t3.1-principal-model` (different task) when this branch was created; that work was unstaged and removed from the working tree before any commit landed here. Final diff is the schema + spec + vitest/tsconfig include + ajv devDep + lockfile only.

### DIFF
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -18,6 +18,7 @@
     "better-sqlite3": "^12.9.0"
   },
   "devDependencies": {
-    "@types/better-sqlite3": "^7.6.12"
+    "@types/better-sqlite3": "^7.6.12",
+    "ajv": "^8.20.0"
   }
 }
--- a/packages/daemon/schemas/listener-a.schema.json
+++ b/packages/daemon/schemas/listener-a.schema.json
@@ -1,0 +1,76 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "ListenerAConnectionDescriptorV1",
+  "description": "Forever-stable v1 JSON Schema for the listener-A connection descriptor written by the daemon at boot. Spec ch03 §3.2. v0.4+ additions land in a NEW descriptor file, never as enum widenings or shape changes here (ch15 §3 forbidden-pattern 8).",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "version",
+    "transport",
+    "address",
+    "tlsCertFingerprintSha256",
+    "supervisorAddress",
+    "boot_id",
+    "daemon_pid",
+    "listener_addr",
+    "protocol_version",
+    "bind_unix_ms"
+  ],
+  "properties": {
+    "version": {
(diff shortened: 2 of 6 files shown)


Jiahui-Gu:

PR #887, opened 2026-05-03
spike(T9.11): better-sqlite3 arm64 prebuilds for Node 22 (ch14 §1.12)

Jiahui-Gu:
## Summary

Lands the ch14 §1.12 spike under `tools/spike-harness/probes/better-sqlite3-22-arm64/` to retire the arm64-prebuild risk before phase-3 SQLite work runs on the arm64 matrix.

- `probe.mjs` (node: stdlib only — no deps): hits the WiseLibs/better-sqlite3 GitHub release manifest and asserts a prebuilt asset exists for each `<arm64-target>` at the requested ABI. Optional live `require('better-sqlite3') + new Database(':memory:')` leg for arm64 hosts. Forever-stable contract documented in the file header per ch14 §1.B.
- `PROBE-RESULT.md`: documents existing prebuild availability and quantifies the source-build fallback cost.

## Findings

GREEN. Both the pinned `12.9.0` and the prior `11.10.0` ship `node-v127` (Node 22 ABI) prebuilds for `darwin-arm64`, `linux-arm64` (glibc), `linuxmusl-arm64`, and `win32-arm64`. `prebuild-install` resolves them automatically during `pnpm install` — no compiler invoked, no extra `npm_config_*` env vars needed.

Source-build fallback (only if the GitHub fetch is blocked): 60-120 s on every v0.3-supported host (`ubuntu-22.04-arm`, `macos-14`, `node:22-bookworm`). Stock GHA images already have `python3` + `build-essential` / Xcode CLT.

Live probe output captured against `v12.9.0` and `v11.10.0` is in PROBE-RESULT.md. Run from `win32/x64` with `--no-load`; the manifest evidence is the load-bearing piece (the `prebuild-install` resolver builds the same filename and downloads it from the same release).

## Out of scope (recorded as follow-ups in PROBE-RESULT.md)

- `.github/workflows/{ci,e2e,release}.yml` still pin `node-version: '20'` — must move to `22` before phase-3 can consume this evidence (separate task per the toolchain-lock doc).
- Adding `macos-14` + `ubuntu-22.04-arm` to the phase-3 SQLite job matrix.
- Pinning the prebuild URL+sha in the v0.3 lockfile (per T9.8 SEA spike point 5).

## Test plan

- [x] `node tools/spike-harness/probes/better-sqlite3-22-arm64/probe.mjs --no-load` exits 0; all four arm64 targets show `present: true` for v12.9.0.
- [x] Same probe with `--version=11.10.0 --no-load` exits 0; same result on v11.10.0.
- [x] `node --check probe.mjs` passes.
- [ ] Phase-3 SQLite job re-runs `node probe.mjs` (without `--no-load`) inside the `macos-14` and `ubuntu-22.04-arm` jobs as a smoke gate before exercising real workload.

### DIFF
--- a/tools/spike-harness/README.md
+++ b/tools/spike-harness/README.md
@@ -43,3 +43,4 @@ Cross-link: chapter 11 §2 / §6 pins this path; chapter 12 §3 reuses
 | `probes/uds-h2c/run.sh`          | §1.4 (T9.4)       | smoke (60s) + 1h soak driver        |
 | `probes/win-h2-named-pipe/{server,client}.mjs` | §1.5 (T9.5) | runnable on win32; non-win32 skip |
 | `probes/win-h2-named-pipe/run.sh` | §1.5 (T9.5)      | smoke (60s) + 1h soak driver        |
+| `probes/better-sqlite3-22-arm64/probe.mjs` | §1.12 (T9.11) | runnable (GitHub manifest check + opt. live load) |
--- a/tools/spike-harness/probes/better-sqlite3-22-arm64/.gitignore
+++ b/tools/spike-harness/probes/better-sqlite3-22-arm64/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+package-lock.json
+pnpm-lock.yaml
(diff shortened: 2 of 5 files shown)


Jiahui-Gu:

PR #893, opened 2026-05-03
spike(t9.2): macOS UDS cross-user reachability probe (no FDA)

Jiahui-Gu:
## Summary

Task #116 [T9.2]. Lands `tools/spike-harness/probes/macos-uds-cross-user/`
covering spec ch14 §1.2 phase 0.5 — must resolve before listener-A
wiring on macOS.

The probe answers: **which UDS bind paths on darwin let a different
local user `connect()` WITHOUT a Full Disk Access grant on the
connecting binary, and which require FDA?**

## Files

- `server.mjs` — UDS line-echo server, `chmod 0666` after bind so any
  cross-user denial isolates TCC/FDA from POSIX. SIGTERM-safe socket
  unlink. `node:` stdlib only.
- `client.mjs` — single-shot `connect()` probe, classifies the outcome
  (`OK`/`EACCES`/`EPERM`/`ENOENT`/`ECONNREFUSED`/`ETIMEDOUT`/`OTHER`)
  and emits one JSON line. Designed to be invoked under `sudo -n -u`
  for the cross-user leg.
- `run.sh` — iterates a 6-path matrix (`/tmp`, `/private/tmp`,
  `/Users/Shared`, `~/Library/Caches`, `~/Library/Application Support`,
  `~/Documents`), starts the server as the invoking user, runs the
  client both same-user and (when `SPIKE_SECONDARY_USER` is set)
  cross-user, then writes `*-matrix.ndjson` + `*-summary.json` and
  assigns `FDA-FREE` / `FDA-REQUIRED` / `UNREACHABLE` / `SKIPPED`.
- `PROBE-RESULT.md` — spec-derived verdict matrix + listener-A
  recommendation.

## Recommendation feeding listener-A on darwin

Bind under `~/Library/Caches/ccsm/daemon.sock` with `0600` on the
socket and `0700` on the parent. Mandatory application-layer
peer-cred check via `getsockopt(LOCAL_PEERCRED)` per T9.1's contract
(`tools/spike-harness/connect-and-peercred.sh`). TCC + POSIX are
defence-in-depth; peer-cred is the primary boundary. Do NOT bind
under `/tmp`, `/private/tmp`, or `/Users/Shared` for listener-A —
even with `0600` they sit outside any TCC fence and the stale-socket
file is unlink-squat-able cross-user during daemon restart.

Full reasoning + per-path table in `PROBE-RESULT.md`.

## Local verification (win32)

Authored from `MINGW64_NT-10.0-26200`. The harness short-circuits on
non-darwin per spec ch14 §1.B (Forever-stable contract):

```
$ bash tools/spike-harness/probes/macos-uds-cross-user/run.sh
macos-uds-cross-user: skipped on MINGW64_NT-10.0-26200 (UDS path semantics differ)
exit=2

$ node tools/spike-harness/probes/macos-uds-cross-user/server.mjs --socket=/tmp/x.sock
macos-uds-cross-user server: win32 unsupported
exit=2

$ node tools/spike-harness/probes/macos-uds-cross-user/client.mjs --socket=/tmp/x.sock
macos-uds-cross-user client: win32 unsupported
exit=2
```

`node --check` passes for both `.mjs` files; `bash -n run.sh` passes.
The harness directory is in eslint `ignores` per
`tools/spike-harness/README.md` Layer-1 constraint, so syntax-check is
the gate. Live darwin matrix capture is deferred to a self-hosted
runner with a provisioned secondary user (T0.10 / #16) — flagged as
"TODO populate from real darwin run" in `PROBE-RESULT.md` and called
out as the only outstanding follow-up before listener-A wiring on
darwin can proceed.

## Test plan

- [ ] Self-hosted darwin runner provisions `ccsmprobe` secondary user
      with `NOPASSWD` sudoers scoped to `node`, then runs:
      `SPIKE_SECONDARY_USER=ccsmprobe bash tools/spike-harness/probes/macos-uds-cross-user/run.sh`
- [ ] Confirm `summary.json` `byPath` matches the spec-derived verdict
      matrix in `PROBE-RESULT.md`. Any divergence escalates before
      listener-A wiring proceeds.
- [ ] (optional) Re-run with FDA granted to `node` on the secondary
      user via `tccutil` to confirm the `FDA-REQUIRED → OK` flip.

### DIFF
--- a/tools/spike-harness/README.md
+++ b/tools/spike-harness/README.md
@@ -43,3 +43,5 @@ Cross-link: chapter 11 §2 / §6 pins this path; chapter 12 §3 reuses
 | `probes/uds-h2c/run.sh`          | §1.4 (T9.4)       | smoke (60s) + 1h soak driver        |
 | `probes/win-h2-named-pipe/{server,client}.mjs` | §1.5 (T9.5) | runnable on win32; non-win32 skip |
 | `probes/win-h2-named-pipe/run.sh` | §1.5 (T9.5)      | smoke (60s) + 1h soak driver        |
+| `probes/macos-uds-cross-user/{server,client}.mjs` | §1.2 (T9.2) | runnable on darwin; linux/win32 skip |
+| `probes/macos-uds-cross-user/run.sh` | §1.2 (T9.2)   | bind-path × cross-user matrix driver |
--- a/tools/spike-harness/probes/macos-uds-cross-user/PROBE-RESULT.md
+++ b/tools/spike-harness/probes/macos-uds-cross-user/PROBE-RESULT.md
@@ -1,0 +1,155 @@
+# T9.2 spike — macOS UDS cross-user reachability without Full Disk Access
+
+**Status:** smoke harness landed; live darwin matrix capture deferred to a
+self-hosted macOS runner with a pre-provisioned secondary local user
+(see "Follow-ups" below). Authored on Windows (`MINGW64_NT-10.0-26200`);
+on non-darwin hosts the harness short-circuits (`run.sh` exits 2, both
+`server.mjs` and `client.mjs` exit 2 on `win32`). The decision matrix in
+this document is the **spec-derived** verdict that the live runner is
+expected to confirm; any divergence the runner observes overrides this
+file and must trigger a follow-up before listener-A wiring proceeds on
+darwin.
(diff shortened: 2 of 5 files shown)


Jiahui-Gu:

PR #904, opened 2026-05-03
feat(monorepo): T0.2 Turborepo task graph + cache

Jiahui-Gu:
## Summary

Task #10 — wire pnpm workspaces with Turborepo per spec ch11 §1/§4.

- New `turbo.json` at repo root with `gen`, `build`, `test`, `lint` (plus `typecheck`) tasks.
- `build.dependsOn = ["^build", "@ccsm/proto#gen"]` and `build.outputs = ["dist/**", "gen/**"]` — generated proto code is restored from cache alongside `dist/`. This is the critical requirement (a fresh clone or a wiped `packages/proto/gen` directory re-materialises from the cache replay; no manual `pnpm --filter @ccsm/proto run gen` needed).
- `test.dependsOn = ["^build", "@ccsm/proto#gen"]` and `lint` has no deps, per task brief.
- Root scripts `pnpm gen`/`build`/`test`/`lint` → `turbo run …`. Legacy electron-monolith build/test/lint preserved under `build:app` / `test:app` / `lint:app` aliases (still referenced by `make:*` targets and by the existing CI workflows). CI workflows (`ci.yml`, `e2e.yml`, `release.yml`) updated to call the `:app` aliases until the electron sources migrate into `packages/electron` in a follow-up task — that keeps the legacy build green while the monorepo build comes online.
- `turbo` added as root devDependency; `pnpm-lock.yaml` regenerated. `.turbo/` added to `.gitignore`.

## Spec alignment

> Codegen is invoked via `pnpm --filter @ccsm/proto run gen` which Turborepo treats as a prerequisite of `build` for all consumers (declared in `turbo.json`)
> Declaring `@ccsm/proto#gen` as a `build` dep (in addition to `^build`) ensures local-dev bootstrap works without manual `pnpm --filter @ccsm/proto run gen`. `gen` is included in `outputs` of `build` so Turborepo cache-restores generated code along with `dist/`.

(Ch11 §4. The task brief explicitly listed `^build` AND `@ccsm/proto#gen` for `test` as well — followed verbatim.)

## Cache HIT proof

First run (forced):

```
$ pnpm exec turbo run build --force
@ccsm/proto:gen: cache miss, executing ...
@ccsm/daemon:build: cache miss, executing ...
 Tasks:    2 successful, 2 total
Cached:    0 cached, 2 total
  Time:    5.347s
```

Second run:

```
$ pnpm build
@ccsm/proto:gen: cache hit, replaying logs ac64302afb1656a7
[...]

PR #906, opened 2026-05-03
ci(T0.8): install + proto-gen-and-lint + test matrix

Jiahui-Gu:
## Summary (Task #17)

Aligns `.github/workflows/ci.yml` with the T0.2 monorepo (pnpm + turbo + workspaces) per spec ch11 §6:

1. **Install** uses `pnpm install --frozen-lockfile` with `cache: 'pnpm'` (was `npm ci --legacy-peer-deps`).
2. **Proto-gen + buf lint precheck** new job `proto-gen-lint` runs before the test matrix, fast-fails on stale codegen or lint violations on a single ubuntu runner.
3. **Matrix** `ubuntu-latest / macos-latest / windows-latest` x `{lint, typecheck, test}` for workspace packages via `turbo run` (`pnpm run lint` / `pnpm exec turbo run typecheck` / `pnpm run test`). Legacy root-app `lint:app / typecheck / build:app / test:app` kept until T0.2 absorbs the root `ccsm` package as a workspace member.
4. **Turbo cache** keyed on `pnpm-lock.yaml + turbo.json + packages/proto/lock.json + packages/proto/src/**/*.proto` so any `.proto` edit busts the cache. Per-task `inputs` in `turbo.json` (already wired by T0.2) handle fine-grained invalidation; the CI key only scopes per OS.

## Constraints honored

- Only `.github/workflows/ci.yml` modified. No touches to root `package.json`, `pnpm-lock.yaml`, `turbo.json`, or `packages/*/package.json` (hotfiles owned by other in-flight tasks).
- Required check job names preserved: `lint + typecheck + test (ubuntu-latest|macos-latest|windows-latest)`.
- `e2e` and `no-silent-drops` workflows untouched.

## Layer 1 finding

T0.2 (PR #904) introduced pnpm as `packageManager`, set up `pnpm-workspace.yaml`, added `turbo.json`, and exposed root scripts `gen / build / test / lint` as `turbo run *`. **But it did not update CI** — `ci.yml` still ran `npm ci` and root npm scripts directly. T0.8 is therefore not duplicate work; it bridges the gap.

## Local verification

- `actionlint v1.7.4 .github/workflows/ci.yml` -> zero issues
- PyYAML parse confirms three jobs with expected step ordering: `changes`, `proto-gen-lint`, `lint-typecheck-test (matrix)`
- `pnpm install` and `turbo run` not executed locally to avoid disturbing the in-flight monorepo state in the worktree; CI is the canonical validator for those steps.

## Test plan

- [ ] CI required checks `lint + typecheck + test (ubuntu-latest|macos-latest|windows-latest)` go green
- [ ] New `proto gen + lint` check goes green
- [ ] `e2e` and `no-silent-drops` remain green (untouched)
- [ ] First-run cold cache <= 15 min; subsequent runs benefit from `nm-pnpm-*` and `turbo-*` cache hits


Jiahui-Gu:
fixed at a36f87c — grandfathered the 5 frozen v0.3 protos in `packages/proto/buf.yaml` via buf v2 `lint.ignore_only` (per-rule, per-file). `pnpm exec buf lint` now exits 0 locally. Rules stay active (`use: STANDARD`) so any FUTURE protos added under `packages/proto/src/` will still be gated against RPC_RESPONSE_STANDARD_NAME / RPC_REQUEST_RESPONSE_UNIQUE.

Jiahui-Gu:
Pushed fix at e0695d1. Two distinct root causes — neither matched the initial diagnosis (codegen was already wired into the matrix step, Node version is 20 not 22).

## 1. `@ccsm/proto:test` lock.spec.ts — CRLF/LF lock drift

Not a "missing codegen" issue. `@ccsm/proto:gen` runs in the matrix job (line 22-28 of CI log: `cache hit, replaying logs`), and the `proto-gen-lint` precheck job's `git diff packages/proto/gen` would have caught a real codegen problem.

Real cause: `packages/proto/scripts/lock.mjs` hashes raw on-disk bytes:
```js
sha256OfFile(path) { return createHash('sha256').update(readFileSync(path)).digest('hex'); }
```

The repo had no `.gitattributes`, so contributors with `core.autocrlf=true` (Windows default) commit CRLF working copies. Whoever generated `lock.json` ran it on a CRLF checkout. Linux/macOS runners check out LF and compute different hashes.

Verified locally: `common.proto` with CRLF -> `7e06b7b67f...` (lock's "expected"); with LF -> `4acceb5752...` (CI's "actual"). All 8 protos same delta.

**Fix:** add `.gitattributes` pinning `*.proto` (and other text formats) to `eol=lf`, regenerate `lock.json` to match LF hashes. The lock.json regen is the unavoidable one-time consequence of switching canonical EOL — but every future Windows commit will now normalize to LF at staging, so this drift class is gone.

## 2. `@ccsm/daemon:test` coalescer.spec.ts 13/13 — better-sqlite3 ABI mismatch

`scripts/postinstall.mjs` rebuilds better-sqlite3 against the **Electron** ABI (NODE_MODULE_VERSION 130 for electron 41) for the app runtime. vitest then runs under **Node 20** (NODE_MODULE_VERSION 115) and the binding fails to load — coalescer.spec.ts is the only daemon test that actually opens a real `better-sqlite3` DB, hence 13/13 fail at import (other daemon specs touching pure path/contract logic pass).

The existing `pnpm rebuild better-sqlite3` step was a **silent no-op**. pnpm 10 ignores build scripts unless the package is allow-listed in `pnpm.onlyBuiltDependencies` (we don't allow-list anything). Logs prove it: the step exited in <1s with no output between `##[group]` and `##[endgroup]`.

**Fix:** switch to `npm rebuild better-sqlite3 --build-from-source`. npm CLI (bundled with `actions/setup-node`) ignores pnpm's allowlist and rebuilds against the active Node ABI. Reverse-verified locally:
- after `node scripts/postinstall.mjs` (Electron ABI): coalescer 13/13 FAIL at import
- after `npm rebuild better-sqlite3`: coalescer 13/13 PASS

## Files
- `.gitattributes` (new) — eol=lf for source/text, eol=crlf for `.bat`/`.ps1`, binary for assets
- `.github/workflows/ci.yml` — replace `pnpm rebuild` with `npm rebuild` (with comment explaining why)
- `packages/proto/lock.json` — regenerated against LF hashes

No source code, no `.proto` files, no other workflows touched.

Jiahui-Gu:
Pushed cross-platform CRLF fix at 145fcad.

**Root cause:** packages/daemon/src/db/migrations/001_initial.sql had no .gitattributes rule. Index stored LF, but Windows working trees got CRLF (autocrlf=true is the Windows git default). The migration-lock self-check (T5.4 / T10.1) hashes raw on-disk bytes; the entry 70838422... in src/db/locked.ts was the CRLF hash but LF hash on Linux/macOS runners is f76859... — every test that opens the DB cascaded to FAIL on the macos+ubuntu matrix legs.

**Fix:** Mirror the existing `*.proto` rule which solved the identical problem for packages/proto/scripts/lock.mjs:
1. Add `*.sql text eol=lf` to .gitattributes
2. `git add --renormalize .` + force re-checkout so working tree flips to LF on Windows
3. Update MIGRATION_HASHES['001_initial.sql'] to the LF SHA

**Verified locally on Windows** (file now LF on disk, sha256sum matches new lock entry, 33 tests pass):
```
$ npx vitest run test/db/migration-lock.spec.ts test/crash/pruner.spec.ts test/sessions/SessionManager.spec.ts
Test Files  3 passed (3)
     Tests  33 passed (33)
```

Jiahui-Gu:
## Rebase status: rebased on #192 (8ecbbf0); CI partially green

**Latest SHA**: `6cc480c`

### What landed in this PR (T0.8 scope)
- ci.yml install + proto-gen-and-lint precheck + 3-OS matrix (orthogonal additions to #192's setup-node baseline)
- buf.yaml grandfather frozen v0.3 protos
- Listener slot mutation lint fix
- LF normalisation for *.sql migration files (already covered by #192's `* text=auto eol=lf`)

### Conflict resolution against #192
- `node-version-file: .nvmrc` retained from #192; my pnpm setup-node block uses it (was hardcoded `20`)
- `npm rebuild better-sqlite3 --build-from-source` retained (pnpm 10 silently skips build scripts without `pnpm.onlyBuiltDeps` allowlist, confirmed in PR #906 first run)
- `check:native` verify step from #192 added after rebuild — fast-fail on ABI mismatch
- Cache key bumped to `node22` per #192

### Real product/test bugs surfaced by the matrix
The CRLF/ABI/Node-baseline failures from prior runs are gone (proto-gen+lint and ubuntu-latest are green). Remaining failures are pre-existing cross-platform test issues that the matrix uncovered:

1. **Fixed in this PR (transport bug, real)**: `Http2Server.close()` waits for active sessions; Connect-node clients keep them alive → afterEach hung 10s × 4 tests on linux/macos. Fix: track + `session.destroy()` before `close()` in both h2c-loopback and h2c-uds. Commit `537ecbe`.

2. **Fixed in this PR (windows test cost, env)**: `recovery.spec.ts` integrity-check case — `PRAGMA integrity_check` on a 200-row corrupt DB takes 12+ s on windows-2022 vs ~50ms elsewhere. Surgical per-test 30s timeout bump with comment. Commit `6cc480c`.

3. **NOT fixed — pushing back**: 
   - **macos-latest**: `build/__tests__/sign-scripts.spec.ts` runs `sign-linux.sh` on macOS; the script invokes `debsigs`/`rpm`/`gpg` which don't exist on darwin. Test needs an `if (platform() === 'linux')` skip OR the test should mock out the binaries.
   - **windows-latest**: `test/crash/crash-raw-recovery.spec.ts` "100 entries SIGKILL replay" times out at 5s. SIGKILL semantics on Windows are different (`child.kill('SIGKILL')` maps to TerminateProcess but the buffered fs writes / flush behaviour differs). Likely a test-side flush race, possibly a real product `crash-raw appender` portability issue.

### Recommendation
T0.8 ci.yml work is complete. Failures #3 are out of T0.8 scope and need separate tracking tasks for proper investigation. Suggest:
- Open task: "fix(test): sign-scripts test platform guard for macOS"  
- Open task: "fix(crash): crash-raw replay SIGKILL semantics on Windows"

Either skip those two tests temporarily with TODOs (so this PR can land) or hold #906 until the real fixes land.

Jiahui-Gu:

PR #922, opened 2026-05-03
electron(T6.7): renderer Hello + boot_id verification + reconnect backoff

Jiahui-Gu:
Task #75. Spec ref: docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md ch03 §3.3 (boot_id verification) + ch08 §6 (renderer error contract) + ch08 §6.1 (cold-start UX).

## Summary

- `src/renderer/connection/reconnect.ts` — pure decider for the locked boot reconnect schedule (100/200/400/800/1600/3200ms cap 5000ms) plus `runWithReconnect` driver. `shouldRetry` hook lets callers mark a class as fatal (used for version mismatch).
- `src/renderer/connection/hello.ts` — `performHello(deps)` re-fetches the descriptor on every attempt (spec ch03 §3.3 step 5), builds clients via T6.3 `createClients()`, calls `SessionService.Hello`, treats Connect `FailedPrecondition` as typed `HelloVersionMismatchError` (non-retryable). Defensive client-side proto-floor check.
- `src/renderer/connection/use-connection.tsx` — `<ConnectionProvider>` + `useConnection()` hook. State machine: `connecting` → `connected` → `reconnecting` → `connected` (or `version-mismatch` terminal). On bootId change across reconnect, calls `queryClient.invalidateQueries({ queryKey: ['ccsm'] })` per T6.3 cache-key shape and fires `onDaemonRestart` event for the toast. `retryNow()` powers the ch08 §6.1 cold-start modal.

## Design notes

**boot_id sourcing — important divergence from spec narrative.** Spec ch03 §3.3 step 4 reads "if the descriptor's boot_id does NOT equal the Hello response's boot_id...". As of v0.3 freeze, proto `HelloResponse` (`packages/proto/src/ccsm/v1/session.proto`) does NOT carry a `boot_id` field — it's only on `SupervisorHelloResponse` (UDS-only, unreachable from renderer per ch02 §4) and in the on-disk descriptor.

Mechanism implemented:
- Every reconnect re-fetches `app://ccsm/listener-descriptor.json`.
- Electron main's `protocol.handle('app', ...)` reads from disk on every request with `Cache-Control: no-store` (verified in `src/main/protocol-app.ts`), so the renderer sees the current daemon's `boot_id` every time.
- Successful Hello against the descriptor's address proves the daemon is reachable + version-compatible; descriptor `boot_id` change across reconnect → daemon restarted → cache nuke + toast.
- Foreign-process spoofing on the bind address is blocked at the bridge layer (peer-cred middleware on Listener A for UDS / named-pipe; `:authority` enforcement for loopback-TCP).

When proto v2+ adds `HelloResponse.boot_id` (additive per ch15 §3 forever-stable rule), `hello.ts` flips one line — the public `HelloResult.bootId` field is named precisely so call sites do not care which side produced it.

## Wiring

`<ConnectionProvider>` is composed by Task #70 (T6.6) boot wiring around `<ClientsProvider>`/`<QueryClientProvider>` in the renderer entry. No renderer entry file exists yet (T6.6 owns it); shipping a stub here would conflict with #70.

Depends on:
- PR #915 (T6.3): `createClients`, `<ClientsProvider>` — used by `performHello`.
- PR #911 (T6.2): `transport-bridge` — supplies the bridge URL the descriptor exposes; `buildTransport` is injected by T6.6 boot.
- PR #898 (T6.1): `protocol-app.ts` `DescriptorV1` shape + `DESCRIPTOR_URL`.

## Test plan (vitest UT — pure, no electron / no daemon)

- [x] `test/renderer/connection/reconnect.spec.ts` — schedule equality (100/200/400/800/1600/3200/5000), cap past schedule, `shouldRetry: false` is non-retryable, AbortSignal cancels mid-sleep, `onBackoff` event shape.
- [x] `test/renderer/connection/hello.spec.ts` — success returns descriptor `bootId`, `FailedPrecondition` → `HelloVersionMismatchError`, daemon `protoVersion < floor` → `HelloVersionMismatchError`, `Unavailable` propagates as ConnectError (caller retries), descriptor fetch failure wrapped as `DescriptorFetchError`, descriptor re-fetched per attempt (no in-memory cache).
- [x] `test/renderer/connection/use-connection.spec.tsx` — connected state with bootId pin + `onConnected` fired, `version-mismatch` terminal (no infinite loop), bootId mismatch across `retryNow()` invalidates `['ccsm']` queries + fires `onDaemonRestart`, unchanged bootId does NOT invalidate, missing-provider guard throws clear error.

## Local quality gates (all green)

```
pnpm lint                                    -> 0 errors / 0 warnings
pnpm --filter @ccsm/electron run typecheck   -> pass
pnpm --filter @ccsm/electron test            -> 95 pass / 2 pre-existing skipped
```

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -48,6 +48,8 @@ export default [
         crypto: 'readonly',
         URL: 'readonly',
         AbortController: 'readonly',
+        AbortSignal: 'readonly',
+        DOMException: 'readonly',
         KeyboardEvent: 'readonly',
         CustomEvent: 'readonly',
         Event: 'readonly',
--- a/packages/daemon/test/integration/pty-soak-shared.ts
+++ b/packages/daemon/test/integration/pty-soak-shared.ts
@@ -1,5 +1,4 @@
 // packages/daemon/test/integration/pty-soak-shared.ts
-/* global AbortSignal */
 //
 // Shared driver for the two PTY soak harnesses:
 //   - pty-soak-1h.spec.ts   — ship-gate (c) per chapter 12 §4.3 (60 min, nightly).
--- a/packages/electron/src/renderer/connection/hello.ts
+++ b/packages/electron/src/renderer/connection/hello.ts
@@ -1,0 +1,307 @@
+// T6.7 — Renderer Hello flow + boot_id verification.
+//
+// Spec ref: docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md
+// chapter 02 §3 (startup ordering) + chapter 03 §3.3 (Hello-echo boot_id
+// verification) + chapter 08 §6 (renderer error contract).
+//
+// Sequence (one connect attempt):
+//
+//   1. Re-read the descriptor from `app://ccsm/listener-descriptor.json`
+//      (NOT a cached in-memory copy — the spec at ch03 §3.3 step 5 is
+//      explicit: every reconnect re-reads the file so a daemon restart
+//      with a new boot_id is detected on the very first attempt).
+//   2. Build a Connect transport against the descriptor's `address`
+//      (the bridge endpoint the main process injected per ch08 §4.1+§4.2).
+//   3. Construct typed clients via `createClients` from T6.3.
+//   4. Call `SessionService.Hello` with the renderer's `proto_min_version`.
+//      This is the FIRST RPC over the connection per ch03 §3.3 step 3.
+//   5. On `FAILED_PRECONDITION` → version mismatch; surface as the
+//      blocking-modal-worthy error (caller decides UX). NOT retried.
+//   6. On any other Connect error → throw; caller's reconnect loop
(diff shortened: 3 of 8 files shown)


Jiahui-Gu:
Lint fixed at 3cb62f2: added AbortSignal + DOMException to root eslint.config.js base globals (alongside existing AbortController), and rewrote children-prop in use-connection.spec.tsx (children as React.createElement positional arg, not prop). Also removed a now-redundant /* global AbortSignal */ comment in pty-soak-shared.ts that would otherwise trip @typescript-eslint/no-redeclare.

Reverse-verify: git stash -> npm run lint:app reports the same 10 errors (8 no-undef + 1 children-prop, plus the redeclare that surfaces only with my partial fix); git stash pop -> 0 errors / 101 pre-existing warnings.

PR #923, opened 2026-05-03
daemon(T5.7): corrupt-DB recovery + healthz modal flag

Jiahui-Gu:
## Summary

Task #60 — implements ch07 §6 boot-time SQLite corrupt-DB recovery for the daemon.

- New `packages/daemon/src/db/recovery.ts` runs `PRAGMA integrity_check` (full, not quick_check per spec) at boot. On non-`"ok"` result it renames the DB + `-wal` / `-shm` siblings to `<db>.corrupt-<unix_ms>`, appends a NDJSON line to `crash-raw.ndjson` via the existing #62 appender (`source = sqlite_corruption_recovered`, `owner_id = daemon-self`), fsyncs the parent dir, and sets an in-memory `RecoveryFlag` shared with the supervisor server. When SQLite cannot even open the file, the NDJSON detail carries `{status:"unrecoverable", error:"..."}`.
- Supervisor `server.ts` now exposes `recovery_modal: { pending, ts_ms, corrupt_path }` on `/healthz` (forever-stable golden + contract spec updated to match the spec extension) and adds `POST /ack-recovery` behind the same peer-cred admin allowlist as `/hello`/`/shutdown` (#22). Non-admin callers get the existing 403 `peer-cred-rejected` shape and the flag stays set.
- Daemon entrypoint `index.ts` calls `checkAndRecover()` during the `OPENING_DB` phase before any other code opens the DB; logs the recovery outcome and notes the deferred #56 migration re-run TODO.

## Layer-1 notes

- Spec ch07 §6 mandates `recovery_modal: { pending, ts_ms, corrupt_path }` (nested object), not the flat `recovery_modal_pending: boolean` the task prompt suggested. Followed spec to avoid v0.4 rework — the Electron client already renders the corrupt path / timestamp from this struct.
- Sidecar: spec routes the audit trail through the existing `crash-raw.ndjson` (FOREVER-STABLE per ch09 §1), not a new `<db>.corrupt-<ts>.recovered.ndjson` file. Reusing the #62 appender keeps the line shape identical to every other crash event and inherits the boot-replay path into `crash_log` (rule §1: don't reinvent existing modules).
- `sqlite.ts` left untouched per its own SRP header ("crash/recovery (#59) all live in separate tasks and MUST NOT land here") — recovery is invoked from the entrypoint instead.
- #56 migration runner not landed; task brief says skip+log in that case, which is what the entrypoint does.

## Test plan

- [x] vitest UT: corrupt-DB rename + NDJSON sidecar emit (`recovery.spec.ts`)
- [x] vitest UT: healthy DB recovery is no-op
- [x] vitest UT: `/healthz` exposes `recovery_modal.pending=true` after recovery, false after `/ack-recovery`
- [x] vitest UT: `/ack-recovery` rejects non-admin callers with 403 (peer-cred allowlist)
- [x] vitest UT: forever-stable contract spec covers the new `recovery_modal` sub-shape
- [x] `pnpm lint` (workspace) — green
- [x] `pnpm --filter @ccsm/daemon run typecheck` — green
- [x] `pnpm --filter @ccsm/daemon test` — 312/312 of the affected files pass; one pre-existing unrelated descriptor/schema CRLF failure on this Windows worktree, present on `origin/working` as well (verified via `git stash` baseline).

### DIFF
--- a/packages/daemon/src/db/__tests__/recovery.spec.ts
+++ b/packages/daemon/src/db/__tests__/recovery.spec.ts
@@ -1,0 +1,205 @@
+// packages/daemon/src/db/__tests__/recovery.spec.ts
+//
+// Boot-time corrupt-DB recovery (T5.7 / Task #60). Spec ch07 §6.
+
+import {
+  closeSync,
+  existsSync,
+  mkdtempSync,
+  openSync,
+  readFileSync,
+  rmSync,
+  writeSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import Database from 'better-sqlite3';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+import { checkAndRecover, makeRecoveryFlag } from '../recovery.js';
+
+describe('checkAndRecover (T5.7 — ch07 §6)', () => {
+  let tmpDir: string;
+  let dbPath: string;
+  let crashRawPath: string;
+
+  beforeEach(() => {
+    tmpDir = mkdtempSync(join(tmpdir(), 'ccsm-daemon-recovery-'));
+    dbPath = join(tmpDir, 'sessions.db');
+    crashRawPath = join(tmpDir, 'crash-raw.ndjson');
+  });
+
+  afterEach(() => {
+    rmSync(tmpDir, { recursive: true, force: true });
+  });
+
+  // --------------------------------------------------------------------- //
+  // first boot — DB file does not exist                                   //
+  // --------------------------------------------------------------------- //
+
+  it('is a no-op when the DB file does not exist (first boot)', () => {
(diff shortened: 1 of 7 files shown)


Jiahui-Gu:
Rebased on origin/working at d1d52f9; new HEAD = 5c33453. Conflict in packages/daemon/src/index.ts resolved keeping both sides: T5.7 checkAndRecover() now runs at the start of OPENING_DB phase BEFORE openDatabase()/runMigrations(); listener-A wiring + watchdog + router from #918/#919/#920/#924 untouched. recoveryFlag plumbed through runStartup signature. Local: pnpm install --frozen-lockfile OK, tsc build OK, src/db/__tests__/recovery.spec.ts 8/8 pass, src/supervisor 27/27 pass.

PR #924, opened 2026-05-03
daemon(T1.8): graceful shutdown

Jiahui-Gu:
## Summary

Implements graceful shutdown orchestrator per spec ch02 §4 (Task #25).

**Order**: stop accepting → drain RPC (5s budget) → SIGTERM claude children → SIGKILL after 3s → close pty-host children → WAL checkpoint TRUNCATE → close DB → close Supervisor (last so `/healthz` survives the drain window).

**Files**:
- `packages/daemon/src/shutdown.ts` (NEW) — `Shutdown.run(ctx)` orchestrator + `installShutdownHandlers` SIGTERM/SIGINT wiring + `InFlightTracker` shape (T2.x interceptor populates impl).
- `packages/daemon/src/shutdown.spec.ts` (NEW) — 21 unit tests covering every spec branch.
- `packages/daemon/src/index.ts` — wires the orchestrator + signal handlers; mutable `ctxRef` so later tasks (T1.4 listener, T5.1 db, T4.x pty children) register their handles as they land.

## Design notes

- **SRP**: `shutdown.ts` is a pure decider. Every side effect is an injected sink (`Listener.stop`, `SupervisorServer.stop`, `PtyHostChildHandle.closeAndWait`, `walCheckpointTruncate`, `db.close`). Test seam = the same trait shapes (#28 / Listener trait, supervisor server trait).
- **Re-use, not re-invent**: `walCheckpointTruncate` from #58 (`packages/daemon/src/db/sqlite.ts`); `PtyHostChildHandle.closeAndWait` (already implements SIGTERM→SIGKILL escalation). The orchestrator only sequences them.
- **Error containment**: each step's failure is captured into `ShutdownResult.errors`; the sequence never aborts mid-way (spec ch02 §4 requires the daemon to exit even messily).
- **Idempotent**: concurrent + sequential `run()` calls return the cached result; no side effect re-fires.
- **Second-signal escape hatch**: a second SIGTERM/SIGINT triggers `process.exit(1)` after a 50ms grace; disposer cancels the timer for clean teardown.

## Layer 1

- Considered `npm:terminus`/`lightship` — none model our ordered staircase, would still need every hook. Skipped.
- Considered building per-RPC tracker here — out of scope; T2.x owns the Connect interceptor wiring. Shipped the `InFlightTracker` shape so the seam is stable.

## Test plan

Local unit tests (`pnpm --filter @ccsm/daemon exec vitest run src/shutdown.spec.ts`):

```
 Test Files  1 passed (1)
      Tests  21 passed (21)
   Duration  2.05s

 ✓ Shutdown.run() — ordered sequence (5)
   ✓ returns clean result when nothing is wired
   ✓ stops every listener and the supervisor (data-plane first, supervisor last)
   ✓ drains in-flight RPC within budget and reports zero
   ✓ continues past stuck RPC when budget exceeded (5s spec budget)
   ✓ default budgets match spec ch02 §4 (5s + 3s)
 ✓ Shutdown.run() — pty children (SIGTERM → SIGKILL after 3s) (5)
   ✓ calls closeAndWait with the spec-locked claude SIGKILL window
   ✓ honours a caller-supplied claudeSigkillMs override
   ✓ classifies a SIGKILL exit as reason=sigkill (not "crashed")
   ✓ records per-child closeAndWait throws into errors but continues
   ✓ fans out child closes in parallel (not sequentially)
 ✓ Shutdown.run() — WAL checkpoint + DB close (3)
   ✓ runs walCheckpointTruncate then db.close when a db is present
   ✓ skips checkpoint + close when db is null (pre-OPENING_DB failure path)
   ✓ records pragma failure into errors but still attempts db.close()
 ✓ Shutdown.run() — error containment + idempotency (4)
   ✓ captures listener stop throws but proceeds through every later step
   ✓ captures supervisor stop throws into errors
   ✓ is idempotent: second concurrent run() returns the same result
   ✓ is idempotent: third sequential run() also returns the cached result
 ✓ installShutdownHandlers (4)
   ✓ triggers exactly once on first signal and is removable after dispose
   ✓ reports first signal kind to the onSignal callback
   ✓ reports second signal kind on a re-emit and forces exit(1)
   ✓ exports SIGTERM + SIGINT as the default shutdown signal set
```

Quality gates (all pass):
- `pnpm lint` — pass (workspace).
- `pnpm --filter @ccsm/daemon run typecheck` — pass.
- `pnpm --filter @ccsm/daemon test` — 339 pass / 21 of mine + 318 pre-existing. One pre-existing failure on Windows (`test/descriptor/schema.spec.ts` CRLF golden mismatch) is unrelated to this PR — confirmed by checking out origin/working and re-running the same case (same failure). CI on Linux will pass.

## Scope notes

- **Not** wired into a running daemon yet because no listeners/db/pty children are constructed by the T1.1 startup stub. Later tasks (T1.4 listener, T5.1 db, T4.x pty) populate `ctxRef` as they land.
- **Not** wired to the Supervisor `/shutdown` RPC `onShutdown` callback yet — that wiring lives in the entrypoint when the supervisor server is constructed (currently a TODO at the entrypoint's STARTING_LISTENERS phase). One-line addition when supervisor instantiation lands.

### DIFF
--- a/packages/daemon/src/index.ts
+++ b/packages/daemon/src/index.ts
@@ -8,7 +8,6 @@
 //   - SQLite open + migrations             → T5.1
 //   - Session restore + orphan-uid check   → T3.4
 //   - Supervisor /healthz                  → T1.7
-//   - Graceful shutdown                    → T1.8
 //
 // Wired-in:
 //   - Listener A bind                      → T1.4 (`makeListenerA`)
@@ -17,6 +16,18 @@
 //                                                  `makeRouterBindHook`)
 //   - Descriptor file write                → T1.6 (separate concern; not
 //                                                  yet called from here)
+//   - systemd watchdog (main thread)       → T5.13 (`startSystemdWatchdog`)
+//   - Graceful shutdown                    → T1.8 (this file installs the
+//                                                  SIGTERM/SIGINT handlers
+//                                                  that call `Shutdown.run`)
+//
+// In scope (T1.8): SIGTERM / SIGINT handler installation. The actual
+// shutdown sequence is `Shutdown.run(...)` (./shutdown.ts); this file
+// only wires (a) the signal trap that calls it, and (b) the empty
+// `ShutdownContext` we can populate today (no listeners / no db / no
+// pty children are constructed by the T1.1 startup stub yet — later
+// tasks register their handles with `register*OnEntrypoint` helpers
+// when they land).
 
 import { mkdirSync } from 'node:fs';
 import { dirname, join } from 'node:path';
@@ -28,6 +39,13 @@ import { Lifecycle, Phase } from './lifecycle.js';
 import { makeListenerA } from './listeners/factory.js';
 import type { Listener } from './listeners/types.js';
 import { makeRouterBindHook } from './rpc/bind.js';
+import {
+  Shutdown,
+  installShutdownHandlers,
+  noopInFlightTracker,
+  type ShutdownContext,
+  type ShutdownResult,
+} from './shutdown.js';
 import {
   startSystemdWatchdog,
   type SystemdWatchdogHandle,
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
rebased on working at 1dfa11a3373f1c585d8274113b224bbd6bc19676; conflicts in packages/daemon/src/index.ts resolved keeping all sibling features (T2.2 ConnectRouter wiring, T1.4 listenerA factory, T5.13 systemd watchdog start/stop) and adding T1.8 shutdown.run + installShutdownHandlers. Order: listener bind → router register → recovery + migrations + db open → watchdog start → shutdown handlers installed LAST. Validated: turbo build green; shutdown tests 21/21 pass.

PR #965, opened 2026-05-03
wave1: T4.5 snapshot-codec package

Jiahui-Gu:
## [LIBRARY-ONLY] Task #207 / WAVE1-4.5

New leaf package `@ccsm/snapshot-codec` implementing the SnapshotV1
inner-codec layer per design spec ch06 §2 (`Codec rules`). No consumers
wired up — Wave 3 #44/#46 (T4.6/T4.7) owns that.

## Summary

- `encode(buf, codec)` -> `[codec_byte][compressed_bytes]` (sync)
- `decode(buf)` -> dispatches on leading byte; returns the decompressed payload
- Codec 1 = zstd, codec 2 = gzip
- Both codecs implemented via Node's built-in `node:zlib` sync helpers
  (`zstdCompressSync` / `zstdDecompressSync` / `gzipSync` / `gunzipSync`)
- **Zero runtime dependencies** — `engines.node` bumped to `>=22.15.0`
  in this package only (where the zstd helpers stabilized in Node 22)
- Public errors: `EmptyBufferError`, `UnknownCodecError`

## Why no third-party zstd dep

Manager spec listed `zstd-napi` / `@mongodb-js/zstd` as candidates. I
evaluated both — they are native modules (prebuild-install + `.node`
binaries) which add SEA-bundle complexity, async-only API
(`@mongodb-js/zstd`), and a postinstall step. Node 22.15+ exposes a
synchronous zstd in the standard library; using it gives:

- zero deps (matches dev protocol §2 "标准库一行能解决 → 不要包一层")
- sync API matches the task spec signature exactly
- no SEA bundling work for daemon (T7.1 already complex)
- forward-compat: swap to a native lib later only requires API
  preservation, no consumer change

If the daemon's bundled SEA Node ever lags behind 22.15, we can swap
the implementation behind the same public surface in one PR.

## Why this is a thin layer (only the codec byte, not the full wire frame)

Spec ch06 §2 defines the full `SnapshotV1Wire` framing:

```
outer_magic[4] "CSS1" | codec[1] | reserved[3]=0 | inner_len[u32] | inner[...]
```

This package intentionally implements **only** the `codec[1] | inner[...]`
inner-dispatch layer. The outer magic + reserved + inner_len framing
lives in the daemon snapshot serializer (T4.6/T4.7) so:

- the wire format and the codec dispatch can evolve independently
- v0.4 wire-format extensions (e.g., chunked snapshots via the reserved
  bytes) don't touch the codec package
- the codec package remains usable for any other byte-prefix codec
  dispatch use case that emerges (delta payload compression, IPC
  payload compression, etc.)

This split is forever-additive — adding new codec values (`codec = 3`
zstd-with-dictionary per spec ch06 §3 "Forward-additive") only extends
the `Codec` union here; no consumer change.

## Tests

26 cases, all pass locally:

- encode/decode roundtrip across 5 fixtures (tiny ascii / repetitive
  4 KiB / multibyte unicode / full 0..255 binary / empty input) x 2
  codecs = 10 cases
- frame-magic assertions: pins algorithm choice without coupling to
  Node's bundled zlib library version (only the magic prefix is
  asserted)
- intra-process determinism: v0.4 ship-gate (c) byte-equality requires
  this — same input twice MUST produce identical bytes
- hex-pinned golden decode for both codecs: decode is portable across
  zlib versions because both gzip and zstd frame formats are fully
  specified, so a hex-pinned blob from any conformant encoder MUST
  decode to the known plaintext on any other conformant Node
- error paths: empty buffer, unknown codec bytes (0x00, 0x03, 0xff),
  underlying zlib propagation on garbage input
- API surface sanity: codec constants, fresh-buffer guarantee

```
Test Files  1 passed (1)
     Tests  26 passed (26)
```

## Verification

- `pnpm install` -> OK (only adds @types/node devDep entry to lockfile)
- `pnpm --filter @ccsm/snapshot-codec build` -> OK
- `pnpm --filter @ccsm/snapshot-codec test` -> 26/26 pass
- `pnpm --filter @ccsm/snapshot-codec lint` -> clean
- `pnpm --filter @ccsm/snapshot-codec typecheck` -> clean
- `pnpm typecheck` (root) -> clean
- `pnpm run lint:app` (root sweep) -> 0 errors (66 pre-existing warnings)
- `pnpm run build` (turbo) -> picks up the new package automatically;
  3/3 tasks successful

## Out of scope (deferred to other tasks per task brief)

- Wire-up to daemon snapshot path -> Wave 3 T4.6/T4.7 (#44/#46)
- Full `SnapshotV1Wire` outer framing -> daemon snapshot serializer
- `SnapshotV1Inner` byte layout encoder -> Wave 3 / spec ch06 §2
- Property-based round-trip spike `[snapshot-roundtrip-fidelity]` ->
  spec ch16 §1.8

## Test plan

- [x] Local test/lint/typecheck/build all green
- [x] Root typecheck unaffected
- [x] Root lint sweep finds 0 new errors
- [ ] CI green

### DIFF
--- a/packages/snapshot-codec/eslint.config.js
+++ b/packages/snapshot-codec/eslint.config.js
@@ -1,0 +1,35 @@
+// Per-package ESLint config. Inherits the repo root flat config and adds
+// boundary rules: snapshot-codec is a leaf library (no project deps), so
+// any cross-package import is a smell.
+import rootConfig from '../../eslint.config.js';
+
+export default [
+  ...rootConfig,
+  {
+    files: ['src/**/*.ts'],
+    languageOptions: {
+      globals: {
+        // Web platform encoders are part of the standard Node global scope
+        // since Node 11; the root config doesn't declare them but our
+        // tests and (future) consumers rely on them.
+        TextEncoder: 'readonly',
+        TextDecoder: 'readonly',
+      },
+    },
+    rules: {
+      'no-restricted-imports': ['error', {
+        patterns: [
+          {
+            group: ['@ccsm/*'],
+            message:
+              '@ccsm/snapshot-codec is a leaf library — must not depend on other workspace packages. Codec primitives only (spec ch06 §2).',
+          },
+          {
+            group: ['electron', 'electron/*', 'react', 'react-dom'],
+            message: '@ccsm/snapshot-codec is a pure-Node library — UI deps forbidden.',
+          },
+        ],
+      }],
+    },
+  },
+];
--- a/packages/snapshot-codec/package.json
+++ b/packages/snapshot-codec/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@ccsm/snapshot-codec",
+  "version": "0.0.1",
+  "private": true,
+  "type": "module",
+  "description": "CCSM SnapshotV1 codec layer (Task #207, T4.5). Provides synchronous encode/decode that prefixes a single codec byte (1 = zstd, 2 = gzip) to a compressed payload, per design spec ch06 §2 'Codec rules'. The full SnapshotV1Wire framing (CSS1 magic, reserved bytes, inner_len) is layered on top by the daemon snapshot serializer (T4.6/T4.7) — this package is intentionally the inner compression-dispatch primitive, nothing more.",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "engines": {
+    "node": ">=22.15.0 <23"
+  },
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
+    "lint": "eslint \"src/**/*.ts\"",
+    "typecheck": "tsc -p tsconfig.json --noEmit",
+    "test": "vitest run"
+  },
+  "devDependencies": {
+    "@types/node": "^22.10.0"
+  }
+}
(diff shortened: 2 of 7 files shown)


Jiahui-Gu:
## Review verdict: APPROVED

### Layer 1 (Direction): RIGHT

- LIBRARY-ONLY scope matches Wave 1 task brief; outer `SnapshotV1Wire` framing correctly deferred to T4.6/T4.7 daemon serializer (#44/#46/#52). Spec ch06 §2 explicitly permits this split — "Codec rules" defines codec dispatch as a discrete concern from the outer wire struct.
- Wire-up check PASS: `packages/daemon/test/pty/snapshot-codec.spec.ts` already names `@ccsm/snapshot-codec` as the future production importer for T4.6/T4.7/T4.8.
- Zero-deps via `node:zlib` instead of `zstd-napi` / `@mongodb-js/zstd` is the right call — matches dev protocol "standard library one-liner > wrapper", avoids SEA bundling complexity for T7.1, sync API matches the spec signature.

### Layer 2 (Implementation): clean

- 26 tests cover the right axes: roundtrip x 5 fixtures (incl. empty + binary 0..255 + multibyte unicode) x 2 codecs, frame-magic pin (algorithm choice without coupling to zlib version), intra-process determinism (v0.4 ship-gate (c) byte-equality), hex-pinned golden decode for both codecs, error paths (empty / unknown codec / garbage payload propagation), fresh-buffer guarantee.
- `engines.node: ">=22.15.0 <23"` correctly pins the Node version where `zstdCompressSync` landed. CI Node is 22.18.0 (`.nvmrc`) so this is satisfied.
- TypeScript exhaustiveness guard on `Codec` union prevents silent codec drift.
- Per-package ESLint config bans `@ccsm/*` and UI imports — leaf-package discipline enforced at lint time.
- `decode` uses `subarray` (no copy) — correct since `node:zlib` does not mutate input.

### Risk noted (non-blocking)

- Node 22's `zstdCompressSync` is **Stability 1 - Experimental** per Node docs. The thin wrapper design means swap to `zstd-napi` (or to the stable Node API once promoted) is a one-PR change behind the same public surface. Acceptable for v0.3 ship; flag for v0.4 review.

### CI

All green: macos / ubuntu / windows lint+typecheck+test PASS, no-silent-drops PASS, detect-changes PASS.

Squash-merging.

PR #966, opened 2026-05-03
wave1: T4.1 daemon-boot-e2e + 3 wire-up TODOs

Jiahui-Gu:
## Task

Wave-1 Task #208 — daemon-boot-e2e + fix 3 wire-up TODOs in
`packages/daemon/src/index.ts` (NOTE: task brief said `apps/daemon/`;
the real path is `packages/daemon/` after the apps→packages migration).

### Three TODOs fixed in `packages/daemon/src/index.ts`

| Line (pre) | Was | Now |
|---|---|---|
| 174 | `makeRouterBindHook({})` — every RPC returns `Unimplemented` and PEER_INFO is never deposited so even Hello would fail Unauthenticated | `makeRouterBindHook({ helloDeps, interceptors: [bearerToPeerInfoInterceptor, peerCredAuthInterceptor] })` — wires the real T2.3 SessionService.Hello handler + the auth chain on production listener bind |
| 169 | `// TODO(T1.6): write listener-a.json descriptor after bind succeeds` | `await writeDescriptor(env.paths.descriptorPath, payload)` with a full v1 `DescriptorV1` derived from the bound listener address (atomic tmp+fsync+rename per ch03 §3.1) |
| 182 | `log('TODO(T1.7): flip Supervisor /healthz to 200')` | `makeSupervisorServer({...}).start()` after lifecycle reaches READY; handle plumbed into the shutdown context for graceful teardown |

### Same `runStartup` also wires
- `replayCrashRawOnBoot` after migrations open the DB (ch09 §6.2 boot replay).
- `installCaptureSources` after replay (ch09 §1 — uncaughtException + sqlite errors + claude exit + listener bind + watchdog miss).
- `bearerToPeerInfoInterceptor` (new, exported) — a tiny Authorization-header → `LoopbackTcpPeer` PeerInfo bridge for the loopback transport. UDS / named-pipe peer-cred extraction stays out of scope (separate hardening pass); spec ch03 §4 documents loopback as the v0.3 dev/test transport.

### New e2e spec — `packages/daemon/test/integration/daemon-boot-end-to-end.spec.ts`

Asserts six wire-up invariants against one real `runStartup` invocation (in-process — same exported function `main()` calls):

1. `~/.ccsm/listener-a.json` (per-test tmp dir) exists and validates against the frozen v1 schema in `packages/daemon/schemas/listener-a.schema.json`.
2. `SessionService.Hello` over loopback h2c returns a populated `HelloResponse` (NOT `Code.Unimplemented`).
3. `Hello` with no Authorization header returns `Code.Unauthenticated` (proves the auth chain is wired in the production bind hook).
4. Supervisor UDS `GET /healthz` returns 200 with `ready=true` after phase READY.
5. `result.captureSourcesInstalled === true` (probe flag from `runStartup`).
6. `result.crashReplayResult` is populated; when the seed succeeds (writable per-OS root), `inserted >= 1` and the file is truncated post-replay.

Note on assertion #2 vs the task brief: the brief said "Hello → returns boot_id, version". The current `HelloResponse` proto has no `boot_id` field (see `packages/daemon/src/rpc/hello.ts` header comment — adding it would require a coordinated `lock.json` + PROTO_VERSION + electron client mirror bump, intentionally out of scope here). The assertion is on `daemonVersion`, `protoVersion`, `listenerId`, and the echoed `meta.requestId` — the actual Hello contract.

## Reverse-verify

Removing the descriptor write (line 169 fix) makes exactly assertion #1 fail; restoring turns all 6 green.

```
$ # remove writeDescriptor call
$ pnpm --filter @ccsm/daemon exec vitest run \
    test/integration/daemon-boot-end-to-end.spec.ts
Tests  1 failed | 5 passed (6)
FAIL ... > writes listener-a.json and validates against v1 schema
  Error: ENOENT ... 'listener-a.json'

$ # restore
$ pnpm --filter @ccsm/daemon exec vitest run \
    test/integration/daemon-boot-end-to-end.spec.ts
Tests  6 passed (6)
```

## Local gates

| Command | Result |
|---|---|
| `pnpm --filter @ccsm/daemon typecheck` | PASS |
| `pnpm --filter @ccsm/daemon lint` | 0 errors (warnings pre-existing) |
| `pnpm --filter @ccsm/daemon build` | PASS |
| `pnpm --filter @ccsm/daemon test` | 995 passed / 23 skipped / 5 todo |
| `pnpm run lint:no-ipc` | PASS |

## Migration-lock side fix

`packages/daemon/src/db/locked.ts` ships the wrong SHA256 for `001_initial.sql` on `origin/working` (entry was computed under CRLF; `.gitattributes` now normalises `*.sql` to LF). With the broken hash, ANY test that calls `runMigrations` is blocked at OPENING_DB — including this new daemon-boot e2e. The matching standalone fix lives on `dev/t0.8-ci-matrix` (PR #906, commit `e29e056`, still DRAFT). To make this PR mergeable + verifiable standalone, the locked.ts hash is updated here. When PR #906 lands the diff will be a no-op merge.

## v0.3 zero-rework

This e2e file is the foundation for Wave 3 wire-up regression catch (Task #225 will layer additional assertions). The shape (one boot, multiple `it()`, probe-flag on `RunStartupResult`) is the long-term shape — Wave 3 extends `it()` blocks against the same boot, no new harness.

## Test plan

- [x] `pnpm --filter @ccsm/daemon typecheck`
- [x] `pnpm --filter @ccsm/daemon test`
- [x] Reverse-verify: remove one wire-up fix, exactly the related assertion fails

### DIFF
--- a/packages/daemon/src/db/locked.ts
+++ b/packages/daemon/src/db/locked.ts
@@ -50,7 +50,7 @@ export interface MigrationLock {
  * reshape into multi-line object literals — that would break both consumers.
  */
 const MIGRATION_HASHES = {
-  '001_initial.sql': '70838422d5fdc82e39818e8722f365ee789f340951485547545c9de5200a128d',
+  '001_initial.sql': 'f76859d5ad478a54f78754b6bd2874495452826a6430ab102534275979e0b06c',
 } as const;
 
 /**
--- a/packages/daemon/src/index.ts
+++ b/packages/daemon/src/index.ts
@@ -1,45 +1,58 @@
 // Daemon entrypoint. Spec ch02 §3 startup order: this file walks phases
 // 1–5 in sequence, logs each transition, and exits non-zero on any phase
-// failure. The actual step bodies (DB open, session restore, listener
-// bind, descriptor write, /healthz flip) live in later tasks — T1.1 just
-// wires the lifecycle skeleton with TODO stubs.
+// failure. Wave 1 wire-up (Task #208) plumbed the Listener-A descriptor
+// write, the SessionService.Hello handler, the Supervisor UDS server,
+// and the crash-raw boot replay + capture-sources install — i.e. the
+// previously-stubbed `runStartup` is now end-to-end.
 //
-// Out of scope (per task brief):
-//   - SQLite open + migrations             → T5.1
-//   - Session restore + orphan-uid check   → T3.4
-//   - Supervisor /healthz                  → T1.7
+// Out of scope (per task brief / future tasks):
+//   - Session restore + orphan-uid check       → T3.4
+//   - PtyHost / SessionManager wiring          → T6.x
 //
-// Wired-in:
-//   - Listener A bind                      → T1.4 (`makeListenerA`)
-//   - HTTP/2 transport adapters            → T1.5 (h2c-uds / h2c-loopback / h2-named-pipe)
-//   - ConnectRouter + stub services        → T2.2 (this file consumes
-//                                                  `makeRouterBindHook`)
-//   - Descriptor file write                → T1.6 (separate concern; not
-//                                                  yet called from here)
-//   - systemd watchdog (main thread)       → T5.13 (`startSystemdWatchdog`)
-//   - Graceful shutdown                    → T1.8 (this file installs the
-//                                                  SIGTERM/SIGINT handlers
-//                                                  that call `Shutdown.run`)
-//
-// In scope (T1.8): SIGTERM / SIGINT handler installation. The actual
-// shutdown sequence is `Shutdown.run(...)` (./shutdown.ts); this file
-// only wires (a) the signal trap that calls it, and (b) the empty
-// `ShutdownContext` we can populate today (no listeners / no db / no
-// pty children are constructed by the T1.1 startup stub yet — later
-// tasks register their handles with `register*OnEntrypoint` helpers
-// when they land).
+// Wired here:
+//   - DaemonEnv build                          → env.ts
+//   - SQLite open + migrations + recovery      → T5.1 / T5.7
+//   - Crash boot replay (ch09 §6)              → crash/raw-appender.replayCrashRawOnBoot
+//   - Crash capture-sources install (ch09 §1)  → crash/sources.installCaptureSources
+//   - Listener A bind                          → T1.4 (`makeListenerA`)
+//   - HTTP/2 transport adapters                → T1.5 (h2c-uds / h2c-loopback / h2-named-pipe)
+//   - ConnectRouter + real Hello handler       → T2.2 + T2.3 (`makeRouterBindHook({ helloDeps })`)
+//   - bearer-token → PeerInfo bridge for       → this file (`bearerToPeerInfoInterceptor`)
+//     loopback transport
+//   - Listener-A descriptor file write         → T1.6 (`writeDescriptor`)
+//   - Supervisor UDS HTTP server (/healthz,    → T1.7 (`makeSupervisorServer`)
+//     /hello, /shutdown, /ack-recovery)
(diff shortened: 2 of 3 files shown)


Jiahui-Gu:
## Review (Task #275 — Wave 1 anchor)

### Layer 1 — Direction: right

Right scope for the wave-1 anchor: 3 wire-up TODOs in the production `runStartup`, plus ch09 §6.2 boot replay + ch09 §1 capture-sources install, plus the e2e gate that catches all of it. Single PR per concern, no over-reach into T6.x / T3.4 wiring (correctly deferred). The probe-flag shape on `RunStartupResult` (`captureSourcesInstalled`, `crashReplayResult`) is the right long-term seam for Wave 3 to extend without reshaping the harness — zero-rework rule honored.

### Layer 2 — Implementation findings

1. **Interceptor order (line 306)** — `[bearerToPeerInfoInterceptor, peerCredAuthInterceptor]` is correct. `createDaemonNodeAdapter` prepends `requestMetaInterceptor`, yielding `[meta, bearer-deposit, peer-cred]`. Connect-ES composes left-to-right; bearer deposits `PEER_INFO_KEY` before `peerCredAuthInterceptor` reads it. Verified against `rpc/router.ts:264-281` and `auth/interceptor.ts:159-198`.

2. **Descriptor schema parity** — `DescriptorV1` (descriptor.ts:64-74) is field-for-field identical to `schemas/listener-a.schema.json` required[]. v1 contract intact.

3. **Supervisor shutdown plumbing** — `ctxRef.supervisor = result.supervisor` (index.ts:494); `shutdown.ts:351-359` calls `ctx.supervisor.stop()` as the last step. No leak on SIGTERM. Crash-pruner / capture-sources / watchdog teardown ordering in `triggerShutdown` (index.ts:453-468) is also correct (timers cancelled before drain).

4. **captureSources + crash-replay ordering** — replay runs after `runMigrations` (line 215) — correct, needs the schema. Capture-sources install runs after replay — correct, otherwise an `uncaughtException` during replay would race the file-truncate. Skip env-flag `CCSM_DAEMON_SKIP_CRASH_CAPTURE=1` is appropriate for the test mode.

5. **Reverse-verify** — Spot-checked the 6 assertions:
   - #1 schema validate via Ajv against frozen schema file → real gate (would fail on field rename).
   - #2 `daemonVersion === '0.3.0-e2e-test'` + `protoVersion` + `listenerId === 'A'` → real (Unimplemented response would not have these populated; Connect would throw).
   - #3 `Code.Unauthenticated` with no `authorization` header → real (verified via auth interceptor flow).
   - #4 `/healthz` 200 + `ready=true` → real, depends on `lifecycle.isReady()`.
   - #5/#6 probe flags surfaced through `RunStartupResult` → real (a wire-up regression flips them).
   - PR body documents removing line 169 fix → assertion #1 fails. Plausible.

6. **Side fix `db/locked.ts`** — Identical hash `f76859d5...` matches PR #906 exactly (`gh pr diff 906` confirmed). Without it, `runMigrations` fails at `OPENING_DB`, blocking the new e2e. Justified inclusion; no-op when #906 lands.

### Notes (non-blocking)

- Assertion #2 substitutes `daemonVersion`/`protoVersion`/`listenerId` for the brief's "boot_id" — well-documented in PR body (HelloResponse has no `boot_id` field; adding it requires coordinated proto bump). Reasonable scope cut.
- POSIX-side replay seed often falls into the `fileMissing` branch (EACCES on `/var/lib/ccsm`); assertion #6 documents this and still validates the code path executed. Acceptable trade-off; tightening would require T5.3-style env override.

### CI
All 5 checks green: `lint+typecheck+test` ubuntu/macos/windows + `detect-changes` + `no-silent-drops`.

### Verdict

**APPROVED.** Merging.

PR #976, opened 2026-05-03
wave0e-1(#289): cut src/stores/persist.ts from window.ccsm.* to localStorage (e2e fix)

Jiahui-Gu:
## Wave 0e-1 — first slice of renderer cutover

Wave 0b (PR #948) deleted `window.ccsm.loadState/saveState` from preload but the renderer cutover (Wave 0e) was [DEFERRED]. Result: every PR's e2e harness-ui run has been failing because the renderer keeps calling phantom IPCs. User overruled the deferral tonight — e2e must be fixed, not muted (PR #975 closed).

This PR is the **first slice**: persist.ts only. Sibling tasks for drafts.ts, xtermSingleton, usePtyAttach, UpdatesPane, sessionCrudSlice, AppearancePane, NotificationsPane are tracked separately and will land as their own PRs.

## Why localStorage and not daemon RPC

Per the rpc-stub-gap audit (`docs/superpowers/specs/2026-05-04-rpc-stub-gap-audit.md`, #228 sub-task 9), SettingsService is a stub. v0.3 transitional posture: localStorage now, daemon RPC when SettingsService is real. A comment block at the top of `persist.ts` documents the cutover plan.

## Wire-up evidence (#218 PR template)

**Before** (broken since Wave 0b PR #948):
- `loadPersisted()` calls `window.ccsm.loadState(STATE_KEY)` — undefined → throws → returns null → app boots empty every time.
- `schedulePersist()` / `flushNow()` call `window.ccsm.saveState(...)` — undefined → unhandled rejection / silent persist loss.
- harness-ui `startup-paints-before-hydrate` reports `loadStateWrapped: false` (the harness's `originalLoad = ccsm.loadState` returns undefined, so its 800ms wrap is skipped).

**After** (this PR):
- `loadPersisted()` reads `localStorage.getItem(STATE_KEY)`, parses JSON, returns the curated `PersistedState`. Function still `async` so callers (`store.ts:hydrateStore`) are untouched.
- `schedulePersist()` / `flushNow()` write `localStorage.setItem(STATE_KEY, JSON.stringify(snap))`. `flushNow()` is now genuinely synchronous (sync setItem) — the old IPC race window where beforeunload couldn't await the IPC is gone.
- `setPersistErrorHandler` still fires on setItem throws (quota exceeded, private mode).

[...]

PR #948, opened 2026-05-03
wave0b(#216): purge v0.2 ipc/preload/registrar — pure delete

Jiahui-Gu:
Task #216 [WAVE0b]. Pure DELETE of v0.2 ipc/preload/registrar/db files in top-level `electron/`. No REWIRE, no MOVE, no new files (except the `ipcGuards` → `pathGuards` rename).

Per electron/ survey **a321fb44** §1 DELETE list + decisions (Electron zero sqlite, frameless default chrome).

## Deleted (17 files)

- `electron/ipc/` — `dbIpc.ts`, `sessionIpc.ts`, `systemIpc.ts`, `utilityIpc.ts`, `windowIpc.ts` + 4 tests
- `electron/preload/` — `index.ts`, `bridges/{ccsmCore,ccsmPty,ccsmSession,ccsmNotify,ccsmSessionTitles}.ts`
- `electron/ptyHost/ipcRegistrar.ts` + its `__tests__/ipcRegistrar.test.ts`

## Renamed

- `electron/security/ipcGuards.ts` → `pathGuards.ts` (+ test). Removed `fromMainFrame()` (the v0.2 IPC defense). Kept `isSafePath` + `resolveCwd`. Wave 0d will move `pathGuards` to the daemon.
- Updated importer `electron/ptyHost/cwdResolver.ts`. The other `ipcGuards` importers were in deleted `ipc/*` files, so they vanish naturally.

## main.ts edits

- Drop `import { ipcMain }`.
- Drop every `register*Ipc(...)` call site, `installUpdaterIpc()`, `registerPtyHostIpc(...)`.
- Drop renderer-pushed mirrors (`activeSidFromRenderer`, `sessionNamesFromRenderer`).
- `createWindow.getActiveSid` → `() => null` and notify pipeline `getNameFn` → `() => null` until Wave 0c sources both from the daemon.
- **Kept**: `app.whenReady`, `acquireSingleInstanceLock`, `initSentry`, `applyAppMenuLocale`, `createWindow`, `ensureTray`, `registerLifecycleHandlers`, `badgeManager` + notify pipeline construction (cheap to keep, avoids a Wave 0c churn revert).

## ptyHost/index.ts edits

- Removed `registerPtyIpc` import + the `registerPtyHostIpc(ipcMain, getMainWindow)` exported wrapper. Wave 1 will wire pty lifecycle through the daemon's Connect-RPC.

## Deviation from task spec

[...]

PR #986, opened 2026-05-03
hotfix(#311): restore paint-gate via rAF before hydrateStore (#976 followup)

Jiahui-Gu:
## Root cause (Task #306 research)

PR #976 (Wave 0e persist.ts cutover, commit `1ed581c`) accidentally collapsed the renderer boot paint-gate.

- **Pre-#976**: `loadPersisted()` in `src/stores/persist.ts` awaited `window.ccsm.loadState(STATE_KEY)` — real async IPC, multiple event-loop turns. This gave React a paint frame after `root.render()` to commit the skeleton before hydration replaced it.
- **Post-#976**: `localStorage.getItem(STATE_KEY)` is sync inside the async wrapper, so the await resolves on the next microtask.
- Sibling `hydrateDrafts()` ALSO sync-bails at `src/stores/persist.ts:26` (`if (!window.ccsm) return;` — `window.ccsm` was nuked in Wave 0b PR #948).
- Net: `void hydrateStore()` at `src/index.tsx:68` completes within the same microtask as `root.render()`. Skeleton never paints. All 8 harness-ui cases waiting on skeleton/hydrated transitions / theme-init order / opacity transitions time out.

PR #985 probe trace evidence: `loadStateWrapped:true, hydrated:true, hasSidebarSkeleton:false`.

## Fix (Option B from research, architectural)

One-line change at `src/index.tsx`: wrap `void hydrateStore()` in `requestAnimationFrame(() => ...)`. This defers hydration past the next paint, restoring the pre-#976 paint-gate behavior.

- No `persist.ts` touch. Keeps `persist.ts` semantically truthful (sync IO described as sync).
- Locates the paint-gate at the boot wiring layer where the explanatory comment already lives.
- Comment block at `src/index.tsx:42-50` extended with 5 lines explaining why rAF is required after #976.
- Includes a `typeof requestAnimationFrame === 'function'` guard for non-DOM unit-test environments (jsdom has rAF, but defensive).

## Expected recovery (8 harness-ui e2e cases)

- settings-updates-pane
- titlebar
- theme-toggle
- import-empty-groups
- rename
- move-to-group-excludes-own-group
- sidebar-long-name-truncates
- startup-paints-before-hydrate

## Local checks

- lint: passing (0 errors, warnings unrelated)
- typecheck: passing (exit 0, both `tsc --noEmit` configs)
- build: passing (`npm run build:app` exit 0, webpack 5.106.2 + tsc -p tsconfig.electron.json)
- unit tests: 821 passing / 10 failing — **all 10 failures are `better-sqlite3` `NODE_MODULE_VERSION 145 vs 137` env mismatch**, unrelated to this change. vitest exit 0.
- e2e: **not run locally — harness-ui requires Linux+xvfb, this hotfix was developed on Windows**. Trusting CI Linux runner for the e2e signal per task instructions.


Jiahui-Gu:
## Diagnosis (Task #311 round 4)

Followed the coordinator's diagnose-not-guess directive. Checked out 1ed581c~1 and 1ed581c, diffed src/stores/persist.ts, cross-referenced with hydrateStore + harness reset path. Findings:

### PR #976 SHAPE diff: nothing changed semantically

`loadPersisted()` returns the exact same `PersistedState` shape pre/post #976:
- Same `STATE_KEY = 'main'`
- Same `JSON.parse` pipeline
- Same `version: 1` gate
- Same field set: `sessions, groups, activeId, sidebarWidth, theme, fontSize, fontSizePx`
- Same `schedulePersist` payload via `JSON.stringify(snap)`

The diff is purely the IO backend: `window.ccsm.loadState/saveState` (electron IPC → sqlite `app_state` table) → `localStorage.getItem/setItem`. Function signatures preserved. No keys dropped or restructured.

### REAL bug: e2e harness clears the OLD backend, not the NEW one

`scripts/probe-helpers/reset-between-cases.mjs` step 4 (lines 102-120) wipes the sqlite `app_state` table between cases:

```js
const db = new Database(dbPath);
db.exec('DELETE FROM messages;');
db.exec('DELETE FROM app_state;');  // <-- pre-#976 persist target
```

But post-#976, persist no longer writes to `app_state` — it writes to renderer-local `localStorage`. The reset path was never updated. So:

1. Case N runs, mutates store (groups, sessions, drafts, sidebar width, etc.)
2. `schedulePersist` debounces 250ms then writes the snapshot to `localStorage['main']`
3. `resetBetweenCases` wipes the sqlite table (no-op — empty already) and resets the in-memory store to clean baselines
4. Case N+1 launches in the SAME renderer (shared app, no relaunch) — but if there IS a relaunch, hydration reads `localStorage['main']` and re-applies stale state from case N

This explains the failure pattern in round 3:
- `import-empty-groups` "synthesized group header not rendered" → group from prior case persists, blocks the synthesized one
- `rename`, `move-to-group-excludes-own-group`, `sidebar-long-name-truncates` → 20s `seedStore` waitForFunction timeouts → store seeding races against hydration replay of stale persisted state
- `settings-updates-pane` → dialog opens with leftover tab selection from a prior case
- `theme-toggle` → leftover `theme` key in localStorage hydrated, then setTheme races
- `titlebar` → `<main>/<aside>` not appearing in 10s → likely related to leaked dialogOpen state forcing skeleton path

### Fix (round 4 commit)

Updated `scripts/probe-helpers/reset-between-cases.mjs` step 4 to ALSO clear localStorage entries for both persist keys:
- `localStorage.removeItem('main')` (src/stores/persist.ts STATE_KEY)
- `localStorage.removeItem('drafts')` (src/stores/drafts.ts STATE_KEY)

Runs in `win.evaluate()` so it executes in the renderer context where localStorage is partitioned per app session. Wrapped in try/catch — degraded environments without localStorage are non-fatal.

Kept the sqlite wipe as a defensive backstop (schema may carry residue from pre-#976 builds, and we're paying the DB open cost anyway).

Also kept the round 2 (double rAF in src/index.tsx) and round 3 (`__ccsm_hydrated` flag + harness-runner gate) changes — those are necessary independent fixes for the paint-gate and probe-race-against-hydration issues respectively. The localStorage leak was a third, unrelated regression hidden behind the first two.

Jiahui-Gu:
## Round 6 ErrorBoundary payload

Round 5 dump (commit `ec646d6`) confirmed Task #312's hypothesis: hydration completes (`hydratedFlag=true`, `storeTheme="dark"`), then App throws before mount, ErrorBoundary fallback replaces the tree (`appMounted=false`, `errorBoundaryShown=true`, `htmlClasses="theme-light"`).

Round 6 (commit `258dc79`) wires Sentry's `ErrorBoundary` `onError` into a `window.__ccsm_error` global so the harness snapshot can capture the actual throw.

### Result — all 7 e2e cases fail with the same root cause:

```
TypeError: window.ccsm?.getVersion is not a function
```

Failing cases (identical error in each): `theme-toggle`, `rename`, `move-to-group-excludes-own-group`, `import-empty-groups`, `settings-updates-pane`, `sidebar-long-name-truncates`, `titlebar`.

### Full theme-toggle dump (ubuntu-latest, run 25299805307)

```json
{
  "themeClassDark": false,
  "themeClassLight": true,
  "dataTheme": "light",
  "sidebarBg": "oklch(0.985 0.002 240)",
  "fg": "oklch(0.22 0.01 240)",
  "contrast": 0.765,
  "bgLum": 0.985,
  "hydratedFlag": true,
  "storeRef": true,
  "storeTheme": "dark",
  "persistedRaw": null,
  "appMounted": false,
  "errorBoundaryShown": true,
  "daemonModalOpen": false,
  "htmlClasses": "theme-light",
  "bodyTextHead": "Something went wrong. The error was reported to the developer.",
  "ccsmError": {
    "message": "window.ccsm?.getVersion is not a function",
    "name": "TypeError",
    "stack": "TypeError: window.ccsm?.getVersion is not a function\n    at file:///home/runner/work/ccsm/ccsm/dist/renderer/bundle.js:2:1295906\n    at nl (...:586519)\n    at wc (...:606473)\n    ...",
    "componentStack": "\n    at GA (.../bundle.js:2:1295781)\n    at div\n    at div\n    at div\n    at .../bundle.js:2:831554\n    ..."
  }
}
```

(`stack` and `componentStack` truncated for readability — full text in CI log.)

### What this means

- NOT a paint-gate / hydration race / persisted-shape issue. Rounds 1-4 fixes were addressing the wrong symptom.
- The renderer expects `window.ccsm.getVersion` to exist (likely from preload / `contextBridge.exposeInMainWorld('ccsm', {...})`) but it's either undefined or not a function in the harness BrowserWindow.
- `componentStack` top frame is `GA` — minified production bundle, but it's the FIRST component in the boot path (renders inside `RendererBoot > <div>...<div>`). Looks like a top-level header / titlebar / version-display component.

### No `seedStore` timeout dumps

Zero `[seedStore] hydration gate timeout` lines in any of the 7 failures — all cases successfully hydrate, the throw happens during the first render after hydration completes. So `seedStore`'s round-5 catch block has not yet been triggered (kept in for future regressions).

### Next round

Round 7 = the actual fix:
1. Find the source-side caller of `window.ccsm.getVersion` (grep `getVersion` in `src/`).
2. Determine whether the harness preload is missing the export, or whether the call needs a `typeof === 'function'` guard.
3. Fix the producer (preload) or the consumer (component) per spec.

Jiahui-Gu:

PR #1054, opened 2026-05-05
chore(proto): regenerate lock.json — sync 5 .proto SHA256

Jiahui-Gu:
## Why

PR #1051 (Task #427, ch15 §3 enforcement wire) CI caught that `packages/proto/lock.json` on working baseline holds **Windows-CRLF SHA256** for 5 `.proto` files, while CI Linux runners hash **LF-normalized** bytes (per `.gitattributes`: `packages/proto/**/*.proto text eol=lf`). Drift is pre-existing on working, **not** introduced by #1051.

This tiny PR regenerates `lock.json` against LF-normalized `.proto` content so the SHAs match what CI computes. **Blocks PR #1051 (Task #427) merge.**

## Affected files (SHA only)

- `src/ccsm/v1/crash.proto`
- `src/ccsm/v1/draft.proto`
- `src/ccsm/v1/notify.proto`
- `src/ccsm/v1/settings.proto`
- `src/ccsm/v1/supervisor.proto`

PR diff is **1 file** (`packages/proto/lock.json`), 5 SHA fields changed, no other fields.

## Local checks (host: windows-11)

- `pnpm --filter @ccsm/proto run lock` → `wrote ... (8 files)`
- `pnpm --filter @ccsm/proto run lock-check` → `lock-check: OK (8 .proto files match lock.json)`
- `git diff packages/proto/lock.json` → only 5 SHA value changes, all matching the "actual" hashes from PR #1051 CI failure log.

Lint / typecheck / build / unit / e2e skipped: change is metadata-only (lock.json), no `.ts/.tsx` touched. lock-check is the only relevant gate and passes.

## Test plan

- [ ] CI `lock-check` job (Linux runner) on this PR passes.
- [ ] After merge, rebase #1051 onto working — its `ch15 §3 enforcement` job goes green.

(Task #447)

PR #1056, opened 2026-05-05
fix(proto): LF-normalize .proto SHA in lock.mjs/lock-check.mjs (#450)

Jiahui-Gu:
## Summary

Both `packages/proto/scripts/lock.mjs` and `lock-check.mjs` hashed `.proto` files as raw bytes. On Windows checkouts with CRLF line endings, this produced different SHAs than Linux and caused spurious `lock.json` drift.

Task #447 / PR #1054 fixed the working baseline symptom but not the root cause — the next `.proto` edit + Windows regenerate would have re-introduced the drift.

This PR reads as UTF-8 and normalizes `\r\n` -> `\n` before hashing in both scripts (writer + checker stay in lockstep).

## Verification

- On this Windows host, `pnpm --filter @ccsm/proto run lock-check` -> `OK (8 .proto files match lock.json)` against Linux-computed `lock.json` (proves normalization is consistent across OS).
- Re-running `pnpm --filter @ccsm/proto run lock` produces zero diff to existing `lock.json`.

- lint: ok (0 errors, 66 preexisting warnings)
- typecheck: ok (exit 0)
- build / unit / e2e: skipped — change is to two `.mjs` build scripts only, no `.ts`/`.tsx` touched, no runtime impact (per dev.md §3 step 2 skip rule for non-runtime config).
- lock-check: OK (8 .proto files match lock.json)
- lock regen: zero diff

## Out of scope

- `lock.json` regeneration if any drift surfaces (separate downstream task).

Refs Task #450

### DIFF
--- a/packages/proto/scripts/lock-check.mjs
+++ b/packages/proto/scripts/lock-check.mjs
@@ -39,7 +39,10 @@ function toPosix(p) {
 }
 
 function sha256OfFile(path) {
-  return createHash('sha256').update(readFileSync(path)).digest('hex');
+  // Mirror lock.mjs: UTF-8 read + CRLF -> LF normalization so Windows and
+  // Linux compute identical hashes regardless of git autocrlf behavior.
+  const content = readFileSync(path, 'utf8').replace(/\r\n/g, '\n');
+  return createHash('sha256').update(content, 'utf8').digest('hex');
 }
 
 function main() {
--- a/packages/proto/scripts/lock.mjs
+++ b/packages/proto/scripts/lock.mjs
@@ -4,10 +4,11 @@
 //
 // Regenerate packages/proto/lock.json from packages/proto/src/**/*.proto.
 //
-// Walks the src tree, computes SHA256 of each .proto file (raw bytes — no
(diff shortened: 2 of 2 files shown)


Jiahui-Gu:
LGTM — Layer 1 right, Layer 2 clean.

Verified on Windows host (pool-2):
- `pnpm --filter @ccsm/proto run lock-check` -> `OK (8 .proto files match lock.json)`
- `pnpm --filter @ccsm/proto run lock` -> zero diff against existing `lock.json`

Reverse-verify: `.gitattributes` enforces `*.proto text eol=lf`, so on this host the proto files are physically LF and dropping the `.replace(/\r\n/g, '\n')` did not flip lock-check red. Confirmed the normalize line still bites via synthetic CRLF probe:
- LF      sha: `7c2e34...906d4`
- CRLF raw: `6c899d...35abe` (mismatch — what the bug looks like on a host without .gitattributes honored)
- CRLF + normalize: `7c2e34...906d4` (matches LF — fix is correct)

Diff is exactly 2 files (writer + checker stay in lockstep), comments explain the why. CI on `working` has no required checks — merging.

PR #451, opened 2026-04-28
chore(cleanup): delete dead notify subsystem (~900 LOC orphan)

Jiahui-Gu:
## Summary
Per audit `a69b7a3a`, the entire `electron/notify*` pipeline was orphaned: zero `src/` callers, zero `ccsm-notify` package dep (the package was removed long ago), and under the new ttyd architecture the event source has shifted enough that a fresh design will be cleaner than retro-fitting the old toast pipeline.

Final pre-commit grep confirmed no production caller for `showNotification` / `bootstrapNotify` / `dispatchNotification` / `notification:show` / `window.ccsm.notify(`.

### Files removed (5)
- `electron/notifications.ts` (195 LOC)
- `electron/notify.ts` (218 LOC)
- `electron/notify-bootstrap.ts` (314 LOC)
- `electron/__tests__/notifications.test.ts` (9 cases)
- `electron/__tests__/notify-bootstrap.test.ts` (3 cases)

### Wiring removed
- `electron/main.ts`
  - `showNotification` / `bootstrapNotify` / `createDefaultToastActionRouter` / `autoSetupAumid` imports
  - `bootstrapNotify(...)` call in `app.whenReady()`
  - `autoSetupAumid()` (the dev `.lnk` auto-installer for toast routing — moot without toasts)
  - `ipcMain.handle('notification:show', ...)`
  - the dev-only `__ccsmDebug = { notify, notifyBootstrap }` probe seam (the only members were the dead modules; whole `if (!app.isPackaged)` block dropped)
- `electron/preload.ts`: `window.ccsm.notify` method
- `src/global.d.ts`: `notify` type declaration

### Kept intact
- `app.setAppUserModelId('com.ccsm.app')` — taskbar identity, decoupled from toasts
- `scripts/setup-aumid.ps1` — standalone AUMID tool, useful for any future Win32 surface

### Tombstone comment cleanup (no functional change)
- `src/components/SettingsDialog.tsx` (lines 23-28, 287-290) — comments said "toggles removed because no one emits"; now that the receiver is also gone, those comments are stale
- `src/stores/store.ts` (line 71) — same
- `src/App.tsx` (lines 39-42, 127) — same
- `scripts/harness-ui.mjs` (1186-1190, 1262-1264, 1289-1292) — same
- `src/global.d.ts` i18n comment — softened "OS notifications come out in the matching language" since there are no OS notifications anymore

## Net change
**-1384 LOC** (`12 files changed, 17 insertions(+), 1401 deletions(-)`)

[...]

PR #1057, opened 2026-05-05
fix(daemon): plumb CCSM_STATE_DIR through to db + crashRaw paths (Task #446)

Jiahui-Gu:
## Summary
- index.ts:196 called `statePaths()` with no args, which only honours `PROGRAMDATA` (paths.spec.ts FORBIDDEN env keys per spec ch07 §2). `CCSM_STATE_DIR` (resolved by `buildDaemonEnv` into `env.paths.stateDir`) was silently dropped — db, recovery, crash-raw replay, and crash-capture sinks all wrote under the per-OS default.
- Adds `statePathsFromRoot(root, platform?)` to `state-dir/paths.ts` (pure, no env reads — paths.spec FROZEN contract preserved).
- index.ts now calls `statePathsFromRoot(env.paths.stateDir)` so the env-resolved root flows through to all downstream paths.

## Test plan
- [x] Adds linux + win32 env-consistency specs proving `CCSM_STATE_DIR` flows to `sp.db` / `sp.crashRaw` / `sp.descriptor`.
- [x] Negative assertion: bare `statePaths()` still ignores `CCSM_STATE_DIR` (proves bypass surface lives in the call site, not in paths.ts — paths.spec.ts FROZEN contract intact).

- lint: exit 0 (warnings only, no errors — preexisting)
- typecheck: exit 0
- state-dir specs: 23 passed (includes 2 new Task #446 cases)
- baseline note: `test/descriptor/schema.spec.ts:187` golden mismatch is pre-existing on origin/working (verified by checking out clean), unrelated to this PR

## Scope
Followup to dogfood phase 1.5 finding. Surgical 3-file change: paths.ts (add new export, no behaviour change to existing `statePaths`), index.ts (one call-site swap), env-consistency.spec.ts (regression cases).

### DIFF
--- a/packages/daemon/src/index.ts
+++ b/packages/daemon/src/index.ts
@@ -59,7 +59,7 @@ import { makeRouterBindHook } from './rpc/bind.js';
 import { upsertSettingsBoot } from './rpc/settings/store.js';
 import { SessionManager, type ISessionManager } from './sessions/SessionManager.js';
 import { makeProductionAttachPtyHost } from './pty-host/attach-on-create.js';
-import { statePaths } from './state-dir/paths.js';
+import { statePathsFromRoot } from './state-dir/paths.js';
 import {
   Shutdown,
   installShutdownHandlers,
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
Reviewer LGTM (cannot self-approve via API).

**Direction: right** — surgical fix for #446 silent-drop. `statePathsFromRoot(root, platform)` is pure (no env reads, root injected), `statePaths()` delegates to it preserving paths.spec.ts FROZEN contract. `index.ts:200` swap rebases db / crashRaw / descriptor onto `env.paths.stateDir` (single source of truth in env.ts). 3 files only.

**Layer 2 findings:**
- diff: exactly 3 files (paths.ts +helper, index.ts 1-line swap, env-consistency.spec.ts +2 cases). No stray inclusions despite pre-#451 stash incident.
- pure-function check: `statePathsFromRoot` no `process.env` reads, takes `root` arg, `path.win32.join` / `path.posix.join` for separator correctness.
- backward compat: paths.spec.ts FROZEN golden untouched (16 passed locally).
- new specs (linux + win32): assert db/crashRaw/descriptor flow under `CCSM_STATE_DIR`, plus negative bare `statePaths()` still ignores env.
- **reverse-verify (reviewer-side)**: temporarily broke `statePathsFromRoot` to ignore `root` arg -> both new cases FAIL with `expected '/tmp/test1' to be '/var/lib/ccsm'`. Restored, 23/23 pass. Tests bite.
- CI: all green incl. `lint + typecheck + test (windows-latest)` 6m40s.
- baseline note on `descriptor/schema.spec.ts:187`: ran origin/working locally -> 8/8 pass; either already fixed upstream or local-flake on dev side. Not blocking, unrelated to this PR's diff.

Squash-merging.

PR #1066, opened 2026-05-05
fix(daemon/db): inline migrations for SEA bundle (P0 SHIP-BLOCKER, Task #463)

Jiahui-Gu:
## Summary

P0 ship blocker. Daemon SEA exe crashed at OPENING_DB before any listener bound. dev #457 round-3 traced the crash to `dist/bundle.cjs:6961` — esbuild rewrites `import.meta.url` to a `{}` stub when bundling to CJS for the SEA carrier, so `fileURLToPath(undefined)` threw the moment the bundled `runner.ts` tried to compute `MIGRATIONS_DIR`.

Manager-selected option A: inline every `*.sql` payload at build time.

- New `prebuild` hook (`build/inline-migrations.mjs`) reads `*.sql` files and emits `src/db/migrations/inlined.ts` (gitignored, like `gen/ts/`). Each entry is a base64-encoded literal that decodes to the exact on-disk bytes (preserves the trailing newline that the FOREVER-STABLE SHA-256 was computed over).
- `runner.ts` now reads SQL from `MIGRATION_SQL[filename]` instead of `readFileSync`. Zero filesystem dependency — works in dev (tsc), tests (vitest), and SEA. SHA-256 verification is preserved exactly: the runner hashes the inlined bytes and compares against `MIGRATION_LOCKS[].sha256` (still the source of truth in `src/db/locked.ts`, FOREVER-STABLE per ch07 §4 + ch15 §3 #4 — the `tools/check-migration-locks.sh` grep regex over `locked.ts` is untouched).
- New `MissingInlinedMigrationError` (acceptance criterion #6) for the case where `MIGRATION_LOCKS` references a filename absent from the inlined bundle.

Also fixed a second `import.meta.url` boot-path crash: `index.ts:isDirectRun` used to throw under SEA (catch returned false → `main()` was silently never invoked). It now short-circuits to `true` under `isSea()` before touching `import.meta.url`.

Regression gate: `test/integration/daemon-sea-boot.spec.ts` greps the bundled `dist/bundle.cjs` for any `fileURLToPath(import_meta_N.url)` call OUTSIDE the two known-safe call sites (post-READY pty-host spawn + the now-`isSea()`-gated `isDirectRun`). Bundle is produced by a `pretest` helper (`build/bundle-for-sea-spec.mjs`) that mirrors `build-sea.{sh,ps1}`'s tsc + esbuild steps, so the spec asserts the exact artifact CI / release ships.

## dev #457 round-3 background

dev #457 round-3 (the round that produced this task) traced the crash to:
- bundle: `dist/bundle.cjs:6961` — `fileURLToPath(import_meta.url)`
- source: `packages/daemon/src/db/migrations/runner.ts:32` — `dirname(fileURLToPath(import.meta.url))`
- root cause: esbuild rewrites `import.meta` → literal `{}` when `--format=cjs`, so `import_meta.url` is `undefined`.

## Local self-verify (host: windows-11)

Built SEA exe locally and ran it once before/after the fix to confirm the migration crash actually disappears. Output below.

### After fix — daemon advances PAST migrations

```
$ ./dist/ccsm-daemon.exe
[ccsm-daemon] phase -> LOADING_CONFIG
[ccsm-daemon] env loaded: mode=dev bootId=2bc2b46b-... version=0.3.0-dev
[ccsm-daemon] phase -> OPENING_DB
[ccsm-daemon] STARTUP FAILED at phase OPENING_DB: Database2 is not a constructor
```

Daemon now reaches OPENING_DB phase — proves migrations runner is no longer the blocker.

### Reverse-verify (unit-level, equivalent to acceptance #4 + #6)

```
# revert runner.ts → run runner.spec.ts
$ git checkout -- src/db/migrations/runner.ts
$ pnpm exec vitest run src/db/migrations/__tests__/runner.spec.ts
 Test Files  1 failed (1)
      Tests  3 failed | 5 passed (8)
# Tests covering MissingInlinedMigrationError, migrationSql helper, and the
# inline-bytes SHA equality assertion fail because the old runner does not
# export MissingInlinedMigrationError / migrationSql.

# reapply fix → tests green
$ git apply /tmp/task-463-runner.patch
$ pnpm exec vitest run src/db/migrations/__tests__/runner.spec.ts
 Test Files  1 passed (1)
      Tests  8 passed (8)
```

The MissingInlinedMigrationError + MigrationLockMismatchError unit tests are the acceptance #4 / #6 equivalents — running them at the SEA-exe level requires modifying `inlined.ts` and rebuilding the bundle, which only repeats the unit-level assertion through more layers.

## Out-of-scope (next P0 in the SEA path)

The daemon now reaches OPENING_DB but immediately throws `Database2 is not a constructor`. This is a SEPARATE SEA bug in `native-loader.ts` (Task #83 surface): the SEA path `loadNative('better_sqlite3')` does `require('./better_sqlite3.node')` directly, returning the raw napi binding. The published `better-sqlite3` package wraps the binding in `lib/index.js` (`module.exports = require('./database')`) — that JS wrapper never runs in SEA. Needs a follow-up task to either bundle the wrapper into the SEA blob or stage it next to `native/`.

This PR cleanly resolves the migrations-runner P0 (Acceptance #1 — daemon does not crash on the migration code path; Acceptance #4 / #6 — SHA mismatch + missing payload detection). Remaining acceptance items (#2 listener-a bound, #3 spec passes in CI) need the native-loader follow-up to land first.

- daemon lint: passes (0 errors, 66 pre-existing warnings)
- daemon typecheck: passes
- daemon build: passes
- daemon unit tests: 1560 passed | 7 pre-existing failed (baseline-equal — same 7 failures on plain `origin/working` HEAD: 3x integration `*-wired.spec.ts` Code.Unimplemented timeouts, 3x `rpc/__tests__/integration.spec.ts` Hook timed out, 1x `descriptor/schema.spec.ts` golden mismatch). My new specs (`daemon-sea-boot.spec.ts` 2 tests + `runner.spec.ts` 8 tests + `locked.spec.ts` 5 tests) all pass.
- root `npm run lint`: passes (turbo, 3 successful)
- root `npm run typecheck`: passes
- root `npm run build`: passes (turbo, 3 successful)
- root `npm test`: 1 pre-existing failure (`@ccsm/proto:test` lock.json mismatch — unrelated to this PR; same fail on `origin/working` HEAD)
- `npm run probe:e2e`: SKIPPED — e2e is electron-app surface; daemon is not yet wired into the electron app in v0.3 (RPC-only contact would happen later). The change set is daemon-internal SQLite migration plumbing + SEA bundle gate; no electron-side ripple. CI runs e2e as remote verification.

## Test plan

- [x] daemon migration runner unit tests pass
- [x] new `daemon-sea-boot.spec.ts` regression gate detects future `import.meta.url`-on-boot regressions
- [x] reverse-verify: revert `runner.ts` → 3 tests fail; reapply → 8 pass
- [x] daemon SEA exe locally advances past migration crash to OPENING_DB


Jiahui-Gu:
Reviewer (Task #479) — APPROVED (post-comment, GH blocks self-approve so noting verdict here).

**Direction: right.** P0 ship-blocker, manager-selected option A. Scope 限 daemon migrations runner + isSea() boot gate; native-loader followup = Task #480 (not this PR).

**Layer 2 7-item:**
- inline-migrations.mjs idempotent — locally re-ran 2x, second prints `up-to-date`, no write.
- isSea() short-circuit at index.ts:784-789 runs BEFORE fileURLToPath(import.meta.url).
- bundle.cjs static grep (daemon-sea-boot.spec.ts) allowlists only defaultChildEntrypoint (post-READY) + isDirectRun (isSea-gated); bundle rebuilt via pretest hook = same artifact CI ships.
- SHA-256 verify preserved (runner.ts:185-189 hashes inlined bytes vs MIGRATION_LOCKS[].sha256); FOREVER-STABLE locked.ts untouched.
- MissingInlinedMigrationError (acceptance #6): runner.ts:90-100 + spec coverage.
- inlined.ts gitignored (.gitignore:6).
- Reverse-verify PR-body段 quote 改前 3 fail / 改后 8 pass — probe bites.

Local-checks 段完整, baseline-equal failures 对齐 origin/working HEAD; probe:e2e SKIPPED 理由合理 (daemon 未 wired 入 electron app)。

Squash-merging now.

Jiahui-Gu:
Round-2 rebase complete. Resolved conflict with PR #1062 (proto-gen prebuild hook) by chaining both prebuild + pretypecheck scripts in `packages/daemon/package.json`:

- `prebuild`: `node ../../scripts/gen-proto.mjs && node build/inline-migrations.mjs`
- `pretypecheck`: same chain (typecheck needs both `gen/ts/` and `src/db/migrations/inlined.ts` since runner.ts imports the latter)

Order: gen-proto first (no dependency between the two; matches existing PR #1062 ordering for consistency).

New head: `ff3759d` (was `d0fa157`).

- lint: exit 0 (66 warnings, all pre-existing)
- typecheck: exit 0
- build: exit 0 — prebuild chain executed both scripts in order
- targeted tests (`test/integration/daemon-sea-boot.spec.ts` + `src/db/migrations`): 10 passed

(Full `npm test` shows 8 pre-existing flake/Windows-CRLF failures unrelated to this fix — same set red on `working` HEAD; round-1 already had reviewer APPROVED with the same baseline.)

## Reverse-verify (rebase did not break round-1 fix)
1. `rm src/db/migrations/inlined.ts` → `npm run build` → inlined.ts auto-regenerated by chained prebuild ✓
2. `rm src/db/migrations/inlined.ts && rm -rf packages/proto/gen/ts/` → `npm run build` → both prebuild steps ran, both artifacts regenerated ✓
3. SEA crash reproduction was verified in round-1 PR body (round-2 rebase only touched script chaining, not the runner.ts inlined-bytes resolution).

Re-requesting review.

Jiahui-Gu:
Round-2 review: REQUEST_CHANGES (CI block).

**CI fail**: `lint + typecheck + test (windows-latest)` exited 1 with **12 lint errors**, all from the two newly-added Node scripts:

```
packages/daemon/build/bundle-for-sea-spec.mjs
  64:5, 67:3, 99:5, 102:3   error  'console' is not defined  no-undef
  72:23, 105:5              error  'process' is not defined  no-undef

packages/daemon/build/inline-migrations.mjs
  49:5, 58:5, 131:7, 138:3  error  'console' is not defined  no-undef
  50:5, 59:5                error  'process' is not defined  no-undef
```

Root cause: root `npm run lint:app` runs `eslint . --ext .ts,.tsx` from repo root, but the flat config picks up `packages/daemon/build/*.mjs` and the env lacks `globals.node`. The daemon-package `lint` script greps `src/**/*.ts + build/**/*.ts` (no `.mjs`), so your local `pnpm --filter @ccsm/daemon lint` was clean — that's why the PR body says 0 errors. CI's root lint catches it.

Pick whichever fix is cleanest:
1. Add `/* eslint-env node */` header to both `.mjs` files, or
2. Add an override in the root flat config for `packages/*/build/*.mjs` -> `languageOptions.globals: { ...globals.node }`.

(1) is local + minimal blast radius — recommended.

Round-1 7-point checklist still stands; this is purely the new-files-not-linted-as-Node-env regression.

Direction: right (round-2 just rebases and chains prebuild/pretypecheck — `gen-proto && inline-migrations` ordering is fine, the two scripts don't share state). Layer-2 finding: lint fail is a real CI-blocking regression introduced by this PR, must be fixed before merge.

Push fix → reply here, I'll re-watch CI and merge once the windows job goes green.

PR #1072, opened 2026-05-05
fix: daemon SEA exe Database constructor wrapper (T480)

Jiahui-Gu:
Task #480 — P0-SHIP-BLOCKER follow-up to Task #463 (PR #1066).

## Problem

After PR #1066 inlined SQL migrations and unblocked the daemon SEA exe past `LOADING_CONFIG`, the daemon advanced to `OPENING_DB` but crashed with:

```
TypeError: Database2 is not a constructor
```

Root cause (per dev #463 round-3 / round-4 reports): `--external:better-sqlite3` in the esbuild invocation left only the raw napi `.node` addon in `bundle.cjs`. That addon exposes `addon.Database` (a constructor on the addon object) but is **not itself** a top-level constructor — the user-facing `Database` constructor lives in better-sqlite3's JS wrapper (`lib/database.js`), which never made it into the SEA bundle. `new Database(path)` therefore exploded.

## Spec choice — A (bundle the JS wrapper)

A: bundle/inline `lib/index.js` wrapper into bundle.cjs (analogous to the #463 inline-migrations approach) — chosen.
B: tiny adapter in native-loader that reimplements the wrapper API around the raw napi addon — rejected.

Why A over B: the wrapper is ~400 lines of public API surface (`prepare`, `transaction`, `pragma`, `backup`, `serialize`, `function`, `aggregate`, `table`, `loadExtension`, `exec`, `close`, `defaultSafeIntegers`, `unsafeMode`, `SqliteError`, plus an 8-arg `addon.Database(filename, ..., buffer)` splat). Reimplementing it inline would be 400 lines of upstream API duplication that breaks on every better-sqlite3 patch. Bundling the upstream wrapper costs +28kb in `bundle.cjs` (470kb → 498kb) and zero ongoing maintenance — clearly the right side of dev.md §1 step 2 "don't reinvent" gradient (tier 4: "抄开源").

## Implementation

- **`packages/daemon/build/bundle-for-sea-spec.mjs` + `build/build-sea.{sh,ps1}`** — drop `--external:better-sqlite3`; only `bindings` + `node-pty` + `*.node` stay external. The wrapper is pure JS so esbuild bundles its 12 source files into `bundle.cjs`.
- **`packages/daemon/src/native-loader.ts`** — add a value `import * as betterSqlite3Module from 'better-sqlite3'` (esbuild needs a static import to follow; `req('better-sqlite3')` would not work in SEA because `createRequire` cannot resolve npm packages once postjected). In the SEA branch of `loadNative('better_sqlite3')`, force-inject the SEA-loaded raw napi addon as the wrapper's `nativeBinding` option on every construction, bypassing the wrapper's `bindings('better_sqlite3.node')` fallback (which fails in SEA — the `bindings` package walks the filesystem for `build/Release/`). Preserve `Database.SqliteError` static + prototype chain so `instanceof Database` keeps matching.
- **dev/test mode is unchanged** — the new branch is gated on `isSea() && name === 'better_sqlite3'`; non-SEA callers go through the original `req(DEV_PACKAGE_NAME[name])` path.

`node-pty` has the same wrapper/raw-addon split and will need the same treatment when SEA boot reaches its phase. Per Task #480 spec, that is intentionally deferred to a separate task.

## Files

- `packages/daemon/src/native-loader.ts` — value-import + SEA branch with `nativeBinding` injection.
- `packages/daemon/build/build-sea.sh` — `--external:better-sqlite3` → `--external:bindings`.
- `packages/daemon/build/build-sea.ps1` — same.
- `packages/daemon/build/bundle-for-sea-spec.mjs` — same (so the test bundle matches the SEA build artifact).
- `packages/daemon/test/integration/daemon-sea-boot.spec.ts` — new case asserting (1) the wrapper is bundled (`Misspelled option "readOnly" should be "readonly"` — wrapper-only string), (2) `nativeBinding:` force-injection compiled into bundle, (3) raw `./better_sqlite3.node` literal preserved for runtime resolution from `<install-dir>/native/`.

- lint: ✓ (exit 0, 67 pre-existing warnings, 0 errors)
- typecheck: ✓ (exit 0, both `tsconfig.json` + `tsconfig.test.json`)
- build: ✓ (exit 0, `turbo run build` — all 3 packages)
- daemon unit + integration tests: `Tests  6 failed | 1727 passed | 20 skipped | 5 todo (1758)`

The 6 failures are **pre-existing** on `origin/working` and unrelated to #480. Verified by `git diff > /tmp/t480.patch && git checkout -- . && pnpm --filter @ccsm/daemon test` before re-applying the patch:

```
baseline:  Tests  6 failed | 1726 passed
this PR:   Tests  6 failed | 1727 passed   (+1 = the new daemon-sea-boot case)
```

The 6 failing specs (`crash-getlog-wired`, `crash-watch-wired`, `watch-sessions-wired`, and 3 cases in `rpc/__tests__/integration.spec.ts`) all fail with the same `Hook timed out in 10000ms` / Listener-A binding symptom and are independent of the native-loader change.

## Reverse-verify

```
$ sed -i "s/'--external:bindings'/'--external:better-sqlite3'/" build/bundle-for-sea-spec.mjs
$ rm dist/bundle.cjs && node build/bundle-for-sea-spec.mjs && npx vitest run test/integration/daemon-sea-boot.spec.ts
...
FAIL  test/integration/daemon-sea-boot.spec.ts > better-sqlite3 JS wrapper is inlined so SEA boot reaches READY (Task #480)
   180|     ).toContain('Misspelled option "readOnly" should be "readonly"');
       |       ^

Tests  1 failed | 2 passed (3)
```

```
$ <restore fix> && rm dist/bundle.cjs && node build/bundle-for-sea-spec.mjs && npx vitest run test/integration/daemon-sea-boot.spec.ts
...
Test Files  1 passed (1)
     Tests  3 passed (3)
```

## Test plan

- [x] daemon unit + integration tests
- [x] new `daemon-sea-boot.spec.ts` wrapper-bundled case
- [x] reverse-verify (drop fix → spec FAILs; restore → PASSes)
- [ ] CI matrix (ubuntu/macos/windows) — pending

### DIFF
--- a/packages/daemon/build/build-sea.ps1
+++ b/packages/daemon/build/build-sea.ps1
@@ -31,7 +31,8 @@ try {
     throw 'dist/index.js missing after tsc'
   }
 
-  # Step 2: esbuild bundle.
+  # Step 2: esbuild bundle. Task #480: better-sqlite3 wrapper bundled in
+  # (only `bindings` external, unreachable in SEA — see build-sea.sh).
   Step 'esbuild bundle'
   npx --yes esbuild dist/index.js `
     --bundle `
@@ -39,7 +40,7 @@ try {
     --target=node22 `
     --format=cjs `
     --outfile=dist/bundle.cjs `
-    --external:better-sqlite3 `
+    --external:bindings `
     --external:node-pty `
     --external:*.node
   if ($LASTEXITCODE -ne 0) { throw "esbuild exited $LASTEXITCODE" }
--- a/packages/daemon/build/build-sea.sh
+++ b/packages/daemon/build/build-sea.sh
@@ -57,6 +57,10 @@ fi
 
 # Step 2: esbuild bundle. `--platform=node` + `--format=cjs` so sea's main
 # can `require()` it. Native modules (.node) are external (per T7.2 plan).
+# Task #480: better-sqlite3 wrapper (pure JS) is BUNDLED in (no longer
+# external) so the SEA daemon has a real Database constructor; only the
+# `bindings` shim is external (unreachable in SEA mode — native-loader
+# force-injects nativeBinding so wrapper never calls bindings()).
 echo "[build-sea] esbuild bundle"
 ( cd "$PKG_DIR" && npx --yes esbuild dist/index.js \
   --bundle \
@@ -64,7 +68,7 @@ echo "[build-sea] esbuild bundle"
   --target=node22 \
   --format=cjs \
   --outfile=dist/bundle.cjs \
-  --external:better-sqlite3 \
+  --external:bindings \
   --external:node-pty \
   --external:*.node )
 
--- a/packages/daemon/build/bundle-for-sea-spec.mjs
+++ b/packages/daemon/build/bundle-for-sea-spec.mjs
@@ -112,7 +112,14 @@ function bundleIfStale() {
       '--target=node22',
       '--format=cjs',
       '--outfile=dist/bundle.cjs',
-      '--external:better-sqlite3',
+      // Task #480 — better-sqlite3 wrapper (lib/index.js → lib/database.js
+      // + methods/*) is now bundled INTO bundle.cjs so the SEA daemon has
+      // a real `Database` constructor at runtime. The native-loader (SEA
+      // mode) calls `require('better-sqlite3')` and force-injects the raw
+      // .node addon as `nativeBinding`, so the wrapper's `bindings()`
+      // fallback never runs. Keep `bindings` external (esbuild emits a
+      // `require("bindings")` stub that is unreachable in SEA mode).
+      '--external:bindings',
       '--external:node-pty',
       '--external:*.node',
     ],
--- a/packages/daemon/src/native-loader.ts
+++ b/packages/daemon/src/native-loader.ts
@@ -27,6 +27,18 @@ import * as path from 'node:path';
 import { isSea } from 'node:sea';
 
 import type BetterSqlite3 from 'better-sqlite3';
+// Task #480 — value import of the better-sqlite3 JS wrapper. esbuild bundles
+// this into `dist/bundle.cjs` for the SEA carrier, so the SEA daemon has a
+// real `Database` constructor at runtime (the napi `.node` only exposes a
+// raw `addon.Database` — without the wrapper `new Database(path)` throws
+// `Database2 is not a constructor`, dev report from PR #1066). The wrapper
+// itself does NOT load the .node addon at module-eval time — it lazily
+// resolves it inside `Database()` only when `nativeBinding == null`. We
+// always pass `nativeBinding` in SEA mode (see `loadNative`), so the
+// wrapper's `bindings('better_sqlite3.node')` fallback (which fails inside
+// SEA — `bindings` walks the FS for `build/Release/`) is never executed.
+// In dev/test the same wrapper resolves the .node via `bindings` normally.
+import * as betterSqlite3Module from 'better-sqlite3';
 // `node-pty` types are not a hard dep yet (T4.2+). Keep the type-side soft so
 // this module compiles before that lands.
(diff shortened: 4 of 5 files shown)


Jiahui-Gu:
REQUEST_CHANGES — silent test deletion outside PR scope

## Direction: better-approach-X (split scope)

The better-sqlite3 wrapper bundling (native-loader.ts SEA branch + esbuild externals + new daemon-sea-boot.spec.ts case) is sound and matches Spec choice A. The SEA branch is correctly gated (`isSea() && name === 'better_sqlite3'`), preserves `Database.SqliteError` + prototype chain, and the static-analysis assertions in the new spec case are reasonable given the dev's reverse-verify. CI sea-smoke (ubuntu/macos) is green, which is the acceptance signal #480 needs.

## Blocker

This PR also deletes `packages/daemon/src/__tests__/index.spec.ts` (-602 lines, 14 `it()` cases — full file). That file:

- Was introduced **yesterday** in PR #1070 (Task #471) as the co-located coverage spec for `runStartup` + `bearerToPeerInfoInterceptor` so `src/index.ts` is included in `vitest.config.coverage.ts`'s numerator (chapter 12 §6 invariant).
- Is **not** mentioned in this PR's body (`## Files` lists only native-loader / build-sea / bundle-for-sea-spec / daemon-sea-boot.spec).
- Is **not** part of Spec choice A (bundle wrapper into `bundle.cjs`).
- Removes coverage for `runStartup` wired-up boot, skip-env paths, error/recovery paths, and bearerToPeerInfoInterceptor regex branches.

The PR body's "6 pre-existing fail" claim does not square with the math either:
- baseline working = 1726 passed
- this PR = 1727 passed (+1 sea-boot case)
- but 14 it() cases were deleted → expected delta is 1726 - 14 + 1 = 1713

Either (a) the deleted file's tests were all failing on working and dev silently dropped them instead of fixing, or (b) the baseline measurement was taken with the file already removed locally. Both are §0 共识铁律 violations ("禁止跳过测试" includes deleting a passing/needed spec without PR-body justification).

## Action

1. Restore `packages/daemon/src/__tests__/index.spec.ts` (revert that delete from this PR).
2. If the file genuinely interferes with the SEA bundle work, file a separate task explaining why and link it; do NOT remove coverage for `src/index.ts` in this PR.
3. If the file's tests are genuinely failing on working independent of this PR, fix them in a separate PR or document the failure mode in PR body's `## Local checks` so it is not silently absorbed.

Re-push after restore + confirm `daemon test` count math reconciles, and I'll re-review.

Jiahui-Gu:
Reviewer #496 自纠 — 上一条 REQUEST_CHANGES 是 false-positive。

reviewer 在 pool-13 上 git checkout 到 task-480 branch 时未 fetch 最新 origin/working (该 pool 之前是 #1066 时代的 detached HEAD), 所以 `git diff origin/working..HEAD` 显示出错位的 base, 把 #1070 (Task #471) 已 merged 的 `packages/daemon/src/__tests__/index.spec.ts` 当成本 PR 的删除。

实际 PR diff 只有 5 文件:
- packages/daemon/build/build-sea.ps1 (+3 -2)
- packages/daemon/build/build-sea.sh (+5 -1)
- packages/daemon/build/bundle-for-sea-spec.mjs (+8 -1)
- packages/daemon/src/native-loader.ts (+87 -2)
- packages/daemon/test/integration/daemon-sea-boot.spec.ts (+48)

Dev 报告的 baseline 1726 → 1727 (+1) 数学正确, reviewer 推 "应该 1713" 是基于错位 diff。

CI 全绿 (10/10 SUCCESS as of 14:32 UTC), 等 windows lint+typecheck+test (in_progress) 收尾后正常 merge。

Direction: right (Spec choice A 正确), Layer 2 SEA boot spec 合理。**APPROVE — pending CI green**.

PR #1090, opened 2026-05-05
ship-gate(#464): CheckClaudeAvailable RPC real impl + renderer wire

Jiahui-Gu:
## Summary
- Daemon: real `PtyService.CheckClaudeAvailable` handler that reuses the existing `claudeResolver` (spawnSync `where`/`which` probe) via an injected `ResolveClaudeFn` dep. Wired into `registerPtyService` alongside Attach+AckPty (single combined registration). Production binding in `index.ts` uses `force:true` so each RPC re-probes the OS — closes the v0.3 ship-gate where the stub `Code.Unimplemented` stranded users on `<ClaudeMissingGuide>` even when claude IS on PATH.
- Renderer: `src/App.tsx` + `src/components/ClaudeMissingGuide.tsx` cut over from `window.ccsmPty.checkClaudeAvailable` (legacy preload shim — v0.4 deletion target) to `PtyService.CheckClaudeAvailable` over Connect RPC. New `useOptionalClients()` in `queries.ts` is the soft variant of `useClients()` for the renderer-boot race window; 3 s daemon-unreachable budget falls back to `available=false` so the guide+Re-check surface instead of an indefinite spinner.
- Spec: `packages/daemon/test/integration/check-claude-available.spec.ts` — 3 wire legs (path → `available=true`, null → `errorCode=ENOENT`, cache-bypass wiring honored end-to-end). Cache primitive itself has unit coverage in `ptyHost/__tests__/claudeResolver.spec.ts`.

Direction A per manager (zero-rework: no preload shim added; B was a v0.4 throwaway). Renderer changes confined to the `useCheckClaudeAvailable` / `window.ccsmPty` window — does NOT cutover the rest of `window.ccsm*` (#519's scope).

## Acceptance (from task spec)
- [x] daemon handler returns real boolean (mock claude → true; null → false)
- [x] `ClaudeMissingGuide` renders/hides per RPC result (boot probe + Re-check button both go through Connect)
- [x] spec covers both legs + cache-bypass wiring (daemon-restart model)
- [x] dogfood first scenario unblocked: e2e `terminal-pane-mounted` PASS via the 3 s fallback path; on a real daemon with claude on PATH the response is `available=true` and `<TerminalPane>` mounts directly

- lint: PASS (exit 0; pre-existing warnings only)
- typecheck: PASS (exit 0)
- build: PASS (exit 0)
- unit tests: **1741 passed**, 8 pre-existing failures (descriptor schema CRLF mismatch on Windows + 4 `*-wired` `Hook timed out` in `afterEach` + 3 router `__tests__/integration` `Hook timed out`). Reverse-verified on baseline `origin/working` via patch-file flow (per dev.md §2): same 8 failures reproduce, unrelated to #464.
- new spec: 3/3 PASS, 5.93s
- e2e (full):
  - `harness-ui`: 14/14 PASS (including `terminal-pane-mounted`)
  - `harness-dnd`: PASS
  - `harness-real-cli`: pre-existing timeout (baseline-reproduced; exercises real claude binary, unrelated)

## Test plan
- [x] Unit: handler decider table (path / null / empty / Windows .cmd shim)
- [x] Wire: in-process Connect harness, 3 legs (incl. cache-bypass)
- [x] Renderer: `terminal-pane-mounted` e2e exercises the boot probe path
- [ ] Manual smoke on macOS / Linux — out of scope for the Windows dev machine, CI matrix covers

## Notes
- Hot-file conflict with #519 (Wave 0e-5): I only touch the `useCheckClaudeAvailable`/`window.ccsmPty` block in `App.tsx` + the re-check button in `ClaudeMissingGuide.tsx`. #519's `window.ccsm*` mass cutover lands after this per manager's blockedBy ordering.
- `useOptionalClients()` is purely additive (does not touch `useClients()` strict semantics) — opens a documented seam for any future component that MUST render during the `<ClientsProvider>`-not-yet-mounted window.


Jiahui-Gu:

PR #567, opened 2026-04-29
refactor(main): extract createWindow + tray (Task #731 Phase A2)

Jiahui-Gu:
## Summary
Continue the SRP slim-down of `electron/main.ts` that #561 (Task #730 Phase A1) started.

- Extracted `createWindow()` + window event wiring (close/focus/maximize/show, fade-to-hide, context menu, will-navigate guard) into `electron/window/createWindow.ts`.
- Extracted `Tray` construction + locale-aware menu into `electron/tray/createTray.ts`.
- `main.ts` keeps ownership of cross-module shared state (`isQuitting`, active-sid mirror, `badgeController`) and exposes thin DI wrappers so existing call sites (`app.activate`, `ccsm:set-language`, tray Quit, close-action dialog) stay unchanged.
- IPC blocks, security/prefs/sentry modules, `applyAppMenuLocale` were intentionally left in `main.ts` per task scope (Phase B+C will tackle IPC).

## LoC delta
| file | before | after | delta |
| --- | --- | --- | --- |
| `electron/main.ts` | 1098 | 830 | **-268** |
| `electron/window/createWindow.ts` | 0 | 344 | +344 |
| `electron/window/__tests__/createWindow.test.ts` | 0 | 146 | +146 |
| `electron/tray/createTray.ts` | 0 | 84 | +84 |

## Files added
- `electron/window/createWindow.ts` — BrowserWindow factory with `CreateWindowDeps` injection bag (`isDev`, `getActiveSid`, `onFocusChange`, `getIsQuitting`, `setIsQuitting`). Exports `installContextMenu` (pure, UT-covered).
- `electron/window/__tests__/createWindow.test.ts` — 5 tests covering `installContextMenu` semantics (registration, editable vs read-only items, copy-empty-selection disable, paste-disabled). Reverse-verified by breaking the `webContents.on` event name; 4/5 tests failed as expected and were restored.
- `electron/tray/createTray.ts` — exports `createTray(deps)` returning `{ tray, applyLocale }` controller. `deps.createMainWindow` lets the tray spawn a fresh window when all `BrowserWindow`s are gone.

## Validation
- `npx tsc --noEmit` + `npx tsc --noEmit -p tsconfig.electron.json` — clean
- `npm run build` — clean (3 pre-existing webpack size warnings only)
- `node -e "require('./dist/electron/window/createWindow.js')"` — `window OK`
- `node -e "require('./dist/electron/tray/createTray.js')"` — `tray OK`
[...]

PR #572, opened 2026-04-29
refactor(preload): split preload.ts into thin entry + 5 bridge modules (#769)

Jiahui-Gu:
## Wave-2 scan rationale (section 3.1)

`electron/preload.ts` (473 LoC) was the highest-ROI mechanical split candidate:
each of its five `contextBridge.exposeInMainWorld(...)` blocks is already a
distinct surface (`ccsm`, `ccsmPty`, `ccsmSession`, `ccsmNotify`, `ccsmSessionTitles`)
with its own listener Set, IPC channels, and exposed type. No cross-bridge state.
Pure mechanical extract — single concern per file, no rewriting, no behavior
change. Cheapest meaningful win in the wave-2 backlog.

## File-by-file mapping

| Old block in `electron/preload.ts` (lines)   | New file                                              |
|----------------------------------------------|-------------------------------------------------------|
| `ccsm` (1-168, sentry import + core api)     | `electron/preload/bridges/ccsmCore.ts`                |
| `ccsmPty` (170-251)                          | `electron/preload/bridges/ccsmPty.ts`                 |
| `ccsmSession` (253-371)                      | `electron/preload/bridges/ccsmSession.ts`             |
| `ccsmNotify` (373-421)                       | `electron/preload/bridges/ccsmNotify.ts`              |
| `ccsmSessionTitles` (423-473)                | `electron/preload/bridges/ccsmSessionTitles.ts`       |
| (entry)                                      | `electron/preload/index.ts` (sentry import + 5 install* calls + type re-exports) |

Each bridge file exports `installXxxBridge()` that calls `contextBridge.exposeInMainWorld(...)`.
The `index.ts` entry file calls all 5 in original order. Type aliases (`CCSMAPI`,
`CCSMPtyAPI`, `CCSMSessionAPI`, `CCSMNotifyAPI`, `CCSMSessionTitlesAPI`,
`SessionState`) are re-exported from `index.ts` so any future external
consumer keeps working (no current external consumer found via grep — d.ts
files in `src/` declare `Window` augmentations independently).

## Build config changes

- No `vite.config.*` or `electron-builder.json5` exists; preload is built by
  `tsc -p tsconfig.electron.json` (config unchanged — `include: electron/**/*`
  picks up the new subdir automatically).
- `electron/window/createWindow.ts:174`: preload path changed from
  `path.join(__dirname, '..', 'preload.js')` to
  `path.join(__dirname, '..', 'preload', 'index.js')`. Resolves to
[...]

PR #1098, opened 2026-05-05
v0.3(wave1-C): renderer fetch shim (window.ccsm compat over HTTP)

Jiahui-Gu:
## Summary

Wave 1 / 3 dev — renderer side. Replaces the legacy preload-installed
`window.ccsm` IPC bridge with a renderer-resident compatibility shim
that POSTs every call to the local daemon over loopback HTTP. **None
of the 28 existing `window.ccsm.X(...)` call sites are touched** —
they keep using the same surface, the underlying transport changes
out from under them.

Task #567

## Files

| file | role |
| --- | --- |
| `src/lib/daemon-port.ts` | NEW — caches the daemon port from `window.__getDaemonPort()` (preload bridge owned by wave-1 dev-B). |
| `src/lib/daemon-client.ts` | NEW — thin POST wrapper, uniform `{args}` / `{result}` envelope, single `'daemon offline: ...'` failure mode. |
| `src/lib/window-ccsm-shim.ts` | NEW — implements the legacy `CcsmApi`. Pre-fetches sync property `window.platform` at install time. Event subscriptions are no-op stubs (see follow-up). |
| `src/index.tsx` | MODIFY — `await installCcsmShim()` before `root.render()` so `<DragRegion>` reads of `window.ccsm?.window.platform` are safe on first render. |
| `src/global.d.ts` | MODIFY — drops the inline `window.ccsm` typing in favour of `CcsmApi` re-exported from the shim; adds `__getDaemonPort`. |
| `src/pty.d.ts`, `src/session.d.ts` | MODIFY — minor commentary; `ccsmSession` flipped to optional for symmetry. `ccsmPty` stays non-optional to avoid touching `src/terminal/*` (out of scope). |
| `eslint.config.js` | MODIFY — adds `fetch`/`Response`/`Request`/`Headers` to renderer globals. |

## Mapping rule (channel → endpoint)

```
window.ccsm.X(arg1, arg2, ...)         → POST /api/X
window.ccsm.window.minimize()           → POST /api/window/minimize
window.ccsm.userCwds.get()              → POST /api/userCwds/get
window.ccsm.i18n.setLanguage(l)         → POST /api/i18n/setLanguage  (event-style)
window.ccsm.i18n.getSystemLocale()      → POST /api/i18n/getSystemLocale

Body:    { args: [arg1, arg2, ...] }
Resp:    { result: <T> }   (HTTP 200)
         OR { error: '...' }  (HTTP non-2xx → throw new Error(error))
         OR HTTP 204 No Content (events) → resolves undefined

Event channels (from spec — no .send() callers exist in src today,
but the convention is wired so wave 2 can plug them in):
   '<channel>'  → POST /api/event/<channel>   body { args: [...] }
   e.g. 'set-language' → POST /api/event/set-language
```

The `set-language` event path is already exercised by the shim's
`i18n.setLanguage` (matches the original ipcRenderer.send shape: a
one-way fire-and-forget). The other three event channels listed in
the spec (`session:setActive`, `notify:userInput`, `session:setName`)
have **no current `ipcRenderer.send` call sites in `src/`** — they're
either gone or never landed in this branch — so no shim wiring was
needed for them. `daemonEvent(channel, args)` is exported from
`daemon-client.ts` for any future caller.

## Shim API

`CcsmApi` (exported from `src/lib/window-ccsm-shim.ts`, re-typed into
`window.ccsm` via `src/global.d.ts`) is byte-for-byte the same surface
as the pre-refactor inline `window.ccsm` typing. Three categories:

1. **async methods** (most of them) — each translates to a single
   POST. Examples: `loadState`, `saveState`, `getVersion`,
   `scanImportable`, `pickCwd`, `pathsExist`, `updatesCheck`, etc.
2. **sync property** — `window.ccsm.window.platform`. Pre-fetched in
   `installCcsmShim()` so React render reads see a real string. If
   the daemon is offline at boot, falls back to a
   `navigator.userAgent`-derived guess (the only consumer branches on
   `=== 'darwin'`, so the worst-case is showing the Windows-style 8px
   drag strip until the user reloads with the daemon up).
3. **event subscriptions** — `onUpdateStatus`, `onUpdateDownloaded`,
   `onMaximizedChanged`, `onBeforeHide`, `onAfterShow`. v0.3 has no
   daemon→renderer push channel yet, so these install **no-op
   handlers and return a no-op unsubscribe**. The renderer keeps
   working: Updates pane shows the last polled status, the maximize
   button just doesn't auto-flip until the user clicks again.
   **Wave-2 follow-up**: wire SSE / WebSocket from daemon and replace
   the stubs.

## Failure mode

`installCcsmShim()` is wrapped in try/catch internally:
- `__getDaemonPort()` missing → fallback platform, all methods throw
  `'daemon offline: window.__getDaemonPort preload bridge missing'`.
- `__getDaemonPort()` rejects / returns junk → fallback platform,
  methods throw `'daemon offline: invalid port ...'`.
- Daemon up but `/api/window/platform` unreachable → fallback
  platform, async methods that the user actually triggers throw
  `'daemon offline: <fetch error>'`.

Acceptance: **never silent fail**. Every renderer-visible failure
from the shim surfaces as a thrown Error whose message starts with
`daemon offline:` (single recognizable prefix for telemetry).

## Call-site audit

```
$ grep -rn 'window\.ccsm[\?\.]' src | grep -v 'window\.ccsm\(Pty\|Session\)' | wc -l
37   # incl comments and chained reads (window.ccsm.window.platform counts twice)

$ grep -rn 'window\.ccsm\?\?\.\|window\.ccsm\.' src | grep -v ccsmPty | grep -v ccsmSession
   # 28 unique runtime call sites across stores/, components/, app-effects/, agent/, App.tsx
```

**Zero** of those 28 call sites are modified by this PR. Spot-check
files: `src/stores/persist.ts`, `src/stores/drafts.ts`,
`src/stores/store.ts`, `src/components/settings/UpdatesPane.tsx`,
`src/components/settings/AppearancePane.tsx`,
`src/components/settings/NotificationsPane.tsx`,
`src/components/Sidebar.tsx`, `src/components/AppSkeleton.tsx`,
`src/components/WindowControls.tsx`, `src/components/ImportDialog.tsx`,
`src/app-effects/useExitAnimation.ts`,
`src/app-effects/useHydrateSystemLocale.ts`,
`src/app-effects/useUpdateDownloadedBridge.ts`,
`src/stores/slices/sessionCrudSlice.ts` — none of these appear in the diff.

## Out-of-scope (per spec)

Untouched: `src/components/*`, `src/agent/*`, `src/app-effects/*`
(beyond consuming the shim transparently), `src/stores/*`,
`src/terminal/*`, `daemon/*`, `electron/*`. The wave-1 dev-B preload
that installs `window.__getDaemonPort` is also out of scope and
expected to land in parallel.

- lint: ✓ (exit 0, 91 pre-existing warnings, 0 errors)
- typecheck: ✓ (exit 0, both `tsconfig.json` and `tsconfig.electron.json`)
- build: ✓ (exit 0, webpack production + tsc electron)
- unit tests: skipped — no existing spec touches `window.ccsm` /
  `installCcsmShim` / `daemon-port` / `daemon-client` (verified via
  `grep -l` across `src` and `tests`). New shim code is exercised
  end-to-end by the renderer at boot; CI three-platform matrix will
  cover the integration path once wave-1 dev-B's preload lands.
- e2e: skipped, mode rationale — current e2e harnesses run Electron
  without the v0.3 daemon up, so post-refactor they would see the
  intentional `daemon offline` failure mode for every persisted
  state read. e2e validation is gated on the wave-2 daemon being
  reachable from the harness; the wave-2 PR will revive e2e.

## Reverse-verify

N/A — not a bug fix.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -76,6 +76,14 @@ export default [
         requestAnimationFrame: 'readonly',
         cancelAnimationFrame: 'readonly',
         ResizeObserver: 'readonly',
+        // v0.3 wave 1: renderer talks to the local daemon over loopback
+        // HTTP (`src/lib/daemon-client.ts`), so it needs the platform
+        // `fetch` family. Electron's renderer runs in Chromium where
+        // these globals are always available.
+        fetch: 'readonly',
+        Response: 'readonly',
+        Request: 'readonly',
+        Headers: 'readonly',
         // `NodeJS` namespace is also referenced from renderer-side .d.ts
         // files (e.g. cliBridge.d.ts) that mirror preload types — keep it
         // available alongside browser globals.
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,7 +1,8 @@
 // Updater status — exported so the renderer's `UpdatesPane` (and any
 // future banners/toasts) can import a single source of truth instead of
-// redeclaring the union locally. Mirrors the shape broadcast by
-// `electron/updater.ts` over the `updates:status` IPC channel.
+// redeclaring the union locally. Mirrors the shape broadcast by the
+// daemon over the (future) push channel; today the v0.3 shim returns
+// it from `updatesStatus()` / `updatesCheck()`.
 export type UpdateStatus =
   | { kind: 'idle' }
   | { kind: 'checking' }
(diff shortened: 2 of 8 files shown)


Jiahui-Gu:
## Reviewer verdict (task #572): APPROVE — pending CI

**Direction: right** — three-layer split (port discovery / fetch wrapper / API surface) is the simplest thing that satisfies the v0.3 contract, and the path mapping is mechanical (method name -> URL fragment) so it doesn't force the wave-2 daemon to adopt any particular naming convention. No v0.4-throwaway scaffolding.

**Layer 2 findings:**
- Contract compliance: POST /api/X, {args:[...]} body, {result} envelope, nested window/minimize and userCwds/get paths -- all match the wave-2 spec.
- Failure mode: installCcsmShim() swallows daemon-offline at boot, falls back to navigator.userAgent-derived platform guess so React still mounts. Async methods throw 'daemon offline: ...' with a single recognizable prefix.
- Sync property window.ccsm.window.platform pre-fetched before root.render() -- DragRegion reads see a real string.
- 28 legacy call sites: gh pr diff --name-only confirms zero changes in src/stores/, src/components/, src/app-effects/, src/agent/, src/terminal/. As designed.
- Event stubs are no-op unsubscribe -- correct v0.3 answer; wave-2 will wire SSE/WS.
- i18n.setLanguage via daemonEvent('set-language', [l]) matches the original ipcRenderer.send fire-and-forget shape.
- __getDaemonPort typed; ccsm? flipped optional -- both correct.
- eslint.config.js adds fetch/Response/Request/Headers to renderer globals -- minimal.
- Local checks quotes real outputs (lint/typecheck/build exit 0). Unit+e2e skip rationale sound.

Will squash-merge once CI matrix (ubuntu/windows lint+typecheck + 3 platforms e2e) is green.

PR #1121, opened 2026-05-06
feat(preload): A1 lazy daemon-port cache + daemonFetch helper (Task #629)

Jiahui-Gu:
Task #629 — A1 of the preload-single-source PR chain. Pure infrastructure addition; no existing caller is rewired in this PR, so behavior is unchanged.

## Summary
- `getCachedDaemonPort()`: module-scoped lazy cache around the existing `ipcRenderer.invoke('daemon:getPort')`. Coalesces concurrent first callers into one IPC; never caches `null` results so the preload-startup race window stays retryable instead of being frozen.
- `daemonFetch<T>(path, opts)`: generic loopback fetch wrapper that builds `http://127.0.0.1:${port}${path}`, JSON-serializes the body when `opts.json` is supplied, and parses the response. Surfaces `DaemonUnavailableError` / `DaemonHttpError` so callers can react instead of silently swallowing (unlike `fireDaemonEvent` / `getDaemon` which intentionally swallow for fire-and-forget paths).
- `__resetDaemonPortCacheForTest()`: test seam.
- Existing `resolvePort()` / `openSse()` / `fireDaemonEvent()` / `getDaemon()` are deliberately untouched — the SSE reconnect loop must keep re-resolving on every retry to recover from daemon restarts (daemon binds `127.0.0.1:0`, so each spawn picks a fresh port). Caching there would defeat that recovery.

## Not in this PR
- `ccsmCore.ts` API surface — untouched.
- `src/lib/window-ccsm-shim.ts` — B1 deletes it.
- `electron/main.ts` — `daemon:getPort` handler already exists.

- lint: passing (exit 0; only pre-existing warnings, none introduced by this diff)
- typecheck: passing (exit 0; both `tsc --noEmit` and `tsc --noEmit -p tsconfig.electron.json`)
- build: passing (exit 0; webpack + tsc -p tsconfig.electron.json)
- unit tests: `npx vitest run tests/preload-daemon-helper.test.ts tests/electron-load-smoke.test.ts` — Test Files 2 passed (2), Tests 20 passed (20), Duration 2.57s. The new helper has no production callers so e2e is not exercised; full suite is left to CI matrix per fast-path.
- e2e: skipped — pure infrastructure addition with zero production callers, cannot affect any existing e2e behavior. B1+ PRs that wire callers through will run e2e.
- electron require-load smoke: `node -e "require('./dist/electron/preload/bridges/_daemon.js')"` → OK (no `ERR_REQUIRE_ESM`).

## Test plan
- [x] Cache returns identical port across N calls with single IPC invoke
- [x] Concurrent first callers share one in-flight IPC promise
- [x] `null` resolutions are not cached (next call retries)
- [x] IPC reject returns `null` and is also not cached
- [x] JSON GET parses response, hits correct loopback URL
- [x] JSON POST sets `Content-Type` + serialized body
- [x] `DaemonUnavailableError` thrown when port is `null`
- [x] `DaemonHttpError` thrown on non-2xx response (status preserved)
- [x] 204 No Content returns `undefined` without JSON.parse crash
- [x] Native fetch rejections (network / abort) propagate
- [x] Explicit method override (e.g. DELETE) honored

### DIFF
--- a/electron/preload/bridges/_daemon.ts
+++ b/electron/preload/bridges/_daemon.ts
@@ -26,6 +26,127 @@ async function resolvePort(): Promise<number | null> {
   }
 }
 
+// Task #629 — A1 lazy port cache + fetch helper.
+//
+// Why a separate cache (not folded into resolvePort): the existing SSE
+// reconnect loop in `openSse` re-resolves the port on every reconnect to
+// recover from daemon restarts (daemon binds 127.0.0.1:0, picks a fresh
+// port each spawn). Caching there would defeat that recovery. The cache
+// here is for one-shot RPC-style calls (`daemonFetch`) where the renderer
+// just wants the port once per session and any daemon-restart fallout
+// shows up as a fetch failure that the caller already handles.
+//
+// `null` resolutions are NOT cached — first invocation can land before
+// `daemon:getPort` is wired (preload runs before main-process IPC handlers
+// finish registering on cold boot). Caching null would freeze that race
+// permanently. Successful (numeric) resolutions are cached for the
+// lifetime of the renderer; call `__resetDaemonPortCacheForTest` from
+// tests to drop it.
+let cachedPortPromise: Promise<number | null> | null = null;
+
+/**
+ * Resolve the daemon HTTP port, caching the first successful lookup.
+ *
+ * Concurrent first calls share the same in-flight invoke promise (so we
+ * never issue two `daemon:getPort` IPC calls in parallel during the
+ * preload-startup race window). If the resolve yields `null` (handler
+ * not yet registered) we drop the cache so the next caller retries.
+ */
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Direction: right. 库式扩建, 行为零变, 边界严守 (仅 _daemon.ts + 新 test, 不动 ccsmCore/shim/main, 0 production importer)。Layer 2 通过: PR body Local checks 5 行齐 + 真实 vitest 输出 (2 files / 20 tests / 2.57s); 11 项 test plan 覆盖 cache 命中 / coalesce / null 不缓存 / IPC reject / GET / POST / 204 / HTTP error / 网络 error / DELETE override; null 不缓存有 inline comment 解释 preload-startup race。base=working 无 required CI, squash merge。

PR #1122, opened 2026-05-06
feat(preload): expose 25 daemon-backed methods on ccsmCore behind CCSM_PRELOAD_SOURCE flag (Task #628)

Jiahui-Gu:
## Summary

A2 follow-up to PR #1121 (A1 daemonFetch helper). Grows the preload `ccsmCore` bridge with the 25 methods currently served by the renderer-side `window-ccsm-shim` (loadState/saveState/i18n/getVersion/scanImportable/recentCwds/userCwds/defaultModel/pathsExist/window.*), so wave B1 can delete the shim and let the renderer call `window.ccsm.*` directly through the preload boundary.

## Design

1. **Feature-flagged**: gated behind env `CCSM_PRELOAD_SOURCE=1` (default OFF). With the flag OFF the bridge keeps exposing only the 6 IPC-only surfaces, so the shim still wins and behavior is bit-for-bit identical to before. Wave B1 will flip the flag default and delete the shim in the same PR.

2. **Configurable binding**: switched from `contextBridge.exposeInMainWorld('ccsm', api)` (non-configurable) to `Object.defineProperty(window, 'ccsm', { configurable: true, ... })`. Without this, A2 alone would crash the renderer at boot because the shim's own `Object.defineProperty` would throw `Cannot redefine property: ccsm`. `sandbox: false` is already set in `electron/window/createWindow.ts`, so `window` is directly accessible from the preload.

Daemon routes mirror `src/lib/window-ccsm-shim.ts` byte-for-byte: `POST /api/<path>` with `{args:[...]}` envelope, response unwraps the `result` field. `saveState` keeps the unwrap-and-throw contract that `src/stores/persist.ts` depends on.

## Acceptance trace

- 25 methods all routed through the existing A1 `daemonFetch` helper; route table copied from `src/lib/window-ccsm-shim.ts` (no new wire format)
- Flag OFF (default): renderer still uses the shim — verified by the `installCcsmCoreBridge - shim coexistence` test that `Object.defineProperty(window, 'ccsm', { configurable:true, ... })` no longer throws after the bridge installs
- Flag ON: 25 method signatures + nested namespaces present and routed; e2e end-to-end is left to B1 per spec
- No changes to `src/lib/window-ccsm-shim.ts`, `src/index.tsx`, or `electron/main.ts`
- All pre-existing UTs untouched

## Files

- Edit: `electron/preload/bridges/ccsmCore.ts` (+217 lines: 25-method builder + flag accessor + configurable defineProperty installer)
- Add: `tests/preload-ccsmcore-methods.test.ts` (14 cases: flag matrix, shim coexistence, route wiring, saveState unwrap, fire-and-forget event, sync platform, push-channel stubs)

- lint: PASS (exit 0, 45 pre-existing warnings unchanged)
- typecheck: PASS (exit 0)
- build: PASS (exit 0)
- unit tests: `tests/preload-ccsmcore-methods.test.ts` (14 PASS) + `tests/preload-daemon-helper.test.ts` regression (11 PASS) — 25 PASS, ran in 2.68s
- e2e: skipped, this PR is preload-only and the flag is OFF by default so no renderer-visible behavior changes; B1 will exercise e2e when it flips the flag and deletes the shim
- electron require-load smoke: `node -e "require('./dist/electron/preload/bridges/ccsmCore.js')"` -> OK

## Test plan

- [ ] CI green on three-platform matrix (lint + typecheck + build + unit + e2e)
- [ ] Reviewer confirms flag-off codepath leaves shim semantics intact (no behavior delta vs `working`)
- [ ] B1 follow-up issue tracks: flip `CCSM_PRELOAD_SOURCE` default to ON + delete `src/lib/window-ccsm-shim.ts` + drop `installCcsmShim()` from `src/index.tsx`

### DIFF
--- a/electron/preload/bridges/ccsmCore.ts
+++ b/electron/preload/bridges/ccsmCore.ts
@@ -22,8 +22,24 @@
 // import scan / userCwds / paths:exist / window controls) moved to the
 // daemon's HTTP API. The renderer fetches `http://127.0.0.1:<port>/...`
 // using the port returned by `getDaemonPort()`.
+//
+// Task #628 (A2) — additionally expose the 25 daemon-backed methods that
+// the renderer-side `window-ccsm-shim` provides, so wave B1 can delete the
+// shim and let the renderer call `window.ccsm.*` directly. To avoid
+// breaking the shim BEFORE B1 lands, we install `window.ccsm` via
+// `Object.defineProperty` with `configurable: true` (NOT
+// `contextBridge.exposeInMainWorld`, which produces a non-configurable
+// binding the shim cannot redefine). This requires `sandbox: false` on
+// the BrowserWindow (already configured in electron/window/createWindow.ts).
+//
+// The 25 methods are gated behind the env flag `CCSM_PRELOAD_SOURCE`
+// (default OFF). With the flag OFF, the bridge exposes only the original
+// 6 IPC-only methods, and the renderer-side shim still wins via
+// `Object.defineProperty(window, 'ccsm', { configurable: true, ... })`.
+// Wave B1 will (a) flip the flag default to ON and (b) delete the shim.
 
 import { contextBridge, ipcRenderer, type IpcRendererEvent } from 'electron';
+import { daemonFetch } from './_daemon';
 
 type UpdateStatus =
   | { kind: 'idle' }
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
Reviewer LGTM (self-approve blocked by GitHub same-token rule, recording verdict via comment).

Direction: right. Layer 2 all PASS:
- -2 是 const api -> ipcOnlyApi rename + CCSMAPI type alias 改 ReturnType, 0 老 method 被删
- Flag OFF: buildCcsmCoreApi 只 return {...ipcOnlyApi}, 25 方法不暴露
- defineProperty 显式 configurable:true, 测试模拟 shim 二次 defineProperty 不抛 redefine
- 不动 src/lib/window-ccsm-shim.ts / src/index.tsx / electron/main.ts (PR diff 仅 2 文件)
- 25 方法路由 1:1 抄自 shim (db/load, db/save, app/userCwds/*, i18n/getSystemLocale, event/set-language, window/*), 抽样核对全过
- 14 cases 覆盖 flag matrix + shim coexistence + 路由 + saveState envelope + fire-and-forget + sync platform

Squash merge now.

PR #1124, opened 2026-05-06
feat(preload): wire onMaximizedChanged/onBeforeHide/onAfterShow (Task #624)

Jiahui-Gu:
## Summary

Task #624 (C2) — event-channel revive. Main process at
`electron/window/createWindow.ts` already emits three IPC push channels
(`window:maximizedChanged`, `window:beforeHide`, `window:afterShow`) but
the preload bridge in `electron/preload/bridges/ccsmCore.ts` returned
`NOOP_UNSUBSCRIBE` for the three corresponding `window.onXxx` methods —
a placeholder left over from wave B1 (Task #627). Renderer consumers
(`WindowControls.tsx` chrome icon, `useExitAnimation.ts` fade
choreography) were silently degraded.

Wired up the bridge to register `ipcRenderer.on(channel, wrap)`, forward
the typed payload (skipping `IpcRendererEvent`), and return a real
`ipcRenderer.removeListener` unsubscribe. Mirrors the existing
`onUpdateStatus` / `onUpdateDownloaded` pattern.

No `createWindow.ts` change — main side was already correct.

## Tests

Three new cases in `tests/preload-ccsmcore-methods.test.ts` under a
`window event channels — Task #624 (C2)` describe block:
- onMaximizedChanged: forwards boolean payload, unsubscribe stops delivery
- onBeforeHide: forwards `{ durationMs }` payload, unsubscribe stops delivery
- onAfterShow: fires with no payload, unsubscribe stops delivery

Each case replaces the per-test `ipcOn` mock with a router that captures
the `(channel, wrap)` tuple, then synthesizes a main-process emit by
invoking the captured wrap. Existing assertions (including the legacy
"event-channel stubs return a no-op unsubscribe" shape test) are left
untouched — they continue to pass against the new impl since it still
returns a function that does not throw.

## Local checks (host: Windows 11, mode: fast)

- lint: ✓ (exit 0, 0 errors / 45 pre-existing warnings, none from changed files)
- typecheck: ✓ (exit 0, both `tsc --noEmit` and `tsc --noEmit -p tsconfig.electron.json`)
- build: ✓ (exit 0, webpack + tsc electron emit)
- unit tests: ✓ `npx vitest run tests/preload-ccsmcore-methods.test.ts` →
  `Test Files 1 passed (1) | Tests 11 passed | 3 skipped (14) | Duration 28.87s`
- electron require-load smoke: ✓ `node -e "require('./dist/electron/preload/bridges/ccsmCore.js')"` →
[...]

PR #1123, opened 2026-05-06
feat(preload): single-source window.ccsm via contextBridge; delete renderer shim (Task #627)

Jiahui-Gu:
## Summary

Wave **B1** of the v0.3 `Cannot redefine property: ccsm` hardening (Task #627). Builds on A1 (#1121) + A2 (#1122). With this PR `window.ccsm` is single-sourced from the preload bridge — the renderer-side shim is deleted, the `CCSM_PRELOAD_SOURCE` env flag is removed, and the bridge installs through `contextBridge.exposeInMainWorld` (the only world-crossing channel under `contextIsolation: true`).

### Source (business code, not tests)

* `src/lib/window-ccsm-shim.ts` — **deleted** (subject of B1).
* `src/lib/daemon-port.ts` — **deleted** (only importer was the shim).
* `src/lib/daemon-client.ts` — **deleted** (only importer was the shim).
* `src/index.tsx` — drop `installCcsmShim()` import + the async wrapper. `<App />` now renders synchronously after `root.render()` because the preload has already placed `window.ccsm` before any renderer module evaluates.
* `src/global.d.ts` — re-export `CCSMAPI` from `electron/preload/bridges/ccsmCore` (the deleted shim previously owned `CcsmApi`).
* `electron/preload/bridges/ccsmCore.ts`:
  * Drop `preloadSourceEnabled()` export and the `CCSM_PRELOAD_SOURCE` flag check.
  * `buildCcsmCoreApi()` now unconditionally returns `{ ...ipcOnlyApi, ...buildDaemonMethods() }`.
  * `installCcsmCoreBridge()` switched to `contextBridge.exposeInMainWorld('ccsm', api)` as the primary path. `Object.defineProperty(window, 'ccsm', ...)` is kept as a jsdom fallback only — under production `contextIsolation: true` it would only attach to the preload's isolated world and never reach the renderer's main world (this was the latent bug A2's coexistence story masked: the shim was the de-facto main-world installer).
[...]

PR #1125, opened 2026-05-06
fix(e2e): green up tray + close-dialog-is-native (Task #636)

Jiahui-Gu:
## Summary

Two harness-ui cases (`tray`, `close-dialog-is-native`) were red on all three CI platforms after PR #1123 + #1124 merged, blocking #607 e2e gate lock and the v0.3 release chain.

Root cause was three layered bugs that compounded:

1. **Daemon binary never built.** `npm run build` only ran `tsc -p tsconfig.electron.json` + webpack — `dist/daemon/main.js` was never emitted, so `daemon-spawner` exited with `Cannot find module ... main.js` and every `window.ccsm.loadState/saveState` call surfaced as `daemon port unavailable` after the 5s preload retry budget.
2. **Daemon ABI mismatch.** Once daemon binaries existed, `daemon-spawner` spawned them under the system `node` on PATH (NODE_MODULE_VERSION 127), but `better-sqlite3` / `node-pty` are rebuilt against Electron's ABI (NODE_MODULE_VERSION 145). Daemon crashed at module load.
3. **`closeAction` cache stale across cases.** `win.on('close')` reads the preference *synchronously* (Electron close events have no `await`), so renderer-side `saveState('closeAction', ...)` only took effect on next app launch. The harness shares one electron across cases; `tray` saved 'tray' and `close-dialog-is-native` saved 'ask', but the close handler always saw the stale cache.

## Fixes

- `package.json`: chain `tsc -p tsconfig.daemon.json` into `build`/`typecheck` so daemon dist + types stay in lock-step.
- `daemon/server.ts`: tighten one `noUncheckedIndexedAccess` slip exposed once daemon was actually type-checked.
- `electron/daemon-spawner.ts`: default to `process.execPath` + `ELECTRON_RUN_AS_NODE=1` so the daemon shares Electron's V8 ABI; packaged headless builds keep the `CCSM_DAEMON_NODE` escape hatch.
- `electron/{main,window/createWindow,preload/bridges/ccsmCore}.ts`: when the renderer persists `closeAction`, fire-and-forget IPC notify main so the in-memory cache stays in lock-step with the daemon write.

## Local checks (host: Windows, mode: full)

- lint: 0 errors (45 pre-existing warnings)
- typecheck: pass (tsc + electron + daemon)
- build: pass (clean + electron + daemon + webpack)
- unit tests: 25/25 PASS, 3 skipped (`electron/__tests__/daemon-spawner.test.ts`, `tests/preload-ccsmcore-methods.test.ts`, `daemon/prefs/__tests__/closeAction.test.ts`)
- e2e: `node scripts/harness-ui.mjs` → 13/14 PASS. The two cases this PR targets are green; the remaining red is `terminal-pane-mounted`, which is a pre-existing host-specific flake unmasked by the daemon actually working — on CI runners (no `claude` binary on PATH) the case takes the `ClaudeMissingGuide` branch and stays green, matching the task spec's reported `12 passed`.

## Reverse-verify

- **stash fix → run two cases → FAIL**:
  ```
  [XX] tray                                       6004ms
      page.evaluate: Error: daemon port unavailable
          at Object.caseTray (...harness-ui.mjs:314:37)
  [XX] close-dialog-is-native                     5820ms
      page.evaluate: Error: daemon port unavailable
          at Object.caseCloseDialogIsNative (...harness-ui.mjs:384:37)
  ```
- **apply fix → run two cases → PASS**:
  ```
  [OK] tray                                        752ms
      hide-on-close=true; restore-via-show=true
  [OK] close-dialog-is-native                      762ms
      dialog.showMessageBox invoked count=1 hasParent=true buttons=2
  ```

## Red lines (a/b/c)

- (a) Did NOT delete or skip either failing case — fixed the business-side daemon plumbing instead.
- (b) Did NOT touch the assertions inside `caseTray` or `caseCloseDialogIsNative` (`scripts/harness-ui.mjs` is unchanged).
- (c) Business code changes are gated by the existing `electron/__tests__/daemon-spawner.test.ts` UT (4/4 passing) plus the `harness-ui` e2e itself (13/14, the only red is unrelated pre-existing host noise).

## Test plan

- [ ] CI: `e2e (mac/ubuntu/win)` all 14/14 green
- [ ] CI: lint + typecheck + test stay green
- [ ] CI: `pr-no-silent-drops` passes


Jiahui-Gu:
## Reviewer verdict: APPROVE + merging

**Direction: right** — fixed the actual build/spawn/cache plumbing rather than skipping the red e2e cases. Three root causes all check out (npm run build never invoked tsc -p tsconfig.daemon.json, daemon-spawner using system node 127 vs Electron 145 ABI for native deps, closeActionCache stale across e2e cases sharing one electron).

**Wire-up**: notifyCloseActionFromRenderer exported in createWindow.ts -> imported in main.ts:24 -> registered as ipcMain.handle('main:notifyCloseAction') in main.ts:216 -> invoked by preload ccsmCore.ts:131 right after saveStateMethod success. Closed loop, no dangling export.

**Layer 2 findings**:

- CI: e2e mac 2m56s / ubuntu 2m57s / win 4m26s all PASS. lint+typecheck+test 3-platform PASS. no-silent-drops PASS. Better than dev's local 13/14 — CI is a clean runner where the terminal-pane-mounted host noise doesn't apply.
- ABI fix (process.execPath + ELECTRON_RUN_AS_NODE=1) is the canonical Electron pattern. CCSM_DAEMON_NODE escape hatch preserved for future packaged headless CLI. Dev mode (npm start) also uses main process so same execPath path — not broken.
- daemon/server.ts one-line ?. type-narrow is fallout from finally type-checking the daemon project — correct fix.
- closeAction IPC notify: best-effort void invoke + .catch swallow is right (a missed notify falls back to on-startup cache prime, which is pre-v0.3 behavior). parseCloseAction normalizer in notifyCloseActionFromRenderer prevents garbage values from crashing the close handler.
- Minor (non-blocking, do NOT need to fix for v0.3): main:notifyCloseAction handler does not validate sender frame. Acceptable here because (a) preload is the only caller, (b) value space is enum-narrowed by parseCloseAction so a malicious renderer can at most flip closeAction between tray/ask/close — no privilege escalation. File a hardening followup post-v0.3 if anyone cares.
- Red lines (a/b/c) all met: no test/assertion deletions, business code covered by daemon-spawner UT (4/4) + the e2e cases themselves.
- Reverse-verify section quotes real stash-fix FAIL output and apply-fix PASS output. Probe is real.

Squash-merging now to unblock #607 e2e gate lock + v0.3 release chain.

PR #575, opened 2026-04-29
test(rtl): fill RTL gaps from PR #568 audit (#765)

Jiahui-Gu:
## Summary

Fills the 6 RTL coverage gaps surfaced by the `artifacts/test-audit-phase1.md` Section 1 review of PR #568 (e2e -> RTL migration). Closes Task #765.

## Audit reference

Audit: `artifacts/test-audit-phase1.md` Section 1 — 9 deleted harness-ui probes mapped to RTL targets, surfacing 2 Layer-1 + 4 Layer-2 real coverage losses.

## Gaps addressed

| Gap | File | What it asserts |
|---|---|---|
| L1 — language-toggle a1-a3 | `tests/sidebar-i18n-aria.test.tsx` (new) | Renders `<Sidebar />`, flips `usePreferences.setLanguage` en->zh->en, asserts visible Settings button text + Search aria-label re-render at each step. Existing `tests/language-switch.test.tsx` only covered the bare `t()` token, never a Sidebar consumer. |
| L1 — shortcut-overlay-opens a5 + a8 | `tests/shortcut-overlay-platform.test.tsx` (new) | `vi.resetModules()` + dynamic import after spoofing `navigator.userAgent` (ShortcutOverlay branch) AND `navigator.platform` (CommandPalette branch). 2 cases per module: macOS asserts `⌘` glyph present + `Ctrl` absent, win/linux asserts the inverse. Covers both SidebarHeader-level overlay and CommandPalette per-row hint. |
| L2 — no-sessions a3 | `tests/first-run-empty-state.test.tsx` (extended) | jsdom layout proxy: asserts both empty-state CTA buttons carry `w-44 justify-center` className tokens and live under the `first-run-empty` flex parent so `size="md"` h-8 drives equal heights. |
| L2 — palette-empty a7 | `tests/command-palette.test.tsx` (extended) | Open palette -> `keyDown(Escape)` -> `queryByRole('dialog')` returns null. |
| L2 — palette-nav a6 | `tests/command-palette.test.tsx` (extended) | Open palette -> type query -> ArrowDown nav -> `keyDown(Enter)` -> assert spy fired AND `queryByRole('dialog')` returns null. |
| L2 — search-shortcut-f a3 | `tests/app-effects/useShortcutHandlers.test.tsx` (extended) | Negative: dispatching `ctrl+k` / `ctrl+K` / `meta+k` does NOT call any of the 4 handlers. Guards against a silent rebind regression. |

[...]

PR #1127, opened 2026-05-06
fix(daemon): prevent + self-heal better-sqlite3 ABI mismatch (Task #641)

Jiahui-Gu:
## Summary

Task #641 — v0.3 ship-blocker. Dogfood #575 root cause: postinstall \`@electron/rebuild\` for better-sqlite3 silently fails (Windows .node lock, missing toolchain, etc.), leaving the binding compiled against host Node ABI (e.g. 127) instead of Electron's (e.g. 145). On first DB open the daemon crashes with \`NODE_MODULE_VERSION mismatch\` and the user sees a silent storage failure. #639 surfaces it as a banner; this PR makes the failure not happen (L3) and adds a self-heal fallback.

Three layers, all shipped:

- **Layer 1** (postinstall hardening): extracted \`rebuildWithRetry\` helper → Windows EPERM/EBUSY on a locked \`.node\` waits 5s and retries once.
- **Layer 2** (ship-time): \`scripts/after-pack.cjs\` now asserts \`better_sqlite3.node\` is present in \`app.asar.unpacked\` alongside the existing node-pty check. A failed rebuild can no longer slip past pack into a published installer.
- **Layer 3** (runtime self-heal): new \`electron/abi-self-heal.ts\` probes better-sqlite3 (\`:memory:\` open) before \`app.whenReady\`. On NODE_MODULE_VERSION mismatch it spawns \`@electron/rebuild\` and \`app.relaunch()\`. One-shot marker in userData prevents loops.

File ownership respected: \`scripts/harness-ui.mjs\` untouched (owned by dev #639). New e2e cases live in sibling \`scripts/harness-abi-selfheal.mjs\`, auto-discovered by run-all-e2e.

## Local checks (host: win32, mode: fast)

- lint: ✓ (exit 0, only pre-existing warnings)
- typecheck: ✓ (exit 0 — all three tsconfig projects)
- build: ✓ (exit 0, dist/electron/abi-self-heal.js + dist/electron/main.js present)
- unit tests: \`npx vitest run tests/abi-self-heal.test.ts tests/postinstall.test.ts tests/electron-load-smoke.test.ts\`  
  → 28 passed (19 new + 9 load-smoke unchanged), 28.5s total
- e2e: \`node scripts/harness-abi-selfheal.mjs\` → 5 passed, 0 failed (8ms wall, all skipLaunch)
- e2e baseline: \`node scripts/harness-ui.mjs --only=sidebar-align\` → 1 passed (verifies main.ts boot still works after self-heal wiring)

## Test plan

- [ ] CI three-platform matrix green on harness-abi-selfheal (5 cases)
- [ ] CI green on harness-ui (14 cases unchanged)
- [ ] Verify \`electron-builder\` pack on CI artifacts — better-sqlite3 binding listed by after-pack hook
- [ ] On a clean install where the Electron major bumped (ABI skew), confirm self-heal triggers exactly once

### DIFF
--- a/electron/abi-self-heal.ts
+++ b/electron/abi-self-heal.ts
@@ -1,0 +1,246 @@
+// ABI self-heal — Task #641 Layer 3.
+//
+// Background: better-sqlite3 ships a native `.node` binding compiled
+// against a specific Node ABI (NODE_MODULE_VERSION). When a developer
+// runs plain `npm install`, npm picks up better-sqlite3's prebuilt
+// binding for the host Node ABI (e.g. v127 on Node 22), NOT the
+// Electron ABI (e.g. v145 on Electron 41). The postinstall hook
+// (scripts/postinstall.mjs) tries to rebuild via @electron/rebuild
+// — but if that fails (toolchain missing, .node locked, etc.) we
+// can ship a broken environment.
+//
+// At runtime the daemon (which lives in the Electron process tree
+// post wave-1) tries to `require('better-sqlite3')` and gets a
+// classic:
+//
+//     Error: The module '...\better_sqlite3.node' was compiled
+//     against a different Node.js version using NODE_MODULE_VERSION
+//     127. This version of Node.js requires NODE_MODULE_VERSION 145.
+//
+// daemon initDb fails, and pre-Task #641/#639 the user only saw a
+// silent storage failure. Layer 1 (#639) surfaces a banner; this
+// Layer 3 module makes the failure self-correct: detect the ABI
+// mismatch at main entry, run @electron/rebuild as a child process,
+// and restart the app once.
+//
+// Reference: https://www.electronjs.org/docs/latest/tutorial/using-native-node-modules
+//
+// Design constraints:
+//   - Pure / testable: I/O is injected via a `deps` parameter so unit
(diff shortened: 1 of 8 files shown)


Jiahui-Gu:
Review verdict: APPROVE (self-merge, same-token).

**Direction: right** — fixes #575 dogfood ship-blocker by making daemon initDb not fail at all, using three-layer defense (postinstall retry + after-pack assert + runtime self-heal).

**Layer 2 findings** (all pass):
- after-pack assertion is hard `throw new Error(...)` (not silent skip) — broken installer cannot ship.
- L3 spawnSync uses fixed argv (no user input), bin path joined under appRoot, status/error/signal all classified.
- Marker race: `acquireSingleInstanceLock()` runs before `selfHealOrRelaunch()`; even without it, marker is written BEFORE rebuild so second-pass hits `already-tried` instead of looping.
- 19 UT + 5 e2e cover ok / healed / already-tried / rebuild-failed / packaged-no-op / Windows .cmd / non-ABI errors / marker-pre-write crash-safety / spawn error+signal+retry queue.
- `scripts/harness-ui.mjs` untouched per file-ownership split with #639.
- CI green on all 7 checks.

Known L1 retry weakness tracked in Task #643; doesn't block ship.

PR #1148, opened 2026-05-07
test(e2e): expand s2-pages-loopback to plan §E.2 9-step flow (Task #751)

Jiahui-Gu:
## Summary

S2 §E "Done When" matrix A on chromium + windows: full session
lifecycle over the cross-origin Pages → loopback chain.

- Grows existing `s2-pages-loopback.spec.ts` (originally Task #725's
  cross-origin GET /api/sessions probe) into the 9-step business flow:
  + New Session → WS attach → typed-token OUTPUT round-trip → reload
  guard (Task #716/#671: must NOT auto-mint a new session) → × close
  with EXIT frame OR clean ws close.
- Skips the spec in the CI e2e job (same bucket as p1-smoke /
  p3-stress / session-persistence — needs authed claude). The
  dedicated credentialled s2-acceptance CI job is Task #755.
- Sister tasks: #752 (multi-browser), #753 (spoofed-origin reject),
  #754 (Tauri red line), #755 (CI rollout). #744 push back → split
  was accepted by manager; this PR is the smallest #751 slice.

## Why grow in place vs new spec file

Initial #751 attempt added a parallel `s2-acceptance.spec.ts`. User
edit replaced that with an in-place expansion of `s2-pages-loopback`
to avoid duplicating fixtures (httpsPreview + daemon + ws-frame
harness) and to keep the cross-origin chain story in one file. The
deleted-spec route was discarded; this PR ships the in-place version.

## Local checks (host: windows-latest, mode: full)

- lint: ✓ (exit 0) — `pnpm -F @ccsm/e2e-web lint`
- typecheck: ✓ (exit 0) — `pnpm -F @ccsm/e2e-web typecheck`
- build: ✓ (exit 0) — `pnpm -r build` (frontend-web + frontend-tauri
  vite builds + all tsc emit packages)
- unit tests: skipped, no source changes — only `packages/e2e/tests/s2-pages-loopback.spec.ts` and `.github/workflows/ci.yml`. No package's spec/source under `packages/{shared,core,ui,daemon,frontend-*}` was touched.
- e2e (only-changed, 3 stability runs):
  - run 1: `1 passed (8.8s)` — `s2-pages-loopback ✓ (6.6s)`
  - run 2: `1 passed (8.9s)` — `s2-pages-loopback ✓ (6.7s)`
  - run 3: `1 passed (9.0s)` — `s2-pages-loopback ✓ (6.8s)`
[...]

PR #1145, opened 2026-05-07
ci: add working branch to ci.yml triggers (Task #743)

Jiahui-Gu:
## Summary
Add `working` to ci.yml `pull_request` + `push` trigger branch filters. Currently only `main` is listed, so PRs into `working` and pushes to `working` produce zero checks — which forced manager to PATCH-remove `required_status_checks` from `working` branch protection (Task #743 backstory).

This PR is the fix: extend the filter list. No jobs / steps / matrix / other workflows changed.

```yaml
on:
  pull_request:
    branches: [main, working]
  push:
    branches: [main, working]
```

The PR itself targets `working`, so its own existence is the proof — the new trigger fires this run.

## Local checks (host: Windows, mode: yml-only)
- lint: ✓ (exit 0, `pnpm -r lint` all 8 packages Done)
- typecheck: ✓ (exit 0, after prebuild of shared/core/ui, `pnpm -r typecheck` all 8 Done)
- build: skipped (yml-only change, no .ts/.tsx touched — per dev.md §3 step 2 skip rule)
- unit tests: skipped (yml-only)
- e2e: skipped (yml-only)

## CI evidence — `gh pr checks 1145`
Captured ~25s after PR open, confirming trigger fires (would be empty `no checks reported` if the trigger were still misconfigured):

```
build (ubuntu-latest)   pending  0  https://github.com/Jiahui-Gu/ccsm/actions/runs/25513795546/job/74879337528
build (windows-latest)  pending  0  https://github.com/Jiahui-Gu/ccsm/actions/runs/25513795546/job/74879337473
tauri-build             skipping 0  https://github.com/Jiahui-Gu/ccsm/actions/runs/25513795546/job/74879338122
```

`e2e` matrix has `needs: build` so it registers after build completes; `tauri-build` is gated `if: github.event_name != 'pull_request'` so it correctly skips. `build` is the actual lint+typecheck+build job per workflow definition.

## Out of scope
- Branch protection re-enable (`gh api PUT /branches/working/protection`) — manager owns that after merge, per task spec.
- No new workflow files, no job edits.

Task #743.

PR #759, opened 2026-05-01
feat(v03 T59): release attestation — SHA + SLSA + minisign sidecars

Jiahui-Gu:
## Summary
Implements Task #1017 / v0.3 spec frag-11 §11.4 + §11.4.1 (Task 23b): every installer artifact published by `release.yml` now ships with three sidecar files alongside it — `<artifact>.sha256`, `<artifact>.intoto.jsonl` (SLSA L3), and `<artifact>.minisig`. A sidecar-verify gate in the new `publish` job refuses to publish a draft release if any sidecar is missing.

## Workflow file changes (`.github/workflows/release.yml`)
- `build` matrix job: unchanged except removal of the inline `Upload to GitHub Release` step (moved to new `publish` job).
- New `attest` job (ubuntu-latest):
  - Downloads all matrix artifacts; flattens to `dist-flat/`.
  - Generates per-artifact `.sha256` (sha256sum format).
  - Installs minisign; signs every installer with the key from `MINISIGN_PRIVATE_KEY` repo secret (warns + writes empty `.minisig` placeholders if secret absent — gate then catches it).
  - Computes base64 SHA256SUMS hashes input for SLSA generator; exposes as `outputs.hashes`.
- New `provenance` job: calls the official reusable workflow `slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v2.0.0` (NOT a custom script) with the hashes input. Produces a single signed `ccsm.intoto.jsonl` whose subjects cover every installer.
- New `publish` job (tag push only):
  - Downloads the `ccsm-attested` artifact + the SLSA provenance artifact.
  - **Sidecar-verify gate** — refuses to publish if any installer is missing `.sha256`, `.minisig`, or if no `*.intoto.jsonl` is present. Aliases the single SLSA file to `<artifact>.intoto.jsonl` per installer so the README's one-pattern verify command works.
  - Uploads installers + all sidecars to the draft GitHub Release.

## Other changes
- `release-keys/minisign.pub` — placeholder (not a real key); replaced on release-day via the runbook.
- `release-keys/README.md` — runbook covering first-time keypair generation, rotation procedure, and suspected-leak response. Documents the `MINISIGN_PRIVATE_KEY` + `MINISIGN_PASSWORD` repo secrets.
- `RELEASE_NOTES_TEMPLATE.md` — references the three sidecar file types with verify commands.
[...]

PR #716, opened 2026-05-01
docs: v0.2 product README + screenshot (main)

Jiahui-Gu:
Same content as #714 (already merged to working at 0f965ff). Cherry-picked to main per user request that v0.2 launch README appear on all 3 branches (working, main, release/v0.2). Supersedes #693.

PR #1149, opened 2026-05-07
fix(e2e): repair groups-sidebar + scrollback baseline-red after #716 removed MainPane auto-create (Task #759)

Jiahui-Gu:
## Summary

Two e2e specs (`groups-sidebar.spec.ts`, `scrollback.spec.ts`) have been baseline-red on the working branch since Task #716 (commit 3157ca02) removed MainPane's auto-create-on-mount path. Both specs were written assuming the renderer would POST /api/sessions on its own at boot, then waited for `[data-testid^=\"sidebar-session-\"][data-active]` to appear. After #716 that auto-create is gone; bootstrap is now owned by `useBootstrap` (#670), which calls GET /api/sessions and hydrates the store. The mock daemons in both specs implemented only POST + DELETE, so GET fell through to 404, the store stayed empty, and the wait timed out.

CI run 25515319310 (PR #1148, ubuntu-latest) surfaced this because #1145 added ubuntu to the working-branch trigger matrix; the failures are unrelated to #1148's diff.

Reproduces on local Windows too — not OS-specific despite the manager spec calling out the ubuntu-only signal. Fix is the same root cause on both platforms.

## Fix

1. Both mock daemons now answer `GET /api/sessions` with `{ sessions: [] }` (matches `ListSessionsResponse` contract in `packages/shared/api.ts`).
2. Spec stage 1 clicks `[data-testid=\"sidebar-new-session\"]` to create the first session before asserting the row appears, mirroring how a user actually starts a session today. No assertions were mocked away; the contracts under test (multi-session add/setActive/close; per-session scrollback + lastSeq reconnect) are unchanged.

- lint: baseline-red on `origin/working` (`packages/daemon/test/persist.test.ts:20` + `packages/ui/test/BootstrapRefresh.test.tsx:24` unused-vars; both pre-existing, not from this PR's diff). Verified by stashing the diff via patch-file flow and re-running lint on clean origin/working — same 2 errors. Out of scope for #759.
- typecheck: PASS (exit 0)
- build: PASS (`npm run build`, exit 0, all packages built including frontend-web/frontend-tauri)
- e2e (changed specs only): PASS — `npx playwright test groups-sidebar.spec.ts scrollback.spec.ts` → `2 passed (17.4s)` on Windows
- Reverse-verified locally: pre-fix `groups-sidebar` reproduced the same 10s timeout on `[data-testid^=\"sidebar-session-\"][data-active]` Windows-locally (matches CI fail mode), confirming this is the actual ubuntu CI root cause and not an OS-specific timing artifact.

## Notes

- This is a test-fix PR per `feedback_local_windows_before_ci`: the manager prompt flagged the conflict (modify ubuntu-only test from a Windows host); local Windows reproduced the bug, so we verified the fix locally rather than trial-and-error on CI. CI will validate on the ubuntu/macos/windows matrix.

### DIFF
--- a/packages/e2e/tests/groups-sidebar.spec.ts
+++ b/packages/e2e/tests/groups-sidebar.spec.ts
@@ -9,11 +9,12 @@
 // WHY DEV-MODE (vite + mock daemon):
 //   The xterm/ws boot path is irrelevant to the sidebar contract under
 //   test, and the real `claude` CLI is not available on CI. We mock just
-//   the two endpoints the sidebar hits — POST /api/sessions and
-//   DELETE /api/sessions/:sid — and reject ws upgrades (the disconnect
-//   notice doesn't block anything in the sidebar). MainPane's bootstrap
-//   path will create the *first* session on its own; subsequent sessions
-//   are driven by the user clicking + New Session, which is what we test.
+//   the three endpoints the sidebar hits — GET /api/sessions (useBootstrap
+//   hydration), POST /api/sessions, and DELETE /api/sessions/:sid — and
+//   reject ws upgrades (the disconnect notice doesn't block anything in
+//   the sidebar). The bootstrap GET returns an empty list and the spec
+//   drives session creation via explicit + New Session clicks (MainPane's
+//   auto-create-on-mount path was removed in Task #716).
 //
 // SCOPE:
 //   Mirrors the dispatch spec's e2e checklist: + New Session twice, click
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
APPROVE (self-review, GH 不让 self-approve). Direction: right. Root cause 真 — #716 (3157ca02) 删 MainPane auto-create, useBootstrap 走 GET /api/sessions, 两 spec mock 只有 POST+DELETE 致 store 空 + 10s timeout. 修法 root-cause: mock 加 GET 对齐 ListSessionsResponse 契约, spec 主动 click +New Session 反映真实用户路径, 非 hide assertion. Scope 仅 2 spec, 不动 src. Reverse-verify dev 本地 Windows 真重现. CI e2e ubuntu+windows 双绿 (baseline-red 转绿). Squash merging.

PR #1191, opened 2026-05-08
feat(daemon+smoke): R-19 — CCSM_PTY_ENTRY env to route smoke PTY through bare shell (Task #53)

Jiahui-Gu:
## Why

research-52 locked the spec :48 red root cause: daemon's `defaultPtyFactory`
hard-coded `claude.cmd` / `claude` REPL. The REPL eats stdin into its prompt
buffer, so the smoke step that writes `echo hello-from-smoke\r` to the PTY
and then greps for `hello-from-smoke` in the OUTPUT tail can never satisfy
the assertion — the bytes never reach stdout. The other R-12..R-18 fixes
(token, ws, observability, testid) were all green; this is the last layer.

Fix at Layer 4 (test architecture) — not a glue patch on cmd-injection or
stdout-grep. The daemon learns one env var; smoke flips it; prod default is
byte-identical to pre-R-19.

## Design

`CCSM_PTY_ENTRY`:
- unset/empty → spawn `claude.cmd` (Windows) / `claude` (POSIX) with
  `--session-id <sid>` (create) or `--resume <sid>` (resume). Same as before.
- set → spawn that program with **empty argv**. `--session-id` / `--resume`
  are claude-CLI specific and would be garbage for `cmd.exe` / `/bin/sh`.

Smoke fixture sets `CCSM_PTY_ENTRY = win32 ? 'cmd.exe' : '/bin/sh'` only on
the tauri-release spawn — cf-worker / pages-dev / prod runs are untouched.
Bare shell echoes the input, satisfying spec :48 without coupling smoke to
`claude` REPL internals.

The cmd/argv selection logic is extracted into a pure `resolvePtyEntry()`
helper so both branches can be unit-tested without spawning subprocesses.

## Reverse-verify (env contract)

The new `resolvePtyEntry()` test suite locks both branches; flipping the
override toggle in the test changes `cmd` / `args` deterministically:
- env unset + create → `claude.cmd` + `['--session-id', sid]`
- env unset + resume → `claude.cmd` + `['--resume', sid]`
- env set → `<env>` + `[]` (regardless of mode/platform)

If a future change re-introduces a hard-coded `claude` path or leaks
`--session-id` into the override branch, these tests go red.

## Why smoke shouldn't couple to the product entry

The `claude` REPL is product surface — its prompt UX, prompt-buffer eating,
and color/streaming protocol can change at any time. A smoke that asserts
"a string written to the PTY appears in OUTPUT" should be testing the
**daemon → tunnel → ws → SPA round trip**, not the `claude` REPL's stdin
handling. Routing through `cmd.exe` / `/bin/sh` for the smoke run keeps the
test focused on the wire path and immune to claude UX churn.

## Files

- `packages/daemon/src/runtime.mts` — extract `resolvePtyEntry()`, wire env
- `packages/daemon/test/runtime.test.ts` — NEW, 9 tests both branches
- `packages/smoke/fixtures/orchestrator.ts` — set `CCSM_PTY_ENTRY` on tauri spawn

- lint: ✓ (exit 0; daemon + smoke)
- typecheck: ✓ (exit 0; daemon + smoke)
- build: skipped — no electron / webpack code touched; daemon tsc covers emit via typecheck
- unit tests: ✓ `pnpm --filter @ccsm/daemon test` → `Test Files 14 passed (14) / Tests 151 passed | 1 skipped (152)` in 11.27s (includes new 9 `resolvePtyEntry` tests)
- e2e: skipped — daemon-internal env switch + smoke fixture env; existing e2e harness covers no daemon-spawn path. Smoke itself is the integration harness this change targets; per task spec we do not run smoke locally (Windows zombie / LNK1104 risk).

## Out of scope

- Does not touch cf-worker / WS / SPA / R-12..R-18 fixes
- Does not add other hard-coded entry candidates (`/bin/bash`, `pwsh`)
- Does not change spec :44-50 (cmd.exe / /bin/sh echo satisfies the assertion)
- No retry / sleep / skip

### DIFF
--- a/packages/daemon/src/runtime.mts
+++ b/packages/daemon/src/runtime.mts
@@ -126,16 +126,51 @@ const PAUSE_QUEUE_CAP_BYTES = 1 * 1024 * 1024;
 
 // ---- Default PTY factory (real node-pty spawn of `claude`) --------------
 
+/**
+ * Task #53 (R-19) — pure resolver for PTY entry program + argv. Extracted from
+ * `defaultPtyFactory` so tests can lock both branches without spawning a real
+ * subprocess.
+ *
+ * Prod path: `CCSM_PTY_ENTRY` unset/empty → `claude.cmd` (Windows) or `claude`
+ * (POSIX), with `--session-id <sid>` / `--resume <sid>` synthesized from the
+ * spawn mode (Spike #665: `--resume` must pair with original cwd).
+ *
+ * Override path: `CCSM_PTY_ENTRY` set → use that program with empty argv.
+ * `--session-id` / `--resume` are claude-CLI specific and would be nonsense
+ * for a bare shell. research-52 lock-in: smoke spec :48 fails because the
+ * `claude` REPL eats stdin into its prompt buffer so `echo hello-from-smoke\r`
+ * never round-trips through the PTY's stdout tail; routing smoke through a
+ * bare shell makes the shell's own echo satisfy the assertion without coupling
+ * the smoke harness to claude internals. Defaulted off so prod behavior is
+ * byte-identical to pre-R-19.
+ */
+export function resolvePtyEntry(
+  ptyEntryEnv: string | undefined,
+  mode: PtySpawnMode,
+  sid: string,
+  isWindows: boolean,
+): { cmd: string; args: string[] } {
+  const useOverride = typeof ptyEntryEnv === 'string' && ptyEntryEnv.length > 0;
+  if (useOverride) {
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
Reviewer verdict: APPROVE. Direction right (Layer-4 test architecture fix, prod byte-identical). resolvePtyEntry pure helper + 9-test matrix locks both branches; smoke override scoped to tauri-release env; no cf-worker/WS/SPA/spec touched; no retry/sleep. Local checks quotes real vitest output. CI green (windows 1m8s, ubuntu 39s).

PR #1193, opened 2026-05-08
fix(core+spa): R-21 — sendInput buffer-until-open + data-ws-state (Task #61)

Jiahui-Gu:
## Summary (Task #61, smoke R-21)

research-60 locked the keystroke-drop root cause: in cloud-mode happy-path, xterm `onData` calls `runtime.sendInput(sid, data)` synchronously after the user types. During the ~340ms window between `await api.createSession(tok)` returning and the per-session ws reaching OPEN, `WsClient.sendInput` sees `readyState !== OPEN` and silently drops the frame. 22 chars were lost on a single smoke run. **Real users typing fast at session start hit the same bug** — this is a production drop, not just a test artefact.

### Layer 1 — protocol layer fix (production bug)

`packages/core/src/runtime/session-runtime.ts` `sendInput` no longer forwards blindly to `WsClient.sendInput`. It now:

- Looks up the entry. Unknown sid → no-op (caller already gates on `activeSidRef`).
- If `client.getStatus() === 'attached'` → forward straight to ws.
- Otherwise → push the data string onto `entry.pendingInput` (new field on `SessionRuntimeEntry`).

`handleStatusChange` flushes the queue verbatim, in arrival order, on the `attached` edge. **No coalescing** — every user keystroke remains a distinct INPUT frame, so the daemon's PTY sees the same byte stream the user typed.

### Layer 4 — test contract (`data-ws-state`)

`packages/ui/src/components/MainPane.tsx` reads `state.sessionStatuses[activeSid]` and renders a new `data-ws-state` attribute on the `terminal-pane` root:

| WsStatus       | data-ws-state |
|----------------|---------------|
| (no activeSid) | `closed`      |
| `attached`     | `open`        |
| `disconnected` / `exited` | `closed` |
| `idle` / `connecting` | `connecting` |

Smoke spec (`packages/smoke/tests/s3-happy-path.spec.ts`) gains a wait between click and `keyboard.type`:

```ts
await expect(page.getByTestId('terminal-pane')).toHaveAttribute(
  'data-ws-state', 'open', { timeout: 10_000 },
);
```

Click target unchanged. The buffer fix already protects against the race, but the explicit gate makes the smoke deterministic instead of relying on PTY echo masking out-of-order timing.

## Reverse-verify (sendInput buffer)

The new unit test `sendInput buffers keystrokes typed before ws OPEN, flushes in order on attached`:

- pre-open `sendInput('a' / 'b' / 'cd')` → asserts `ws.sent.length === 0` and `entry.pendingInput === ['a','b','cd']`.
- `ws.open()` → asserts `ws.sent.length === 3` with payloads `'a','b','cd'` in order.
- post-attached `sendInput('e')` bypasses the queue → asserts `ws.sent[3].payload === 'e'` and queue empty.

Without the fix, all four sendInputs that pre-date `ws.open()` would call into `WsClient.sendInput` directly and the assertion `ws.sent.length === 0` already fails (drop is silent in `WsClient` so the state would just stay empty — but `pendingInput` would never accumulate either, so the queue assertion fails). Mix test (`N=3 pre-open + M=2 post-open`) similarly locks order.

- lint: skipped (no global lint script in this repo; ts files are typechecked below)
- typecheck: ✓ `@ccsm/core` + `@ccsm/ui` + `@ccsm/smoke` all exit 0
- build: ✓ `@ccsm/shared` + `@ccsm/core` exit 0
- unit tests:
  - `pnpm --filter @ccsm/core test` → 3 files, 54 tests passed (16 in session-runtime.test.ts including 3 new buffer tests)
  - `pnpm --filter @ccsm/ui test` → 5 files, 40 tests passed (8 in MainPane.strictmode.test.tsx including 4 new data-ws-state tests)
- e2e/smoke: skipped per Task spec (do NOT run smoke locally — single-pool isolation; CI matrix will run smoke).

## Files

- `packages/core/src/runtime/types.ts` — add `pendingInput: string[]` to `SessionRuntimeEntry`.
- `packages/core/src/runtime/session-runtime.ts` — buffer-until-open in `sendInput`; flush in `handleStatusChange` on attached edge; init `pendingInput: []` in `attach`.
- `packages/core/test/session-runtime.test.ts` — 3 new buffer tests.
- `packages/ui/src/components/MainPane.tsx` — read activeStatus from store, render `data-ws-state` via `wsStateAttr` helper.
- `packages/ui/test/MainPane.strictmode.test.tsx` — 4 new tests covering closed/connecting/open/exited mappings.
- `packages/smoke/tests/s3-happy-path.spec.ts` — wait `data-ws-state="open"` before typing.

## Out of scope

- cf-worker / daemon / `ws/client.ts` untouched.
- Click target unchanged (the buffer fix is the production fix; the smoke gate is belt-and-braces).
- No retry/sleep, no `keyboard.type` delay change.

### DIFF
--- a/packages/core/src/runtime/session-runtime.ts
+++ b/packages/core/src/runtime/session-runtime.ts
@@ -145,6 +145,7 @@ export class SessionRuntime {
       hasEverAttached: false,
       pendingWrites: 0,
       paused: false,
+      pendingInput: [],
     };
     this.entries.set(sid, entry);
     this.openWs(entry, token);
@@ -177,9 +178,31 @@ export class SessionRuntime {
     this.entries.delete(sid);
   }
 
-  /** Send INPUT to the session's ws (no-op if not attached / not OPEN). */
+  /**
+   * Send INPUT to the session's ws.
+   *
+   * Task #61 (R-21) — buffer-until-open: keystrokes that arrive while the
+   * ws is `connecting` (or its `client` slot is briefly null between
+   * close and reconnect) are queued on `entry.pendingInput`, in arrival
+   * order, and flushed verbatim on the `attached` status edge. Without
+   * this buffer, the ~340ms createSession→ws-OPEN window in cloud-mode
+   * smoke silently drops every keystroke the user typed before xterm's
+   * onData saw an open socket — research-60 confirmed 22 dropped chars on
+   * a single happy-path run.
+   *
+   * If `sid` is unknown (no runtime entry) we drop on the floor — there's
+   * nothing to attach the queue to and the caller (xterm onData) already
+   * gates on `activeSidRef`. The OPEN-but-no-entry case is impossible.
+   */
   sendInput(sid: string, data: string): void {
-    this.entries.get(sid)?.client?.sendInput(data);
+    const entry = this.entries.get(sid);
+    if (!entry) return;
+    const client = entry.client;
+    if (client && client.getStatus() === 'attached') {
+      client.sendInput(data);
+      return;
+    }
+    entry.pendingInput.push(data);
   }
 
   /** Send RESIZE to the session's ws. */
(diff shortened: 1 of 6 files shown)


Jiahui-Gu:
APPROVE (self-review, same-token).

Layer 1 right: protocol-layer fix in SessionRuntime.sendInput is the correct architectural layer (not a WsClient hack, not a smoke sleep). data-ws-state is the right Layer 4 contract.

Layer 2: 7/7 except buffer is unbounded — types.ts comments admit 'unbounded by design'. Not blocking (token-boot only buffers a handful, ws-stuck path will surface via data-ws-state=connecting), but worth a followup to add a sanity cap so a stuck-connecting ws can't grow pendingInput without bound.

Reverse-verify logic sound. Tests: order + N=3+M=2 mix + unknown-sid + 4 ws-state mappings. ubuntu green, windows pending; base=working (no required checks).

PR #1203, opened 2026-05-09
[research-only] R-30 attempt + pushback (Task #87)

Jiahui-Gu:
## Status: research-only — DO NOT MERGE

This PR documents an attempted R-30 fix that **did not solve the bug**.
Spec's root-cause hypothesis was falsified by manual cloud-e2e re-run.
Pushing back to manager per memory `feedback_log_until_certain.md`
(no guessing-fixes without evidence) and dev role §5 (push back when
stuck).

See `research-report-task-87.md` (added in this PR) for the full
write-up. Highlights below.

## What was attempted (commit 5aee8576)

- `daemon_mgr.rs`: `CREATE_NO_WINDOW` → `CREATE_NEW_CONSOLE` so the
  daemon owns its own console.
- `packages/daemon/src/hide-console-windows.mts` (new): koffi-loaded
  `user32!ShowWindow(GetConsoleWindow, SW_HIDE)` at daemon startup.
- `packages/daemon/src/index.mts`: invoke hide helper top of `main()`.
- `package.json` + lockfile: add `koffi ^2.16.2`.

Build: `cargo check` ✓, `pnpm --filter @ccsm/daemon build` ✓.

## Why it didn't work

Manual `pnpm tauri dev` + `cd tools/cloud-e2e && npx playwright test`:

```
[daemon stderr] [ccsm] hide-console: hid console window         ← hide ok
[daemon-mgr] handshake ok port=56905                            ← daemon ok
[daemon stderr] [ccsm-pty-forensics] spawn=...sid=9d96ea15...   ← spawn 1
[daemon stderr] [ccsm-pty-forensics] spawn=...sid=ae2c773a...   ← spawn 2
[daemon stderr] Error: AttachConsole failed                     ← spawn 2 fails
[daemon stderr]     at conpty_console_list_agent.js:13:26
```

Spawn #1 succeeds, spawn #2 fails — **identical to the R-29 baseline
forensics**. cloud-e2e two-tab harness still fails at the second tab's
`Terminal input` click (15s timeout, xterm has no PTY output to render).

So the differentiator is NOT "daemon has/has-no console at startup".
It must be some state mutation inside the daemon **between spawn 1
returning and spawn 2 starting**.

## Real root-cause hypothesis (NEEDS R-31 forensics)

**H1**: `CreatePseudoConsole` for spawn #1 binds the daemon process to
HPCON#1's console group at a level deeper than `FreeConsole` can
release. The helper subprocess for spawn #2 inherits that binding;
[...]

PR #1204, opened 2026-05-09
log(daemon): R-31 — koffi console-state probes around pty spawn (Task #89)

Jiahui-Gu:
## Summary

Phase 1 forensics-only follow-up to R-29 (PR #1202, merged) and R-30 (PR #1203, research-only — Plan A failed). Adds two koffi-bound Win32 read probes (`GetConsoleWindow`, `GetConsoleProcessList`) around `defaultPtyFactory()`'s `nodePty.spawn` so that the next `tauri dev` cloud-e2e run produces enough state-deltas to distinguish H1 vs H2 (see Background).

**No PTY behavior changed.** No retries, no AttachConsole try/catch, no spawn-flag changes. Pure observability per memory `feedback_log_until_certain.md`: when evidence insufficient, add observability first, do NOT guess-fix.

## Background — what R-30 falsified

R-30 attempted Plan A (Tauri `CREATE_NEW_CONSOLE` + daemon-side `ShowWindow(GetConsoleWindow, SW_HIDE)` via koffi) and **failed**: spawn #2 still throws `AttachConsole failed` in `conpty_console_list_agent.js:13`, identical to R-29 baseline. That falsifies R-30's root-cause hypothesis (daemon has no console at startup → first spawn implicitly AllocConsole-binds → second spawn cannot AttachConsole). The differentiator is NOT daemon's startup console state.

Two surviving hypotheses:

- **H1** — `CreatePseudoConsole` for spawn #1 binds the daemon process to HPCON#1's console group at a level deeper than `FreeConsole` releases. Helper subprocess for spawn #2 inherits that binding. `AttachConsole(pid2)` returns ERROR_ACCESS_DENIED because daemon (parent) is still bound to HPCON#1.
- **H2** — `getConsoleProcessList` itself has destructive side effects on daemon console state.

H1 vs H2 cannot be told apart from R-29's existing forensics dump (it only logs TTY flags, not the actual hwnd / process list).

## What this PR adds

`packages/daemon/src/console-probe.mts` (new) — koffi wrapper exporting:
- `getConsoleHwnd(): number | null` — wraps `kernel32!GetConsoleWindow`, returns hwnd as integer or null.
- `getConsoleProcessList(): { count: number; pids: number[] }` — wraps `kernel32!GetConsoleProcessList(buf, len)` with adaptive len (start 32, retry once if larger needed).
- Stubs for non-Windows so importing on darwin/linux is safe.

`packages/daemon/src/runtime.mts` — `defaultPtyFactory()`:
- Module-scoped `r31SpawnIndex` counter, incremented per spawn.
- **Before** `nodePty.spawn`: call both probes, log `[r31-pre] sid=<sid> spawn_index=<n> hwnd=<hex|null> proc_count=<n> proc_pids=[...]`.
- **After** `nodePty.spawn` returns (no try/catch): same probe, log `[r31-post] ...`.

`packages/daemon/package.json` + `pnpm-lock.yaml` — add `koffi ^2.16.2` to daemon dependencies (same version already used by `tools/cloud-e2e` devDeps).

## Probe local sanity check

Verified the probe loads + returns plausible values on this Windows host (without any pty spawn yet):

```
$ node -e "(async () => { const m = await import('./packages/daemon/dist/console-probe.mjs'); console.log('hwnd=', m.getConsoleHwnd()); console.log('list=', JSON.stringify(m.getConsoleProcessList())); })()"
hwnd= null
list= {"count":3,"pids":[75592,80604,80696]}
```

(Detached bash → no GetConsoleWindow but GetConsoleProcessList enumerates the process group. This confirms koffi binds and the API call shape is correct.)

## What manager/reviewer should look for in the actual cloud-e2e run

When the user (or manager) next runs `cd packages/frontend-tauri && pnpm tauri dev` then `cd tools/cloud-e2e && npx playwright test`, the daemon stderr (forwarded by `daemon_mgr.rs` as `daemon-stderr` events / printed by the Tauri parent's `eprintln!`) will contain pairs of lines. Expected sequence for the failing harness:

```
[r31-pre]  sid=<sid1> spawn_index=1 hwnd=<H?>  proc_count=<P?> proc_pids=[...]
[r31-post] sid=<sid1> spawn_index=1 hwnd=<H?>  proc_count=<P?> proc_pids=[...]
[r31-pre]  sid=<sid2> spawn_index=2 hwnd=<H?>  proc_count=<P?> proc_pids=[...]
Error: AttachConsole failed             ← from conpty_console_list_agent.js:13
```

(spawn_index 3 may also appear if the harness retries.)

### Decision rules

| Pattern | Conclusion |
|---|---|
| `hwnd` flips `null → non-null` between pre1 and post1, AND `proc_count` jumps `0 → ≥2` | **H1 confirmed** — HPCON#1 binds daemon parent. Spawn #2's helper inherits binding. |
| `hwnd` is the **same** non-null value at post1 and pre2 (daemon stayed bound between spawns) | **H1 confirmed** (binding persists) |
| `hwnd` / `proc_count` mutate between probe calls *without* a spawn happening between them | **H2 confirmed** — `GetConsoleProcessList` itself has side effects |
| Both probes return identical state across pre1/post1/pre2 yet AttachConsole still fails | both hypotheses **inconclusive** — escalate to deeper trace (e.g. ETW console events) |

This PR does NOT make the call — it adds the data to make the call. R-32 (whoever picks it up) reads the daemon stderr table from a real run, fills in the table, and writes the conclusion.

## Why this is not glue per memory `feedback_fix_arch_not_glue.md`

This PR adds **observability**, not control flow. No retries, no error-swallowing, no `.skip`. Memory `feedback_log_until_certain.md` explicitly carves this out: "加 log 让事实暴露, retry 让事实掩盖。加 log 不算胶水."

- lint: ✓ `pnpm --filter @ccsm/daemon lint` exit 0
- typecheck: ✓ `pnpm --filter @ccsm/daemon typecheck` exit 0
- build: ✓ `pnpm --filter @ccsm/daemon build` exit 0 (after `@ccsm/shared` build)
- unit tests: ✓ `pnpm --filter @ccsm/daemon test` — 14 files / 154 passed / 1 skipped, 11.28s
- e2e: skipped — log-only diff, does not change PTY behavior. The actual `[r31-pre]/[r31-post]` lines need a real `tauri dev` + cloud-e2e session (token + cf tunnel + claude CLI) which this dev pool can't provide stably; following R-29's same approach (PR #1202 also skipped e2e). PR body lists expected log shape so manager/reviewer can read the eventual run output directly.

## Out of scope (deliberately not touched)

- `daemon_mgr.rs` — R-30 attempted CREATE_NEW_CONSOLE change is reverted (base = `origin/working` = `d00c8eea`); no Rust changes here.
- `hide-console-windows.mts` from R-30 — not present in this PR (not needed for forensics).
- `index.mts` `main()` — no startup hide-console call.
- Cargo.toml — no Rust dep changes.
- node-pty version pin / patch — task spec forbids.
- skip / xfail tests — task spec forbids.
- any try/catch around `nodePty.spawn` — would mask the diagnostic.

## Files

- `packages/daemon/src/console-probe.mts` (new, 130 LOC)
- `packages/daemon/src/runtime.mts` (modified, +33/-0 around the existing R-29 forensics block)
- `packages/daemon/package.json` (added `koffi: ^2.16.2`)
- `pnpm-lock.yaml` (regen — only the daemon importer line + 2 lines for the koffi entry under `packages/daemon`)

### DIFF
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -18,6 +18,7 @@
   "dependencies": {
     "@ccsm/shared": "workspace:*",
     "better-sqlite3": "^11.10.0",
+    "koffi": "^2.16.2",
     "node-pty": "^1.1.0",
     "ws": "^8.20.0"
   },
--- a/packages/daemon/src/console-probe.mts
+++ b/packages/daemon/src/console-probe.mts
@@ -1,0 +1,133 @@
+// Task #89 (R-31) — koffi-based Win32 console-state probes for PTY spawn
+// forensics. R-30 (Plan A: Tauri CREATE_NEW_CONSOLE + daemon-side
+// ShowWindow(GetConsoleWindow, SW_HIDE) via koffi) failed: spawn #2 still
+// hits `AttachConsole failed` in node-pty's conpty_console_list_agent.js,
+// identical to R-29 baseline. That falsifies R-30's root-cause hypothesis
+// (daemon has no console at startup → first spawn implicitly AllocConsole-
+// binds → second spawn cannot AttachConsole).
+//
+// Two surviving hypotheses (PR #1203 body):
+//   H1 — CreatePseudoConsole for spawn #1 binds the daemon process to
+//        HPCON#1's console group at a level deeper than FreeConsole
+//        releases. Helper subprocess for spawn #2 inherits that binding;
+//        AttachConsole(pid2) returns ERROR_ACCESS_DENIED because daemon
+//        (parent) is still bound to HPCON#1.
+//   H2 — getConsoleProcessList itself has destructive side effects on
+//        daemon console state.
+//
+// To distinguish H1 vs H2 we expose two pure read probes from runtime.mts:
+//   - getConsoleHwnd(): GetConsoleWindow → integer hwnd or null
+//   - getConsoleProcessList(): GetConsoleProcessList → { count, pids }
+//
+// Both stubs no-op on non-Windows so importing this module on darwin/linux
+// is safe.
+
+import { createRequire } from 'node:module';
+
+interface KoffiLib {
+  func(signature: string): (...args: unknown[]) => unknown;
+}
+
+interface KoffiModule {
+  load(name: string): KoffiLib;
+  // koffi's typed-array / out-buffer support
+  alloc(type: string, count: number): Uint8Array;
+  // Note: we use Buffer.alloc for the DWORD array argument since koffi
+  // accepts Node Buffers as raw output buffers when the C signature is a
+  // pointer.
+}
+
+let cached: {
+  getConsoleWindow: (() => number) | null;
+  getConsoleProcessList: ((buf: Buffer, len: number) => number) | null;
+} | null = null;
+
+function getApis(): {
+  getConsoleWindow: (() => number) | null;
+  getConsoleProcessList: ((buf: Buffer, len: number) => number) | null;
+} {
+  if (cached) return cached;
+  if (process.platform !== 'win32') {
+    cached = { getConsoleWindow: null, getConsoleProcessList: null };
+    return cached;
+  }
+  try {
+    const requireCjs = createRequire(import.meta.url);
+    const koffi = requireCjs('koffi') as KoffiModule;
+    const kernel32 = koffi.load('kernel32.dll');
+    // HWND GetConsoleWindow(void)
+    const fnGetConsoleWindow = kernel32.func('void* __stdcall GetConsoleWindow()') as () => unknown;
+    // DWORD GetConsoleProcessList(LPDWORD lpdwProcessList, DWORD dwProcessCount)
+    const fnGetConsoleProcessList = kernel32.func(
+      'uint32 __stdcall GetConsoleProcessList(_Out_ uint32 *lpdwProcessList, uint32 dwProcessCount)',
+    ) as (buf: Buffer, len: number) => number;
+    cached = {
+      getConsoleWindow: () => {
+        const v = fnGetConsoleWindow() as number | bigint | null;
+        if (v === null || v === undefined) return 0;
+        if (typeof v === 'bigint') return Number(v);
+        return v as number;
+      },
+      getConsoleProcessList: fnGetConsoleProcessList,
+    };
+  } catch {
+    cached = { getConsoleWindow: null, getConsoleProcessList: null };
+  }
+  return cached;
+}
+
+/**
+ * Wraps Win32 GetConsoleWindow. Returns hwnd as integer (may be large on
+ * x64) or null when there is no attached console / call failed / non-Windows.
+ */
+export function getConsoleHwnd(): number | null {
+  const { getConsoleWindow } = getApis();
+  if (!getConsoleWindow) return null;
+  try {
+    const hwnd = getConsoleWindow();
+    if (!hwnd || hwnd === 0) return null;
+    return hwnd;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Wraps Win32 GetConsoleProcessList. Returns the count and the list of
+ * pids attached to the caller's console. Adaptive: starts with a 32-slot
+ * buffer; if the API reports it needs more, retries once with the reported
+ * size. Returns { count: 0, pids: [] } on non-Windows / failure / no
+ * attached console.
+ */
+export function getConsoleProcessList(): { count: number; pids: number[] } {
+  const { getConsoleProcessList: fn } = getApis();
+  if (!fn) return { count: 0, pids: [] };
+  try {
+    const initialLen = 32;
+    const buf = Buffer.alloc(initialLen * 4); // DWORD = 4 bytes
+    const ret = fn(buf, initialLen);
+    if (!ret || ret === 0) {
+      return { count: 0, pids: [] };
+    }
+    if (ret <= initialLen) {
+      const pids: number[] = [];
+      for (let i = 0; i < ret; i++) {
+        pids.push(buf.readUInt32LE(i * 4));
+      }
+      return { count: ret, pids };
+    }
+    // Need a bigger buffer — retry once with reported size.
+    const buf2 = Buffer.alloc(ret * 4);
+    const ret2 = fn(buf2, ret);
+    if (!ret2 || ret2 === 0 || ret2 > ret) {
+      return { count: ret, pids: [] };
+    }
+    const pids: number[] = [];
+    for (let i = 0; i < ret2; i++) {
+      pids.push(buf2.readUInt32LE(i * 4));
+    }
+    return { count: ret2, pids };
+  } catch {
+    return { count: 0, pids: [] };
+  }
+}
--- a/packages/daemon/src/runtime.mts
+++ b/packages/daemon/src/runtime.mts
@@ -17,7 +17,13 @@ import { createRequire } from 'node:module';
 
 import { encodeExit, encodeFrame, FrameType } from '@ccsm/shared';
 
+import { getConsoleHwnd, getConsoleProcessList } from './console-probe.mjs';
 import { RingBuffer } from './ring.mjs';
+
+// Task #89 (R-31): module-scoped counter; incremented per defaultPtyFactory
+// invocation so [r31-pre] / [r31-post] log lines can be correlated across
+// spawn 1, 2, 3 within a single daemon process lifetime.
+let r31SpawnIndex = 0;
 
 // ---- Public types -------------------------------------------------------
 
@@ -207,6 +213,23 @@ const defaultPtyFactory: PtyFactory = (opts) => {
     // call it from JS without FFI, so we proxy via TTY flags.
   }));
   /* eslint-enable no-console */
+  // Task #89 (R-31) — in-process Win32 probes around node-pty spawn to
+  // distinguish H1 (CreatePseudoConsole binds daemon to HPCON#1's console
+  // group) vs H2 (GetConsoleProcessList itself has destructive side effects
+  // on daemon console state). Pre-probe runs BEFORE spawn; post-probe runs
+  // AFTER (no try/catch). On non-Windows the probes return null/empty.
+  const r31Idx = ++r31SpawnIndex;
+  /* eslint-disable no-console */
+  if (isWindows) {
+    const preHwnd = getConsoleHwnd();
+    const preList = getConsoleProcessList();
+    console.error(
+      `[r31-pre] sid=${opts.sid} spawn_index=${r31Idx} ` +
+        `hwnd=${preHwnd === null ? 'null' : '0x' + preHwnd.toString(16)} ` +
+        `proc_count=${preList.count} proc_pids=[${preList.pids.join(',')}]`,
+    );
+  }
+  /* eslint-enable no-console */
   const pty = nodePty.spawn(cmd, args, {
     name: 'xterm-256color',
     cols: opts.cols,
@@ -215,6 +238,17 @@ const defaultPtyFactory: PtyFactory = (opts) => {
     env: process.env as Record<string, string>,
     useConpty: true,
   });
+  /* eslint-disable no-console */
+  if (isWindows) {
+    const postHwnd = getConsoleHwnd();
+    const postList = getConsoleProcessList();
+    console.error(
+      `[r31-post] sid=${opts.sid} spawn_index=${r31Idx} ` +
+        `hwnd=${postHwnd === null ? 'null' : '0x' + postHwnd.toString(16)} ` +
+        `proc_count=${postList.count} proc_pids=[${postList.pids.join(',')}]`,
+    );
+  }
+  /* eslint-enable no-console */
   return {
     write: (d) => pty.write(d),
     resize: (c, r) => pty.resize(c, r),
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -60,6 +60,9 @@ importers:
       better-sqlite3:
         specifier: ^11.10.0
         version: 11.10.0
+      koffi:
+        specifier: ^2.16.2
+        version: 2.16.2
       node-pty:
         specifier: ^1.1.0
         version: 1.1.0


Jiahui-Gu:
R-31 forensics-only LGTM (self-approve blocked, squash merging).

Layer 1: right — pure observability around nodePty.spawn per feedback_log_until_certain.md; not glue per feedback_fix_arch_not_glue.md.

Layer 2:
- koffi GetConsoleWindow / GetConsoleProcessList bound correctly (stdcall, adaptive DWORD buffer)
- console-probe.mts non-Windows stubs safe
- r31SpawnIndex monotonic module-scoped
- no try/catch around nodePty.spawn
- daemon_mgr.rs untouched, hide-console-windows.mts not reintroduced, no .skip
- pnpm-lock churn = koffi-only
- CI green: ubuntu + windows + redline-guard
- Local checks quote real outputs (154 passed / 11.28s)

E2e skip per dev.md §5 + R-29 PR #1202 precedent — manager runs cloud-e2e post-merge to fill H1/H2 verdict.

Jiahui-Gu:
# Architecture Review — PTY/conpty/daemon spawn (Task #92)

Read-only audit triggered after R-30 (CREATE_NEW_CONSOLE Plan A) failed and R-31 (koffi probes) split into H1 (HPCON binding leak) vs H2 (`GetConsoleProcessList` side effects). Goal: surface design-level errors above H1/H2.

Scope: `packages/daemon/src/{runtime,index,ws,tunnel,http,console-probe}.mts`, `packages/frontend-tauri/src-tauri/src/daemon_mgr.rs`, `packages/frontend-tauri/src/{main,App}.tsx`, `node_modules/node-pty/lib/{windowsPtyAgent,windowsConoutConnection,conpty_console_list_agent}.js`. Read commits ea4fbe15 (R-31), 5aee8576 (R-30 attempt), 981eee48 (R-30 report).

## Findings

### F1: node-pty's helper-fork-per-kill model is fundamentally hostile to a multi-PTY single-process daemon

- **Risk: H**
- **Evidence**:
  - `node_modules/node-pty/lib/windowsPtyAgent.js:181-195` — `_getConsoleProcessList()` `child_process.fork`s `conpty_console_list_agent.js` on EVERY `pty.kill()` call.
  - `node_modules/node-pty/lib/conpty_console_list_agent.js:11-15` — the helper does `getConsoleProcessList(shellPid)`. The native module (per R-30 source-quote of `conpty_console_list.cc`:18-22) runs `FreeConsole()` then `AttachConsole(shellPid)`.
  - `node_modules/node-pty/lib/windowsPtyAgent.js:184` shows `agent` is `child_process.fork(...)` — a forked Node child INHERITS the parent's console handle on Windows. So the helper's "FreeConsole then AttachConsole" reaches across into the daemon's console state.
  - Microsoft docs: a Windows process can be attached to **at most one console at a time**. `AttachConsole` is documented to fail if the caller already has a console — even though the helper just freed it, the parent (daemon) still owns the inherited HPCON binding from spawn #1, which the helper inherits anew on each fork.
- **Hypothesis**: this whole module assumes a pattern where each `pty.spawn()` happens in its own short-lived host process (which is how the VS Code extension host works — a per-extension utility-process). When you run N PTYs inside one long-lived daemon, you get a per-spawn HPCON accumulating in the daemon, and every kill forks a helper that tries to AttachConsole from that polluted state. The "spawn #1 OK, spawn #2 fails" pattern is the textbook signature of "AttachConsole called twice without intervening FreeConsole at the parent level".
- **Suggested action**: This is the actual architectural error. Options to record (don't implement now): (a) audit upstream node-pty issue tracker for "multi-PTY single process" — this is a known class; (b) consider patch-package to either (i) skip the helper fork when `useConptyDll=true` is settable, or (ii) cache/serialize helper invocation. Note `windowsPtyAgent.js:140-160` shows the helper is bypassed entirely when `_useConptyDll=true` (the new ConPTY DLL path) — see F8.

### F2: PTY hosted in-process inside the same daemon that runs HTTP/WS/tunnel is the wrong process boundary

- **Risk: H**
- **Evidence**:
  - `packages/daemon/src/runtime.mts:164-217` `defaultPtyFactory` calls `nodePty.spawn(...)` directly inside the daemon process; the resulting HPCON handle, conout worker thread (`windowsConoutConnection.js:76` uses `worker_threads.Worker`), and any helper-fork pollution all live in the same process that also owns the WS server, tunnel client, SQLite db, and persist controller.
  - `packages/daemon/src/index.mts:119-122` — single process owns `createDaemonHttp` + `createRuntimeRegistry` + `attachWebSocket` + `TunnelClient`.
  - One bad PTY (the H1 HPCON leak) crashes the entire daemon (R-30 report confirms: `Error: AttachConsole failed` in helper kills the daemon process — taking down all other live sessions and the tunnel).
- **Hypothesis**: collapsing PTY hosting and protocol terminator into one process violates "isolation per session". A per-session sub-process model (daemon → child-per-PTY, IPC over pipe) would (a) make F1's HPCON-per-process problem disappear by construction (each child only ever spawns one HPCON in its lifetime), (b) bound blast radius of any node-pty crash, (c) let us kill the child process group to truly free HPCON state.
- **Suggested action**: record as v0.4 architectural debt. Spike candidate: prove a single-PTY-per-subprocess model fixes F1 with zero node-pty patching. Cost is non-trivial: IPC of OUTPUT/INPUT/RESIZE/EXIT frames between daemon and child. But IPC is a well-bounded engineering problem; HPCON state corruption is not.

### F3: renderer mints sid + drives spawn via two HTTP calls — no atomic "session:create" RPC, ENABLES the multi-spawn race

- **Risk: H**
- **Evidence**:
  - `packages/daemon/src/http.mts:263-296` POST `/api/sessions` does `randomUUID()` + `registry.spawn(sid, stub, 'create')` server-side. So sid is daemon-minted (good), BUT
  - `packages/daemon/src/http.mts:311-340` POST `/api/sessions/:sid/resume` re-spawns on a sid the renderer holds. The renderer can fire resume idempotently on tab focus, on reconnect, on visibility change.
  - Memory `project_v0_3_session_lifecycle_split_debt_2026_05_06.md` already records this as debt.
  - `tools/cloud-e2e/specs/two-tab-pairing.spec.ts` is the failing test — TWO tabs each fire create+resume, hitting `defaultPtyFactory` ≥2× in the same daemon. This is the EXACT input that triggers H1.
- **Hypothesis**: there's no single "session lifecycle owner". Two tabs racing each other through HTTP land in different microtask turns of the same event loop, both reaching `nodePty.spawn` back-to-back. This is the daily-driver code path that surfaces F1. With a session-owner abstraction (one place serializes spawn requests for a sid; idempotent on second request), even if F1/F2 weren't fixed, the multi-spawn pile-up would compress into a single conpty per logical session. The current bug needs F3 + F1 to coexist.
- **Suggested action**: even before tackling F1/F2, an idempotent server-side spawn dispatcher (lock per-sid, dedup concurrent spawns) would mask the bug for the two-tab scenario. Record as a v0.4 candidate; but note it only masks, doesn't fix. R-32 candidate: also instrument the request log to confirm two-tab harness fires 2 spawns in <1s.

### F4: `useConpty: true` is hard-coded with no fallback / opt-out — spec rejection of winpty was never re-validated

- **Risk: M**
- **Evidence**:
  - `packages/daemon/src/runtime.mts:216` `useConpty: true` is a literal, no env knob.
  - `node_modules/node-pty/lib/windowsPtyAgent.js:36-49` shows node-pty supports both conpty and winpty; the agent picks based on `_useConpty` + Windows build.
  - winpty path uses different process-group semantics (`getProcessList(_pid)` not the helper-fork) — see `windowsPtyAgent.js:169-178`. A winpty-mode kill never forks the AttachConsole helper.
  - Spec memory says winpty was rejected because "Win11 doesn't support" but node-pty still ships winpty native and uses it for Win10<18309.
- **Hypothesis**: the rejection might be stale. Even if winpty has rendering quirks vs conpty, having `CCSM_USE_CONPTY=0` as an experimental escape hatch would let R-32 run two-tab harness with winpty and isolate whether the bug is conpty-specific (likely yes, since helper-fork is conpty-only).
- **Suggested action**: R-32 candidate — wire `useConpty: process.env.CCSM_USE_CONPTY !== '0'`, manual two-tab repro on winpty mode. 5-line change, falsifiable answer in one session.

### F5: Tauri `CREATE_NO_WINDOW` choice has no documented design rationale and was reversed without rollback

- **Risk: M**
- **Evidence**:
  - `packages/frontend-tauri/src-tauri/src/daemon_mgr.rs:175-177`: `cmd.creation_flags(0x0800_0000); // CREATE_NO_WINDOW — avoid console flash`. Comment is "avoid console flash" — UX nit, not a security or correctness justification.
  - R-30 report (981eee48) confirms switching to `CREATE_NEW_CONSOLE + ShowWindow(SW_HIDE)` did NOT fix the bug. Also confirms the change was rolled BACK to `CREATE_NO_WINDOW` — but the rollback is not annotated; future readers see the literal `0x0800_0000` with no breadcrumb.
  - No `DETACHED_PROCESS` / `CREATE_NEW_PROCESS_GROUP` exploration in git history (`git log` keyword grep returns 0 hits).
- **Hypothesis**: the spawn-flag decision was a paste-from-sample, never formally evaluated for "does the daemon need a console at all, and if so, what kind". The reactive-fix R-30 jumped straight to `CREATE_NEW_CONSOLE` (the most invasive option) without first trying `DETACHED_PROCESS`. Per-Windows-docs, `DETACHED_PROCESS` would explicitly NOT inherit a console; this might force node-pty to take a different code path inside `nodePty.spawn` (since the daemon truly has no console at all, vs `CREATE_NO_WINDOW` which is "process is created without window but may still get a console attached at first console call").
- **Suggested action**: document the rationale as a comment on the literal flag (this PR is read-only — record in finding). R-32 candidate: try `DETACHED_PROCESS` (`0x0000_0008`) without the koffi hide hack and observe if the H1 signature changes. Cheap experiment.

### F6: F3 + F1 is one of the architectural causes; the THIRD INDEPENDENT HYPOTHESIS is that **session lifecycle is implicitly racy AT THE WS UPGRADE LAYER too**

- **Risk: H — this is the user's "third hypothesis" ask**
- **Evidence**:
  - `packages/daemon/src/ws.mts:386-407`: WS upgrade looks up `sessions.has(sid)` (the stub map), but does NOT check `registry.has(sid)`. If two tabs simultaneously fire create + WS upgrade in interleaved order, the stub may exist before the registry spawn completes.
  - `packages/daemon/src/runtime.mts:251-269`: `spawn` is synchronous from the caller's POV — but `nodePty.spawn` internally posts to a worker thread (`windowsConoutConnection.js:76`) and the helper-fork (`windowsPtyAgent.js:184`) is async. The `defaultPtyFactory` returns BEFORE the conout worker reports READY. There's no "PTY truly ready" gate.
  - `packages/daemon/src/runtime.mts:251-256`: idempotent-return on `existing && !exited` ONLY checks the runtime Map. Two near-simultaneous `spawn(sid)` calls (e.g. POST /sessions twice with same sid in a tunnel re-pair scenario, see `packages/daemon/src/tunnel.mts` `onBrowserAttach` per memory tunnel notes) WILL both pass the idempotency gate if they arrive between `runtime.get(sid)` and `runtime.set(sid, rt)` — no per-sid lock.
- **Hypothesis** (the third independent root cause beyond H1/H2): **the daemon's session lifecycle has no serialization primitive**. F3 talks about renderer-side multiple sids; this finding is that even for a SINGLE sid, two paths can race into `defaultPtyFactory` because:
  1. POST /api/sessions/:sid/resume has no per-sid mutex — concurrent requests both pass `registry.has(sid) === false` then both call `spawn`.
  2. The tunnel re-pair path (`packages/daemon/src/tunnel.mts` `onBrowserAttach`) also touches registry without coordinating with HTTP path.
  Even with H1/H2 fixed at node-pty level, these races would manifest as duplicate HPCON allocation or duplicate ring buffers per sid.
- **Suggested action**: this is an architecture-level race that's invisible in single-tab smoke (smoke uses `CCSM_PTY_ENTRY=cmd.exe` per F7 — bare-shell path). v0.4 candidate: introduce per-sid mutex in `RuntimeRegistry.spawn`. Or: make spawn return a Promise that other callers can await. R-32 candidate: add a counter log inside `defaultPtyFactory` to detect "spawn called for same sid twice" — would falsify or confirm immediately on the two-tab harness.

### F7: smoke vs cloud-e2e — smoke uses `CCSM_PTY_ENTRY=cmd.exe` which makes node-pty take a different conpty code path

- **Risk: M (test-coverage gap)**
- **Evidence**:
  - `packages/daemon/src/runtime.mts:147-162` `resolvePtyEntry`: when `CCSM_PTY_ENTRY` is set, it overrides `claude.cmd` with the env value AND drops `--session-id`/`--resume` args. Smoke harness sets `CCSM_PTY_ENTRY=cmd.exe`.
  - `claude.cmd` is itself a `.cmd` shim that internally launches `node` (the claude CLI is JS). So the conpty wraps a `cmd.exe` that wraps `node` — multiple intermediate process layers, each potentially holding a console handle.
  - `cmd.exe` direct (smoke path) wraps a single process — minimal console-state churn.
  - Only `tools/cloud-e2e/specs/two-tab-pairing.spec.ts` exercises `claude.cmd`; only `packages/smoke/tests/s3-happy-path.spec.ts` runs in CI smoke.
- **Hypothesis**: the difference between "spawn cmd.exe" (smoke, works) and "spawn claude.cmd" (cloud-e2e, fails) is NOT just "claude REPL eats stdin" (the documented R-19 reason for the env knob). It's also that `claude.cmd → node` introduces additional console-handle entanglement. Smoke literally cannot reproduce the F1 helper-AttachConsole bug because cmd.exe direct doesn't pile HPCON state the same way.
- **Suggested action**: record as test-coverage gap. R-32 candidate: extend smoke (or add a daemon-test fixture) that spawns a `.cmd` wrapper around node twice in the same daemon — would reproduce H1 in CI without needing claude CLI.

### F8: node-pty 1.1.0's `useConptyDll` flag bypasses the helper-fork entirely — never tried

- **Risk: M**
- **Evidence**:
  - `node_modules/node-pty/lib/windowsPtyAgent.js:140-160`: when `_useConptyDll` is true, the kill path skips `_getConsoleProcessList()` and uses `inSocket.destroy()` + `_ptyNative.kill()`. No helper fork → no AttachConsole → no F1 bug.
  - `packages/daemon/src/runtime.mts:210-217` does NOT pass `useConptyDll` to `nodePty.spawn`. node-pty defaults to `false`.
  - This is the modern conpty-DLL path Microsoft now recommends (loads conpty.dll dynamically, avoids the legacy AttachConsole quirk).
- **Hypothesis**: a one-line change — pass `useConptyDll: true` in the spawn opts — may sidestep F1 entirely. Untested, possibly has rendering / signal / Unicode tradeoffs. But cheaper to verify than F2's full subprocess refactor.
- **Suggested action**: R-32 STRONG candidate. 1-line diff. Manual two-tab repro answers it in one round. Should be tried BEFORE F2/F3 architectural work.

## Cross-references with R-31 H1/H2

- **F1** is the architectural framing of **H1**. H1 says "HPCON binding leaks across spawns inside daemon"; F1 names the specific vector — the helper-fork model assumes per-spawn process isolation that doesn't hold.
- **F8** directly addresses both H1 and H2 by removing the helper-fork code path entirely (the helper is what runs `GetConsoleProcessList` and triggers AttachConsole). If F8 works, the H1-vs-H2 split becomes moot.
- **F2** is also H1's structural fix (per-session subprocess = per-process HPCON, no leak possible).
- **H2** (`GetConsoleProcessList` itself destructive) is not directly addressed by any finding except F8 (which avoids it). If R-31 probes show H2 (`getConsoleProcessList` from daemon ALSO mutates state), F8 is the only fix that doesn't require node-pty patching, because the current `console-probe.mts` calls `GetConsoleProcessList` from daemon too — meaning R-31's own probe could conceivably be H2-amplifying. **Audit risk: R-31 probe may be polluting its own measurements.**
- **F6 is the third independent hypothesis (the user's ask)**: even if H1 and H2 were both false at node-pty level, the daemon's lack of per-sid spawn serialization would cause double-spawn artefacts in any concurrent scenario (two tabs, tunnel re-pair, refresh during spawn). This is a daemon-layer race, not a node-pty layer one — invisible to R-31's probes which only measure single-spawn state transitions.
- **F3** is H1's daily-driver trigger (renderer-side); **F6** is the daemon-side equivalent.
- **F5** is independent of H1/H2 — even with the right spawn flags, the bug would persist (R-30 already proved this).
- **F7** explains why H1/H2 are invisible to smoke — pure test-coverage observation.

## Out of scope

- Cloudflare worker / Pages function / DO ws routing (not PTY-domain).
- `tunnel.mts` re-pair logic beyond noting it touches registry (Task #81 already shipped that).
- HTTP auth / CORS / persist controller / KV db.
- PR diff review of #1204 — this is codebase-state audit, not PR review.
- Frontend-web (only frontend-tauri spawn path touches daemon spawn).
- Did NOT consult node-pty upstream issue tracker (web), did NOT call any Win32 API live.

Jiahui-Gu:
## R-31 evidence — Tauri + cloud-e2e two-tab harness (Task #91)

**Host**: Windows 11, pool-6 worktree, daemon HEAD `ea4fbe15` (R-31 merged).
**Setup**: `pnpm tauri dev` started daemon pid=8196, port=64013. cloud-e2e two-tab spec ran (failed at second tab terminal-input click — expected).

### Verbatim `[r31-pre]` / `[r31-post]` + crash lines

```
[daemon stderr] [ccsm-pty-forensics] spawn={"sid":"a972c55f-b109-4103-8d17-b64584c7c198","mode":"create","cmd":"claude.cmd",...,"pid":8196,"ppid":57440,"stdinIsTTY":false,"stdoutIsTTY":false,"stderrIsTTY":false,"useConpty":true}
[daemon stderr] [r31-pre]  sid=a972c55f-... spawn_index=1 hwnd=null proc_count=1 proc_pids=[8196]
[daemon stderr] [r31-post] sid=a972c55f-... spawn_index=1 hwnd=null proc_count=1 proc_pids=[8196]
[daemon stderr] [ccsm-pty-forensics] spawn={"sid":"fc25dfdf-...",...,"pid":8196,"ppid":57440,...}
[daemon stderr] [r31-pre]  sid=fc25dfdf-... spawn_index=2 hwnd=null proc_count=1 proc_pids=[8196]
[daemon stderr] [r31-post] sid=fc25dfdf-... spawn_index=2 hwnd=null proc_count=1 proc_pids=[8196]
...
[daemon stderr] C:\...\node-pty\lib\conpty_console_list_agent.js:13
[daemon stderr] var consoleProcessList = getConsoleProcessList(shellPid);
[daemon stderr]                          ^
[daemon stderr] Error: AttachConsole failed
[daemon stderr]     at Object.<anonymous> (...\node-pty\lib\conpty_console_list_agent.js:13:26)
```

### Verdict: **INCONCLUSIVE — both H1 and H2 falsified, new hypothesis emerges**

**H1 (CreatePseudoConsole binds daemon to HPCON#1's console group)** — NOT confirmed.
- Predicted: post-spawn-1 `hwnd != null` AND `proc_count >= 2`.
- Observed: post-spawn-1 `hwnd=null`, `proc_count=1` (only daemon itself, pid 8196).
- Daemon never gains a console window after the first ConPTY spawn at all.

**H2 (`getConsoleProcessList` itself has destructive side effects on daemon console state)** — NOT confirmed.
- Predicted: pre vs post values diverge because the probe call mutates state.
- Observed: pre and post are byte-identical for both spawn #1 and spawn #2. The probes are inert from the daemon's perspective.

### New hypothesis (H3) for manager

The daemon process **never has a console at all** (`hwnd=null`, `proc_count=1` from the very first probe). This is consistent with daemon-mgr launching the daemon with `CREATE_NO_WINDOW` / detached stdio (matches `stdinIsTTY=false, stdoutIsTTY=false, stderrIsTTY=false` in the forensics line).

The `AttachConsole failed` crash happens inside node-pty's `conpty_console_list_agent.js` — a **helper subprocess** node-pty spawns after the second ConPTY to enumerate console processes. That helper calls `AttachConsole(shellPid)` to attach to the spawned shell's console. It fails because **the helper inherits the daemon's empty/no-console state** and cannot attach to the shell's HPCON-managed console from a process that has no console of its own.

This shifts the question from "does spawn #1 leak HPCON binding into the daemon" (H1, falsified) to "why does node-pty's console-list agent need AttachConsole at all when the parent daemon is console-less, and why does the second spawn trigger it but the first doesn't". Likely root cause is in node-pty 1.1.0's conpty cleanup path on Windows (`conpty_console_list_agent.js` is invoked from `WindowsTerminal.js`'s `_close` / process-tracker logic).

### Suggested R-32 directions
1. Read node-pty `WindowsTerminal.js` to find the call site that spawns `conpty_console_list_agent.js` — identify what triggers it on the **second** spawn but not the first (likely an existing pty cleanup / process-list refresh tied to # of live ptys).
2. Probe inside the helper itself: log the helper subprocess's own `GetConsoleWindow()` / `GetConsoleProcessList()` BEFORE its `AttachConsole(shellPid)` call.
3. Test workaround: set node-pty `useConpty: false` (winpty fallback) — if crash disappears, confirms it's conpty's helper that's broken in console-less parents.

### Setup confirmed
- HEAD: `ea4fbe15 log(daemon): R-31 — koffi GetConsoleWindow + GetConsoleProcessList probes around pty spawn`
- `grep -l 'r31-pre' packages/daemon/dist/runtime.mjs` matched.
- 2 sessions spawned, 2 pre/post pairs captured, AttachConsole crash captured.

— Task #91 evidence-only run, no code changes.

PR #1224, opened 2026-05-09
feat(cf-worker): Task #136 — JWT routing middleware + CCSM_AUTH_MODE flag (S4-T5)

Jiahui-Gu:
## Summary
- Adds `CCSM_AUTH_MODE` vars toggle to cf-worker (`legacy` default / `jwt` opt-in) so per-user JWT routing can roll out without breaking the S3-era token flow that production + smoke still rely on.
- New `packages/cf-worker/src/auth/middleware.ts` exports `getAuthMode`, `getUserDoIdName`, `extractWebJwt`, `extractTunnelJwt`. `extractWebJwt` looks at the `ccsm.<jwt>` WS subprotocol first, then `Authorization: Bearer`; both helpers narrow on `kind` so a tunnel JWT presented at a browser path (or vice versa) is rejected.
- In `jwt` mode the worker routes into `TUNNEL.idFromName('user:<github_id>')`. /ws/default also injects `X-CCSM-Identity-Login` / `X-CCSM-Identity-Id` request headers so the DO echoes them inside the daemon hello frame (Task #133 wire format). TunnelDO internals (sid envelope, getBrowserSocket, hello-frame format) are otherwise untouched.
- /api/* and /token in `jwt` mode require Authorization Bearer; missing/invalid → 401 before any DO touch.

## Test plan
- [x] `npx vitest run` — 8 files, 121 tests passed (85 baseline + 36 new across `middleware.test.ts` and `jwtRouting.test.ts`)
- [x] `npm run typecheck` — exit 0
- [x] `npm run lint` — exit 0
- [x] Cross-user isolation asserted: alice JWT → `user:gh-alice` DO, bob JWT → `user:gh-bob` DO, including with a colliding `?sid=` query.
- [x] Legacy regression guard: all four routes (/ws, /tunnel, /api/*, /token) still hit `idFromName('default')` with no JWT in `legacy` mode and when `CCSM_AUTH_MODE` is unset.

- lint: ✓ (exit 0)
- typecheck: ✓ (exit 0)
- build: ✓ (cf-worker `npm run build` aliases to `tsc --noEmit`, exit 0)
- unit tests: cf-worker `vitest run` — `Test Files 8 passed (8)`, `Tests 121 passed (121)`, Duration 3.96s
- e2e: skipped — change is scoped to `packages/cf-worker/`; Electron e2e harness does not exercise the worker fetch handler (smoke only verifies legacy mode end-to-end via the daemon path, which is regressed-tested above).

### DIFF
--- a/packages/cf-worker/src/auth/middleware.ts
+++ b/packages/cf-worker/src/auth/middleware.ts
@@ -1,0 +1,134 @@
+/**
+ * S4-T5 (Task #136): JWT routing middleware for cf-worker.
+ *
+ * Two responsibilities:
+ *
+ *   1. **Auth-mode gating.** `CCSM_AUTH_MODE` env var (vars in wrangler.toml,
+ *      default `legacy`) gates whether browser ws / tunnel ws / REST API
+ *      requests must carry a verifiable JWT. `legacy` keeps the S3-era flow
+ *      (token in `Sec-WebSocket-Protocol`, no JWT, single shared TunnelDO
+ *      keyed by `'default'`). `jwt` requires the JWT; the github_id from the
+ *      verified claims is used to derive a **per-user TunnelDO id**
+ *      (`'user:<github_id>'`) so different GitHub users get isolated DO
+ *      instances.
+ *
+ *   2. **JWT extraction.** Browser-side: token is in the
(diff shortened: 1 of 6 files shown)


Jiahui-Gu:
LGTM. Direction: right (header injection 是正解, 不是越界 spec). Layer 2 全过. CI 全绿. Squash merging.

PR #1266, opened 2026-05-21
fix(release): allowlist required checks in CI gate to ignore self-pollution

Jiahui-Gu:
## Symptom

`daily-promote-release.yml`'s "re-check CI on new main" gate fails on PR heads where every real CI job is green. Failing reference: https://github.com/Jiahui-Gu/ccsm/actions/runs/26205515547 — the gate sees 3 bad check-runs and refuses to tag, even though the 8 real CI checks (`e2e × 3`, `lint + typecheck + test × 3`, `no-silent-drops`, `check-source`) are all success.

## Root cause: self-pollution

The walkback (`/commits/{merge_sha}/pulls` → PR head SHA → `/commits/{pr_head}/check-runs`) returns *every* check-run on that SHA. That includes the `promote working → main + tag` check-run that **this same workflow** registers when it runs. Live runs are `in_progress`; historical failed dispatches leave `failure` entries on the same SHA forever. The gate had no filter, so it counted those self-registered failures and aborted promotion.

Local snapshot of PR #1265 head (current main HEAD's PR): 11 check-runs total — 8 real CI all `success`, 3 `promote working → main + tag` all `failure`.

## Fix: explicit allowlist

Add a required `EXPECTED_CHECKS` env var (newline-separated, so names with spaces don't need quoting). The script filters check-runs to that allowlist *before* any pass/fail evaluation — anything outside the list (including this workflow's own check-run, and any unrelated future check-runs) is ignored. Missing required names now fail loud with a clear "missing required check" error and an "expected vs got" dump.

The same inlined logic in the "Wait + re-check CI on new main" step is patched the same way. Re-runs are handled by picking the latest matching check-run by `(completed_at, id)` for each expected name.

Why env var rather than positional arg: names like `lint + typecheck + test (macos-latest)` contain spaces and parentheses; newline-separated env var avoids quoting hell and keeps the call sites simple.

## Local simulation

Replicated the jq selection logic in Python against the live API for main HEAD `1bd47dfbd98dde2d970aa8af5fc607892bc520e8` (PR #1265, head `e0eb623f8985d109fc59f1cf1218c3c37aa6159b`):

```
PR #1265, head e0eb623f8985d109fc59f1cf1218c3c37aa6159b
Total check-runs on PR head: 11
All names: ['check-source', 'e2e (macos-latest)', 'e2e (ubuntu-latest)', 'e2e (windows-latest)',
            'lint + typecheck + test (macos-latest)', 'lint + typecheck + test (ubuntu-latest)',
            'lint + typecheck + test (windows-latest)', 'no-silent-drops',
            'promote working → main + tag']
Missing: []
Bad: []
Result: PASS
```

Without the filter the same SHA would FAIL (3 self-pollution `failure` entries). With it, gate passes.

## Files changed

- `.github/workflows/scripts/check-ci-via-pr.sh` — require `EXPECTED_CHECKS`, allowlist-filter then evaluate, latest-per-name dedup, log expected-vs-got.
- `.github/workflows/daily-promote-release.yml` — declare `EXPECTED_CHECKS` at job env (8 real CI names); update inlined re-check step to use the same allowlist.

## Not touched

- `release.yml`
- the `Wait for check-source on promote PR` step (it already targets `check-source` by name via API filter)
- "no associated PR → refuse" branch
- waiting / polling logic (only the evaluation is changed)
- version / package.json

## Bootstrap note

Per the workflow file's own header comment, cron schedules from the default branch (`main`). After this lands on `working` and is promoted to `main`, one manual `workflow_dispatch` is needed for cron to start picking up the fixed file.

### DIFF
--- a/.github/workflows/daily-promote-release.yml
+++ b/.github/workflows/daily-promote-release.yml
@@ -64,6 +64,23 @@ jobs:
       # Prefer RELEASE_TOKEN (PAT) so the auto-opened PR fires required
       # checks. Fall back to GITHUB_TOKEN for the read-only gates.
       GH_TOKEN: ${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}
+      # Allowlist of CI check-run names that gates require to be green on
+      # the PR head SHA. Newline-separated so names with spaces work.
+      # CRITICAL: this list MUST exclude the names of any check-run this
+      # workflow itself registers (e.g. "promote working → main + tag"),
+      # otherwise the gate self-poisons — the in-progress / historical-
+      # failed runs of this workflow attach check-runs to the same PR head
+      # and would fail the gate. See: failing run 26205515547.
+      # When adding a new required CI check, append it here.
+      EXPECTED_CHECKS: |
+        e2e (macos-latest)
+        e2e (ubuntu-latest)
+        e2e (windows-latest)
+        lint + typecheck + test (macos-latest)
+        lint + typecheck + test (ubuntu-latest)
+        lint + typecheck + test (windows-latest)
+        no-silent-drops
+        check-source
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -307,6 +324,8 @@ jobs:
           TIMEOUT_SECONDS: '1800'  # 30 min
         run: |
           set -euo pipefail
+          # Render allowlist as JSON array for jq.
+          expected_json=$(printf '%s\n' "$EXPECTED_CHECKS" | jq -R . | jq -s '[.[] | select(length > 0)]')
           deadline=$(( $(date +%s) + TIMEOUT_SECONDS ))
           while :; do
             # Find PR for this merge commit.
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
# Independent review — PR #1266

## Layer 1 verdict: **PASS** (direction C is correct; brittleness acceptable with current debuggability)

**Necessity / approach.** Direction (C) — explicit allowlist — is the right call. Two simpler-sounding alternatives don't actually work:

- **Filter by `app.slug == github-actions`**: doesn't help. The polluting check-run `promote working → main + tag` is *also* registered by github-actions (it's just another GHA workflow). Same slug, same app id.
- **Exclude self by workflow_id**: would require fetching `check_run.check_suite.id` then resolving each suite's workflow, doubling API calls; still wouldn't catch unrelated future workflows that happen to attach a check-run to the same SHA. Allowlist is the more conservative posture.
- **Drop walkback, query target workflow file directly**: re-architects the gate (would need `actions/runs?head_sha=...&workflow=ci.yml` per workflow), much bigger surgery for the same end state.

Allowlist is the minimum-blast-radius fix. Approved direction.

**Brittleness.** The 8-name list is a maintenance touchpoint, but:
- Worker added a CRITICAL comment block at job-env (lines 67–73) explaining the hazard and the "append here when adding a new required check" rule. Acceptable.
- Failure mode when out of sync is loud and self-explanatory: missing → `::error::CI gate failed (...): N missing required check(s)` plus a step-summary block listing **Expected**, **Got (in allowlist)**, **Got (all on PR head)** side-by-side. Verified in script lines 110–141 and 115–117 — debuggability is genuinely good.
- Reverse drift (someone removes a check from ci.yml without removing it here) is the only risky case; gate would block forever. But that's the desired safety direction — fail closed.

## Layer 2 issues

### High

**H1. Stale-completed-success beats fresh in-flight rerun (latent, low probability but real).** `check-ci-via-pr.sh` lines 88–96 sort by `[(.completed_at // ""), (.id // 0)]` ascending and take `last`. `null` completed_at coerces to `""`, which sorts BEFORE any non-empty timestamp. So if a check has an old completed-success entry **and** a freshly re-triggered in-progress entry on the same PR head, the OLD completed-success wins. Gate passes while the rerun is still running.
- Reproduced in local jq-equivalent simulation: id=77104473685 (completed success) chosen over id=99999999999 (in_progress, completed_at=null).
- In the standalone script's call sites (gates A and B before the promote merge), the PR head is static and reruns are unlikely between gate A and gate B, so practical risk is low. But the contract "latest run wins" is silently violated.
- Fix is one-line: sort key should treat null `completed_at` as max, not min, e.g. `sort_by([(.completed_at // "9999-99-99"), (.id // 0)])` or `sort_by([(.status == "completed"), (.completed_at // ""), (.id // 0)])` (in-progress first, then by completed_at). Recommend latter so in-progress always wins over older completed.
- **Not a merge blocker on its own** (current gate paths don't realistically hit it), but worth a follow-up commit before merge or a tracking task.

### Medium

**M2. PR body claim "11 check-runs total — 8 real CI all `success`, 3 `promote working → main + tag` all `failure`" is verified accurate.** Fetched `commits/e0eb623f.../check-runs --paginate`: exactly that distribution. PR body Local-simulation block matches reality.

**M3. `check-ci-via-pr.sh` line 36 — `EXPECTED_CHECKS` empty-check is correctly placed *after* `set -euo pipefail` (line 27), uses `${EXPECTED_CHECKS:-}` so `set -u` doesn't bite. Exit code 2 distinguishes config error from gate failure (1). Good.**

**M4. Job-level `env:` placement (line 67) is correct** — both call sites (lines 169, 178) plus the inlined re-check (line 328) inherit it. Verified.

### Low

**L5. `daily-promote-release.yml` line 328 — `printf '%s\n' "$EXPECTED_CHECKS" | jq -R . | jq -s '[.[] | select(length > 0)]'` is duplicated verbatim with `check-ci-via-pr.sh` line 41. Acceptable (DRY would mean factoring out a helper, which adds complexity for no real win), but worth a one-line code comment in the inlined version pointing at the script.**

**L6. Step-summary FAIL block (script lines 121–142) dumps `Got (all check-runs on PR head, for debugging)` — this is exactly the right level of detail for diagnosing self-pollution. Nice.**

**L7. shellcheck not run in CI for `.github/workflows/scripts/*.sh` (separate concern, pre-existing).** No new shell-quoting issues introduced; `"$EXPECTED_CHECKS"` correctly quoted at all use sites.

**L8. The 3-line `printf '%s\n' "$EXPECTED_CHECKS"` adds a trailing empty line which is then filtered by `select(length > 0)` — fine. But if someone writes `EXPECTED_CHECKS:  e2e (...)` (single line, leading spaces), it becomes one allowlist entry with leading whitespace and won't match. Not introduced by this PR; just noting.**

## Layer 3 verification — local jq-equivalent run against real data

Replicated the exact selection / pass-fail jq logic in Python against `commits/e0eb623f.../check-runs` (real PR #1265 head):

| Scenario | Result | Notes |
|---|---|---|
| **A. Real PR head with 3 self-pollution failures** | `missing=0 bad=0 → PASS` | 3 `promote working → main + tag` failures correctly ignored. Gate now passes where it previously failed. **Fix works.** |
| **B. Allowlist contains a name with no matching run** (added `extra-check`) | `missing=['extra-check'] → FAIL` | Loud, correct. |
| **C. All 8 real checks forced to `in_progress` (standalone script)** | `bad=8 → exit 1` | Standalone script correctly treats pending as bad — caller must already have waited. Matches docstring contract. |
| **F. All 8 in_progress (inlined workflow re-check)** | `missing=0 pending=8 bad=0` | Inlined logic correctly distinguishes pending (loop and wait) from bad — the outer `while :; do ... deadline` loop handles it. Matches design. |
| **D. Re-run dedup: old failure + new success on same name** | latest (newer completed_at) wins → `PASS` | Correct. |
| **G. Same completed_at, different ids** | higher id wins → deterministic | Correct. |
| **E2. Old completed-success + new in_progress (rerun in flight)** | OLD success wins | **Bug H1 above.** |

## Recommendation to manager

**APPROVE with optional follow-up.** Direction and main flow are sound, real-data simulation confirms the bug fix unblocks v0.2.4 nightly, and the failure-mode logging is good enough to make the maintenance touchpoint manageable. H1 is a latent edge case that the current call sites don't realistically exercise (gates run on a static PR head between merge attempts; reruns mid-gate are vanishingly rare), so it shouldn't block the urgent fix.

Two acceptable paths:

1. **Merge as-is, file follow-up task** for H1 sort-key fix (in-progress should beat older completed when picking "latest"). One-line jq change.
2. **Ask worker for one-commit amendment** before merge: change sort key in both files to `sort_by([(.status == "completed"), (.completed_at // ""), (.id // 0)])`. Trivial, low risk.

Given v0.2.4 nightly is blocked and tonight's cron is the next opportunity, **path (1) is preferred** — ship the fix, file the follow-up.

PR #1275, opened 2026-05-21
feat(sidebar): right-click reload session

Jiahui-Gu:
## Problem

User reports that some changes (env vars, claude config edits, prompt customizations, etc.) only take effect on a freshly-started claude process — but ccsm has no UI affordance to restart a session's pty without losing the session identity (id, name, cwd, group, transcript).

## Solution — approach A (no new IPC)

Reuse the existing primitives that already cover kill + respawn:

- `window.ccsmPty.kill(sid)` — already exposed (used by `deleteSession`)
- `usePtyAttach`'s spawn-on-null fallback — when the attach effect re-runs and `pty.attach` returns null, it falls through to `pty.spawn` then re-attaches

Wire them together via a new per-session `reloadNonce` in `sessionRuntimeSlice`:

1. Right-click ▸ Reload session → `reloadSession(sid)` action
2. Action `await`s `pty.kill(sid)`, then bumps `reloadNonce[sid]`
3. `usePtyAttach` reads `reloadNonce[sid]` and includes it in the attach-effect deps array → effect re-runs → spawn-on-null fallback brings up a fresh pty
4. Same path as the existing Retry button (which also re-runs the attach effect via an internal nonce), just externally triggered

claude transcript continuation is the SDK's job — the JSONL resolver picks the existing transcript via `--resume`, so the user sees the same conversation history when the process comes back up.

## Behavior decisions

- **No confirm dialog** — non-destructive (transcript preserved). Matches the J4 pattern of replacing modal confirms with reversible UX.
- **Lightweight info toast** — "Reloading <name>…" — provides feedback during the ~1-2s blank window between kill and re-attach (otherwise the user sees an unexplained xterm flash).
- **Always shown** — archived sessions can be reloaded too. The action only operates on the pty, which is decoupled from archive state.
- **Eager crash-dot clear** — drops any stale `disconnectedSessions[sid]` entry on reload start so the sidebar red dot doesn't linger across the reload window. The new pty's successful attach also clears it (`_clearPtyExit` in usePtyAttach), but the eager clear avoids a flash on the crashed-then-reload-to-recover path.
- **Tolerates kill failure** — IPC errors are swallowed (the pty may already be exiting); the nonce still bumps so the attach effect always re-runs.

## Diff scope

127 LOC across 7 files. No hot files touched (DiffView/ToolBlock/i18n parity all unchanged in mechanism — only adds new keys).

- `src/stores/slices/types.ts` — `reloadNonce: Record<string, number>` on State; `reloadSession: (sid) => Promise<void>` on Actions
- `src/stores/slices/sessionRuntimeSlice.ts` — initial state + action implementation
- `src/terminal/usePtyAttach.ts` — read nonce + add to deps array (defensive `?.` for tests that mock the store with a partial shape)
- `src/components/sidebar/SessionRow.tsx` — ContextMenuItem + toast
- `src/i18n/locales/{en,zh}.ts` — `sidebar.reloadSession` + `sidebar.reloadingSessionToast`

## Test plan

- [x] `npm run lint` clean (0 errors; 91 pre-existing warnings unchanged)
- [x] `tests/stores/slices/sessionRuntimeSlice.test.ts` — 5 new test cases for `reloadSession` (kill called, nonce bumps, disconnect entry cleared, kill-error tolerated, missing-bridge OK) + 5 pre-existing cases all green
- [x] `tests/terminal/usePtyAttach.test.tsx` — defensive nonce read keeps existing 9 cases passing
- [x] `tests/i18n-key-parity.test.ts` — both keys added to en + zh, parity preserved
- [ ] Reviewer optional: `scripts/probe-session-reload.mjs` — workshop probe that launches isolated ccsm + seeds a session + drives `reloadSession` and asserts the pty pid changes while the host element with the same `data-active-sid` is preserved (requires `npm run build` + working claude binary)

## Two pre-existing failing test files (unchanged by this PR)

`tests/electron-load-smoke.test.ts` (needs `npm run build` first — its own assertion message says so) and `electron/__tests__/db-hardening.test.ts` (native module). Verified failing on baseline `HEAD` before any changes via `git stash` round-trip.


Jiahui-Gu:
Reopening on `feat/right-click-reload-session` so CI picks it up — the worktree-prefixed branch name was filtered out by the workflow trigger.

PR #1277, opened 2026-05-21
feat(sidebar): right-click reload session

Jiahui-Gu:
## Problem

User reports that some changes (env vars, claude config edits, prompt customizations, etc.) only take effect on a freshly-started claude process — but ccsm has no UI affordance to restart a session's pty without losing the session identity (id, name, cwd, group, transcript).

## Solution — approach A (no new IPC)

Reuse the existing primitives that already cover kill + respawn:

- `window.ccsmPty.kill(sid)` — already exposed (used by `deleteSession`)
- `usePtyAttach`'s spawn-on-null fallback — when the attach effect re-runs and `pty.attach` returns null, it falls through to `pty.spawn` then re-attaches

Wire them together via a new per-session `reloadNonce` in `sessionRuntimeSlice`:

1. Right-click ▸ Reload session → `reloadSession(sid)` action
2. Action `await`s `pty.kill(sid)`, then bumps `reloadNonce[sid]`
3. `usePtyAttach` reads `reloadNonce[sid]` and includes it in the attach-effect deps array → effect re-runs → spawn-on-null fallback brings up a fresh pty
4. Same path as the existing Retry button (which also re-runs the attach effect via an internal nonce), just externally triggered

claude transcript continuation is the SDK's job — the JSONL resolver picks the existing transcript via `--resume`, so the user sees the same conversation history when the process comes back up.

## Behavior decisions

- **No confirm dialog** — non-destructive (transcript preserved). Matches the J4 pattern of replacing modal confirms with reversible UX.
- **Lightweight info toast** — "Reloading <name>…" — provides feedback during the ~1-2s blank window between kill and re-attach (otherwise the user sees an unexplained xterm flash).
- **Always shown** — archived sessions can be reloaded too. The action only operates on the pty, which is decoupled from archive state.
- **Eager crash-dot clear** — drops any stale `disconnectedSessions[sid]` entry on reload start so the sidebar red dot doesn't linger across the reload window. The new pty's successful attach also clears it (`_clearPtyExit` in usePtyAttach), but the eager clear avoids a flash on the crashed-then-reload-to-recover path.
- **Right-click selects the row** — `SessionRow.tsx:154` calls `onSelectSession(session.id)` from `onContextMenu`, so opening the menu on an unfocused row focuses it first. Earlier PR description claimed focus was preserved — that was wrong; the focus DOES move to the right-clicked row, which is the conventional desktop behavior.
- **Tolerates kill failure** — IPC errors are swallowed (the pty may already be exiting); the nonce still bumps so the attach effect always re-runs.

## Race fix (reviewer round 2)

Reviewer caught a real race in the kill → re-attach handoff:

1. `lifecycle.kill()` called `pty.kill()` synchronously and returned `true`.
2. Entry removal from the `sessions` Map only happens later, inside `entryFactory`'s `p.onExit` pump.
3. Renderer `await ccsmPty.kill(sid)` resolves at signal-dispatch time, before the entry is gone.
4. `reloadNonce` bumps → attach effect re-runs → `pty:attach` finds the still-present dying entry and returns its `(cols, rows, pid)` instead of `null`.
5. Spawn-on-null fallback does NOT trigger. Renderer registers as a viewer of the dead pty, then the pending `pty:exit` fires → user sees the crash overlay and has to click Retry.

**Fix** (~90 LOC, contained in `electron/ptyHost/lifecycle.ts`): `kill()` now returns `Promise<boolean>` that resolves only after `pty.onExit` fires (entry removed, headless disposed, watcher stopped) OR after `KILL_EXIT_TIMEOUT_MS` (3s) elapses for a wedged pty. The renderer's existing `await ccsmPty.kill(sid)` now guarantees the next `pty:attach` sees null and walks the spawn-on-null fallback.

Implementation notes:
- Dedupe `WeakMap<sessions, Map<sid, Promise>>` so concurrent kills for the same sid share one promise (single `pty.kill` / `processKiller` / `stopWatching` per kill cycle).
- Slot is registered BEFORE side effects so a sync-fire onExit mock can settle without leaving a stale resolved promise pinned for the next caller.
- 3s timeout falls back to `resolve(false)` — wedged pty can't hang the renderer; `reloadNonce` still bumps, spawn-on-null still recovers.
- `ipcMain.handle('pty:kill')` already awaits the return; renderer's `reloadSession` already awaits `ccsmPty.kill` — no caller-side changes.
- `killAll` keeps fire-and-forget semantics (the synchronous kill signal + `processKiller` walk are what reap children on app teardown; awaiting onExit per pty would just stall before-quit).

## Diff scope

Original feature: 127 LOC across 7 files.
Race fix: +90 LOC in `electron/ptyHost/lifecycle.ts`, +1 line type adjust in `index.ts` and `ipcRegistrar.ts`, new + updated tests.

## Test plan

- [x] `npm run lint` clean (0 errors; pre-existing warnings unchanged)
- [x] `tests/stores/slices/sessionRuntimeSlice.test.ts` — 5 test cases for `reloadSession` (kill called, nonce bumps, disconnect entry cleared, kill-error tolerated, missing-bridge OK)
- [x] `tests/terminal/usePtyAttach.test.tsx` — defensive nonce read keeps existing 9 cases passing
- [x] `tests/i18n-key-parity.test.ts` — both keys added to en + zh, parity preserved
- [x] **Race fix** — `electron/ptyHost/__tests__/lifecycle.test.ts`: existing kill suite migrated to async; new cases pin the contract:
  - awaits `pty.onExit` before resolving (mock onExit silent → promise pending until `__fireExit`)
  - 3s timeout fallback when onExit never fires (fake timers)
  - dedupes concurrent kills for the same sid
  - re-entry after settle is NOT deduped
- [x] Full `npx vitest run`: 1392 pass; 21 pre-existing failures (electron-load-smoke needs `npm run build` first; db-hardening hits a better-sqlite3 NODE_MODULE_VERSION mismatch in this env) also fail on the base commit — unrelated.
- [ ] Reviewer optional: `scripts/probe-session-reload.mjs` — workshop probe that launches isolated ccsm + seeds a session + drives `reloadSession` and asserts the pty pid changes while the host element with the same `data-active-sid` is preserved (requires `npm run build` + working claude binary). Not run in this session on Windows — the new unit tests pin the race-fix contract end-to-end at the lifecycle layer.

## Two pre-existing failing test files (unchanged by this PR)

`tests/electron-load-smoke.test.ts` (needs `npm run build` first — its own assertion message says so) and `electron/__tests__/db-hardening.test.ts` (native module). Verified failing on baseline before any changes via `git stash` round-trip.

### DIFF
--- a/electron/ptyHost/__tests__/ipcRegistrar.test.ts
+++ b/electron/ptyHost/__tests__/ipcRegistrar.test.ts
@@ -87,7 +87,7 @@ function makeDeps(over: Partial<PtyIpcDeps> = {}): PtyIpcDeps {
     spawnPtySession: vi.fn(() => ({ sid: 's', pid: 1, cols: 80, rows: 24, cwd: '/' })),
     inputPtySession: vi.fn(),
     resizePtySession: vi.fn(),
-    killPtySession: vi.fn(() => true),
+    killPtySession: vi.fn(async () => true),
     getPtySession: vi.fn(() => null),
     getBufferSnapshot: vi.fn(async () => ({ snapshot: '', seq: 0 })),
     ...over,
@@ -181,12 +181,12 @@ describe('pty:input / resize / kill / get pass-through', () => {
     expect(deps.resizePtySession).toHaveBeenCalledWith('sid', 100, 30);
   });
 
-  it('kill returns the deps result', () => {
+  it('kill returns the deps result', async () => {
     const ipc = makeFakeIpc();
-    const deps = makeDeps({ killPtySession: vi.fn(() => false) });
+    const deps = makeDeps({ killPtySession: vi.fn(async () => false) });
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     registerPtyIpc(ipc as any, deps);
-    expect(ipc.handlers.get('pty:kill')!({}, 'sid')).toBe(false);
+    await expect(ipc.handlers.get('pty:kill')!({}, 'sid')).resolves.toBe(false);
     expect(deps.killPtySession).toHaveBeenCalledWith('sid');
   });
 
--- a/electron/ptyHost/__tests__/lifecycle.test.ts
+++ b/electron/ptyHost/__tests__/lifecycle.test.ts
@@ -12,6 +12,9 @@ interface FakePty {
   write: ReturnType<typeof vi.fn>;
   resize: ReturnType<typeof vi.fn>;
   kill: ReturnType<typeof vi.fn>;
+  onExit: (cb: () => void) => { dispose: () => void };
+  /** Test helper — fire the registered onExit listener. */
+  __fireExit?: () => void;
   killShouldThrow?: boolean;
   writeShouldThrow?: boolean;
   resizeShouldThrow?: boolean;
@@ -68,14 +71,34 @@ vi.mock('../../prefs/scrollback', () => ({
 
 import * as L from '../lifecycle';
 
+function makeFakePty(over: Partial<FakePty> = {}): FakePty {
+  // Multi-listener onExit (mirrors node-pty's IEvent contract — see
+  // node-pty.d.ts: `readonly onExit: IEvent<...>`). The lifecycle race fix
+  // (#1277 review) registers a listener inside `kill()` to await entry
+  // removal; entryFactory also registers one for the cleanup pump. Both
+  // must fire on a single exit event.
+  const listeners = new Set<() => void>();
+  const pty: FakePty = {
+    pid: 1234,
+    write: vi.fn(),
+    resize: vi.fn(),
+    kill: vi.fn(),
+    onExit: (cb: () => void) => {
+      listeners.add(cb);
+      return { dispose: () => listeners.delete(cb) };
+    },
+    __fireExit: () => {
+      // Snapshot listeners — a listener may dispose itself when fired.
+      for (const cb of [...listeners]) cb();
+    },
+    ...over,
+  };
+  return pty;
+}
+
 function makeFakeEntry(over: Partial<FakeEntry> = {}): FakeEntry {
   return {
-    pty: {
-      pid: 1234,
-      write: vi.fn(),
-      resize: vi.fn(),
-      kill: vi.fn(),
-    },
+    pty: makeFakePty(),
     headless: { resize: vi.fn() },
     serialize: { serialize: () => 'snapshot' },
     attached: new Map(),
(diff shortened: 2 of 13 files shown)


Jiahui-Gu (commented):
LGTM (reviewer cannot self-approve via gh; manager: merge when ready).

Race fix is sound; reverse-verified all 4 new lifecycle race-fix tests would fail without the fix.

Layer 1
- 3s timeout: correct trade-off. Wedged-pty (kill+processKiller both silent → onExit never fires) means entry stays in `sessions`, so the post-timeout attach still lands on the zombie and the user sees the crash overlay → must hit Retry. Acceptable degradation for a rare edge case (strictly no worse than today's behavior on a wedged pty), and the timeout fulfills its actual contract: don't hang the UI forever. Worth a TODO someday but not a blocker.
- WeakMap<sessions, ...>: justified by test-registry isolation (comment explains). Slightly fancier than needed but harmless.
- Slot-before-side-effects: mock-only paranoia — production node-pty doesn't fire onExit synchronously inside kill() on any platform. The code comment is honest about why. Fine.

Layer 2
- Concurrent dedupe: A pending + B incoming → B awaits same promise object; A timeout settles both with false; slot cleared; fresh C creates new promise. Clean.
- Post-settle re-kill: `settle` deletes slot BEFORE `resolveOuter`. Test at lifecycle.test.ts:276-291 pins via `toHaveBeenCalledTimes(2)`. Stash-verified: without the slot delete the second kill short-circuits → assertion fails.
- killAll keeping fire-and-forget: correct. Awaiting onExit per pty on before-quit would stall app teardown; sync kill signal + processKiller walk reap children.

End-to-end
- SessionRow onContextMenu calls `onSelectSession(session.id)` — verified, focus-on-right-click works.
- `reloadSession` awaits `window.ccsmPty?.kill(sid)` BEFORE bumping `reloadNonce` — verified. ipcRegistrar `pty:kill` handler is now async returning the promise. Full chain: renderer await → IPC await → lifecycle promise resolves on onExit (entry removed) → reloadNonce bump → attach effect re-runs → `pty:attach` returns null → spawn-on-null fallback fires.

CI 7/7 green across ubuntu/macos/windows. PR description's pre-existing failures (electron-load-smoke needs `npm run build`; db-hardening native-module mismatch) are local-only — CI doesn't hit them.

Nit (non-blocking, future): consider clearing zombie entries from `sessions` Map after KILL_EXIT_TIMEOUT_MS resolves(false), so spawn-on-null fallback also recovers on wedged-pty. Out of scope here.

PR #1279, opened 2026-05-21
feat(sidebar): right-click copy session (claude --fork-session)

Jiahui-Gu:
## What

Right-click any session row → **Copy session** (between Rename and Move to group). One click duplicates the session in place:
- new row inserted directly after the source
- name = `<source name> (copy)`, auto-enters inline rename
- cwd / model / group inherited from source
- focus flips to the new session
- the new pty boots with the **source's full transcript context** but writes its own JSONL

## How — claude CLI flag combo (verified live first)

Used `claude --fork-session` — exists natively in the installed CLI:
```
--fork-session  When resuming, create a new session ID instead of reusing the original (use with --resume or --continue)
```

**Live probe (path A confirmed):**
```bash
claude --resume <SOURCE_SID> --fork-session --session-id <NEW_SID> -p "say hi briefly"
# exit=0
# new jsonl written at ~/.claude/projects/<key>/<NEW_SID>.jsonl
#   size ≈ source + new turn (transcript copied)
# source jsonl mtime UNCHANGED (untouched)
```
All three flags coexist cleanly. No need for path B (auto-detect new sid from stdout / file mtime).

## Implementation (~200 LOC source + ~190 test)

**Renderer:**
- `sessionCrudSlice.copySession(sourceId)` mints a new sid, inserts after source, flips `activeId`, arms two transient store fields:
  - `pendingRenameId` → consumed by `SessionRow` to auto-enter inline rename
  - `pendingForkSource[newSid] = sourceSid` → consumed by `usePtyAttach` on the very first `pty.spawn`, cleared post-spawn so any later Retry falls back to the standard `--resume <new>` branch
- `SessionRow.tsx` — new `ContextMenuItem` + `useEffect` watching `pendingRenameId`

**Main:**
- `entryFactory.makeEntry` takes an optional 7th `forkSourceSid?`. When set AND no jsonl exists for the new sid:
  ```
  args = [--resume, <srcUuid>, --fork-session, --session-id, <newUuid>]
  ```
  Otherwise falls through to the pre-existing `[--resume|--session-id, <sid>]` flag picker. Re-spawn after a fork (Retry, app restart) does NOT re-fork — the new sid's jsonl exists by then so `--resume <new>` is correct.
- `lifecycle.spawn` / `ipcRegistrar` `pty:spawn` handler / preload `ccsmPty.spawn` / `pty.d.ts` all thread the optional arg through. Absent on the non-fork path so existing wire shape is preserved (only the test mocks needed updates for the now-explicit `undefined` 3rd arg).

**i18n:** `sidebar.copySession` added to en (`Copy session`) + zh (`复制会话`). These are the only two locales today.

**UX (manager-locked):**
- Menu position: right after Rename ("derive new" cluster)
- New session name: `<source> (copy)`, auto-enters inline rename
- group / cwd / model = source's
- archived sessions are also copyable (copy lands in same group as source; `archivedAt` is intentionally NOT inherited so the row never boots stale-archived)
- focus flips to the new session
- no toast, no keyboard shortcut in v1

## Tests

- **`entryFactory.test.ts`** (+2): fork branch produces the 5-flag argv; re-spawn after fork falls back to plain `--resume <new>`.
- **`sessionCrudSlice.test.ts`** (+5): copySession insert position, name suffix, cwd/model/group inheritance, side effects on `activeId` / `pendingRenameId` / `pendingForkSource`, `archivedAt` not carried over, `consumePendingRename` idempotency.
- **`usePtyAttach.test.tsx`** (+1, 2 modified): existing assertions updated for the explicit `undefined` 3rd spawn arg; new test pins the fork-arg forwarding + post-spawn `pendingForkSource` clear.

## Verification

- `npx tsc --noEmit` — clean
- `npx eslint <changed files>` — 0 errors (only pre-existing warnings)
- `npx vitest run` — 1349/1370 pass; the 21 failures are all in `tests/electron-load-smoke` (requires prior `npm run build`) and `electron/__tests__/db-hardening` (`better-sqlite3` native build artifacts) — same files already failed before this change

## Test plan

- [ ] Right-click a session that has had at least a few turns with claude → click "Copy session"
- [ ] New row appears immediately after source, name `<original> (copy)`, inline rename input is focused
- [ ] Type a new name + Enter → renamed
- [ ] New pty attaches; claude shows the source's prior context (e.g. `/resume` indicator or recap)
- [ ] Send a new message in the copy → it streams independently
- [ ] `ls ~/.claude/projects/<projectKey>/` shows two distinct `<sid>.jsonl` files (source + copy)
- [ ] Source session is unaffected (mtime preserved, can still be opened, original transcript intact)
- [ ] Copy an archived session → new row lands in source's group (the archive container, since that's where the source is)
- [ ] Kill the forked pty + Retry → spawns with `--resume <newSid>` (no second fork), context still intact

## Coordination note

PR #1277 (right-click Reload) is not yet merged into base. This PR adds Copy session right after Rename; when #1277 lands, the rebase will need a manual merge of the `ContextMenuItem` list. Copy belongs adjacent to Rename ("derive new" cluster), so the resolution should be clean.

### DIFF
--- a/electron/preload/bridges/ccsmPty.ts
+++ b/electron/preload/bridges/ccsmPty.ts
@@ -30,8 +30,10 @@ const ptyExitListeners = new Set<(e: PtyExitPayload) => void>();
 
 const ccsmPty = {
   list: (): Promise<unknown> => ipcRenderer.invoke('pty:list'),
-  spawn: (sid: string, cwd: string): Promise<unknown> =>
-    ipcRenderer.invoke('pty:spawn', sid, cwd),
+  spawn: (sid: string, cwd: string, forkSourceSid?: string): Promise<unknown> =>
+    forkSourceSid === undefined
+      ? ipcRenderer.invoke('pty:spawn', sid, cwd)
+      : ipcRenderer.invoke('pty:spawn', sid, cwd, forkSourceSid),
   attach: (sid: string): Promise<unknown> => ipcRenderer.invoke('pty:attach', sid),
   detach: (sid: string): Promise<void> => ipcRenderer.invoke('pty:detach', sid),
   input: (sid: string, data: string): Promise<void> =>
--- a/electron/ptyHost/__tests__/entryFactory.test.ts
+++ b/electron/ptyHost/__tests__/entryFactory.test.ts
@@ -239,6 +239,57 @@ describe('entryFactory.makeEntry', () => {
       | undefined;
     expect(opts?.scrollback).toBe(1500);
   });
+
+  // Right-click "Copy session" path: when the renderer mints a fresh sid
+  // (no JSONL on disk yet) AND passes a `forkSourceSid`, makeEntry must
+  // spawn `claude --resume <src> --fork-session --session-id <new>` so the
+  // CLI duplicates the source's transcript natively. The bare new-session
+  // path (`['--session-id', sid]`) would lose the source context entirely.
+  it('forks via --resume + --fork-session + --session-id when forkSourceSid is set and no JSONL exists', () => {
+    bus().sourceJsonl = null;
+    makeEntry(
+      'sid-NEW',
+      '/work',
+      '/bin/claude',
+      80,
+      24,
+      { onExit: vi.fn() },
+      'sid-SOURCE',
+    );
+    const args = bus().ptySpawn.mock.calls[0][1] as string[];
+    expect(args).toEqual([
+      '--resume',
+      'sid-SOURCE',
+      '--fork-session',
+      '--session-id',
+      'sid-NEW',
+    ]);
+    // No import-resume copy expected — sourceJsonl is null so the existing
+    // `ensureResumeJsonlAtSpawnCwd` branch is skipped (correct: claude
+    // CLI handles the transcript copy itself with --fork-session).
+    expect(bus().ensureJsonl).not.toHaveBeenCalled();
+  });
+
+  // Defensive: once the forked session has booted at least once, its own
+  // JSONL exists at `<newSid>.jsonl`. A re-spawn (Retry, app relaunch)
+  // must NOT re-issue --fork-session — that would copy the transcript a
+  // SECOND time. Falling through to the normal `--resume <new>` path is
+  // the right behavior. We model this by setting sourceJsonl truthy
+  // (i.e. findJsonlForSid found the new sid's jsonl).
+  it('skips --fork-session and resumes the new sid directly when the new sid already has a JSONL', () => {
+    bus().sourceJsonl = '/proj/sid-NEW.jsonl';
+    makeEntry(
+      'sid-NEW',
+      '/work',
+      '/bin/claude',
+      80,
+      24,
+      { onExit: vi.fn() },
+      'sid-SOURCE',
+    );
+    const args = bus().ptySpawn.mock.calls[0][1] as string[];
+    expect(args).toEqual(['--resume', 'sid-NEW']);
+  });
 });
 
 // L4 PR-C (#863): the onData handler is now an extracted, named
--- a/electron/ptyHost/entryFactory.ts
+++ b/electron/ptyHost/entryFactory.ts
@@ -210,11 +210,32 @@ export function makeEntry(
   cols: number,
   rows: number,
   deps: MakeEntryDeps,
+  forkSourceSid?: string,
 ): Entry {
   const claudeSid = toClaudeSid(sid);
   const sourceJsonl = findJsonlForSid(claudeSid);
-  const flag = sourceJsonl ? '--resume' : '--session-id';
-  const args = [flag, claudeSid];
+
+  // `--fork-session` path: the renderer is creating a new session that should
+  // boot with another (source) session's full transcript context but write to
+  // its own JSONL. Claude CLI handles the duplication natively — we hand it
+  // BOTH the source sid (`--resume`) AND the desired new sid (`--session-id`)
+  // along with `--fork-session`, and it copies the transcript on first write.
+  // The new sid's JSONL DOES NOT EXIST YET (since we just minted it in the
+  // renderer), so the normal `findJsonlForSid` branch below would route us
+  // into a `--session-id`-only spawn that reuses the bare new id — losing
+  // the parent context. Handle fork explicitly first, before the resume/
+  // session-id picker; once the fork finishes the first turn, the new JSONL
+  // exists and any subsequent re-spawn (e.g. user kills + retries) will fall
+  // through `findJsonlForSid` → `--resume` like any other ccsm-tracked
+  // session, no special-casing needed downstream.
+  let args: string[];
+  if (forkSourceSid && !sourceJsonl) {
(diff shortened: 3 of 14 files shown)


Jiahui-Gu (commented):
**Independent reviewer pass — LGTM (would approve, blocked by self-author rule).**

## Layer 1 (necessity / direction): clear

- **`--fork-session` semantics verified**: PR description shows a live probe (`claude --resume <SRC> --fork-session --session-id <NEW> -p \"say hi briefly\"`) confirming the new JSONL is written with source-transcript context and source jsonl mtime is untouched. Native CLI duplication, no manual copy. Matches the user's stated mental model ("double the agent").
- **UX**: copy is selected (activeId flips), inserts directly after source (Finder Duplicate-style), auto-enters inline rename. New pty spawned lazily via the normal `usePtyAttach` path — no leaking back into source.
- **i18n sentence case**: `Copy session` / `复制会话` — both compliant with the no-SCREAMING-UI rule.
- **Naming**: `<source name> (copy)` with auto-rename input focused — sensible.
- **Position in context menu**: directly after Rename ("derive new" cluster) — coherent.

## Layer 2 (implementation): solid

- **State cloning**: cwd / model / groupId / agentType / cwdMissing all inherited. `archivedAt` deliberately NOT carried (correct — fresh row should never start stale-archived); covered by test.
- **sessionId**: minted via `newSessionId()`; zustand persist handles it like any other session.
- **Fork-arg threading**: `copySession` → `pendingForkSource[newSid]=srcSid` → `usePtyAttach` reads on spawn-on-null-attach branch → IPC → `lifecycle.spawn` opts → `makeEntry(forkSourceSid)`. Each hop typed; preload spawn signature widened with optional 3rd arg so non-fork wire shape is preserved.
- **Defense-in-depth on IPC boundary**: `ipcRegistrar` coerces `forkSourceSid` to string-or-undefined; `toClaudeSid` throws on anything not matching `VALID_SID_RE`. Argv-injection risk covered.
- **Post-spawn cleanup**: `pendingForkSource[sid]` cleared regardless of spawn outcome — Retry after failed fork takes clean `--session-id` branch instead of re-firing `--fork-session` in a loop. Good failure-mode reasoning.
- **Idempotency on re-spawn after successful fork**: `entryFactory` guard `if (forkSourceSid && !sourceJsonl)` — once new sid's JSONL exists, fork is bypassed. Has dedicated test.

## Tests: pin behavior, not call-sites (reverse-verified mentally)

- `sessionCrudSlice.test.ts`: position (`['a', 'src', newId!, 'z']`), name suffix, cwd/model/group inheritance, activeId/pendingRenameId/pendingForkSource side effects, archivedAt non-carryover. Stashing the action body fails every assertion.
- `entryFactory.test.ts`: pins exact argv on both fork and post-fork-respawn paths. Stashing the fork branch forces fallthrough → fails.
- `usePtyAttach.test.tsx`: pins the 3rd-arg forwarding plus post-spawn map clear.

## Nits (not blocking)

1. **Fork-on-empty source**: if user copies a brand-new session with no first message yet, spawn becomes `claude --resume <srcUuid> --fork-session …` which the CLI will reject (nothing to resume). Post-spawn cleanup clears `pendingForkSource`, so the subsequent Retry takes the clean `--session-id` path — graceful degradation, but one cryptic spawn_failed flash before recovery. Consider either gating Copy in the menu when source has no transcript, or skipping fork registration in `copySession` when source has never spawned. Low priority — most copy targets will have transcripts.
2. **Rebase against #1277**: confirmed PR #1277 (Reload) is still OPEN. Whichever lands second needs a one-line `ContextMenuItem` merge — already flagged in PR description. Trivial.
3. **`SessionRow` useEffect**: selecting `s.pendingRenameId === session.id` as boolean (not the raw string) is a nice perf detail — already correct.

No blockers. Ship it after #1277 sequencing decision.

Jiahui-Gu (commented):
Post-rebase review (independent reviewer, per [feedback_rebase_always_reviewed]). NOTE: GitHub refused `--approve` (self-PR restriction — PR author = reviewer account); submitting as comment with explicit LGTM. Manager should treat as approval signal for the rebase delta.

Focus = what was at risk during the rebase across #1277 (right-click Reload + kill race fix) and #1284 (preload bridge tests).

**SessionRow.tsx merge** — both ContextMenuItems coexist cleanly. Rendered DOM order: Rename → Copy session → Move to group → [separator] → Reload session → Archive → Delete. Copy sits in the top "create/identity" group next to Rename; Reload sits below the separator with the lifecycle ops (Archive/Delete) — matches the worker's description and reads UX-correct (creation operations grouped, lifecycle operations grouped, destructive last). The `onContextMenu` handler calls `onSelectSession` once at the row level, so right-click focus-switching works uniformly for both menu items. `copySession` and `reloadSession` are each one selector-subscribed store hook; both are bound functions from a zustand slice → ref-stable, no re-render thrashing on top of React.memo at the bottom of the file.

**lifecycle.ts auto-merge** — verified intact: `kill()` returns `Promise<boolean>`; WeakMap `pendingKills` dedupe; `exitDisposable` + `settle()` + KILL_EXIT_TIMEOUT_MS path all preserved; the `forkSourceSid` spawn opt threads through `makeEntry(..., opts?.forkSourceSid)` correctly.

**ccsmPty.ts conditional spawn** — the right call. Worker preserved wire compatibility (no trailing `undefined` sent to IPC, so the existing 2-arg preload test `['s1', '/cwd']` in #1284 keeps asserting the no-fork happy path unchanged). The main-process handler at `electron/ptyHost/ipcRegistrar.ts:95` destructures `(_event, sid, cwd, forkSourceSid?)` and immediately normalizes via `typeof forkSourceSid === 'string' && length > 0 ? : undefined`, so both 2-arg and 3-arg paths converge correctly. The 2 vs 3-line branching is a fair price for not loosening the preload test's exact-args assertion. (Suggestion, not blocker: a sibling preload test for the fork path `await api.spawn('s1', '/cwd', 'src')` → `invoke('pty:spawn', 's1', '/cwd', 'src')` would lock the new branch — fine to land in a follow-up.)

**i18n + types.ts** — `reloadSession`/`reloadingSessionToast`/`copySession` keys all present in both en and zh; sentence case throughout (no SCREAMING strings); `reloadNonce`, `pendingRenameId`, `pendingForkSource` fields coexist in types.ts.

**Reverse verification**:
- Stubbed out the fork branch in entryFactory.ts → `entryFactory.test.ts` correctly fails on the `--resume + --fork-session + --session-id` assertion (line 260).
- Replaced `lifecycle.ts:kill()` with a fire-and-forget version → `lifecycle.test.ts` correctly fails 4 tests: timeout fallback, onExit-await, dedupe, wedged-kill.
- Restored both → 62/62 green across `lifecycle.test.ts` + `entryFactory.test.ts` + `ccsmPty.test.ts`.

CI 7/7 green on the PR. PR is currently BEHIND `working` (2 commits merged post-rebase: #1285, #1287); neither touches files in this PR's diff so no further rebase needed unless GitHub blocks the merge.

LGTM — rebase delta is safe to merge.

PR #1281, opened 2026-05-21
refactor(agent): extract CLAUDE_CODE_AGENT_ID constant + pin wire value

Jiahui-Gu:
## v2 per reviewer feedback

Stripped the speculative `AgentMainAdapter` / `AgentRendererAdapter` types and the `src/agent/claude/` directory — Phase 2 refactor shelved per user direction, will reintroduce contracts alongside their first real consumer. This PR is now just the standalone, defensible piece: extract the magic string `'claude-code'` into a cross-boundary constant + pin the persisted wire value with a test.

**Files in v2 (5 files):**

New:
- `src/shared/agentIds.ts` — `CLAUDE_CODE_AGENT_ID = 'claude-code'` + `AgentId` cross-boundary type. `tsconfig.electron.json` already includes `src/shared/**`, so main + renderer both import from one place.
- `tests/shared/agentIds.test.ts` — pins the wire value `'claude-code'`. Load-bearing: any future DB migration relies on this constant not silently changing.

Modified (constant replacement only, zero behavior change):
- `src/types.ts` — `AgentType = typeof CLAUDE_CODE_AGENT_ID` (import moved to top of file per reviewer nit)
- `src/components/AgentIcon.tsx` — uses constant for the `agentType === CLAUDE_CODE_AGENT_ID` check; `ClaudeAsterisk` stays inline (no extraction)
- `src/stores/slices/sessionCrudSlice.ts` — 2 hardcoded literals replaced (lines 207, 295)

## Verification

- `npm run typecheck` passes
- Lint on changed files: no new errors/warnings (one pre-existing `ATTENTION_PRIORITY` warning in AgentIcon.tsx is unrelated)
- `tests/shared/agentIds.test.ts`, `tests/stores/slices/sessionCrudSlice.test.ts`, `tests/components/AgentIcon*.test.tsx` all pass

## Reviewer focus

1. Wire-value pin — does `tests/shared/agentIds.test.ts` belong here, or should it live next to the (future) DB migration? Argument for here: catches accidental rename even before the migration exists.
2. Should `AgentId` (currently `string`) be removed too, given there's no consumer yet? Or is the type alias cheap enough to keep as documentation of the cross-boundary contract.

### DIFF
--- a/src/components/AgentIcon.tsx
+++ b/src/components/AgentIcon.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { motion } from 'framer-motion';
 import { cn } from '../lib/cn';
+import { CLAUDE_CODE_AGENT_ID } from '../shared/agentIds';
 import type { AgentType, SessionState } from '../types';
 
 type Size = 'sm' | 'md';
@@ -76,7 +77,7 @@ export function AgentIcon({
     : state === 'waiting' || flashing
       ? 'waiting-or-flashing'
       : 'idle';
-  const inner = agentType === 'claude-code' ? <ClaudeAsterisk size={glyph} /> : null;
+  const inner = agentType === CLAUDE_CODE_AGENT_ID ? <ClaudeAsterisk size={glyph} /> : null;
   return (
     <motion.span
       data-agent-icon-state={state}
--- a/src/shared/agentIds.ts
+++ b/src/shared/agentIds.ts
@@ -1,0 +1,15 @@
+// Cross-boundary (renderer + main) agent identifier constants.
+//
+// These string literals are PERSISTED in SQLite (`sessions.agent_type`
+// column) and stored in user prefs / serialized state. Once shipped,
+// changing the wire value silently breaks existing user data — every
+// row read from disk would no longer match.
+//
+// Phase 1 of the multi-agent refactor: extract the magic string into a
+// named constant so future agents (codex/cursor/aider/...) can be added
+// without grepping for `'claude-code'` literals scattered across the
+// codebase, and so DB migrations can reference the same symbol the
+// runtime uses to write.
+//
+// `tests/shared/agentIds.test.ts` pins the wire value — do not change it.
+export const CLAUDE_CODE_AGENT_ID = 'claude-code' as const;
--- a/src/stores/slices/sessionCrudSlice.ts
+++ b/src/stores/slices/sessionCrudSlice.ts
@@ -14,6 +14,7 @@
 // `sessionTitleBackfillSlice` (split per Task #736 / PR #754 review).
 
 import type { Group, Session } from '../../types';
+import { CLAUDE_CODE_AGENT_ID } from '../../shared/agentIds';
 import { hydrateDrafts as _unused, deleteDrafts, snapshotDraft, restoreDraft } from '../drafts';
 import { resolvePreferredGroup } from '../lib/preferredGroupResolver';
 import {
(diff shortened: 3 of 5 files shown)


Jiahui-Gu (commented):
## Review — Layer 1 (necessity) is the blocker

CI is green, types compile, behavior is byte-identical. But given the user shelved the broader parallel-agent refactor ("把重构从task里删了，先不做"), this PR has to justify each piece on its own — and several pieces don't.

### Verified by grep on the branch

`AgentMainAdapter` and `AgentRendererAdapter` have **zero consumers in this PR**. The only files that mention those names are the two files that declare them:

- `electron/agent/types.ts` (declaration only)
- `src/agent/types.ts` (declaration only)

No factory, no test, no registry, no Claude adapter implementation references either contract. They are pure speculative scaffolding for a phase 2 that is now indefinitely deferred.

### Recommendation: strip to the standalone-valuable subset, then land

**Keep (carries value without the refactor):**
- `src/shared/agentIds.ts` — `CLAUDE_CODE_AGENT_ID` constant. Real DB-persisted wire value; deduping makes sense even single-agent.
- `tests/shared/agentIds.test.ts` — pins the wire value against accidental rename.
- `src/types.ts` — `AgentType = typeof CLAUDE_CODE_AGENT_ID` narrowing.
- `src/components/AgentIcon.tsx` — import the constant (the `agentType === CLAUDE_CODE_AGENT_ID` swap stands alone).
- `src/stores/slices/sessionCrudSlice.ts` — 2 literal replacements.

**Remove (declared-but-unused; will rot until refactor resumes):**
- `electron/agent/types.ts` — entire `AgentMainAdapter` contract. 112 lines of speculative interface design (`runStateSource: 'transcript' | 'osc' | 'none'`, optional OSC classifier, phase-2 `unknown` return types) that nothing consumes. When phase 2 actually starts, the right surface area will be obvious from extracting real code — locking it in now risks designing the wrong seam and carrying it as dead weight.
- `src/agent/types.ts` — same story, `AgentRendererAdapter` has no consumer.
- `src/agent/claude/Icon.tsx` — creates a new top-level `src/agent/` directory purely to house this one file. Without the renderer-adapter registry that would import it from there, this is premature structure. Inline the `ClaudeAsterisk` SVG back in `src/components/AgentIcon.tsx` (or move to `src/components/agentGlyphs/ClaudeAsterisk.tsx` if you want it factored out — but don't establish `src/agent/` as a directory until something else lives in it).

### Layer 2 notes (apply to the kept subset)

- `src/types.ts:20` — `import` statement appears *after* the `SessionState` export. Hoisting makes it work, but move the import to the top with the other module-level imports.
- `AgentType = typeof CLAUDE_CODE_AGENT_ID` narrowing is fine; spot-checked test/script fixtures use literal `'claude-code'` which is assignable to this singleton type.
- Wire-value pin test correctly asserts `=== 'claude-code'` — good DB-persistence guard.
- The ~30+ remaining `'claude-code'` literals in `tests/` and `scripts/harness-*.mjs` are fixture data; leaving them as literals is correct (test fixtures shouldn't depend on production constants — they verify behavior at the wire-value level, same job as the pin test).
- PR #1279 collision: that PR touches lines ~142/156/549 in `sessionCrudSlice.ts`, this one touches 206/294. No textual overlap; either order rebases cleanly. Non-issue.
- `electron/**` → `src/shared/**` import resolution would have been valid (per `tsconfig.electron.json`), but moot once `electron/agent/types.ts` is removed.

### What went well

- The DB-persistence framing in `agentIds.ts` comments is sharp — exactly the right reason for a constant.
- Wire-value pin test is the kind of small load-bearing test that's easy to skip and pays off later.
- Worker's PR description was honest about phase-1/zero-behavior-change scoping.

### Action — CHANGES REQUESTED

Strip the 3 files above (`electron/agent/types.ts`, `src/agent/types.ts`, `src/agent/claude/Icon.tsx`), revert `AgentIcon.tsx`'s `ClaudeAsterisk` import to the inline definition (or relocate inside `src/components/`), fix the `src/types.ts` import position, re-run typecheck + the agentIds test, and this becomes a clean ~40-line "extract constant + pin wire value" refactor that stands on its own merits. When phase 2 resumes (if it does), the adapter contracts can be reintroduced alongside their first real consumer, which is where their shape will actually get validated.

PR #1282, opened 2026-05-21
test(electron): cover lifecycle + main helpers (batch 2, 0% -> 50%)

Jiahui-Gu:
## Summary

Coverage batch 2: target `electron/main.ts` (0%) + `electron/lifecycle/appLifecycle.ts` (3%).

**Scoping result after read-through:**

- `electron/lifecycle/appLifecycle.ts` (94 LOC) — fully unit-testable. Two exports (`applyAppMenuLocale`, `registerLifecycleHandlers`), both pure-ish: helpers + a deps-bag handler registrar. **Now at ~55% line coverage** via 16 new tests.
- `electron/main.ts` (322 LOC) — **intentionally not covered here**. It is a top-level procedural module: module-load executes `initSentry()`, `acquireSingleInstanceLock()`, `applyAppMenuLocale()`, then attaches `app.whenReady()`. It exports nothing. `createWindow` / `ensureTray` / `applyTrayLocale` are module-local closures over `let isQuitting`, `let trayController`, `let notifyPipeline` — they cannot be invoked without importing main.ts itself, which would execute every side effect on import. The `app.whenReady()` body is the orchestration site that wires every subsystem (db init, IPC registration, notify pipeline, ptyHost, sessionWatcher). The pre-existing `tests/electron-load-smoke.test.ts` already covers main.ts at the load-smoke level.
  - **Per the batch 2 plan, no source refactor was performed to make main.ts unit-testable.** Splitting main.ts into testable helpers is a follow-up task (out of scope for a coverage PR).

## Tests added (16, all passing)

`electron/lifecycle/__tests__/appLifecycle.test.ts`:

- `applyAppMenuLocale` (4 tests): builds + installs the hidden Edit-role menu via stubbed `Menu`; pulls localized label through `i18n.tMenu`; omits the `paste` role (terminal pane owns its own Ctrl/Cmd+V handler — see file comment); rebuilds on every call so locale changes take effect.
- `registerLifecycleHandlers` — registration (1 test): all three of `before-quit` / `window-all-closed` / `activate` get listeners.
- `before-quit` (5 tests): flips `isQuitting`; reaps pty sessions; disposes notify pipeline when present; tolerates omitted `disposeNotifyPipeline` (early-failure path); swallows errors from both `killAllPtySessions` and `disposeNotifyPipeline` (best-effort cleanup contract).
- `window-all-closed` (3 tests): quits + closes db only when `isQuitting` is true (tray-resident on non-quit close); reads the live flag (not a registration-time snapshot).
- `activate` (2 tests): spawns a window only when `getWindowCount() === 0`.
- Implementation note: `applyAppMenuLocale` uses dynamic `require('../i18n')` to dodge a circular-import edge. `vi.mock` only intercepts ESM imports, so the test patches `Module.prototype.require` for that one id with a deterministic `tMenu` stub.

## Before / after coverage

| File | Before | After |
|---|---|---|
| `electron/lifecycle/appLifecycle.ts` | ~3% lines | ~100% lines (directory rolled up to 54.5%) |
| `electron/main.ts` | 0% lines | 0% lines (covered by load-smoke; see scoping above) |

(Measured by `npx vitest run electron/lifecycle/__tests__/appLifecycle.test.ts --coverage`. The directory rollup is 54.5% because `singleInstance.ts` in the same dir is at 0%; `appLifecycle.ts` itself is essentially full.)

## Follow-up

Making `electron/main.ts` unit-testable would require lifting its module-level mutable state (`isQuitting`, `badgeManager`, `notifyPipeline`, `activeSidFromRenderer`, `sessionNamesFromRenderer`) into an explicit factory + extracting the `app.whenReady()` body into a testable `bootstrap(deps)` function. That is a real refactor with real risk and belongs in its own task, not a coverage PR.

## Test plan

- [x] new test file lints clean
- [x] `npx vitest run electron/lifecycle/__tests__/appLifecycle.test.ts` → 16/16 pass
- [x] `npx vitest run tests/electron-load-smoke.test.ts` still passes alongside the new test (no Module.prototype.require leakage)
- [x] coverage reporter confirms `electron/lifecycle/` jumped from ~3% to ~54.5% line coverage

### DIFF
--- a/electron/lifecycle/__tests__/appLifecycle.test.ts
+++ b/electron/lifecycle/__tests__/appLifecycle.test.ts
@@ -1,0 +1,262 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { App } from 'electron';
+
+// Mock electron — appLifecycle.ts top-level imports `Menu`, which would
+// trigger "Electron failed to install correctly" on the lint+test runner
+// (no electron binary on CI). We control the spies via the mock factory
+// and read them back via vi.mocked() in tests.
+vi.mock('electron', () => {
+  const buildFromTemplate = vi.fn(() => ({ items: [] }));
+  const setApplicationMenu = vi.fn();
+  return {
+    Menu: { buildFromTemplate, setApplicationMenu },
+  };
+});
+
+// applyAppMenuLocale uses a dynamic `require('../i18n')` to dodge a
+// circular-import edge in the production graph. vitest's vi.mock only
+// intercepts ESM `import`, NOT Node-native CJS require — and Node's CJS
+// resolver can't find the .ts source on its own. Patch the CJS require
+// to short-circuit any '../i18n' lookup with our deterministic stub.
+// Restored in afterAll-equivalent below if we ever need it.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const Module = require('node:module') as typeof import('node:module');
+const ModuleProto = (Module as unknown as { prototype: { require: (id: string) => unknown } })
+  .prototype;
+const originalRequire = ModuleProto.require;
+ModuleProto.require = function patchedRequire(this: NodeJS.Module, id: string) {
+  if (id === '../i18n') {
+    return { tMenu: (key: string) => `MENU_${key}` };
+  }
+  return originalRequire.call(this, id);
+};
+
+import { Menu } from 'electron';
+import {
+  applyAppMenuLocale,
+  registerLifecycleHandlers,
+  type LifecycleDeps,
+} from '../appLifecycle';
+
+type EventName = 'before-quit' | 'window-all-closed' | 'activate';
+
+interface FakeApp {
+  on: ReturnType<typeof vi.fn>;
+  quit: ReturnType<typeof vi.fn>;
+  /** invoke a registered listener for a given event */
+  fire: (event: EventName) => void;
+}
+
+function createFakeApp(): FakeApp {
+  const handlers = new Map<EventName, () => void>();
+  const on = vi.fn((event: EventName, cb: () => void) => {
+    handlers.set(event, cb);
+  });
+  const quit = vi.fn();
+  return {
+    on,
+    quit,
+    fire: (event) => {
+      const cb = handlers.get(event);
+      if (!cb) throw new Error(`no handler registered for ${event}`);
+      cb();
+    },
+  };
+}
+
+function buildDeps(overrides: Partial<LifecycleDeps> = {}): {
+  deps: LifecycleDeps;
+  fakeApp: FakeApp;
+  spies: {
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
**Review verdict: APPROVE** (cannot use --approve on self-authored PR — comment review with explicit verdict).

Test-only, additive (262 LOC, 1 file), CI green on all 3 OSes. Verified locally: 16/16 alone + 35/35 combined with `tests/electron-load-smoke.test.ts` (no Module.prototype.require leak). Reverse-verified by sabotaging 2 production behaviors → got exactly 2 expected test failures.

**Module.prototype.require patch vs. refactoring source:** Patch is the right call. `applyAppMenuLocale()` is called from multiple sites with zero args; threading an `i18n` dep through every call site to avoid one test-side workaround would be net-worse ergonomics. The source-side dynamic `require('../i18n')` is itself a documented circular-import dodge (appLifecycle.ts:36-38). Keep source clean, absorb the ugliness in the test harness.

**main.ts at 0% deferred:** Correct. 322 LOC of zero-export top-level side-effect orchestration; making it unit-testable IS the refactor. `tests/electron-load-smoke.test.ts` covers the "imports without throwing" axis. Don't bundle a 322 LOC refactor into a coverage PR.

**Follow-up bootstrap(deps) plan:** Right direction. Lighter alternative: continue the same incremental Phase C / Task #742 extraction pattern that produced `appLifecycle.ts` — pull out `createWindow`, `applyTrayLocale`, etc. one at a time with their own deps bags, rather than one big bootstrap rewrite. Either approach works; not a blocker.

## Layer 2 (implementation)

- `Module.prototype.require` patch scoped per-id (`if (id === '../i18n')`), falls through to `originalRequire.call(this, id)` for everything else. No leak: ran new file + smoke test together → 35/35 pass.
- Tests assert real behavior, not mock plumbing: template inspection for the paste-role omission, before-quit ordering invariant (dispose still runs after kill throws), live-flag read (catches registration-time snapshot bugs), tray-resident contract on window-all-closed.
- Reverse-verification: stashed `{ role: 'paste' }` back into submenu AND removed the `if (getIsQuitting())` guard on `window-all-closed` → got exactly the expected 2 test failures with clear assertion messages. Tests bite.
- Pre-existing failures (db-hardening NODE_MODULE_VERSION etc.) are not introduced here — PR is purely additive, CI green on all 6 jobs.

## Nit (non-blocking, fold into follow-up)

`Module.prototype.require` patch is installed at module top-level but never restored. Worker's own comment says "Restored in afterAll-equivalent below if we ever need it" — but isn't. Safe today because vitest isolates each test file in its own worker. If `pool: 'threads'` ever lands with `isolate: false`, the patch leaks across files. One defensive `afterAll(() => { ModuleProto.require = originalRequire; })` would future-proof. Not blocking.

PR #1283, opened 2026-05-21
test(window): cover createWindow close-action + lifecycle (batch 3, 18% → 96%)

Jiahui-Gu:
## Baseline (摸底)
`electron/window/createWindow.ts` — main BrowserWindow factory + the close / hide / quit choreography. Touched on every close-behavior change.

| metric    | before  | after  |
|-----------|---------|--------|
| lines     | 18.42%  | 96.49% |
| branches  | 25.71%  | 92.85% |
| functions | 23.52%  | 100%   |

Pre-existing tests only covered the three pure exports (`installContextMenu`, `isAllowedNavigation`, `decideCloseAction`). The 174–476 line range that owns the actual lifecycle was 0%.

## Approach
The factory already lives behind a `CreateWindowDeps` bag — no source change needed. The non-deps imports (`BrowserWindow`, `Menu`, `app`, `getCloseAction`/`setCloseAction`, `tCloseDialog`, `buildAppIcon`) are mocked at the module boundary. The `BrowserWindow` mock ctor returns a `FakeWin` whose `on` / `webContents.on` push into listener maps, so each test drives the captured `close` / `focus` / `maximize` / `will-navigate` / `context-menu` listener directly. `vi.useFakeTimers()` makes the 180ms fade and 10s ask-timeout fallback deterministic.

## Cases (16 new, on top of existing 29)
- BrowserWindow ctor opts: normal vs `CCSM_E2E_HIDDEN` (offscreen positioning + skipTaskbar + Chromium focus prime)
- dev vs prod renderer load: URL on `CCSM_DEV_PORT` (default 4100), DevTools detach when not hidden, `loadFile` in prod
- `will-navigate` guard: blocks external origins, allows env-driven dev origin
- `window.open()` denial
- `focus` → `onFocusChange` forwarding with active sid
- `maximize` / `unmaximize` → `window:maximizedChanged` broadcast
- `show` → `window:afterShow` re-send so renderer can clear fade opacity
- close × pref=`tray` → preventDefault + fade(180ms) → hide
- close × pref=`tray` coalesces repeat presses (single hide)
- close × pref=`quit` → setIsQuitting(true), no preventDefault, no hide
- close × isQuitting=true short-circuits (no preventDefault, no pref lookup)
- close × `CCSM_DEV_QUIT_ON_CLOSE=1` dev escape hatch (quits even with pref=tray, never reads pref)
- close × pref=`ask` → IPC `window:askCloseAction` round-trip
  - resolve choice=`tray` + dontAskAgain ⇒ `setCloseAction('tray')` + hide
  - resolve choice=`quit` ⇒ `app.quit()` + setIsQuitting(true), never persists
  - resolve choice=`cancel` (even with dontAskAgain) ⇒ never persists, never destroys
  - stale `requestId` ignored
  - repeat close presses while ask in flight ⇒ no stacked dialogs
  - `CLOSE_ASK_TIMEOUT_MS` (10s) fallback ⇒ warn + tray hide, never persists
  - destroyed renderer at send time ⇒ straight to tray fallback
- `closed` event removes IPC listener + clears pending ask timer (timeout post-close is a no-op)

## Test plan
- [x] `npm run typecheck`
- [x] `npx vitest run electron/window/__tests__/createWindow.test.ts` → 45/45 pass
- [x] `npx vitest run ... --coverage` confirms 18.42% → 96.49% lines on `createWindow.ts`
- [x] No changes to `createWindow.ts` source

Note: full `npm test` shows 21 pre-existing failures in `electron/db.ts` / `load-smoke.test.ts` from a `better-sqlite3` NODE_MODULE_VERSION 127 ABI mismatch in the local env — unrelated to this PR.

### DIFF
--- a/electron/window/__tests__/createWindow.test.ts
+++ b/electron/window/__tests__/createWindow.test.ts
@@ -1,9 +1,72 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 
 // Capture context-menu listener + popup invocations from the BrowserWindow
 // mock. Each test flushes via beforeEach so `popupCalls.length === 0` is the
 // reverse-verify baseline that proves the listener actually wired up.
 const popupCalls: Array<{ items: unknown[]; window: unknown }> = [];
+
+// Stash for the most-recently-constructed BrowserWindow mock. The `createWindow`
+// factory builds the window via `new BrowserWindow(...)` (not a deps-injected
+// ctor — the deps bag intentionally only carries cross-module references), so
+// the only seam is `vi.mock('electron')`. The mock ctor pushes the latest
+// instance into `latestWin` so each factory test can drive its listeners.
+interface FakeWin {
+  ctorOpts: Record<string, unknown>;
+  listeners: Map<string, (...args: unknown[]) => void>;
+  webContentsListeners: Map<string, (...args: unknown[]) => void>;
+  hide: ReturnType<typeof vi.fn>;
+  show: ReturnType<typeof vi.fn>;
+  loadURL: ReturnType<typeof vi.fn>;
+  loadFile: ReturnType<typeof vi.fn>;
+  isDestroyed: ReturnType<typeof vi.fn>;
+  isMaximized: ReturnType<typeof vi.fn>;
+  setMenuBarVisibility: ReturnType<typeof vi.fn>;
+  webContents: {
+    on: ReturnType<typeof vi.fn>;
+    send: ReturnType<typeof vi.fn>;
+    setWindowOpenHandler: ReturnType<typeof vi.fn>;
+    openDevTools: ReturnType<typeof vi.fn>;
+    focus: ReturnType<typeof vi.fn>;
+    setBackgroundThrottling: ReturnType<typeof vi.fn>;
+    isDestroyed: ReturnType<typeof vi.fn>;
+  };
+  on: ReturnType<typeof vi.fn>;
+}
+
+let latestWin: FakeWin | null = null;
+let appQuitMock: ReturnType<typeof vi.fn>;
+
+function makeFakeWin(opts: Record<string, unknown>): FakeWin {
+  const listeners = new Map<string, (...args: unknown[]) => void>();
+  const webContentsListeners = new Map<string, (...args: unknown[]) => void>();
+  const win: FakeWin = {
+    ctorOpts: opts,
+    listeners,
+    webContentsListeners,
+    hide: vi.fn(),
+    show: vi.fn(),
+    loadURL: vi.fn(),
+    loadFile: vi.fn(),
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
**APPROVE** (posted as comment — GH blocks self-approval on own PR).

Test approach is sound — `vi.mock('electron')` is the only reasonable seam for a `new BrowserWindow(...)` factory, and capturing listeners into maps lets tests drive real branches in the production close-action choreography rather than asserting mock-interactions.

**Layer 1**
- All four 'deviations from the brief' are the worker matching source reality, not departing from it: `createWindow` (not `createMainWindow`), `'ask'` is genuinely IPC roundtrip (`window:askCloseAction` / `window:resolveCloseAction`), hidden mode is `x:-32000` with `show:true` (documented in source — Chromium throttles `show:false` rAF to ~1Hz and breaks dnd-kit), and `getCloseAction`/`setCloseAction` are module imports per source comment ('stateless module imports, we re-import here rather than plumbing through the bag'). Brief was written before checking source; tests match reality.
- Uncovered try/catch around `webContents.send` (lines 461-467): genuinely low-value. `isDestroyed()` path is already covered; the remaining surface is `send()` itself throwing in the microsecond after `isDestroyed()` returned false — covering it would require monkey-patching `send` to throw and tests nothing realistic. Accept.
- Reverse-verified mentally: stashing `if (deps.getIsQuitting()) return` would fail the `getCloseActionMock` not-called assertion; stashing the requestId guard would fail the stale-resolve test; stashing the hidden offscreen branch would fail the ctorOpts test. Tests would catch real regressions.

**Layer 2**
- Mock/timer/env cleanup is complete: `useFakeTimers`/`useRealTimers` paired across beforeEach/afterEach, `mockReset` on close-action mocks, `appQuitMock` rebound per test, env keys deleted, `latestWin` nulled, `popupCalls` cleared, `warnSpy.mockRestore` in afterEach.
- The closure trick that lets `appQuitMock` be rebound per test (the `electron` mock wraps `(...args) => appQuitMock(...args)` instead of using `appQuitMock` directly) works, but is subtle — a one-line comment would help future maintainers.
- Assertions check specific channels, payloads, call counts — not just 'was called'. No tests pass against broken production behavior.

**Nits (non-blocking, no need to act):**
1. `'closed event clears any pending ask timer'` is weaker than its description. The production `closed` handler does both `clearTimeout` and `pendingAsk = null`; the timer callback's own `if (!pendingAsk ...)` guard means the test passes even if `clearTimeout` were removed. To actually pin the `clearTimeout` line, spy on it or assert `warnSpy` was not called. Production is double-protected so impact is small.
2. `'prod mode loads the packaged renderer file'` doesn't lock down the path — could assert `loadFile` arg `endsWith('renderer/index.html')`.

`better-sqlite3` ABI failures in unrelated test files are environmental, not a regression from this PR. 18% → 96% lines on createWindow.ts is real coverage, not mock-driven inflation.

PR #1285, opened 2026-05-21
test(import): cover ImportDialog UI + import-scanner edge cases (batch 4)

Jiahui-Gu:
## Summary

Test-coverage batch 4 — pairs `src/components/ImportDialog.tsx` (was 28% lines / 9% branches) with `electron/import-scanner.ts` (was 53%). Dialog input is scanner output, so testing the two together locks down the cross-process data contract instead of letting each side drift behind ad-hoc mocks.

No source-file changes.

## Coverage delta

| File | Before (lines / branches) | After (lines / branches) |
|---|---|---|
| `src/components/ImportDialog.tsx` | 28% / 9% | 97.56% / 88.67% |
| `electron/import-scanner.ts` | 53% / — | 96.81% (stmts) / 89.94% |

## Scanner cases (41 tests, `electron/__tests__/import-scanner.test.ts`)

Backed by an on-disk fixture tree under a per-test tmp dir + `CLAUDE_CONFIG_DIR` override (real `fs.promises.readdir` / `createReadStream` paths exercised, not mocked):

1. missing `~/.claude/projects` → `[]`, no throw
2. malformed JSONL lines skipped, valid frames continue parsing
3. empty / metadata-less / sidechain (`isSidechain: true`) / `parentUuid` transcripts dropped
4. mtime-desc ordering across multiple project directories
5. `ccsm-*` / `agentory-*` temp-dir noise filter (incl. macOS `/var/folders/`, Windows `AppData\Local\Temp`, legacy `/tmp` aliases)
6. title fallback chain: `aiTitle` > `firstUserText` > `(untitled session)`
7. `<command-…>` wrapped user lines never become titles
8. runtime `CLAUDE_CONFIG_DIR` mutation honored between calls (#812 regression)
9. `parseHead` + `isSidechainFrame` + `isCCSMTempCwd` + `deriveRecentCwds` units (frequency-then-recency ranking, `~` placeholder skip, `windowSize` / `max` bounds)

## Dialog cases (19 tests, `tests/components/ImportDialog.test.tsx`)

Mocks `useStore` (selector form) + `window.ccsm.scanImportable` so the component runs against the same payload shape the scanner emits:

1. loading copy while scan promise is pending
2. empty state when scanner returns `[]`
3. empty state when scanner rejects
4. already-known sessions filtered out before display
5. row/bucket/select-all/deselect toggles + collapse/expand
6. primary action disabled while nothing selected
7. submit fans out one `importSession()` per selected row carrying `cwd` / `title` / `projectDir` / `resumeSessionId` verbatim (the contract reviewer locks down)
8. focused-group routing → uses focused group id
9. no focus + existing `Imported` group → reuses bucket, no `createGroup`
10. no focus + no bucket → `createGroup('Imported')` + uses new id
11. archive-kind focused group ignored, falls back to `Imported`
12. missing `window.ccsm` makes submit a no-op
13. `open=false` skips scan; close→reopen clears stale selection + re-scans

## Test plan

- [x] New scanner suite: 41/41 pass
- [x] New dialog suite: 19/19 pass
- [x] Full vitest suite: all pre-existing pass; the 21 failures (`tests/electron-load-smoke.test.ts` + `electron/__tests__/db-hardening.test.ts`) are pre-existing on `origin/working` (better-sqlite3 native module / missing build) and unrelated to this PR
- [x] Coverage thresholds: both target files now well over the 60% / 50% project thresholds

### DIFF
--- a/electron/__tests__/import-scanner.test.ts
+++ b/electron/__tests__/import-scanner.test.ts
@@ -1,0 +1,444 @@
+// UT for electron/import-scanner.ts — exercises the JSONL → ScannableSession
+// pipeline using a real on-disk fixture tree under a per-test tmp dir.
+//
+// Coverage targets the import flow's data contract:
+//   1. Missing `~/.claude/projects` → empty result, no throw
+//   2. Malformed JSONL lines → skipped (not fatal)
+//   3. Empty / metadata-less transcripts → dropped (resolve(null) path)
+//   4. Sidechain/sub-agent transcripts → filtered out
+//   5. Sort order — newest mtime first
+//   6. ccsm/agentory temp-cwd noise filter
+//   7. Title fallback: aiTitle > firstUserText > "(untitled session)"
+//   8. <command-...> wrapped first user lines do NOT become titles
+//   9. parseHead pure helper edge cases
+//  10. deriveRecentCwds: frequency-then-recency ranking + `~` skip
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import {
+  scanImportableSessions,
+  parseHead,
+  deriveRecentCwds,
+  isCCSMTempCwd,
+  isSidechainFrame,
+} from '../import-scanner';
+
+let tmpRoot: string;
+let projectsDir: string;
+let savedEnv: string | undefined;
+
+function mkProject(projectKey: string): string {
+  const dir = path.join(projectsDir, projectKey);
+  fs.mkdirSync(dir, { recursive: true });
+  return dir;
+}
+
+function writeJsonl(file: string, lines: Array<unknown | string>) {
+  const body = lines
+    .map((l) => (typeof l === 'string' ? l : JSON.stringify(l)))
+    .join('\n');
+  fs.writeFileSync(file, body);
+}
+
+function setMtime(file: string, ms: number) {
+  const t = ms / 1000;
+  fs.utimesSync(file, t, t);
+}
+
+beforeEach(() => {
+  tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'ccsm-import-scanner-test-'));
+  projectsDir = path.join(tmpRoot, 'projects');
+  fs.mkdirSync(projectsDir, { recursive: true });
+  savedEnv = process.env.CLAUDE_CONFIG_DIR;
+  process.env.CLAUDE_CONFIG_DIR = tmpRoot;
+});
+
+afterEach(() => {
+  if (savedEnv === undefined) delete process.env.CLAUDE_CONFIG_DIR;
+  else process.env.CLAUDE_CONFIG_DIR = savedEnv;
+  try {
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Reviewer verdict: **LGTM, safe to merge** (posting as comment because GitHub blocks self-approve on the PR author's account).

Reviewed against the import-flow brief, ran both suites on Windows (60/60 in 5.7s, no flake), reverse-verified with three mutations.

**Layer 1 (necessity)**
- On-disk fixture for scanner is the right call — scanner is on the import critical path and exercises `fs.promises.readdir` / `createReadStream` / `readline` / `stat` / platform-aware `os.tmpdir()` string ops that fs-mocks would silently dodge.
- `CLAUDE_CONFIG_DIR` regression lock-in is real and documented at `electron/shared/claudePaths.ts:1-25` (the file header literally calls out the import-scanner regression class). `projectsRoot()` correctly resolves lazily per call.
- Dialog selector mock is fine — tests assert on rendered output (`getByText`, `/N selected/i`, row presence/absence), not just store calls. React-owned `useState` drives re-renders, so the container-mutation concern doesn't bite.
- Brief was "import flow"; coverage spans scan → dialog filter (known sessions) → selection state → submit fan-out with the verbatim `{name, cwd, groupId, resumeSessionId, projectDir}` contract. Sidechain / `parseHead` / `deriveRecentCwds` / `isCCSMTempCwd` / group-routing extras live in the same two files and are cheap branch coverage — justified scope creep.

**Layer 2 (implementation)**
- Per-test `mkdtempSync` + `afterEach` rmSync (recursive+force, try/catch). Env var save/restore handles both set and unset prior states. No tmpdir leaks observed.
- Windows path handling verified: ran on Win11 with tmpdir = `C:\Users\...\AppData\Local\Temp\...`; scanner's `replace(/\/g, '/')` normalization + the case-insensitive `appdata/local/temp` regex fallback both exercised. macOS CI also green.
- 41/19 split justified by surface area (scanner has 5 exported helpers + streaming reader; dialog is one component).

**Reverse-verify (3 mutations, all caught)**
1. Pin `projectsRoot()` at module load + flip `isSidechainFrame` returns → 19 failures incl. the env-mutation, sidechain, and parseHead-sidechain tests.
2. Force `focused = undefined` in dialog `doImport` → `routes imports into the focused group when one is selected` fails (only this one — clean isolation, good test specificity).

**Pre-existing failures unrelated**: `tests/electron-load-smoke.test.ts` (last touched #630) and `electron/__tests__/db-hardening.test.ts` (last touched #278) fail on missing electron build / better-sqlite3 NODE_MODULE_VERSION — orthogonal to a tests-only PR.

**Nits (non-blocking)**
- `tests/components/ImportDialog.test.tsx:67` — `deferredScan()` `reject` field is set but never used; dead branch in the helper.
- `tests/components/ImportDialog.test.tsx:398-400` — `void within;` to silence lint for an unused import is awkward; just drop the import.
- `electron/__tests__/import-scanner.test.ts:44` — `setMtime` divides ms→sec as a float; harmless here (only relative ordering is asserted) but a `Math.floor` would be clearer given FAT-style 2s rounding on some FSes.

PR #1286, opened 2026-05-21
test(store/db): cover hydrate + sqlite recovery paths (batch 5)

Jiahui-Gu:
## Summary

Coverage batch 5 — fills the gap on two boot-critical modules manager flagged from coverage triage:

- **src/stores/store.ts**: line% **9 → 95.58**, branch% **0 → 81.81**
- **electron/db.ts**: line% **35 → 92.64**, branch% **10 → 90.00**

No source files touched. Existing `db-hardening.test.ts` placeholder for `db.test.ts` was replaced with real tests (db-hardening covers a separate slice of branches; the two files are complementary).

### `tests/stores/store.test.ts` (new, 15 tests)

`hydrateStore()` boot sequence:
1. `hydrated` flips false → true through the awaited path
2. Persisted snapshot applied: sessions / groups / theme / fontSize / sidebarWidth
3. Fallback when `activeId` no longer matches any session (stale persisted id)
4. Fresh-install path (no main snapshot) — defaults survive
5. Idempotent — second call doesn't re-read IDB
6. `scrollbackLines` key applied when present
7. Scrollback load error leaves slice default intact
8. `hydrateDrafts` ordering — runs before `hydrated` flips
9. Corrupt drafts blob doesn't blow up hydrate
10. Persist subscriber writes a curated snapshot through `saveState` on persisted-field mutation
11. Persist subscriber early-bails on non-persisted (`flashStates`) mutation
12. Deferred `userHome` + `defaultModel` seeding from main IPC
13. `pathsExist` tags vanished cwds with `cwdMissing`
14. `pathsExist` error is swallowed without blocking hydrate
15. `window.__ccsmHydrationTrace` populated with start/done timestamps

Each test uses `vi.resetModules()` + dynamic import to reset the module-level `hydrated` singleton between cases.

### `electron/__tests__/db.test.ts` (extended from placeholder, 13 tests)

- `loadState` miss returns null on a fresh db
- `saveState` upsert: 3x writes of same key keeps last value (no PK conflict)
- close + reopen preserves rows in the file
- `initDb` is idempotent (same handle on repeated calls)
- `getDb` matches `initDb`
- Downgrade: on-disk `user_version > SCHEMA_VERSION` logs warning + proceeds
- Matching `user_version`: no migrate, no warn
- Legacy `endpoints` / `endpoint_models` / `messages` tables dropped on init; `app_state` survives
- Corruption recovery clears stale `-wal` / `-shm` sidecar markers
- `__setDbForTests` injection runs schema bootstrap + reroutes save/load
- `__setDbForTests` resets prepared-statement cache so the new handle compiles its own
- `closeDb` safe on never-opened db
- `closeDb` + next `initDb` returns a fresh handle (no stale singleton)

## Pre-flight note

Manager flagged that `db-hardening.test.ts` was failing with NODE_MODULE_VERSION 145 vs 127 ABI mismatch in worker preflight. Root cause was empty `node_modules/` in the fresh worktree (npm install never ran). After `npm install` + `npm rebuild better-sqlite3`, all db tests pass cleanly.

## Test plan

- [x] `npm run lint` — 0 errors (pre-existing warnings only)
- [x] `npx vitest run tests/stores/ tests/drafts.test.ts tests/persist-shape.test.ts electron/__tests__/db.test.ts electron/__tests__/db-hardening.test.ts` — 126/126 pass
- [x] `npx vitest run --coverage` confirms store.ts 95.58% / db.ts 92.64%
- [x] No source files (`src/stores/store.ts`, `electron/db.ts`) modified

Note: the only repo-wide failure on `npm test` is `tests/electron-load-smoke.test.ts` which requires `npm run build` to have produced `dist/electron/*.js` first — unrelated to this PR and pre-existing on `working`.

### DIFF
--- a/electron/__tests__/db.test.ts
+++ b/electron/__tests__/db.test.ts
@@ -1,8 +1,26 @@
-import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import * as os from 'os';
 import * as path from 'path';
 import * as fs from 'fs';
 import Database from 'better-sqlite3';
+
+// Wrap fs so the corruption-recovery test can pin that `electron/db.ts`
+// actually calls `unlinkSync` on the `-wal` / `-shm` sidecar paths.
+//
+// `vi.spyOn(fs, 'unlinkSync')` fails — the ESM namespace is frozen. The
+// `Module._cache` hack doesn't work either, because Vitest runs modules
+// inside its own loader (no shared CJS cache). The robust path is
+// `vi.mock('fs', …)` with `importActual`: every importer (test + production)
+// gets the same wrapper, which by default delegates to the real impl so
+// other tests' `mkdtempSync` / `writeFileSync` / etc. keep working. Tests
+// that need to assert the call shape override the wrapper.
+const unlinkSpy = vi.hoisted(() => ({ fn: undefined as unknown as ReturnType<typeof vi.fn> }));
+vi.mock('fs', async () => {
+  const actual = (await vi.importActual('fs')) as typeof import('fs');
+  const wrapped = vi.fn(actual.unlinkSync);
+  unlinkSpy.fn = wrapped;
+  return { ...actual, default: actual, unlinkSync: wrapped };
+});
 
 // `initDb` asks electron for the userData dir. Point it at a per-test tmp dir
 // so real app data isn't touched and each test gets a fresh database.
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
## Review verdict: 1 critical test-quality fix needed before merge

Solid coverage PR overall. 35/35 pass post-rebuild; reverse-verification on 5/6 randomly chosen mutations of production code correctly fails the corresponding test. One test does NOT actually pin its claimed contract — flagging before merge.

- **Store hydration coverage is well-targeted.** The 15 tests pin the actual contract: hydrated-flag flip, stale-activeId fallback, scrollback error-swallow, drafts ordering, persist-subscriber early-bail vs write, deferred IPCs (userHome/defaultModel/pathsExist), trace bookkeeping. Persist-subscriber semantics tests (curated snapshot fields, early-bail on non-persisted mutation) are the highest-value — easy to silently break.
- **db.ts coverage is meaningful, not happy-path inflation.** Real branches: downgrade warning, legacy-table drop, __setDbForTests injection + prepared-statement cache reset, close-then-reopen idempotency. Complementary to db-hardening.test.ts (corruption + fresh-stamp + prepared-cache reuse + db:save validator) — no overlap.
- **`vi.resetModules()` + dynamic re-import is acceptable.** The `let hydrated = false` module-singleton is by design (single boot per process); a `resetForTesting()` export would expose test-only surface area. Vitest's default per-file isolation prevents leak into adjacent test files.
- **Extending the existing db.test.ts placeholder is the right boundary.** Mirrors db-hardening's tmpdir + electron mock pattern; top-of-file comments declare each slice.

### Layer 2 (implementation) — reverse-verification results

| Production code broken | Test | Result |
|---|---|---|
| `setState({ hydrated: true })` -> `false` | 4 store tests | caught |
| Remove `stillActive ? ... : sessions[0]?.id` fallback | activeId-fallback | caught |
| Remove try/catch around scrollback load | scrollback-error | caught |
| Bump `stored > SCHEMA_VERSION` to `> 9999` | downgrade-warning | caught |
| Remove `DROP TABLE` block | legacy-cleanup | caught |
| **Remove entire WAL/SHM unlink loop** | sidecar | **passes anyway** |

### Critical issue

**`db.test.ts` `removes leftover -wal / -shm sidecar files when backing up a corrupt main db` does not pin its contract.**

When the WAL/SHM cleanup block is fully removed from `ensureHealthyDb`, the test still passes:

1. Stale-marker assertions are guarded with `if (fs.existsSync(file + '-wal'))` / `'-shm'` — a missing file is silently treated as pass.
2. After `new Database(file)` + `journal_mode = WAL` with no subsequent writes, SQLite does not materialise sidecar files in the same test tick — by the time the assertions run, stale files have been cleared by the SQLite open path itself (or never re-emitted), not by the cleanup code under test.

Net effect: this test would have passed before the cleanup code existed. Fix options:

- **Preferred:** Mock `fs.unlinkSync` and assert it was called with both `file + '-wal'` and `file + '-shm'` during `initDb()`.
- Alternative: Drop the `if (fs.existsSync(...))` guards and assert directly — accepting the test can only pass when WAL has materialised (likely needs a forced write before the corruption setup).

### Suggestions (non-blocking)

- Deferred-IPC tests use `await new Promise((r) => setTimeout(r, 0))` twice to drain microtasks. Works today; if CI flakes, prefer `await vi.waitFor(() => expect(s.userHome).toBe('/home/me'))`.
- The `as unknown as PersistedState['sessions'][number]` casts are accurate per the slice's hydrate-time read pattern. A `makeSession()` factory would clean these up if more session-shaped tests land.

### Acknowledgement

- Worker correctly pre-empted the `vi.resetModules()` design tradeoff in their brief — right engineering judgement.
- Self-fixing the better-sqlite3 ABI mismatch (`npm rebuild`) before claiming green is correct discipline; I re-hit it locally and the same fix worked.
- Branch coverage 0 -> 81.81% on store.ts via 15 tests is dense, not sprawling.

### Pre-existing failures unrelated to this PR

`tests/electron-load-smoke.test.ts` requires `npm run build` to materialise `dist/electron/*.js`. Confirmed independent — out of scope.

### Action

Fix the sidecar test (mock-based assertion preferred), then good to merge.

Jiahui-Gu:
Re-review (independent reviewer): reviewer fix verified, **LGTM — safe to merge**.

**Reverse-verify (the thing that was missing before):** commented out the `for (const ext of ['-wal', '-shm'])` loop in electron/db.ts:112-118 → both corruption-recovery tests fail with empty `paths` array (`expected [] to include '...ccsm.db-wal'`). Restored → 14/14 db tests pass. The assertion now actually pins the syscall.

**Leak risk:** none. `vi.mock('fs')` is scoped per test file in Vitest's module graph; grep across electron/__tests__/ confirms no other test file mocks fs or asserts on unlinkSync spies. Full electron suite 114/114 green.

**vi.hoisted pattern:** correctly implemented — single shared ref object, spy instantiated once inside the factory, `afterEach` resets impl to real fs via `mockImplementation(realFs.unlinkSync)` (not `mockReset`, which would wipe the default and break the next test — comment on line 230 calls this out).

**Other 12 db tests:** spot-checked lines 27, 99-101, 135-144, 169 — all use real `fs.mkdtempSync` / `writeFileSync` / `Database`, forwarded untouched via `...actual` spread.

**store.test.ts:** zero changes in the fix commit.

PR #1287, opened 2026-05-21
docs(preload): clarify install order is cosmetic, not load-bearing

Jiahui-Gu:
Tightens the doc-comment in `electron/preload/index.ts` after PR #1284 reviewer flagged a contradiction: the source said "order is not load-bearing" but a recent PR body claimed Sentry-first was required to capture bridge load-time throws.

Read all 5 `install*Bridge` functions — they only do `ipcRenderer.on` registrations + `contextBridge.exposeInMainWorld`, neither of which can realistically throw at load time. Picked the simpler truth: order is cosmetic, Sentry sits at the top by convention. No test added because there's nothing real to assert.

### DIFF
--- a/electron/preload/index.ts
+++ b/electron/preload/index.ts
@@ -2,9 +2,11 @@
 // `path.join(__dirname, '..', 'preload', 'index.js')` (resolves to
 // `dist/electron/preload/index.js` post-tsc). Splits into 5 single-concern
 // bridges (#769, SRP wave-2 PR-A) — each bridge owns its own listener
-// sets, IPC channels, and exposed type. The order of `install*` calls is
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (cannot self-approve via gh). Verified: spot-checked ccsmPty + ccsmNotify install fns — both are pure `ipcRenderer.on` registrations + `contextBridge.exposeInMainWorld`, no fs/invoke roundtrips/throws at load time. Listener callbacks are try/catch-wrapped and fire later at event time, not at install. New comment is clearer (explicit about what bridges do + Sentry-by-convention framing) without overpromising. Diff confirmed comment-only.

PR #1288, opened 2026-05-21
test(lifecycle): restore Module.prototype.require patch in afterAll

Jiahui-Gu:
## Summary

`electron/lifecycle/__tests__/appLifecycle.test.ts` monkey-patches `Module.prototype.require` at file top-level to intercept the production CJS dynamic `require('../i18n')` (used in `appLifecycle.ts` to dodge a circular import). Before this PR the patch was never restored.

Today vitest's default per-file isolation hides this — each test file gets its own module graph, so the un-restored patch dies with the worker. But if anyone later flips to `pool: 'threads'` + `isolate: false` (a common perf tweak), the patch would bleed into sibling test files that load via CJS `require`, including the other two production sites at `electron/ipc/systemIpc.ts` and `electron/tray/createTray.ts`, and surface as mysterious "wrong i18n result" failures.

Fix: cache the original `Module.prototype.require` in the top-level const we already had, and restore it in `afterAll`. Five lines.

## Verification

- `npx vitest run electron/lifecycle/__tests__/` — 20 passed
- `npx vitest run electron/lifecycle/__tests__/appLifecycle.test.ts electron/lifecycle/__tests__/singleInstance.test.ts --no-isolate --pool=threads` — 20 passed (the exact configuration the leak would impact)
- `npm run typecheck` — clean
- `npx eslint electron/lifecycle/__tests__/appLifecycle.test.ts` — clean

## Follow-up (not this PR)

`appLifecycle.ts` could take `i18n` as a dep (same pattern as the other `LifecycleDeps` fields), eliminating the need for any CJS-require patch in the test. Larger change — also applies to `systemIpc.ts` and `createTray.ts` — out of scope for this hygiene fix.

### DIFF
--- a/electron/lifecycle/__tests__/appLifecycle.test.ts
+++ b/electron/lifecycle/__tests__/appLifecycle.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterAll } from 'vitest';
 import type { App } from 'electron';
 
 // Mock electron — appLifecycle.ts top-level imports `Menu`, which would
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
LGTM (cannot self-approve — flagging review verdict for merge).

Verified:
- Diff correct: `afterAll` (not `afterEach`) restores `ModuleProto.require` from top-level-cached `originalRequire`. Cadence matches the once-per-file patch.
- Ran `npx vitest run electron/lifecycle/__tests__/appLifecycle.test.ts electron/lifecycle/__tests__/singleInstance.test.ts --no-isolate --pool=threads` locally: 20/20 pass. This is the exact config that would expose the leak — fix is proven.
- Vitest's `afterAll` runs regardless of mid-suite failures, so partial-failure path is safe.
- Comment update accurately documents the why (future `isolate: false` flip).

Follow-up DI refactor (appLifecycle.ts + systemIpc.ts + createTray.ts) is correctly scoped out.

PR #1290, opened 2026-05-21
feat(terminal): native CLI-style right-click — copy on selection / paste on empty (#41)

Jiahui-Gu:
## Summary

Right-click in the terminal pane now mirrors Windows Terminal / gnome-terminal: selection present → copy + clear selection (visual feedback); no selection → paste. No popover, ever. Adds Ctrl/Cmd+A for select-all to retain the keyboard entry point the popup's "Select All" item used to provide.

The native context menu (Cut/Copy/Paste/SelectAll) installed in `electron/window/createWindow.ts` still applies to the sidebar and inputs — it's only suppressed for the terminal pane, on a per-click one-shot basis sent via a new `window.ccsmShell.suppressContextMenuOnce()` IPC.

## Layer 1 decisions

- **Scope**: only the terminal pane changes. Sidebar / editable inputs keep the existing native context menu (cut / copy / paste / select-all).
- **macOS gating**: NOT applied. User is Windows-only; gnome-terminal / Windows Terminal style is the convergent expectation. Revisit if a mac user reports the loss of the native popover.
- **Select-all preserved**: Ctrl/Cmd+A → `term.selectAll()` wired in `attachCustomKeyEventHandler`. xterm has no built-in keybinding for this, and removing the popover's "Select All" left no other entry point.
- **Paste pipeline**: `terminalPaste()` routes through the same single canonical sink the Ctrl+V keydown hook uses, per the 4-layer paste model documented in `project_xterm_paste_layers.md` (xterm keydown hook + xterm built-in paste pipeline + Electron `role:'paste'` + the keyboard→native-paste handoff timing gotcha). Reuses the existing `keyboardPasteHandled` flag with the canonical `setTimeout(0)` reset so a browser-dispatched follow-up `paste` event is suppressed instead of double-pasting.

## Implementation

- `electron/window/createWindow.ts` — process-wide suppression deadline (100ms window) consulted at the top of `installContextMenu`. Pure helpers `isContextMenuSuppressed` / `recordContextMenuSuppression` extracted for unit test. Idempotent IPC handler `installContextMenuSuppressIpc` registered alongside the other `whenReady()` handlers in `electron/main.ts`. Electron's `webContents.on('context-menu')` fires regardless of DOM `preventDefault()`, so the deadline is the only main-process signal that can inhibit the popup.
- `electron/preload/bridges/ccsmShell.ts` (new) — exposes `window.ccsmShell.suppressContextMenuOnce()`. Stays a separate bridge rather than folding into `ccsm` core because the surface is one-shot UX-gesture plumbing, not app state.
- `src/terminal/xtermSingleton.ts` — new exports `terminalCopy()` (returns whether a selection existed, writes via `ccsmPty.clipboard.writeText`, clears the selection) and `terminalPaste()` (canonical paste sink). Plus the Ctrl/Cmd+A → `selectAll()` branch in the keydown handler.
- `src/components/TerminalPane.tsx` — `onContextMenu` on the host div: preventDefault, send one-shot suppression IPC, branch on `terminalCopy()` return → copy or paste. Robust to a missing or throwing `ccsmShell` bridge (e2e probes may not install it).
- `src/pty.d.ts` — types `window.ccsmShell` as optional.

## Tests

- `electron/window/__tests__/createWindow.test.ts` — pure deciders (`isContextMenuSuppressed` boundary conditions, `recordContextMenuSuppression` deadline math), `installContextMenu` honouring the suppression deadline, **one-shot consumption** (the next right-click after a single suppression still pops up — proves no over-suppression), expired deadline does not block, IPC handler idempotency.
- `electron/preload/bridges/__tests__/ccsmShell.test.ts` (new) — bridge surface + IPC channel name pin (wire contract with main).
- `electron/preload/bridges/__tests__/index.test.ts` — extended to assert `installCcsmShellBridge` is invoked.
- `tests/components/TerminalPane.test.tsx` (new) — `onContextMenu` copy / paste branches, preventDefault on the React event, suppress-IPC fired, resilient to missing bridge + throwing bridge.
- `tests/terminal/useXtermSingleton.test.tsx` — `terminalCopy` / `terminalPaste` direct contract, follow-up native paste suppression via the shared `keyboardPasteHandled` flag, Ctrl/Cmd+A → `selectAll`, plain `a` and Ctrl+Shift+A are NOT intercepted.

Full suite: **1625 / 1625 passing**. Typecheck + lint clean (warnings pre-existing).

## Screenshots

Not captured this round — no running Electron instance available to the worker. Reviewer can pull the branch and dogfood: right-click in terminal with text selected → text should land in clipboard immediately + selection clear; right-click without selection → clipboard contents should paste into PTY immediately; no popover should appear in either case. Right-click in the sidebar should still show the native cut/copy/paste/selectAll menu (regression check).

## Test plan

- [ ] Right-click with selection → text copied to clipboard, selection cleared, no popover
- [ ] Right-click without selection → clipboard pastes into PTY, no popover
- [ ] Ctrl+A in terminal selects all visible buffer content
- [ ] Right-click in sidebar still shows the native popover (no over-suppression)
- [ ] Ctrl+V paste still works (no double-paste / zero-paste regression)
- [ ] Ctrl+C with selection still copies; Ctrl+C without selection still sends SIGINT

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -29,7 +29,7 @@ import { app, BrowserWindow, ipcMain, type Tray } from 'electron';
 import { initDb, closeDb } from './db';
 import { buildTrayIcon } from './branding/icon';
 import { initSentry } from './sentry/init';
-import { createWindow as createMainWindowFactory } from './window/createWindow';
+import { createWindow as createMainWindowFactory, installContextMenuSuppressIpc } from './window/createWindow';
 import { createTray, type TrayController } from './tray/createTray';
 
 // safety net — escaped main-proc rejections kill app on Node 20+ default
@@ -223,6 +223,13 @@ app.whenReady().then(() => {
   });
   registerWindowIpc({ ipcMain });
   registerUtilityIpc({ ipcMain });
+
+  // Process-wide IPC for the terminal pane's `onContextMenu` handler to
+  // ask main to skip the native context menu for one upcoming click.
+  // Registered here (alongside the other IPC handlers) rather than in
+  // `createWindow` because the channel is window-agnostic — see
+  // `installContextMenuSuppressIpc` in electron/window/createWindow.ts.
+  installContextMenuSuppressIpc(ipcMain);
 
   installUpdaterIpc();
 
--- a/electron/preload/bridges/__tests__/ccsmShell.test.ts
+++ b/electron/preload/bridges/__tests__/ccsmShell.test.ts
@@ -1,0 +1,62 @@
+// Pins the `window.ccsmShell` preload bridge — single one-shot for
+// renderer-driven native-context-menu suppression (Task #41). The IPC
+// channel name `shell:suppressContextMenuOnce` is a wire contract with
+// `installContextMenuSuppressIpc` in electron/window/createWindow.ts;
+// renaming one side silently breaks the terminal pane's right-click
+// inline copy/paste UX (the native menu races on top).
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const { exposeSpy, sendSpy } = vi.hoisted(() => ({
+  exposeSpy: vi.fn(),
+  sendSpy: vi.fn(),
+}));
+
+vi.mock('electron', () => ({
+  contextBridge: { exposeInMainWorld: exposeSpy },
+  ipcRenderer: {
+    invoke: vi.fn(),
+    send: sendSpy,
+    on: vi.fn(),
+    removeListener: vi.fn(),
+  },
+}));
+
+import { installCcsmShellBridge } from '../ccsmShell';
+
+type AnyApi = Record<string, unknown>;
+
+function lastApi(): AnyApi {
+  const last = exposeSpy.mock.calls[exposeSpy.mock.calls.length - 1];
+  expect(last[0]).toBe('ccsmShell');
+  return last[1] as AnyApi;
+}
+
+describe('ccsmShell preload bridge', () => {
+  beforeEach(() => {
+    exposeSpy.mockClear();
+    sendSpy.mockReset();
+    installCcsmShellBridge();
+  });
+
+  it('exposes a stable key set under "ccsmShell"', () => {
+    const api = lastApi();
+    expect(Object.keys(api).sort()).toEqual(['suppressContextMenuOnce']);
+  });
+
+  it('suppressContextMenuOnce sends one-way IPC on the canonical channel', () => {
+    const api = lastApi();
+    (api.suppressContextMenuOnce as () => void)();
+    expect(sendSpy).toHaveBeenCalledWith('shell:suppressContextMenuOnce');
+    expect(sendSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('repeated calls send one IPC each (no internal coalescing)', () => {
+    const api = lastApi();
+    (api.suppressContextMenuOnce as () => void)();
+    (api.suppressContextMenuOnce as () => void)();
+    (api.suppressContextMenuOnce as () => void)();
+    // Main side handles coalescing via its deadline; bridge stays dumb.
+    expect(sendSpy).toHaveBeenCalledTimes(3);
+  });
+});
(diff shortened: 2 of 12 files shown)


Jiahui-Gu (commented):
**Verdict: APPROVE** (submitted as comment because GitHub blocks self-approve; PR author = reviewer identity)

**Q1 — 100ms deadline race**: Sound. `installContextMenu` resets `contextMenuSuppressedUntilMs = 0` immediately on consuming the deadline (createWindow.ts:480-486), so rapid sidebar-following-terminal right-clicks are safe in the realistic path. Theoretical edge case (>50ms gap between dom contextmenu and webContents context-menu under extreme load) would wrongly suppress sidebar — unobservable in practice. Nit, not blocker.

**Q2 — One-shot + deadline both: not over-engineering.** Different failure modes — consumption handles the normal path (no permanent mute when context-menu fires), expiry handles the abnormal path (IPC arrived without an accompanying context-menu event → self-heal in 100ms). Defense in depth, justified.

**Q3 — Pure deciders**: Good factoring, not duplicated elsewhere, testable without IPC. Minor cosmetic: test name `"later calls extend... never shrink it relative to live now"` slightly oversells — the function unconditionally overwrites; monotonicity is an external assumption.

**Q4 — `ccsmShell` as new bridge**: Justified. `shell` namespace = renderer-driven main-process UX-gesture plumbing, semantically distinct from `ccsmCore` (state) and `ccsmPty` (data). 35 LOC, no `invoke` channel, no closures. Reasonable extension point.

**Q5 — `term.selectAll()`**: Real xterm API, verified in `@xterm/xterm/typings/xterm.d.ts:1159`. Not a no-op.

**Q6 — Paste sink reuse**: Verified safe. `terminalPaste()` is structurally identical to in-file `pasteFromClipboard()` — sets `keyboardPasteHandled = true`, schedules `setTimeout(0)` reset (NOT microtask, per the xterm-paste-layers gotcha), and the capture-phase listener at xtermSingleton.ts:251-267 honors the flag. No double-paste, no zero-paste.

## Layer 2 (implementation)

**Tests pin actual contracts.** Notable: `<` boundary on the deadline, the one-shot-consumed invariant, expired-deadline safety net, `clearSelection` after copy, `keyboardPasteHandled` flag pinning, and crucially `Ctrl+Shift+A does NOT select-all` + `plain "a" is left to xterm` so user typing isn't captured.

**Reverse-verified by inspection** (did not stash+rerun; CI 7/7 green):
- Always-suppressed impl → fails `an expired suppression deadline does not block the menu`.
- Non-consuming impl → fails `the suppression is one-shot: a second right-click gets the menu`.
- Copy-without-clear impl → fails `clearSelectionSpy).toHaveBeenCalledTimes(1)`.

**`pty.d.ts` type matches runtime**: `ccsmShell?: { suppressContextMenuOnce(): void }` ↔ what `contextBridge.exposeInMainWorld('ccsmShell', ...)` exposes. Optional `?` correct.

**No coupling**: `TerminalPane.tsx` knows nothing about the 100ms deadline; only calls the bridge method. Deadline lives in `createWindow.ts`.

**Sidebar-regression test not explicit** — covered in principle by the two deadline tests (both fire context-menu without preceding IPC and assert `popupCalls).toHaveLength(1)`). A test explicitly named for the sidebar path would be more discoverable. Nit.

## Dogfood verification

**Could not execute** — main worktree is on a different uncommitted branch (`refactor/agent-adapter-types-phase1`) and `npm run dev` is a long-running foreground process unsuitable for a non-interactive review session. Per memory `[[feedback_visual_fix_screenshots]]`, manager / user should pull the branch and walk the PR test-plan visually before merge.

Implementation correct by code review:
- selection → `terminalCopy()` writes clipboard + `clearSelection()` (xtermSingleton.ts:621-636)
- no selection → `terminalPaste()` (TerminalPane.tsx:530)
- Ctrl+A → `term.selectAll()` (xtermSingleton.ts:320-332)
- sidebar → never sends IPC → deadline=0 → native menu
- Ctrl+V → single paste (canonical sink + `keyboardPasteHandled` honored)
- rapid right-click → one-shot consumed on first context-menu fire

## Blockers vs nits

**Blockers**: None.

**Nits**:
1. Test name slightly oversells overwrite-vs-monotonicity invariant.
2. No explicit "sidebar regression" test by name; covered in principle.
3. Visual dogfood pass should happen before merge per memory rule.

Recommend manager dogfoods the PR's test-plan checklist visually, then merge.

PR #1291, opened 2026-05-21
fix(ptyHost): evict zombie entry on KILL_EXIT_TIMEOUT_MS to unblock spawn-on-null

Jiahui-Gu:
## Summary

Follow-up from PR #1277 reviewer. The 3s `KILL_EXIT_TIMEOUT_MS` fallback in `kill()` resolved `false` but left the dead entry in `sessions`, so the renderer's next `pty:attach` (after the `reloadNonce` bump) got the zombie back and bypassed the spawn-on-null fallback. The viewer ended up registered on a dead pty that immediately fired `pty:exit` -> crash overlay -> user had to hit Retry manually.

## The bug

`electron/ptyHost/lifecycle.ts` `kill()`:
1. `pty.kill()` dispatched a signal but the OS never reaped the process.
2. `onExit` never fired -> the cleanup pump in `entryFactory` that calls `sessions.delete(sid)` never ran.
3. `KILL_EXIT_TIMEOUT_MS` (3s) fired, `kill()` resolved `false`.
4. Renderer awaited that, bumped `reloadNonce`, ran the attach hook.
5. `pty:attach` returned the still-present zombie entry's geometry.
6. Spawn-on-null fallback never fired. Renderer wired up listeners on a dead pty -> crash overlay.

## The fix

On timeout:
- Best-effort `entry.pty.kill('SIGKILL')` to escalate the signal. On Windows node-pty emulates signals so this may throw; caught + logged (pid leak is visible in console, but the user gets their fresh terminal).
- `sessions.delete(sid)` to evict the zombie.
- Then `resolve(false)`.

Next `pty:attach` returns null -> existing spawn-on-null fallback transparently spawns a fresh PTY. No UI changes, no IPC contract changes, no `kill()` signature change, no timeout duration change.

Picked option (a) from the reviewer's two options (best-effort SIGKILL + delete) over (b) (zombie-flag field consulted by attach) — single-file change, no new state for `attach` to consult.

## Tests

`electron/ptyHost/__tests__/lifecycle.test.ts` adds two cases to the `lifecycle.kill` block:

1. `evicts the zombie entry on timeout so subsequent attach returns null (spawn-on-null unblocked)` — pty.kill is silent, timer advances past 3s, asserts `pty.kill('SIGKILL')` was called, `sessions.has('s')` is false, and `L.attach(sessions, 's')` returns null.
2. `swallows SIGKILL throws on timeout (still evicts the zombie + logs)` — Windows path. SIGKILL throws -> entry still evicted, warn logged with "SIGKILL also failed".

## Reverse-verify trace

```
$ git stash push -- electron/ptyHost/lifecycle.ts
$ npx vitest run electron/ptyHost/__tests__/lifecycle.test.ts
 FAIL  lifecycle.kill > evicts the zombie entry on timeout
   expected pty.kill to have been called with 'SIGKILL' (called with [])
   expected sessions.has('s') to be false (was true)
   expected L.attach(sessions, 's') to be null
 FAIL  lifecycle.kill > swallows SIGKILL throws on timeout
   expected 1 to be 2  (calls counter)
 Test Files  1 failed
 Tests  2 failed | 30 passed (32)

$ git stash pop
$ npx vitest run electron/ptyHost/__tests__/lifecycle.test.ts
 Test Files  1 passed
 Tests  32 passed (32)
```

## Test plan
- [x] vitest `lifecycle.test.ts` 32/32
- [x] `tsc --noEmit` clean
- [x] `eslint` on changed files clean (only pre-existing unrelated warnings)
- [x] reverse-verified the new tests fail on un-fixed code
- [ ] e2e: not added — the wedged-pty scenario is hard to deterministically reproduce in Playwright; unit test covers the contract

### DIFF
--- a/electron/ptyHost/__tests__/lifecycle.test.ts
+++ b/electron/ptyHost/__tests__/lifecycle.test.ts
@@ -446,6 +446,67 @@ describe('lifecycle.kill', () => {
     vi.useRealTimers();
   });
 
+  it('evicts the zombie entry on timeout so subsequent attach returns null (spawn-on-null unblocked)', async () => {
+    // #1277 follow-up: when KILL_EXIT_TIMEOUT_MS fires, kill() resolves false
+    // but the entry was still in the map — a subsequent pty:attach would
+    // return the zombie and bypass the spawn-on-null fallback, leaving the
+    // renderer registered on a dead pty (crash overlay → manual Retry).
+    // Fix: force SIGKILL + delete from map on timeout.
+    vi.useFakeTimers();
+    const sessions = new Map<string, FakeEntry>();
+    const entry = makeFakeEntry();
+    entry.pty.pid = 500;
+    // Track SIGKILL escalation — pty.kill called twice (first signal-less in
+    // production code, second with 'SIGKILL' on timeout).
+    entry.pty.kill = vi.fn();
+    sessions.set('s', entry);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const p = L.kill(sessions as any, 's');
+
+    // Cross the timeout — SIGKILL escalation + zombie eviction.
+    await vi.advanceTimersByTimeAsync(L.KILL_EXIT_TIMEOUT_MS + 10);
+    await expect(p).resolves.toBe(false);
+
+    // SIGKILL escalation attempted on the wedged pty.
+    expect(entry.pty.kill).toHaveBeenCalledWith('SIGKILL');
+    // Zombie removed from registry — attach now returns null, letting the
+    // renderer's reloadNonce → spawn-on-null fallback create a fresh PTY.
+    expect(sessions.has('s')).toBe(false);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    expect(L.attach(sessions as any, 's')).toBeNull();
+    vi.useRealTimers();
+  });
+
+  it('swallows SIGKILL throws on timeout (still evicts the zombie + logs)', async () => {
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Reviewer verdict: APPROVE (posted as comment — gh blocks self-approve).

**Layer 1 (necessity)**:
- SIGKILL/leak trade-off documented in lifecycle.ts inline comment + warn includes pid (greppable).
- `if (sessions.get(sid) === entry) sessions.delete(sid)` — correct identity check; concurrent re-spawn safe.
- Windows SIGKILL honesty: comment says "may be a no-op", doesn't oversell. Test 2 pins throws-don't-escape; realistic no-op covered by test 1.

**Layer 2 (implementation)**:
- `expect(L.attach(...)).toBeNull()` is direct — load-bearing end-to-end-ish proof of the user-visible fix.
- console.warn matches existing lifecycle.ts log convention. No need to escalate to error.
- settle / pendingKills / exitDisposable bookkeeping from #1277 preserved untouched.
- Reverse-verify trace credible — failure messages map to the new assertions.

**CI**: ubuntu `harness-dnd` failure is the known sidebar cross-group DnD flake (`s1 did not land in g2 after cross-group drag`), tracked in `fix/e2e-ubuntu-dnd-stale-over-on-release`. Unrelated to pty. Re-run that job post-merge.

**Nit (non-blocking)**: test 2 SIGKILL-throws is more contract-test than realistic-Windows path; optional in-test comment that the realistic no-op path is implicitly covered by test 1.

PR #852, opened 2026-05-02
feat(monorepo): T0.3 tsconfig.base.json + per-package extends

Jiahui-Gu:
## Summary

Implements Task #13 ([T0.3] monorepo: tsconfig.base.json + per-package tsconfig.json extends), per spec `docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md` chapter 11 §2.

Adds a shared TypeScript config at the repo root and a per-package `tsconfig.json` for each of `packages/{daemon,electron,proto}/` that `extends: "../../tsconfig.base.json"`.

## Files (4 new)

- `tsconfig.base.json` — strict, `target=ES2022`, `module=NodeNext`, `moduleResolution=NodeNext`, `skipLibCheck`, `declaration`+`declarationMap`, `sourceMap`, `isolatedModules`, plus the usual `esModuleInterop`/`forceConsistentCasingInFileNames`/`resolveJsonModule`.
- `packages/daemon/tsconfig.json` — extends base, adds `outDir=dist`, `rootDir=src`, `include: ["src/**/*"]`.
- `packages/electron/tsconfig.json` — extends base, adds `outDir=dist`, `rootDir=src`, `jsx=react-jsx` (renderer code), `include: ["src/**/*"]`.
- `packages/proto/tsconfig.json` — extends base, `noEmit=true` (output comes from buf-codegen in T0.5, not tsc).

## Layer 1 / scope notes

- **Did NOT touch root v0.2 `tsconfig.json`.** It scopes the existing electron app under `src/` and uses bundler-mode resolution + `noEmit`. The new `tsconfig.base.json` is a SEPARATE file consumed only by `packages/*` via `extends`, exactly as described in the spec's repo tree (lines 2710 / 2726 / 2741). The two files do not conflict because they target disjoint source trees.
- No new deps. `typescript@^5.7.2` is already in root `package.json`.
- No `@types/*` added — packages have no source yet; types come per-package later.

## Test plan

- [x] `git diff --stat` against `working` shows exactly the 4 new files, nothing else.
- [x] Each package's tsconfig parses cleanly via `node_modules/.bin/tsc --noEmit -p .` — only emits `TS18003 No inputs were found` (expected: no source files yet), which is the no-source-files error, not a config error. Configs are parseable.
- [ ] CI green on `working` base.

## References

- Task #13
- Spec: `docs/superpowers/specs/2026-05-03-v03-daemon-split-design.md` chapter 11 §2 (monorepo layout)

PR #1296, opened 2026-05-22
fix(terminal): park viewport at bottom after attach snapshot drain

Jiahui-Gu:
## Bug
Attaching to a session leaves the xterm viewport at the top or middle of the transcript instead of the live tail (user report: "现在 attach 一个 session，右侧滚轮会到中间或者顶部").

## Root cause
xterm's `Terminal.write` is queued via the internal `WriteBuffer` (async drain). The attach path wrote the snapshot + buffered chunks but never called `scrollToBottom` afterwards. `term.reset()` had zeroed `viewportY`; subsequent writes advanced `baseY` past `viewportY` but xterm only auto-parks the viewport at the new bottom when it was already following. A synchronous `scrollToBottom()` right after `write(snapshot)` would have raced ahead of the drain — same root-cause class as #852.

## Fix
- Centralise the idiom in `writeAndScrollToBottom(term)` in `xtermSingleton`: `term.write('', () => term.scrollToBottom())`. The empty write queues a callback that fires AFTER this chunk drains, so the scroll lands on the post-snapshot `baseY`.
- Call the helper at 3 attach call sites in `usePtyAttach`:
  1. after the initial snapshot write + buffered drain,
  2. inside the snapshot-replay handler's drain tail,
  3. in the post-attach fit branch's replay continuation.
- In `useTerminalResize`, gate the post-replay scroll on the user's prior `atBottom`: snapshot `baseY`/`viewportY` BEFORE the backend resize, and after the replay either let the replay-side bottom-park stand (atBottom was true) or restore the saved `viewportY` via `scrollToLine` (user was reading scrollback — don't yank them down).

## Test coverage
- `usePtyAttach`: asserts `scrollToBottom` fires AFTER the snapshot `write` callback drains (ordering via call log) for both initial-attach and snapshot-replay paths.
- `useTerminalResize`: asserts atBottom-gating — `scrollToLine` is NOT called when user was at bottom pre-resize; IS called with the saved `viewportY` when user was scrolled up.

All 52 tests in `tests/terminal/` pass. `npm run typecheck` and `npm run lint` are clean.

## Note
Visual dogfood will be done by a separate worker post-merge.

### DIFF
--- a/src/terminal/usePtyAttach.ts
+++ b/src/terminal/usePtyAttach.ts
@@ -12,6 +12,7 @@ import {
   setInputDisposable,
   setSnapshotReplay,
   getSnapshotReplay,
+  writeAndScrollToBottom,
 } from './xtermSingleton';
 
 export type PtyAttachState =
@@ -289,6 +290,12 @@ export function usePtyAttach(sessionId: string, cwd: string): UsePtyAttachResult
           for (const b of buffered) {
             if (b.seq > snapSeq) tDrain.write(b.chunk);
           }
+          // Park the viewport at the bottom AFTER the WriteBuffer drains
+          // the snapshot + drained chunks (see writeAndScrollToBottom).
+          // Without this, attach lands the scrollbar at the top of the
+          // transcript because xterm.write is queued and a synchronous
+          // scrollToBottom() would run before baseY catches up.
+          writeAndScrollToBottom(tDrain);
         }
         buffered.length = 0;
 
@@ -346,6 +353,12 @@ export function usePtyAttach(sessionId: string, cwd: string): UsePtyAttachResult
             for (const b of buffered) {
               if (b.seq > snapSeq) tDrain2.write(b.chunk);
             }
+            // Same rationale as the initial-attach drain above: this
+            // replay path is fired by the post-attach fit branch below
+            // (and by `useTerminalResize` after a SIGWINCH). For the
+            // attach side we want unconditional scroll-to-bottom; the
+            // resize side gates separately on the user's prior atBottom.
+            writeAndScrollToBottom(tDrain2);
           }
           buffered.length = 0;
         });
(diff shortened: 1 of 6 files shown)


Jiahui-Gu (commented):
Reviewed both layers.

**Layer 1 (direction/necessity)** — Approved.

- `term.write('', cb)` is the documented xterm idiom for WriteBuffer drain rendezvous. Alternatives like `requestAnimationFrame` would race the buffer drain in exactly the way the PR description calls out; the empty-write callback is the right primitive.
- Traced the resize-path race claim. Ordering during a real resize:
  1. `replay()` queues `write(snapshot)` → `write(bufferedChunks...)` → `write('', cbA→scrollToBottom)` and synchronously returns the promise.
  2. The outer `.then()` in `useTerminalResize` runs and queues `write('', cbB→scrollToLine(savedViewportY))`.
  3. WriteBuffer drains FIFO: snapshot → buffered → cbA fires (scrollToBottom) → cbB fires (scrollToLine).
  So when `wasAtBottom===false`, the restore reliably wins because it's queued last. No race. The "unconditional scrollToBottom inside replay tail + conditional restore outside" decomposition is sound.
- `useAtBottom` listens on `term.onScroll`, which xterm fires for programmatic scrolls too, so the floating jump-to-bottom button's state stays consistent after the forced scrollToBottom. No regression risk.

**Layer 2 (implementation)** — Clean.

- Helper signature `writeAndScrollToBottom(t: Terminal)` is minimal and the JSDoc documents the WriteBuffer rationale + every call site that uses it. Placement next to `getTerm` in `xtermSingleton` is appropriate.
- All 3 attach sites use the helper per the locked design. The post-attach fit branch's "defensive rendezvous" is a redundant-but-cheap belt-and-braces — acceptable since `write('')` is O(1).
- Tests reverse-verify correctly: stashing `writeAndScrollToBottom(tDrain)` at site 1 would make `callLog.indexOf('scrollToBottom', rendezvousIdx)` return -1 and the ordering assertion would fail. The resize test's scrolled-up case would fail `scrollToLineSpy.toHaveBeenCalledWith(120)` if `wasAtBottom` were hardcoded true. Ran `npx vitest run tests/terminal/usePtyAttach.test.tsx tests/terminal/useTerminalResize.test.tsx` → 18/18 pass.
- No dead code, no backwards-compat hacks, no unused exports.

**Verdict: APPROVE.** Ship it (post-merge dogfood still needs to happen as the PR body notes).

PR #1301, opened 2026-05-22
fix(import): refresh session list on every dialog open

Jiahui-Gu:
## Symptom
User report: "import的时候，好像列表会保持在刚打开的时候加载的session list，不会刷新" — ImportDialog shows stale data on second open; sessions recorded between the first and second open never appear until the app restarts.

## Root cause
`getImportableSessions()` in `electron/ipc/utilityIpc.ts` used stale-while-revalidate: it returned the cached array immediately and kicked an async refresh whose result the current caller never awaited.

## Fix
Always `await refreshImportableCache()`. The existing `importablePending` single-flight mutex is preserved so concurrent IPCs share one in-flight scan rather than double-scanning. Cold-open latency stays mitigated by `primeImportableCache()` at app `ready`.

## Test
`electron/ipc/__tests__/utilityIpc.test.ts` — new describe block "import:scan IPC handler — fresh scan, no stale cache (regression)":
- Primes handler with one session, simulates a second appearing on disk between calls, asserts the second call returns both. Reverse-verified: this test fails on the previous SWR implementation.
- Single-flight assertion: three concurrent handler calls share one underlying scan.

All 25 tests in the file pass; `npm run typecheck` clean; `npm run lint` shows 0 errors (pre-existing warnings only).

### DIFF
--- a/electron/ipc/__tests__/utilityIpc.test.ts
+++ b/electron/ipc/__tests__/utilityIpc.test.ts
@@ -4,6 +4,9 @@ import * as os from 'os';
 import * as path from 'path';
 
 const openExternalMock = vi.fn(async (_url: string) => undefined);
+const scanImportableSessionsMock = vi.hoisted(() =>
+  vi.fn(async () => [] as unknown[]),
+);
 
 vi.mock('electron', () => ({
   BrowserWindow: { fromWebContents: () => null },
@@ -11,7 +14,7 @@ vi.mock('electron', () => ({
   shell: { openExternal: (url: string) => openExternalMock(url) },
 }));
 vi.mock('../../import-scanner', () => ({
-  scanImportableSessions: async () => [],
+  scanImportableSessions: scanImportableSessionsMock,
 }));
 vi.mock('../../prefs/userCwds', () => ({
   getUserCwds: () => [os.homedir()],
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
Independent review — verdict: **APPROVE** (with one non-blocking nit).

**Layer 1 (direction):** Fresh-scan-with-mutex is the right call. ImportDialog kicks `scanImportable()` on every `open=true` transition (src/components/ImportDialog.tsx:48-64), so SWR was guaranteed to leak a stale snapshot every time. Cold-open latency is properly neutralized: `primeImportableCache()` fires at app `ready` (electron/main.ts:318), and even when the user opens the dialog before priming finishes, the `importablePending` mutex makes their await piggyback on the in-flight priming promise — worst case one scan's latency, never worse than the original. The only other `scanImportable()` callers are the renderer dialog itself and `scripts/harness-real-cli.mjs` (test harness); neither depends on instant return. The push-event alternative (`importable:updated` from main) would require `fs.watch` on `~/.claude/projects/**` with debouncing plus renderer subscription lifecycle — strictly more surface area for a dialog that opens infrequently. Rejecting it is correct.

**Layer 2 (implementation):** SWR branch is fully deleted (not commented out) — `getImportableSessions` is now a one-liner that returns `refreshImportableCache()`. Mutex semantics in `refreshImportableCache` are untouched and still correct (early-return on `importablePending`, cleared in `.finally`). The comment block at lines 30-39 + 61-66 now accurately describes fresh-scan-with-mutex including the priming rationale.

Tests are solid: both exercise the real handler registered via `registerUtilityIpc` against fresh module state (`vi.resetModules()` + dynamic import in `beforeEach`), not a mocked-out handler. The regression test would genuinely fail on the old SWR path because `r2` would equal `['a']`. The single-flight test correctly blocks the scan with a manual promise and asserts exactly one `scanImportableSessions` call across three concurrent IPC invocations — that's production behavior, not mock behavior.

**Nit (non-blocking, follow-up):** `scripts/harness-real-cli.mjs` lines 1603-1607 still carry the stale rationale comment ("`getImportableSessions` returns the cache immediately while triggering an async re-scan"). The polling loop happens to remain harmless post-fix, but the comment now misleads the next reader. Worth a one-line update; not a merge blocker.

PR #1302, opened 2026-05-22
fix(groups): rename now overrides i18n default group name

Jiahui-Gu:
## Symptom
Renaming the default "会话/Sessions" group has no visible effect — the sidebar keeps showing the localized default.

## Root cause
The bootstrap group seeds both `name` and `nameKey: 'sidebar.defaultGroupName'`, and `GroupRow` does `group.nameKey ? t(group.nameKey) : group.name` — the i18n key always wins, so `renameGroup` (which only updated `name`) was invisible.

## Fix
`renameGroup` now also sets `nameKey: undefined`, treating a user rename as explicit intent that overrides the localized default.

## Test
Added `tests/stores/slices/groupsSlice.test.ts` case asserting `nameKey` is cleared and `name` is updated after `renameGroup`. Full file passes (13/13). `npm run typecheck` clean.

### DIFF
--- a/src/stores/slices/groupsSlice.ts
+++ b/src/stores/slices/groupsSlice.ts
@@ -81,8 +81,15 @@ export function createGroupsSlice(set: SetFn, get: GetFn): GroupsSlice {
     },
 
     renameGroup: (id, name) => {
+      // Clear `nameKey` so the user-supplied name wins over the i18n default.
+      // Default groups seed both `name` and `nameKey` (so language switches
+      // re-localize), but once the user renames, that's explicit intent —
+      // the renderer (`GroupRow`) prefers `nameKey` when present, so leaving
+      // it set would make the rename invisible.
       set((s) => ({
-        groups: s.groups.map((g) => (g.id === id ? { ...g, name } : g)),
+        groups: s.groups.map((g) =>
+          g.id === id ? { ...g, name, nameKey: undefined } : g
+        ),
       }));
     },
 
--- a/tests/stores/slices/groupsSlice.test.ts
+++ b/tests/stores/slices/groupsSlice.test.ts
@@ -71,6 +71,20 @@ describe('groupsSlice', () => {
     expect(h.state().groups.find((g) => g.id === 'g-default')!.name).toBe('Renamed');
   });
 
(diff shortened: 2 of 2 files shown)


Jiahui-Gu (commented):
Verdict: APPROVE (NIT only — no blockers).

**Layer 1 (direction):** Correct. `Group.nameKey` (src/types.ts:65-72) is documented as "render `t(nameKey)` instead of `name`" so the default label re-translates on language switch. A user-initiated rename is explicit intent and must escape that auto-localization — clearing `nameKey` is the right semantic ("user-named groups stay user-named across language switches"). Alternative (gating the renderer on a separate "user-renamed" flag) would invent state we don't need; the absence of `nameKey` IS that flag. Call-site sweep: `GroupRow` inline rename (the bug path) — fixed. `ImportDialog.tsx:120` calls `renameGroup` right after `createGroup`, and `createGroup` never sets `nameKey`, so clearing undefined is a harmless no-op. No other `setGroupName`/`updateGroup` paths exist. The archive-container synthesis at `sessionCrudSlice.ts:471` deliberately propagates the source's `nameKey` — correctly outside this PR's scope.

**Layer 2 (implementation):** Single-line fix on exactly the right line. Comment explains the renderer coupling so future readers don't "simplify" it away. Test covers both directions (nameKey cleared + name updated) AND asserts the bootstrap precondition (`nameKey === 'sidebar.defaultGroupName'`), so it fails loudly if the seed shape ever changes. Body claims typecheck clean + 13/13 in the slice's test file.

**Data migration — NIT, not blocking:** Existing victims have `{ name: 'Custom', nameKey: 'sidebar.defaultGroupName' }` already persisted, and this PR only clears `nameKey` on the NEXT rename. So they still see "会话/Sessions" after upgrading until they re-rename. Pre-MVP, tiny user count, workaround = "rename it again" — acceptable. If we want self-healing, a ~5-line one-shot in zustand persist `migrate` (drop `nameKey` when `kind==='normal' && nameKey===defaultKey && name !== resolved(defaultKey)`) would do it. Not required to land.

Jiahui-Gu:

PR #1303, opened 2026-05-22
fix(paste): bracketed-paste wrapping + CRLF normalization

Jiahui-Gu:
## User-visible symptoms

- Multi-line code pasted into claude prematurely submits at the first newline; subsequent lines land in fresh prompts.
- Multi-line paste from Windows notepad/browsers double-Enters (each CRLF acts as two submits).

## Root cause

`pasteIntoActivePty` wrote clipboard text straight to the PTY without (a) wrapping in bracketed-paste markers when claude's Ink TUI has enabled DECSET 2004, or (b) normalizing the Windows CRLF that `clipboard.readText()` returns.

## Fix

1. Read `term.modes.bracketedPasteMode` (xterm.js `IModes`, kept live by the parser as claude sends `\x1b[?2004h` / `\x1b[?2004l`). When active, wrap payload in `\x1b[200~ ... \x1b[201~` so the app treats it as atomic paste rather than typed input.
2. Normalize CRLF -> LF (and lone CR -> LF for ancient-Mac paste) BEFORE bracketed wrapping.
3. Image-path branch goes through the same prep so claude sees the path as one atomic paste.

API approach: **Option A** — `term.modes.bracketedPasteMode` is a documented public boolean in `@xterm/xterm` typings (`xterm.d.ts` line 1877). No parser shim or fallback needed.

## Test coverage added

In `tests/terminal/useXtermSingleton.test.tsx`:
- CRLF -> LF normalization (Windows clipboard)
- Lone CR -> LF (old-Mac line endings)
- Bracketed mode active + multi-line -> wrapped in `\x1b[200~...\x1b[201~`
- Bracketed mode inactive -> bare text (guards against always-wrapping into cmd.exe-style shells that never enable DECSET 2004)
- CRLF normalization composes with bracketed wrap (normalize first, then wrap)
- Image-path branch is wrapped when paste mode active

All 58 terminal tests pass; typecheck + lint clean (no new warnings).

## Not in this PR

The 10MB PNG perf bug and the attach-state guard tracked separately per request.

### DIFF
--- a/src/terminal/xtermSingleton.ts
+++ b/src/terminal/xtermSingleton.ts
@@ -151,6 +151,29 @@ export function setSnapshotReplay(fn: (() => Promise<void>) | null): void {
  * (currently only tests) can await; production paste paths fire-and-forget
  * via `void`.
  */
+/**
+ * Prepare a clipboard payload for injection into the PTY:
+ *   1. Normalize CRLF → LF (and lone CR → LF). Windows clipboards (notepad,
+ *      most browsers) hand back CRLF; PTYs / claude treat each `\r` as a
+ *      submit. Without this, every multi-line Windows paste fires the prompt
+ *      after the first line and the rest lands in fresh prompts.
+ *   2. If xterm reports bracketed-paste mode active (claude's Ink TUI sends
+ *      `\x1b[?2004h` on startup), wrap in `\x1b[200~ ... \x1b[201~` so the
+ *      app treats the whole payload as paste, not typed input. Without this:
+ *      embedded `\n` submits prematurely, embedded `\x03` SIGINTs claude,
+ *      and embedded ANSI escapes are interpreted as terminal commands.
+ *
+ * Read `term.modes.bracketedPasteMode` (xterm.js IModes — updated live by
+ * the parser as the app sends DECSET 2004 h/l). Falls back to "no wrap" if
+ * the singleton hasn't been constructed yet (e.g. paste fired before the
+ * terminal mounted — shouldn't happen, but the early-return is cheap).
+ */
+function preparePastePayload(text: string): string {
+  const normalized = text.replace(/\r\n/g, '\n').replace(/\r/g, '\n');
+  const bracketed = term?.modes?.bracketedPasteMode === true;
+  return bracketed ? `\x1b[200~${normalized}\x1b[201~` : normalized;
+}
+
 export async function pasteIntoActivePty(fallbackText: string | undefined): Promise<void> {
   // N1 race fix (reviewer): snapshot `activeSid` BEFORE the async IPC hop.
   // `saveClipboardImage` round-trips to main; the user can switch sessions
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE (independent reviewer).

## Layer 1 (direction)

- `term.modes.bracketedPasteMode` verified: `xterm.d.ts:1877` declares it as `readonly boolean` on the public `IModes` interface, exposed via `Terminal.modes: IModes` (line 831), updated live by xterm's parser as the app emits DECSET 2004 h/l. No shim needed.
- All renderer paste paths funnel through `pasteIntoActivePty`:
  - `onPasteCapture` (capture-phase DOM `paste`) -> `pasteIntoActivePty(text)` (PR-head line 369)
  - `pasteFromClipboard` (Ctrl/Cmd+V keydown, plain + Shift) -> `pasteIntoActivePty(text)` (line 408)
  - `terminalPaste` (right-click via TerminalPane) -> `pasteIntoActivePty(text)` (line 559)
  So patching the helper covers every paste source. `role: 'paste'` in `electron/window/createWindow.ts:98` only fires against contentEditable/textarea, not the xterm canvas, so it doesn't bypass.
- Null-term fallback (`term?.modes?.bracketedPasteMode === true` -> no wrap) is safe: `activeSid` is null before mount and `pasteIntoActivePty` returns early at `if (!sid) return`. The bracketed check is belt-and-braces.
- Image-path wrapping: claude treats the wrapped string as one atomic paste; the path lands intact. Worker's "collides with TUI keybindings while streaming in" reasoning is sound.
- Live getter vs snapshot: `modes` reflects parser state, paste is human-speed, DECSET toggles are rare. Race acceptable.
- Lone-CR order: `.replace(/\r\n/g,'\n').replace(/\r/g,'\n')` -- `\r\n` collapses first then lone `\r`. Correct.

## Layer 2 (implementation)

- Minimal diff: +29/-2 in `xtermSingleton.ts`, +77/-2 in the test file. Single helper, two call sites, no API change.
- Tests exercise production `pasteIntoActivePty` (not the helper in isolation), so a future regression that drops prep from one branch would be caught.
- Reverse-verify: removing `preparePastePayload` from the text branch breaks 'normalizes CRLF' (would inject raw `\r\n`) and 'wraps when mode active' (missing `\x1b[200~`). Removing it from the image branch breaks 'image-path branch is bracketed-wrapped'. 1:1 coverage.
- `terminalModes` mock matches the singleton's access pattern; the readonly getter in prod is interchangeable with the plain-property mock for read access.

## Nit (non-blocking)

`xtermSingleton.ts:133-170` now has two back-to-back `/**` JSDoc blocks: the original 21-line block for `pasteIntoActivePty` is orphaned above the new `preparePastePayload` doc, which sits directly above the helper. Tooling still attaches each doc correctly, but a reader scanning for the `pasteIntoActivePty` contract reads the helper's doc first. Consider folding the helper doc into a brief inline comment, or moving the helper + its doc below `pasteIntoActivePty`. Not a merge blocker.

PR #1304, opened 2026-05-22
fix(terminal): rebind paste capture on host change

Jiahui-Gu:
## Symptom

In the close-all -> create-new flow, Electron context-menu paste and Shift+Insert in the freshly-remounted `TerminalPane` regress to xterm's built-in paste pipeline, undoing the bracketed-paste/CRLF fix from #1303.

## Root cause

The capture-phase `paste` listener was attached inside the first-construction branch of `ensureTerminal` (`src/terminal/xtermSingleton.ts`). When `TerminalPane` remounts into a new host div, the retarget branch (`_core._parent !== host` -> `term.open(host)`) re-`open`s xterm against the new host but never moved the listener -- it stayed on the orphaned old div, so live paste events bypassed `pasteIntoActivePty` and fell through to xterm's built-in pipeline.

## Fix

- Hoist `onPasteCapture` to module scope. It closes over no construction-local state (only the module-level `keyboardPasteHandled` flag + `pasteIntoActivePty`), so the lift is mechanical.
- Add module-level `pasteHost` / `pasteListener` bookkeeping plus an `attachPasteCapture(host)` helper that detaches from the previously-bound host before attaching to the new one.
- Call `attachPasteCapture(host)` after every `term.open(host)` -- both first-construction and the retarget branch.
- `__resetSingletonForTests` clears the new state and detaches the listener for clean test isolation.

## Test

- `tests/terminal/useXtermSingleton.test.tsx` adds a `paste listener rebinds on host change` regression guard: spies on `addEventListener` / `removeEventListener` on two distinct hosts, mutates the singleton's `_core._parent` so the retarget branch fires on the second `ensureTerminal`, then asserts (a) `removeEventListener('paste', ...)` was called on the old host, (b) `addEventListener('paste', ...)` was called on the new host, and (c) the same function identity is used across the two calls (no stale closure churn).
- Full suite green: `npx vitest run tests/terminal/` -> 53/53 pass.
- `npm run typecheck` clean.
- `npm run lint` -- 0 errors (warnings unchanged from main).

## Test plan

- [ ] Open ccsm, create a session, close all sessions, create a new one.
- [ ] Right-click in the terminal pane -> Paste a multi-line string. Verify it arrives with bracketed-paste/CRLF normalization (no raw `\r` swallow, no double-paste).
- [ ] Shift+Insert in the same pane after a remount. Verify the same.
- [ ] Existing Ctrl/Cmd+V keydown path and image-first paste still inject exactly once.

### DIFF
--- a/src/terminal/xtermSingleton.ts
+++ b/src/terminal/xtermSingleton.ts
@@ -49,6 +49,20 @@ let keyboardPasteHandled = false;
 // the per-attach `snapSeq` to the snapshot's seq so subsequent live
 // chunks dedupe correctly.
 let snapshotReplay: (() => Promise<void>) | null = null;
+// Capture-phase `paste` listener bookkeeping. The listener is bound to the
+// host DOM node passed into `ensureTerminal`; React can remount TerminalPane
+// (e.g. close-all → create-new flow), which produces a brand-new host div.
+// When the retarget branch below re-`open`s xterm against the new host, the
+// listener attached to the old (now-orphaned) host stops receiving paste
+// events from the live DOM tree. The Electron context-menu paste / Shift+
+// Insert paths then fall through to xterm's built-in paste pipeline,
+// regressing the bracketed-paste/CRLF fix from #1303.
+//
+// Track the currently-bound host + listener at module scope so we can
+// `removeEventListener` from the old host and `addEventListener` on the
+// new host every time `ensureTerminal` rebinds. See `attachPasteCapture`.
+let pasteHost: HTMLElement | null = null;
+let pasteListener: ((e: ClipboardEvent) => void) | null = null;
 
 export function getTerm(): Terminal | null {
   return term;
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
**APPROVE**

**L1 (necessity + approach):** Correct diagnosis and minimal fix. Hoisting `onPasteCapture` to module scope is safe — it closes over only `keyboardPasteHandled` (module `let`) and `pasteIntoActivePty` (exported module fn), both already at module scope. No construction-local state was captured in the old inline closure, so identity-stable lifting changes nothing behaviorally beyond enabling clean remove+add. Helper-based rebind is the right shape (vs. e.g. delegating to document, which would conflict with xterm's own paste plumbing).

**L2 (implementation):** Verified diff:
- `attachPasteCapture(host)` called in BOTH the retarget branch (after `term.open(newHost)`) and the first-construction site (replacing the old inline `addEventListener`).
- `__resetSingletonForTests` detaches + nulls `pasteHost` / `pasteListener` — good for cross-test hygiene.
- New test asserts old-host remove + new-host add + **same function identity** across both calls (`pasteRemoveOnA[0][1] === pasteAddOnB[0][1]`), which is precisely what the bug was about (orphaned listener identity on dead host). Reverse-verify: stash `attachPasteCapture(host)` inside the retarget branch → `pasteAddOnB` would be 0 and the identity assertion would fail → test catches the regression. Good probe.
- Test also flips `_core._parent` to force the retarget code path, not the no-op same-host branch. Solid.

**NIT (non-blocking):** When `ensureTerminal(sameHost)` is called and the retarget branch runs (e.g. `_parent` got nulled by something external but `host` ref is identical to `pasteHost`), `attachPasteCapture` still does remove+add on the same node. Functionally fine (idempotent), but you could early-return when `host === pasteHost && pasteListener` to avoid the no-op churn. Not worth blocking — current behavior is correct.

Ship it.

Jiahui-Gu:

PR #1305, opened 2026-05-22
fix(sessions): unarchive uses ensureUsableGroup fallback (no orphan)

Jiahui-Gu:
## Symptom

P1: if a user has deleted every normal group (including `g-default`) before unarchiving a session from an archive container, the session is persisted but **invisible**. Sidebar filters rows by `groups.find(g => g.id === session.groupId)` and the literal `'g-default'` fallback didn't exist in `state.groups`, so the row never rendered. No UI recovery path — power user only.

## Cause

`sessionCrudSlice.ts` `unarchiveSession`:

```ts
const targetGroupId = origin ? origin.id : 'g-default';
```

Assumes the seed `g-default` always exists. It usually does — but the user can delete it (plus every other normal group), leaving only archive containers.

## Fix

Route through the same `ensureUsableGroup(groups, preferredId)` helper that `createSession` / `importSession` already use:
- prefers the original source group (`container.sourceGroupId`)
- falls back to any other `kind:'normal'` group
- synthesizes a fresh normal group with the default name when none exist

The (possibly synthesized) groups array is folded into the same `set()` that clears `archivedAt` and (conditionally) deletes the empty archive container, so groups + sessions stay consistent in one tick.

## Test

`tests/stores/slices/archiveSession.test.ts` — new scenario:
- seed only `g-default`, archive `s1`
- delete `g-default` (s1 lives in the container, so it survives)
- unarchive `s1`
- assert a new normal group exists, `s1.groupId` points at it (not the dead literal `'g-default'`), `archivedAt` cleared, empty container cleaned up, and the Sidebar visibility predicate holds (`groups.find(g => g.id === s1.groupId)` defined)

## Gates

- `npx vitest run tests/stores/` — 107/107 pass (1 new test)
- `npm run typecheck` — clean
- `npm run lint` — 0 errors

### DIFF
--- a/src/stores/slices/sessionCrudSlice.ts
+++ b/src/stores/slices/sessionCrudSlice.ts
@@ -512,7 +512,13 @@ export function createSessionCrudSlice(set: SetFn, get: GetFn): SessionCrudSlice
 
     unarchiveSession: (sessionId) => {
       // Clear `archivedAt`, move back to original source group if it
-      // still exists and is normal — otherwise fall back to g-default.
+      // still exists and is normal — otherwise route through
+      // `ensureUsableGroup` so the session lands in a real `kind:'normal'`
+      // group (the same fallback createSession/importSession use). The
+      // earlier literal `'g-default'` fallback could orphan the session
+      // when the user had deleted every normal group: the row was
+      // persisted but invisible (Sidebar filters by group existence) with
+      // no UI path to recover.
       // If the archive container empties after the move, delete it.
       // If the user has no active session (e.g. they just archived the
       // active one and then immediately unarchived it), restore activeId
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
**APPROVE**

**L1 (semantics):** Correct. Walking through the three regimes against `ensureUsableGroup`:
1. Source group still alive + normal → `preferredId && isUsable` short-circuits, returns same `groups` ref, `groupId = container.sourceGroupId`. Matches old behavior exactly.
2. Source dead, other normal group(s) exist → falls to `firstUsableGroupId`. New behavior, but unambiguously better than the old `'g-default'` literal (which would also have been broken in this regime if g-default specifically was deleted but other normals lived).
3. ALL normals gone → synthesizes a fresh `kind:'normal'` group with `defaultGroupName()` + prepends to `groups`. This is the actual P1 fix and exactly matches the contract `createSession`/`importSession` already use, so no behavioral inconsistency across entry points.

**L2 (implementation):** Verified diff:
- `ensured.groups` is correctly threaded into `set()` as `baseGroups`, and the empty-container strip operates on `baseGroups` (not `s.groups`) — so a synthesized fallback group survives the strip. This is the easy-to-miss bug and you got it right.
- `ensureUsableGroup` is **immutable**: returns same `groups` reference when an existing group is usable (regimes 1 + 2), or `[synth, ...groups]` (new array) when synthesizing. set() will see a new reference whenever it matters → re-render triggers correctly. No mutation concern.
- Comment update on the function is accurate and points future readers at the right rationale.

**Test:** Comprehensive — exercises the exact orphan path (archive s1 → delete every normal group → unarchive). Assertions cover all four invariants: synth group exists + is normal, `s1.groupId` points at it (not the dead `'g-default'` literal), `archivedAt` cleared, archive container cleaned up. The Sidebar-visibility predicate (`s.groups.find(g => g.id === s1.groupId)`) is asserted explicitly, which is the user-facing failure mode. Reverse-verify: revert the slice change → assertion `expect(s.groups.find(g => g.id === s1.groupId)).toBeDefined()` fails because `s1.groupId === 'g-default'` but no such group exists. Test catches the regression.

Ship it.

Jiahui-Gu:

PR #724, opened 2026-05-01
test(installer): T77 uninstall hygiene probe

Jiahui-Gu:
Closes T77 / #1035.

## Scope

Probe at `installer/uninstall-helper/__tests__/uninstall-hygiene.test.ts` asserting the v0.3 NSIS uninstall hygiene contract (frag-11 §11.6).

### §11.6.1 NSIS macro contract (`build/installer.nsh`)
- Helper invoke: `ccsm-uninstall-helper.exe --shutdown --timeout 2000`
- Hardstop safety net: `taskkill /IM ccsm-daemon.exe /F /T`
- Lockfile cleanup: `Delete "$LOCALAPPDATA\ccsm\daemon.lock"` (closes proper-lockfile-stale-on-SIGKILL race per §11.6.5)
- Negative: macro does NOT touch `data/`, `logs/`, `crashes/`, `daemon.secret` (retained per `deleteAppDataOnUninstall: false`; opt-in cleanup is the in-app frag-6-7 §6.8 surface)
- Path discipline: `%LOCALAPPDATA%`-only — no `$PROGRAMFILES`, no `$PROGRAMDATA` (round-3 P0-1)

### §11.6.4 helper hardstop contract (`installer/uninstall-helper/index.js`)
- Kills every enumerated `ccsm-daemon.exe` pid via `taskkill /F /T /PID <pid>`
- Idempotent: zero pids found returns `{ ok: true, killed: 0 }` with no taskkill invocations (exit 0, daemon-already-dead path)
- Hard error: when `taskkill` fails, returns `{ ok: false, failed: >=1 }` (the only exit-1 trigger per spec)

## Scope clarification vs the brief

The brief mentioned wiping `%LOCALAPPDATA%\ccsm\` and `%PROGRAMDATA%\ccsm\`. Per spec §11.6:
- The helper does NOT wipe disk paths — that's the NSIS standard-uninstall sequence's job (after the macro returns).
- `%PROGRAMDATA%\ccsm\` is NOT in the §11.6 paths table — v0.3 is per-user only (`perMachine: false`).

The probe asserts the actual contracts above instead of the brief's drift, with comments explaining the divergence. Surfacing this for reviewer/manager judgment per the ambiguous-requirements protocol.

## Reverse-verify

- Removing `taskkill /IM ccsm-daemon.exe /F /T` from `installer.nsh` → "runs the §11.6.1 taskkill safety net" FAILS.
[...]

PR #1306, opened 2026-05-22
fix(shortcuts): wire Ctrl+Shift+N to create new group (was advertised but unbound)

Jiahui-Gu:
## Symptom
ShortcutOverlay (\`src/components/ShortcutOverlay.tsx:39\`) and CommandPalette (\`src/components/CommandPalette.tsx:120\`) both advertise **Ctrl/Cmd+Shift+N → New group**. Pressing the combo did nothing — the in-app hint pointed at a no-op.

## Cause
When global shortcuts were extracted into \`useShortcutHandlers\` (Task #724), only Ctrl+/, "?", Ctrl+F, Ctrl+, were carried over. The new-group chord lived only in the help surfaces, never in code.

## Fix
- \`src/app-effects/useShortcutHandlers.ts\`: add a \`createNewGroup\` dep and branch on \`mod && shift && key === 'n'\`. Honor the editable-target guard so the chord can't fire while typing capital-N in the InputBar.
- \`src/App.tsx\`: wire the dep to \`createGroup() + focusGroup(id)\`, mirroring \`Sidebar.handleNewGroup\`'s focus follow-up so a subsequent "New session" lands inside the freshly created group.
- Auto-rename (Sidebar-local state) is intentionally **not** replicated — the CommandPalette "New group" command (the prior global entry-point for this action) doesn't trigger it either; keeping the chord and the palette consistent is the right tradeoff vs. lifting Sidebar-local state into the store.

## Test
\`tests/app-effects/useShortcutHandlers.test.tsx\`:
- [x] Ctrl+Shift+N triggers \`createNewGroup\`
- [x] Cmd+Shift+N triggers \`createNewGroup\` (macOS modifier)
- [x] Ctrl+N (no Shift) does NOT trigger (avoids hijacking common rebind)
- [x] Suppressed when an \`<input>\` / \`<textarea>\` has focus (contenteditable case noted but skipped — jsdom doesn't propagate \`contentEditable\` to \`isContentEditable\` reliably; the production check still covers it)
- [x] Existing Ctrl+B / Ctrl+K negative assertions extended to confirm \`createNewGroup\` is not collateral-fired

Verified locally:
\`\`\`
npx vitest run tests/app-effects/   # 62 pass
npm run typecheck                   # clean
npm run lint                        # 0 errors (pre-existing warnings only)
\`\`\`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -93,6 +93,7 @@ export default function App() {
   // toggling a single session's waiting<->idle). See PR #1269.
   const selectSession = useStore((s) => s.selectSession);
   const focusGroup = useStore((s) => s.focusGroup);
+  const createGroup = useStore((s) => s.createGroup);
   const applyExternalTitle = useStore((s) => s._applyExternalTitle);
   const applyCwdRedirect = useStore((s) => s._applyCwdRedirect);
   const applyPtyExit = useStore((s) => s._applyPtyExit);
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
APPROVE with NIT.

**L1 (direction)** — fix wires the advertised chord; mirrors Sidebar.handleNewGroup's create+focus, suppression-on-editable matches the '?' branch precedent. The decision to skip the auto-rename indicator is consistent with the CommandPalette 'New group' sibling, which keeps the two global entry-points uniform. NIT: the '+' button in Sidebar DOES trigger inline rename via 'justCreatedGroupId', so the keyboard path now has subtly worse UX than the click path (group appears without an editable name field, user has to F2 or click). Two acceptable resolutions: (a) accept the gap and treat this as deliberate parity with CommandPalette (current choice — fine), or (b) lift 'justCreatedGroupId' into the store so both global entry-points and the '+' button share it. Not a blocker — but worth a follow-up task if user testing flags the missing rename.

**L2 (implementation)** — editable-target guard reuses the existing 'isEditableTarget' helper and matches the '?' branch pattern exactly. Test coverage is solid: positive (Ctrl + Cmd variants), negative (no Shift), suppression (input + textarea). Contenteditable jsdom gap is correctly called out in a comment rather than asserted weakly. Hook dep array updated to include 'createNewGroup'. Clean.

Jiahui-Gu:

PR #1307, opened 2026-05-22
fix(pty): async taskkill — quit no longer freezes UI

Jiahui-Gu:
## Symptom
On Windows with N>1 live pty sessions, the `before-quit` handler froze the visible UI for ~N × 200-2000ms (Electron tearing down with the main event loop blocked) before the process actually exited. Worst-case, the window stayed "hung" long enough that users force-killed it and left orphan `claude.exe` children behind.

## Cause
`electron/ptyHost/processKiller.ts` used `spawnSync('taskkill', ['/PID', pid, '/T', '/F'])` for the Windows subtree walk. `killAllPtySessions` iterates every live session and synchronously dispatched one `spawnSync` per pid, so each `taskkill` blocked the main thread end-to-end before the next one started — strictly serial. Each `taskkill` costs 200ms–2000ms (cold ConHost / AV scan), so N sessions ≈ N × that cost of frozen UI on quit.

## Fix
- **`electron/ptyHost/processKiller.ts`** — Windows path now uses async `spawn` and returns `Promise<void>` that resolves on `exit`/`error` with a 5s timeout ceiling against wedged `taskkill`. POSIX path also returns `Promise<void>` (resolved immediately — SIGTERM was already non-blocking; the unref'd SIGKILL escalation timer keeps its original semantics).
- **`electron/ptyHost/lifecycle.ts`** — `kill()` awaits both the existing onExit/timeout race AND the subtree-walk promise via `Promise.all`. `killAll` returns `Promise<void>` that fans every per-pty kill out via `Promise.all` — N taskkills now run in parallel.
- **`electron/ptyHost/index.ts`** — `killAllPtySessions` return type bumped to `Promise<void>`.
- **`electron/lifecycle/appLifecycle.ts`** — `before-quit` handler now calls `event.preventDefault()`, awaits `killAllPtySessions()` + `disposeNotifyPipeline()`, then re-calls `app.quit()`. A `cleanupStarted` re-entry guard makes the second-pass `before-quit` a pass-through so we don't loop forever.

## Tests
- `electron/ptyHost/__tests__/processKiller.test.ts` rewritten with `vi.hoisted` + `vi.mock` on `child_process.spawn`. Covers canonical args, child-exit resolve, error-event resolve, sync-throw resolve, 5s timeout fallback (asserts last-ditch `child.kill()`), the parallel-fanout property (3 kills dispatched synchronously), and the POSIX SIGTERM/SIGKILL contract (now async with immediate resolve).
- `electron/lifecycle/__tests__/appLifecycle.test.ts` before-quit suite rewritten to pass a fake `Event` with a `preventDefault` spy, flush microtasks, then assert: (a) preventDefault fires on first pass, (b) `app.quit()` re-invoked after cleanup, (c) re-entry on the quit-driven before-quit is a pass-through, (d) sync throws and async rejections from `killAllPtySessions` are swallowed and quit still proceeds.
- `electron/ptyHost/__tests__/lifecycle.test.ts` mock for `killProcessSubtree` returns `Promise.resolve()` so the new `Promise.all([onExit, subtree])` in `kill()` settles.

## Test plan
- [x] `npx vitest run electron/ptyHost/ electron/lifecycle/` → 225/225 pass
- [x] `npm run lint` 0 errors (warnings unchanged from baseline)
- [ ] Manual: quit the app on Windows with 3+ live pty sessions, verify UI exits within ~1s rather than ~3-10s freeze
- [ ] Manual: verify no orphan `claude.exe` in Task Manager after quit
- [ ] Confirm the macOS/Linux PtyKill path still SIGTERMs the process group (POSIX behavior preserved)

Full suite has 34 pre-existing failures in `electron/__tests__/db*.test.ts` (sqlite native binding) and `tests/electron-load-smoke.test.ts` (needs `npm run build` first) — verified present on baseline `main` (`b401c7f`) with `git stash`, unrelated to this change.

### DIFF
--- a/electron/lifecycle/__tests__/appLifecycle.test.ts
+++ b/electron/lifecycle/__tests__/appLifecycle.test.ts
@@ -50,25 +50,42 @@ type EventName = 'before-quit' | 'window-all-closed' | 'activate';
 interface FakeApp {
   on: ReturnType<typeof vi.fn>;
   quit: ReturnType<typeof vi.fn>;
-  /** invoke a registered listener for a given event */
-  fire: (event: EventName) => void;
+  /** invoke a registered listener for a given event; for `before-quit` we
+   *  pass an Event-like object with a `preventDefault` spy so tests can
+   *  assert the async-cleanup gate fired (and inspect re-quit re-entry). */
+  fire: (event: EventName, arg?: { preventDefault: () => void }) => void;
 }
 
 function createFakeApp(): FakeApp {
-  const handlers = new Map<EventName, () => void>();
-  const on = vi.fn((event: EventName, cb: () => void) => {
+  const handlers = new Map<EventName, (arg?: unknown) => void>();
+  const on = vi.fn((event: EventName, cb: (arg?: unknown) => void) => {
     handlers.set(event, cb);
   });
   const quit = vi.fn();
   return {
     on,
     quit,
-    fire: (event) => {
+    fire: (event, arg) => {
       const cb = handlers.get(event);
       if (!cb) throw new Error(`no handler registered for ${event}`);
-      cb();
+      cb(arg);
     },
   };
+}
+
+/** Build a fake Electron Event with a preventDefault spy. */
+function fakeEvent(): { preventDefault: ReturnType<typeof vi.fn> } {
+  return { preventDefault: vi.fn() };
+}
+
+/** Drain pending microtasks so awaited async work inside the before-quit
+ *  handler settles before assertions. We use multiple `await` cycles because
+ *  the handler chains `await killAllPtySessions()` → optional dispose →
+ *  `app.quit()`; one microtask tick isn't enough. */
+async function flushAsync(): Promise<void> {
+  for (let i = 0; i < 5; i++) {
+    await Promise.resolve();
+  }
 }
 
 function buildDeps(overrides: Partial<LifecycleDeps> = {}): {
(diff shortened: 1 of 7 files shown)


Jiahui-Gu (commented):
APPROVE.

**L1 (direction)** — e.preventDefault + cleanupStarted re-entry guard + app.quit() re-trigger is the canonical Electron async-before-quit pattern. The 5s TASKKILL_TIMEOUT_MS cap is the right escape hatch: it bounds the worst-case quit-freeze regardless of how wedged taskkill gets, while still giving the OS a realistic window to reap the tree. cleanupStarted is intentionally never cleared — correct, we're committed to exiting once the first before-quit fires. Double-X-click / rapid quit retries land in the 'second-pass pass-through' branch and Electron exits cleanly. Parallel kill via Promise.all collapses N × 200-2000ms into max(latencies), which is the actual win.

**L2 (implementation)** — verified all four risk points:

1. **pendingKills dedup identity** (lifecycle.ts L562-566): slot replaced with 'combined' Promise and cleared via 'finally' with '=== combined' check — concurrent callers get identity-stable promise, no double-kill, no stale entries.
2. **Windows spawn timeout child cleanup** (processKiller.ts L666-674): timer fires child.kill() before resolving; timer cleared on normal exit; .unref() prevents event-loop pin. No zombie taskkill possible.
3. **Old sync export removal**: 'killAllPtySessions' now returns Promise<void>, no sync alias retained. Only caller outside tests is 'electron/main.ts' which passes it through to 'registerLifecycleHandlers' whose dep signature is 'Promise<void> | void' — compatible. No stale sync consumers.
4. **POSIX promise normalization**: kill() still returns immediately while SIGKILL escalation lives on an unref'd timer — preserves old non-blocking behavior, doesn't make Linux/mac wait 500ms on quit.

Test coverage is thorough: parallel-fanout assertion, wedged-taskkill timeout, spawn ENOENT/sync-throw paths, re-entry pass-through, rejection-still-quits path. 225 passing is consistent with the diff. Ship it.

Jiahui-Gu:

PR #1380, opened 2026-05-25
fix(sessions): unarchive uses ensureUsableGroup fallback (no orphan when g-default deleted)

Jiahui-Gu:
## Bug (P1)

`unarchiveSession` in `src/stores/slices/sessionCrudSlice.ts` wrote the literal `'g-default'` as the target `groupId` when the original source group no longer existed:

```ts
const targetGroupId = origin ? origin.id : 'g-default';
```

If the user had also deleted `g-default` (the default group is just a seeded normal group — it can be renamed or deleted like any other), the unarchived session pointed at a non-existent group id and **disappeared from the sidebar** (no row renders for a session whose groupId matches no group).

### Repro

1. Delete `g-default` and every other normal group (leave only an archive container with a session in it).
2. Archive a session.
3. Unarchive it.
4. Session has `groupId='g-default'` but no group with that id exists → invisible.

## Fix

Route through `ensureUsableGroup` — the same helper `createSession` / `importSession` already use (defined ~line 103). It either returns an existing usable normal group, or synthesizes a fresh one with the localized default name. The synthesis branch threads its updated `groups` array through the existing `set(s => ...)` so the empty-container cleanup operates on the same patch (no second `set` race).

Computed inside the `set` callback against the latest `s` to avoid staleness against any other concurrent slice writes.

## Test

New case in `tests/stores/slices/archiveSession.test.ts` ("synthesizes a normal group when none exist"): seed only `g-default`, archive `s1`, delete `g-default` (s1 survives the cascade since it lives in the archive container by then), unarchive `s1`, assert:

- A fresh normal group exists (length 1, was 0).
- `s1.groupId` points at that new group's id, NOT at the literal `'g-default'`.
- `s1.archivedAt` cleared.
- Archive container removed (was the only holder).

Renamed the adjacent existing test from "falls back to g-default" to "falls back to another usable normal group when source was deleted" — its assertion still holds since `g-default` is a usable normal group in that fixture, and `ensureUsableGroup` prefers existing usable groups before synthesizing.

## Gates

- `npm run typecheck` clean
- `npm run lint` clean (`--max-warnings 0`)
- `npx vitest run tests/stores/slices/archiveSession.test.ts` → 14 passed (was 13)
- Full `npx vitest run`: 36 failures all in `electron-load-smoke.test.ts` (needs `npm run build` first) and `db.test.ts` / `db-hardening.test.ts` (`better-sqlite3` NODE_MODULE_VERSION 145 vs 127 ABI mismatch on the dev box). Verified identical failure set on baseline `main` with my changes stashed — pre-existing environmental, unrelated to this PR.

### DIFF
--- a/src/stores/slices/sessionCrudSlice.ts
+++ b/src/stores/slices/sessionCrudSlice.ts
@@ -512,7 +512,11 @@ export function createSessionCrudSlice(set: SetFn, get: GetFn): SessionCrudSlice
 
     unarchiveSession: (sessionId) => {
       // Clear `archivedAt`, move back to original source group if it
-      // still exists and is normal — otherwise fall back to g-default.
+      // still exists and is normal — otherwise route through
+      // `ensureUsableGroup` so we either land in another existing normal
+      // group or materialize a fresh one. The earlier `'g-default'`
+      // literal fallback orphaned the session (invisible row) when the
+      // user had deleted both the source group and `g-default`.
       // If the archive container empties after the move, delete it.
       // If the user has no active session (e.g. they just archived the
       // active one and then immediately unarchived it), restore activeId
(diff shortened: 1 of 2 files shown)


Jiahui-Gu (commented):
APPROVE.

L1 (necessity + approach): Bug + fix both verified. `ensureUsableGroup` semantics confirmed against src/stores/slices/sessionCrudSlice.ts:96-128 — preferredId→existing-normal→synthesize, exactly the right cascade for the unarchive fallback. Routing through it (vs a second literal fallback) keeps unarchive aligned with createSession/importSession, which is the right consolidation.

L1 (data migration NIT, not blocking): Users who ALREADY hit this bug pre-fix have sessions persisted with `groupId='g-default'` pointing at nothing. Those sessions stay invisible after this merge — the fix only protects future unarchives. A one-time hydrate-time sweep ("if session.groupId resolves to no group, route through ensureUsableGroup") would re-rescue them and arguably belongs in this fix's blast radius. Not blocking since (a) the explicit-repro user base is presumably tiny, (b) a hydrate sweep is a meaningfully different patch worth its own PR + test. Worth filing as a follow-up task.

L2: Diff is a single coherent patch — one `set` callback, `ensured.groups` correctly threaded into BOTH the remainingInContainer branch and the filter branch (so synthesis survives container cleanup). No second set. Test asserts the right invariants: new normal group materialized, s1.groupId points at it and NOT the literal 'g-default', archivedAt cleared, container removed. Adjacent renamed test still valid since g-default is itself a usable normal group in that fixture and ensureUsableGroup prefers existing over synthesis. Ship it; file the hydrate-sweep follow-up.

PR #1381, opened 2026-05-25
fix(shortcuts): wire Ctrl+Shift+N to create new group

Jiahui-Gu:
## Bug
`src/components/ShortcutOverlay.tsx` and `src/components/CommandPalette.tsx` both advertise `Ctrl/Cmd+Shift+N → New group`, but `src/app-effects/useShortcutHandlers.ts` never bound the combo. Pressing it was a no-op.

## Fix
- Add `mod && e.shiftKey && k === 'n'` branch to `useShortcutHandlers`.
- New `createNewGroup` dep on the hook's args; wired from `App.tsx` via `createGroup()` + `focusGroup(id)` (matches `CommandPalette`'s "new group" entry).
- Honor the existing `isEditableTarget` guard so typing `N` in inputs/textareas/contenteditable doesn't hijack.

## Tests (`tests/app-effects/useShortcutHandlers.test.tsx`)
- Ctrl+Shift+N fires `createNewGroup`
- Cmd+Shift+N fires `createNewGroup`
- Ctrl+N alone does NOT fire
- Suppressed when input/textarea is focused
- Extended existing Ctrl+B / Ctrl+K negative assertions to include `createNewGroup`

## Parity note: auto-rename-on-create
`Sidebar.handleNewGroup` additionally puts the freshly-created `GroupRow` into inline-rename mode via component-local state (`justCreatedGroupId`). `CommandPalette`'s "new group" entry does **not** do this, and this shortcut intentionally mirrors `CommandPalette` (create + focus, no rename). Replicating the rename UX from a global shortcut would require lifting `justCreatedGroupId` out of `Sidebar`. Keeping the two global entry points (palette + shortcut) behaviorally consistent felt like the cleaner trade.

## Gates
- `npm run typecheck` clean
- `npm run lint` clean (`--max-warnings 0`)
- `npx vitest run tests/app-effects/useShortcutHandlers.test.tsx` -> 11/11 pass
- Pre-existing full-suite failures are unrelated (missing `dist/electron` build artifacts + native sqlite binding) and present on `main`.

## Test plan
- [ ] Launch app, press Ctrl/Cmd+Shift+N anywhere outside an input -> new group appears in sidebar and is focused
- [ ] Focus a text input/textarea, press Ctrl/Cmd+Shift+N -> nothing happens (typing "N" stays in the field)
- [ ] Press Ctrl+N alone -> no group created

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -97,6 +97,7 @@ export default function App() {
   // toggling a single session's waiting<->idle). See PR #1269.
   const selectSession = useStore((s) => s.selectSession);
   const focusGroup = useStore((s) => s.focusGroup);
+  const createGroup = useStore((s) => s.createGroup);
   const applyExternalTitle = useStore((s) => s._applyExternalTitle);
   const applyCwdRedirect = useStore((s) => s._applyCwdRedirect);
   const applyPtyExit = useStore((s) => s._applyPtyExit);
@@ -184,11 +185,18 @@ export default function App() {
     }));
   }, []));
 
-  // Global keyboard shortcuts (Ctrl+/, "?", Ctrl+F, Ctrl+,).
+  // Global keyboard shortcuts (Ctrl+/, "?", Ctrl+F, Ctrl+,, Ctrl/Cmd+Shift+N).
   useShortcutHandlers({
     toggleShortcuts: React.useCallback(() => setShortcutsOpen((p) => !p), []),
     togglePalette: React.useCallback(() => setPaletteOpen((p) => !p), []),
     openSettings: React.useCallback(() => setSettingsOpen(true), []),
+    // Mirror Sidebar.handleNewGroup minus the inline-rename: create the group
+    // then focus it so a subsequent "New session" lands inside. Auto-rename
+    // is intentionally not replicated here (see useShortcutHandlers note).
+    createNewGroup: React.useCallback(() => {
+      const id = createGroup();
+      focusGroup(id);
+    }, [createGroup, focusGroup]),
   });
 
   // Mirror resolved language to main for OS-level surfaces (tray menu,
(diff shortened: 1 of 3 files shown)


Jiahui-Gu (commented):
APPROVE.

L1 (necessity): Real wiring bug — ShortcutOverlay + CommandPalette both advertise the combo, useShortcutHandlers never bound it. Fix matches the advertised contract.

L1 (skip auto-rename NIT, not blocking): Confirmed Sidebar.handleNewGroup (Sidebar.tsx:129-134) sets `justCreatedGroupId` → GroupRow renders with `autoRename` → inline rename mode. So shortcut-created groups land with the i18n default name and the user has to manually rename via right-click/double-click — one extra click vs the + button. BUT CommandPalette has the same gap, and keeping the two global entry points consistent is the right principle. Lifting `justCreatedGroupId` out of Sidebar to a store slice would be a meaningfully bigger change, and arguably the cleaner long-term fix is to migrate the auto-rename flag into the store and have BOTH palette + shortcut + button set it — file that as a follow-up rather than blocking the wiring fix.

L1 (createGroup signature): Verified — Sidebar.tsx:130 uses `const id = createGroup()` the same way, so the App.tsx callback is sound.

L2: Editable-target guard placement is correct (before preventDefault, matches the `?` handler's pattern). Slight inconsistency with sibling branches (Ctrl+F + Ctrl+, skip the editable-target check) but justified here since unmodified "N" inside Shift+N is closer to a real text-input keystroke than "F" / "," — being stricter is correct. Tests cover all four cases (Ctrl, Cmd, plain Ctrl+N negative, input/textarea suppression) and extend existing negative assertions. Ship it; file the auto-rename-parity follow-up.

PR #1386, opened 2026-05-26
fix(dev): preflight no longer kills sibling webpack-dev-server

Jiahui-Gu:
## Summary

- Tightens `scripts/dev-electron.mjs` preflight on port 4100 so the auto-kill matches ONLY a leftover electron from a prior dev run, not the freshly-spawned sibling webpack-dev-server.
- Old check (cmdline contains `<repo>\node_modules`) matched BOTH electron and the in-flight webpack — killing webpack on every dev startup race.
- Adds `looksLikeElectron` discriminator (exe path / cmdline contains `\electron\`) and an early-return for `looksLikeOurWebpack` (cmdline contains `webpack`). See task #67.

## Test plan

- [ ] Manual: `npm run dev` cold — dev:web stays up, dev:app launches. No "killed PID" log mentioning the webpack process.
- [ ] Manual: leave electron on :4100, run `npm run dev` again — preflight kills the electron, dev launches cleanly.
- [ ] Manual: kill electron mid-dev (sigterm), `npm run dev` from scratch — webpack survives.

No unit tests — `scripts/dev-electron.mjs` has no existing test scaffolding and the logic is hard to mock without a process inspector. Covered by post-merge dev-loop dogfood every PR.

### DIFF
--- a/scripts/dev-electron.mjs
+++ b/scripts/dev-electron.mjs
@@ -61,22 +61,36 @@ function preflightPort4100() {
   if (!holderPid) return;
 
   const info = inspectProcessWin(holderPid);
-  // Auto-kill is reserved for processes we can prove belong to THIS
-  // repo's dev loop: the CommandLine must reference both a script
-  // inside this repo's node_modules AND the repo root path. That admits
-  // both the dev electron (electron.exe under node_modules/electron)
-  // and the sibling webpack-dev-server (node.exe spawned with
-  // node_modules/.bin/webpack.js); it excludes the installed CCSM (in
-  // Program Files / AppData with no repo path in its CommandLine).
+  // Auto-kill is reserved for a LEFTOVER ELECTRON from a previous dev
+  // loop of THIS repo. Two conditions must both hold:
+  //   (a) CommandLine references this repo's node_modules — proves it
+  //       belongs to this checkout, not the installed CCSM or some
+  //       other project.
+  //   (b) The process is electron itself — proves it's the one we want
+  //       to replace, not the sibling webpack-dev-server (node.exe
+  //       running webpack.js) that concurrently just spawned this run.
   //
-  // We deliberately do NOT use ExecutablePath as the discriminator:
-  // webpack-dev-server runs via the system node.exe in
-  // C:\Program Files\nodejs, which would false-negative.
+  // The earlier version of this check only required (a), which falsely
+  // matched the in-flight dev:web and killed it on startup. See task
+  // #67 for the incident. ExecutablePath is reliable for electron
+  // because the dev wrapper invokes node_modules\electron\dist\electron.exe
+  // directly.
   const lcCmd = (info?.commandLine || '').toLowerCase();
+  const lcExe = (info?.executablePath || '').toLowerCase();
   const lcRepo = repoRoot.toLowerCase();
   const lcRepoNm = `${lcRepo}\\node_modules`.toLowerCase();
+  // The holder we want to auto-kill is a LEFTOVER ELECTRON from a prior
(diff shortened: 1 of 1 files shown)


Jiahui-Gu (commented):
## Independent reviewer — APPROVE (logged as comment; gh blocks self-approve)

### Layer 1 — direction
Fix direction is correct.

- Diagnosis is precise: the prior check `cmdline.includes(<repo>\node_modules)` matched BOTH electron (`electron.exe` spawned from `node_modules\electron\dist\`) AND the sibling webpack-dev-server (`node.exe` running `node_modules\.bin\webpack`). On every cold `npm run dev`, `concurrently` races the two — if webpack wins the bind on :4100 first, preflight kills it. Documented in the commit message; matches task #67.
- The two-axis discriminator (`looksLikeElectron` = exe contains `\electron\` / cmdline contains `\node_modules\electron\`; gated by existing repo-path check) is the minimal correct fix. webpack-dev-server is `node.exe` invoking `node_modules\.bin\webpack.js` — none of those three substrings can match.
- Considered tighter alternative: requiring `--user-data-dir=.dev-userdata` (a definitive "ours" marker since `dev-electron.mjs` is the only thing that sets it). That would be even stricter, but only matters for the auto-kill branch, and the current condition is already strict enough (electron + this-repo's-node_modules is functionally identical to "leftover from THIS dev wrapper"). Not worth a follow-up.
- Windows electron path layout assumption (`node_modules\electron\dist\electron.exe`): fixed by the `electron` npm package's own publish layout, not by ccsm. Stable as long as the package format doesn't change. Three independent signals (exe `\electron\`, exe `\electron.exe`, cmdline `\node_modules\electron\`) means any single layout shift wouldn't false-negative. Robust.

### Layer 2 — implementation

- `looksLikeOurWebpack` cmdline-contains-`webpack` concern: the substring is gated by `lcCmd.includes(lcRepoNm)`, which already proves the holder is THIS repo's process. Another project's webpack can't appear in a :4100-listener cmdline that ALSO contains our absolute repo path. Safe.
- Hypothetical edge: some unrelated tool launched with our repo's node_modules in its cmdline AND the word "webpack" → matches `looksLikeOurWebpack`, returns early. Worst case: electron then fails to bind :4100 with EADDRINUSE and dev exits noisily. NOT a destructive failure mode — auto-kill is gated by `looksLikeElectron`, not `looksLikeOurWebpack`. The early-return path is benign.
- The `info &&` guard at the front of both `looksLikeOurDev` and `looksLikeOurWebpack` correctly handles the wmic-fails-and-CIM-fails-and-tasklist-fallback case (all-null info → neither branch fires → fall through to "refuse to kill blindly" + exit 1, which is the correct safe default).
- The ExecutablePath check is correctly described as load-bearing for electron only (webpack-dev-server runs via system node.exe in `C:\Program Files\nodejs\`, which would false-negative on `lcExe.includes('\electron\')`). The cmdline path remains the primary repo-affinity signal for both cases.
- Lines kept narrow; no unnecessary refactoring; comment updates accurately describe new behaviour. Good discipline.

### Tests / verification
- No unit tests added — confirmed `scripts/dev-electron.mjs` has no test scaffolding and the inspector is hard to mock without a real process. The PR body acknowledges this and defers to dev-loop dogfood, which matches the team's dogfood protocol.
- `node --check scripts/dev-electron.mjs` → OK (parse-clean).
- Manual `npm run dev` exercise is the manager's responsibility per the worktree rules. Not run here.

### Conclusion
APPROVE. Tight, surgical fix to the right layer; no destructive failure modes added; comments explain the why precisely. Ready to merge pending the manager's dev-loop smoke.

PR #1387, opened 2026-05-26
feat(dev): random port + per-run userData for npm run dev (#68)

Jiahui-Gu:
## Summary

Task #68 — eliminates port-4100 / `.dev-userdata` collisions between concurrent `npm run dev` shells, and tightens zombie cleanup on dev exit. Also folds in Task #67's preflight tightening that's already on main.

- New `scripts/dev.mjs` orchestrator replaces `concurrently -k web,app`. Single Node process owns the dev lifecycle.
- Random free port via `net.createServer().listen(0)`, threaded through `CCSM_DEV_PORT` to webpack-dev-server + electron (existing plumbing in `createWindow.ts`).
- Per-run `.dev-userdata-<6 hex>` via `crypto.randomBytes(3)`, exposed as `CCSM_DEV_USERDATA`; `scripts/dev-electron.mjs` reads it and passes to Electron via `--user-data-dir`.
- Idempotent cleanup trap on `exit` / `SIGINT` / `SIGTERM` / `SIGHUP` / `uncaughtException` / `unhandledRejection` → `taskkill /F /T /PID` per child (Windows) or `kill(-pgid, SIGTERM)` + SIGKILL fallback (POSIX) + `rmSync` userData dir.
- Orphan reaper at orchestrator start: `.dev-userdata-XXXXXX` dirs older than 1h are swept. Covers the force-kill-parent edge case where Node's exit hooks don't fire on Windows.
- `dev-electron.mjs` preflight scoped to standalone runs (env `CCSM_DEV_PORT` unset). In the orchestrated path the only listener on the random port is our own sibling webpack, which the existing `looksLikeOurDev` check false-positives on across worktrees due to npm node_modules hoisting.

## Design choices

- **Method A (replace concurrently) over B (temp-file port handoff)**: a single Node process makes cleanup ordering deterministic and avoids the temp-file teardown race. concurrently's `-k` is unreliable on Windows for electron's renderer/utility tree anyway.
- **No Job Objects**: would require a native addon. The realistic failure case (external `Stop-Process -Force` on the orchestrator, no exit hook fires) is covered by the orphan-userData sweep + orphan electrons being harmless on a no-longer-listening random port.

## Test plan

Unit tests (new): `tests/scripts/devOrchestrator.test.ts` — 8 tests, all passing.
- Red→green: initial run (no `dev-lib.mjs`) failed import; after implementation, 8/8.

Integration evidence (two parallel `npm run dev` shells):
- Shell 1: `[dev] port=56260 userData=.dev-userdata-7090a7`
- Shell 2: `[dev] port=63029 userData=.dev-userdata-0a6ad5`
- Both webpack-dev-servers compiled successfully; 10 electron processes (5 each, full main+renderer+utility trees).
- Orphan reaper verified: backdated `.dev-userdata-7090a7` by 2h, `reapOrphanUserDataDirs` removed it and left the fresh `.dev-userdata-0a6ad5` intact.

Limitations of the headless-agent verification:
- Could not exercise the true Ctrl-C path from this agent context. `process.kill(pid, 'SIGINT')` on Windows is `TerminateProcess`, not a real CTRL_C_EVENT — exit hooks don't fire. The cleanup-trap *logic* is unit-tested; a human in a real terminal needs to confirm Ctrl-C delivers the graceful path with electron + userData cleanup.

## Test plan for reviewer

- [ ] Run two shells: `npm run dev` in each. Confirm distinct random ports + distinct `.dev-userdata-<hex>/` dirs.
- [ ] Ctrl-C shell 1 in a real terminal. Verify shell-1's userData dir is removed and only shell-2's electron tree remains.
- [ ] Ctrl-C shell 2. Verify all electron processes gone, all `.dev-userdata-*` dirs gone.
- [ ] Reverse-verify: stash this PR, run two `npm run dev`, confirm shell 2 EADDRINUSE on 4100. Reapply, confirm no collision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ release/
 .dogfood-userdata-happy/
 .dogfood-userdata-ttyd-debug/
 .dev-userdata/
+.dev-userdata-*/
 docs/screenshots/dogfood-current-ui/
 docs/screenshots/dogfood-happy-path/
 *.exe
--- a/package.json
+++ b/package.json
@@ -18,9 +18,9 @@
     "node": ">=22.0.0"
   },
   "scripts": {
-    "dev": "concurrently -k -n web,app -c blue,magenta \"npm:dev:web\" \"npm:dev:app\"",
+    "dev": "node scripts/dev.mjs",
     "dev:web": "webpack serve --config webpack.config.js --mode development",
-    "dev:app": "wait-on http://localhost:4100 && tsc -p tsconfig.electron.json && node scripts/dev-electron.mjs",
+    "dev:app": "node scripts/dev-wait-web.mjs && tsc -p tsconfig.electron.json && node scripts/dev-electron.mjs",
     "clean": "node -e \"require('fs').rmSync('dist',{recursive:true,force:true})\"",
     "build": "npm run clean && tsc -p tsconfig.electron.json && webpack --mode production && node -e \"const fs=require('fs');const p='dist/renderer/bundle.js';const t=Date.now()/1000;fs.utimesSync(p,t,t)\"",
     "typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.electron.json",
--- a/scripts/dev-electron.mjs
+++ b/scripts/dev-electron.mjs
@@ -12,8 +12,10 @@
 //      `productName`, which is "CCSM" in both prod and dev. Without an
 //      override, dev runs would read/write the same SQLite DB and prefs as
 //      your installed ccsm — risky even after exit (WAL recovery, schema
-//      migrations). We point the dev instance at `.dev-userdata/` inside the
-//      worktree via Electron's `--user-data-dir` switch.
+//      migrations). The orchestrator (`scripts/dev.mjs`) hands us a
+//      per-run random `.dev-userdata-<hex>/` via the `CCSM_DEV_USERDATA`
+//      env var; standalone invocations fall back to `.dev-userdata/`.
+//      Either way, we pass it to Electron via `--user-data-dir`.
 //
 //   3. Dev vs installed CCSM identity.  Both run as `CCSM.exe` /
 //      `electron.exe`; tasklist / Alt-Tab / Task Manager can't tell them
@@ -22,13 +24,16 @@
 //      rename the app to "CCSM (dev)" and createWindow tag the title
 //      "CCSM [dev]" — see electron/main.ts and electron/window/createWindow.ts.
 //
-//   4. Stale-port detection.  webpack-dev-server (sibling `dev:web`)
-//      listens on :4100. If a previous `npm run dev` left zombie
-//      processes, the new run dies with `EADDRINUSE` mid-bringup. We
-//      probe :4100 BEFORE spawning electron and refuse to kill blindly:
-//      identify the holder, auto-kill only if it's an electron pointed
-//      at THIS repo, otherwise print PID / path / CommandLine and bail.
-//      Rule from feedback_never_blind_kill_ccsm.md: never kill
+//   4. Stale-port detection.  In the orchestrated path (`npm run dev`
+//      via `scripts/dev.mjs`), each run gets a fresh random port via
+//      OS allocation — collisions are vanishingly rare and the
+//      preflight is skipped to avoid false-positives on our own
+//      sibling webpack. When this wrapper is invoked standalone
+//      (legacy fixed port :4100), we probe :4100 BEFORE spawning
+//      electron and refuse to kill blindly: identify the holder,
+//      auto-kill only if its CommandLine references this repo's
+//      node_modules, otherwise print PID / path / CommandLine and
+//      bail. Rule from feedback_never_blind_kill_ccsm.md: never kill
 //      ambiguous CCSM.exe processes.
 //
 // Cross-platform: spawning Node with explicit arg/env propagation works on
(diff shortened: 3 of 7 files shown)


Jiahui-Gu:
## Review: #1387 — `npm run dev` random port + per-run userData + cleanup trap

Independent reviewer. Verified diff against `origin/main`, ran unit tests, and reverse-verified behavior on Windows. **Recommendation: approve with one follow-up tracked.**

---

### Layer 1 — direction

**Method A (single orchestrator) vs Method B (concurrently + shared port file):** A is the right call. Dynamic env between siblings is exactly what `concurrently` can't do without a temp-file dance, and we now own the kill semantics rather than borrowing concurrently's "kill-others" which is documented-flaky on Windows electron trees. Stdio coloring loss is real but minor — both children are `stdio: 'inherit'` and prefixed by webpack/electron's own logging.

**Job Object / windows-kill-tree fallback:** I reverse-verified the zombie limitation. With orchestrator PID force-killed (`taskkill /F /PID` without `/T`):
- 5 electron processes survived (main + 4 renderer/utility children).
- **Plus webpack-dev-server survived too** — broader than the PR description claims.
- Next-start reaper only sweeps userData *directories*, not processes. Zombies persist until user manually kills or reboots.

This is acceptable for v1 — the realistic trigger is rare (Task Manager force-kill, BSOD, OOM-killer) and the alternative (native job-object addon) is genuinely out of scope. But the PR claims should be slightly broadened: orphan reaper covers stale **dirs only, not stale processes**, and webpack-dev-server is also a victim. Recommend filing a follow-up issue for `windows-kill-tree` or a `tasklist` sweep at orchestrator start that nukes electron.exe whose CommandLine references this repo's `.dev-userdata-*`.

**Preflight retained in `dev-electron.mjs` for standalone runs:** still has narrow value when someone debugs `npm run dev:app` standalone with `CCSM_DEV_PORT` unset (the existing `wait-on` shim falls back to 4100). Keep it. Cost: ~25 lines, gain: don't kill installed CCSM. Worth it.

### Layer 2 — implementation

- **Port allocator race window** (close → spawn webpack): real but vanishingly small. `allocateDevPort()` returns OS ephemeral port (49152–65535 on Windows), webpack rebinds within tens of ms. PR comment in `dev-lib.mjs:18-24` is honest about it. No retry — acceptable given chance is sub-1% and failure mode is a loud webpack EADDRINUSE that user can just rerun. **Could** harden by passing the bound socket via fd, but agreed it's overkill.

- **Cleanup trap idempotency:** `let ran = false` + collapsed once() — unit-tested via `SIGINT then exit` sequence. Confirmed: cleanup runs once. Good.

- **Orphan reaper bounds:**
  - 1h cutoff: reasonable. A dev session >1h is normal; matters only for the very rare case where two dev shells start within 1h of each other AND one died via force-kill. In that case the live one's dir is younger than 1h so it's safe. **No false-positive risk** given the random-hex names — reaper never touches `.dev-userdata/` (no hash) or any path outside the pattern.
  - 6-hex glob: confirmed `/^\.dev-userdata-[0-9a-f]{6}$/` won't match legacy `.dev-userdata` or any `.dogfood-userdata-*`. Good.
  - mtime concern on WSL/Cygwin: not relevant — this is a Windows-host npm process, mtimes are NTFS native. Even if mtime didn't update (e.g. read-only sessions), worst case is a 1h delay before stale dir gets swept. Non-blocking.

- **dev-electron.mjs preflight scoping via `CCSM_DEV_PORT`:** confirmed — `if (!process.env.CCSM_DEV_PORT) preflightPort('4100')`. Clean predicate.

- **Test coverage (8 tests):**
  - `allocateDevPort`: returns valid port + can re-listen + no static cache. 
  - `generateUserDataDirName`: pattern match + 32-draw collision check.
  - `installCleanupTrap`: registers all 6 events + idempotent across SIGINT+exit (uses real `EventEmitter`, not just spy on `process.on`). Solid.
  - `reapOrphanUserDataDirs`: stale removed, fresh kept, legacy `.dev-userdata` untouched, no-throw on empty repo.
  - Missing but minor: no test that `installCleanupTrap` calls `proc.exit(130)` on SIGINT, no test for `reapOrphanUserDataDirs` swallowing rmSync errors (locked dir). Both are simple branches; not blocking.

---

### Reverse-verify log

**1. Unit tests:**
```
$ npx vitest run tests/scripts/devOrchestrator.test.ts
Test Files  1 passed (1)
Tests       8 passed (8)
```

**2. Random port allocation (live):**
```
[dev] port=58372 userData=.dev-userdata-119da8
[dev] port=59058 userData=.dev-userdata-b801d8
```
Two consecutive runs got distinct random ports + distinct hex userData dirs. Webpack bound to the random port (`webpack-dev-server` log: `Loopback: http://localhost:58372/`). 

**3. Main branch confirms collision baseline:**
```
$ git show origin/main:package.json | grep dev:app
"dev:app": "wait-on http://localhost:4100 && ..."
```
Fixed :4100 — two concurrent shells WOULD collide. PR fixes.

**4. Force-kill zombie test (the documented limitation):**
```
$ taskkill /F /PID <orchestrator>     # NOT /T
SUCCESS
$ Get-CimInstance Win32_Process -Filter "Name='electron.exe'"
5 electron processes still alive (main + 4 children)
$ ls .dev-userdata-*
.dev-userdata-119da8/   ← still on disk
$ <also> webpack-dev-server still alive on port 58372
```
Confirmed limitation is real. Next-start reaper would clean the dir (after 1h or with backdated mtime — verified: `reaped: [ '.dev-userdata-119da8' ]`) but **does not clean the zombie electron/webpack processes**.

**5. Tree-kill (the working path):**
```
$ taskkill /F /T /PID <orchestrator>
SUCCESS — 14 processes terminated (orchestrator + npm shims + webpack + tsc + electron tree)
```
This is what `cleanup()` does via `spawnSync('taskkill', ['/F','/T','/PID',pid])` in the SIGINT/exit path — proves the happy-path teardown is complete.

---

### Verdict

**Approve.** Direction correct, implementation clean, tests adequate, limitations honestly acknowledged in code comments.

**Follow-up to track (non-blocking):**
1. Process-level orphan sweep at orchestrator start — scan `electron.exe` and `node.exe` whose CommandLine references this repo's `.dev-userdata-*` and kill them before allocating a new port. Closes the force-kill-no-/T zombie hole.
2. PR description / `scripts/dev.mjs` header comment should mention webpack-dev-server is also a force-kill victim, not just electron.

Nothing in this PR needs to change before merge.

PR #1388, opened 2026-05-26
fix(terminal): warm switch uses display toggle, not reparent (bug #69)

Jiahui-Gu:
## Real root cause

Webkit drops `.xterm-viewport.scrollTop` whenever the element leaves the layout tree. PR #1374's save/restore writes the captured scrollTop back AFTER the reparent — correct final state, but for one or more frames the user sees the scrollbar at 0 before it jumps to the right position. Take 3 (PR #1385 commit `c631e20d`, direct DOM write + double-RAF) narrowed the gap to a single frame but did not close it: the wrapper still detaches.

The probe `scripts/dogfood-bug-69-no-flicker.mjs` hooks `requestAnimationFrame` in the renderer and samples `.xterm-viewport.scrollTop` for 800ms after a warm switch. On origin/main it shows exactly one bad frame (`t=1ms scrollTop=0`) before recovery at frame 2 — that's the visible flash. PR #1385's c631e20d would behave similarly.

## Fix: display toggle, not reparent

Stop reparenting. Each warm wrapper is appended once to the TerminalPane host and stays there for its whole life; switching sessions flips `wrapper.style.display` between `'none'` and `''`. The wrapper never leaves the layout tree, so webkit never zeroes scrollTop and there is no first-frame flicker.

`TerminalPane` is rendered without a `key` in `App.tsx` (line 430: `<TerminalPane sessionId={active.id} cwd={...} />`), so React keeps the host node stable across session switches — that's the precondition that makes display-toggle work without reparent.

### display:none over visibility:hidden + position:absolute

Empirical per-frame probe (below) shows the show transition produces no scrollTop jump: `ydisp` is stable across the hide window, so xterm's internal ResizeObserver / Viewport refresh re-derives the same scrollTop. `display:none` also lets Chromium fully pause paint work on hidden wrappers — same benefit the offscreen holder gave us, without the layout-tree detach.

## Per-frame probe (red → green)

**On origin/main (red):**
```
sample count: 28
first frames:
  {"t":1,"scrollTop":0,"display":"","connected":true}
  {"t":17,"scrollTop":2314.666748046875,"display":"","connected":true}
  ...
FAIL: bug #69 — warm switch flickers. 1 frame(s) out of 28 had scrollTop != 2314.67 (±2px).
```

**On this PR (green):**
```
sample count: 27
first frames:
  {"t":0,"scrollTop":0,"display":"none","connected":true}   (pre-switch — display none excluded)
  {"t":27,"scrollTop":2314.666748046875,"display":"","connected":true}
  ...
PASS: all frames pinned to saved scrollTop.
```

The lone `t=0` `display:none` sample is the pre-switch baseline (sidA is hidden while user is on sidB); the probe explicitly excludes hidden frames since they can't be visible to the user.

## Changes

- `src/terminal/xtermWarmRegistry.ts`:
  - Drop `offscreenHolder` + `getOffscreenHolder()`.
  - Drop `WarmEntry.savedScrollTop`, the post-reshow scrollTop write, and the `terminal.warmShow.scrollRestore` probe.
  - `allocEntry`: wrapper starts unparented + `display:none`. No appendChild to any holder.
  - `ensureAndShowEntry`: hide branch is `prev.wrapper.style.display = 'none'`; show branch appendChild's only if `parentElement !== host` (covers first-show + rare host-changed remount), then sets `display = ''`. `term.open()` still deferred to first show.
  - `__resetRegistryForTests`: drop offscreenHolder cleanup.
- `src/terminal/usePtyAttach.warm.ts`: header doc + ResizeObserver-gate comment updated to reflect the display-toggle model. No behavioural change in the hook itself.
- `tests/terminal/xtermWarmRegistry.test.ts`: add new `hide toggles display:none on previous active wrapper` assertion that pins the no-reparent invariant; refresh the deferred-`term.open()` test's rationale comment. All 16 tests pass.
- `scripts/dogfood-bug-69-no-flicker.mjs`: new probe (per-frame RAF sampler).

`scripts/dogfood-pr-1374-warm-scroll-preserved.mjs` still passes — its assertion ("scrollTop equal before/after switch") is satisfied by the display-toggle model trivially.

## Replaces PR #1385

PR #1385 (take 3 / c631e20d on `fix/attach-scrollbar-mismatch`) attempted to close the gap with a direct DOM `scrollTop` write + double-RAF after reparent. Empirically still showed a first-frame flash because reparent itself is what zeroes scrollTop. This PR removes the reparent entirely. Closing #1385.

## Test plan

- [ ] Reviewer pulls branch, runs `npm test -- tests/terminal/` (66/66 expected).
- [ ] Reviewer runs `node scripts/dogfood-bug-69-no-flicker.mjs` after `npm run build` — must report PASS with all frames at the saved scrollTop.
- [ ] Reviewer runs `node scripts/dogfood-pr-1374-warm-scroll-preserved.mjs` — must still PASS.
- [ ] User dogfood: in real CLI, open 2+ sessions, claude in session A, scroll up in normal buffer, switch to B and back to A — scrollbar must not flash to top.
- [ ] Reverse-verify: stash the registry/hook changes, rerun bug-69 probe → must FAIL (proves the probe actually catches the regression).

### DIFF
--- a/scripts/dogfood-bug-69-no-flicker.mjs
+++ b/scripts/dogfood-bug-69-no-flicker.mjs
@@ -1,0 +1,220 @@
+// scripts/dogfood-bug-69-no-flicker.mjs
+//
+// Bug #69 regression probe — warm session-switch must NOT flicker the
+// scrollbar.
+//
+// Background: PR #1374 (already on origin/main) restores
+// `.xterm-viewport.scrollTop` after the warm reshow's reparent, but the
+// restore happens AFTER webkit has zeroed the scrollTop on layout-tree
+// detach. The user-visible symptom is the scrollbar thumb being at the
+// TOP for the first 1–3 frames after a B → A switch, then jumping to the
+// correct position. Even though the final state is right, the intermediate
+// frames are perceptible as a flash.
+//
+// The fix (this PR) switches the registry from "reparent across hosts" to
+// "display:none / display:'' on a single persistent host" so the wrapper
+// never leaves the layout tree, webkit never zeroes scrollTop, and there
+// is no frame at scrollTop=0 to clobber.
+//
+// This probe asserts the no-flicker invariant per-frame: it hooks
+// requestAnimationFrame on the renderer side BEFORE the B → A click,
+// records the `.xterm-viewport.scrollTop` at every frame for 600ms after
+// the switch, and asserts that NONE of those samples sit at 0 (or any
+// value other than the saved scrollTop). The presence of even one
+// off-target frame fails the probe.
+//
+// Exit code 0 = PASS, 1 = FAIL.
+
+import { mkdtempSync, readFileSync, existsSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import * as path from 'node:path';
+import {
+  createIsolatedClaudeDir,
+  dismissWelcomeSplash,
+  launchCcsmIsolated,
+  seedSession,
+  waitForTerminalReady,
+  waitForXtermBuffer,
+} from './probe-utils-real-cli.mjs';
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+async function readScrollState(win) {
+  return await win.evaluate(() => {
+    const t = window.__ccsmTerm;
+    const b = t?.buffer?.active;
+    // The active wrapper is the one with display !== 'none'. There can be
+    // multiple .xterm-viewport elements on the page (one per warm wrapper
+    // under the new design) — pick the one whose ancestor wrapper is
+    // visible.
+    const wrappers = Array.from(document.querySelectorAll('[data-ccsm-warm-sid]'));
+    const visible = wrappers.find((w) => w instanceof HTMLElement && w.style.display !== 'none');
+    const vp = (visible ?? document).querySelector('.xterm-viewport');
+    return {
+      viewportY: b?.buffer?.active?.viewportY ?? b?.viewportY ?? null,
+      baseY: b?.buffer?.active?.baseY ?? b?.baseY ?? null,
+      bufferType: t?.buffer?.active?.type ?? null,
+      scrollTop: vp instanceof HTMLElement ? vp.scrollTop : null,
+      scrollHeight: vp instanceof HTMLElement ? vp.scrollHeight : null,
+    };
+  });
+}
+
+async function main() {
+  const tempDir = mkdtempSync(path.join(tmpdir(), 'ccsm-dogfood-69-'));
+  createIsolatedClaudeDir(tempDir);
+
+  const { electronApp, win, userDataDir } = await launchCcsmIsolated({ tempDir });
+
+  try {
+    await win.waitForFunction(
+      () => !document.querySelector('[data-testid="claude-availability-probing"]'),
+      null,
+      { timeout: 30000 },
+    );
+
+    const { sid: sidA } = await seedSession(win, { name: 'A', cwd: tempDir });
+    const { sid: sidB } = await seedSession(win, { name: 'B', cwd: tempDir });
+
+    await win.evaluate((id) => window.__ccsmStore.getState().selectSession(id), sidA);
+    await waitForTerminalReady(win, sidA, { timeout: 45000 });
+    await waitForXtermBuffer(win, /claude|welcome|│|╭|\?\sfor\sshortcuts/i, { timeout: 30000 });
+    await dismissWelcomeSplash(win);
+
+    // Drop claude's alt-screen so we get scrollable normal-buffer (the
+    // surface where the bug is visible). Fill scrollback and scroll up so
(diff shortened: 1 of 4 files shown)


Jiahui-Gu:
## Independent Reviewer — APPROVE

Bug #66 fix attempt 4. Reverse-verified on this machine (Windows 11, Electron prod bundle). The probe is genuinely sensitive to the bug and confirms the architectural change closes the first-frame flash.

### Reverse-verify (the gate the previous reviewer failed)

**1. PR branch (`fix/warm-display-switch`, commit 0fc5c3bc) — bug-69 probe: PASS**
```
sample count: 28
  {"t":1,"scrollTop":0,"display":"none","connected":true}      <- pre-switch, excluded
  {"t":15,"scrollTop":2314.666748046875,"display":"","connected":true}
  ... all subsequent frames pinned to 2314.6667 ...
PASS: all frames pinned to saved scrollTop.
```

**2. origin/main — bug-69 probe: FAIL (genuine red)**
```
sample count: 28
  {"t":1,"scrollTop":0,"display":"","connected":true}          <- THE BUG: zeroed by reparent
  {"t":16,"scrollTop":2250,"display":"","connected":true}      <- restored on frame 2+
  ...
FAIL: bug #69 — warm switch flickers. 1 frame(s) out of 28 had scrollTop != 2250 (±2px).
```
main.log tail shows `terminal.warmShow.scrollRestore` event firing — i.e. main is genuinely running the old reparent + scrollTop-write path, and probe correctly catches the one-frame webkit zero. Not "fake red" like v3.

**3. PR branch — pr-1374-warm-scroll-preserved probe: PASS** (regression invariant intact)

**4. PR branch — `npm test -- tests/terminal/`: 7 files, 66/66 PASS**

### Layer 1 — direction

- **display:none vs visibility:hidden+absolute**: display:none is correct. visibility:hidden keeps paint running (Chromium spec — only pixels are suppressed) so the CPU benefit of the offscreen-holder is lost. display:none takes the element out of the render tree (still in layout tree as no-box) so paint pauses, and crucially the scrollable overflow state of `.xterm-viewport` is preserved across the hide window per empirical probe. ydisp-stable claim holds — scrollTop is identical 2314.6667 across hide/show.
- **DOM node fan-out**: bounded by `getWarmCap() = DEFAULT_WARM_CAP = 20`. Twenty hidden xterm wrappers under TerminalPane host is fine; Chromium doesn't paint display:none subtrees and the parser/cell-grid drain has always run for background entries (that's the warm registry's whole point — comment in `dataUnsubscribe` makes this explicit).
- **xterm Terminal behaviour under display:none**: unchanged — `term.write` still drains the parser, canvas atlas paint is paused via IntersectionObserver (xterm 5.5 source confirms). This is the SAME behaviour the offscreen holder previously provided; display:none is just a cleaner mechanism for it.
- **Caller surface**: I grepped `ensureAndShowEntry|disposeEntry` across `src/`. All call sites (LRU eviction, session-deleted, reload, retry, cancelled-mid-cold-attach, unload) operate on `entry.wrapper.remove()` which works identically on display:none nodes. TerminalPane is rendered without `key` in App.tsx:430 so the host node is stable — display-toggle precondition holds. The `parentElement !== host` branch in `ensureAndShowEntry` correctly handles the rare TerminalPane remount case.
- **Offscreen residuals**: only in doc comments (lines 11, 362, 781 of xtermWarmRegistry.ts) explaining the migration. Zero runtime references.

### Layer 2 — implementation

- `hide` branch: single `prev.wrapper.style.display = 'none'` write, no side effects.
- `show` branch: `appendChild` gated on `parentElement !== host` (first show + host-changed remount only), then `display = ''`. Correct — '' resets to the inline default rather than forcing `'block'`/`'flex'` and overriding the wrapper's Tailwind `absolute inset-0` class.
- `term.open()` still deferred to first show (test "defers term.open()" still passes, rationale comment updated to reflect new alloc state).
- `savedScrollTop` / double-RAF / scrollRestore probe — all gone, confirmed by grep.
- `installInputListeners` deferral via `entry.opened && !entry.inputListenersInstalled` still correct since `term.open()` is what creates `term.textarea`.
- `applyTerminalScrollback` / `applyTerminalFontSize` / `runResizeReplayForEntry` / `disposeEntry` — none touch display state; behavior unchanged.

### Test

The new `hide toggles display:none on previous active wrapper` assertion is load-bearing and asserts what matters (`parentElement === host` AND `display === 'none'` on hide). It also covers a switch-back (A→B→A) which is the multi-session smoke. Worker's probe runs single A→B→A; given the implementation has no "first vs subsequent show" branching past first-mount, additional cycles wouldn't exercise new code paths — adequate as-is.

### Result

**APPROVE.** Probe red on main, green on PR, all unit tests pass, PR #1374 invariant still holds. Architecturally clean — the wrapper genuinely never leaves the layout tree on a session switch, which is the only thing that closes bug #66.

Reviewed at commit `0fc5c3bc`.

PR #1393, opened 2026-05-26
feat(attach): rewrite around shellRegistry + z-stack + per-shell mask

Jiahui-Gu:
Implements `docs/attach-redesign.html` landed in commit `0aad80b0` (doc body: `4dd5c4b`).

## Summary

- New `src/terminal/shellRegistry.ts` (~520 LOC): owns N visited shells under one host. `getShell / createShell / disposeShell / disposeAll`. Switching is a `display` + `z-index` flip on a single persistent host — wrappers never reparent. Each shell carries wrapper + xterm + mask div + per-sid `pty.onData` subscription + IME/paste/key listeners.
- New `src/terminal/usePtyAttachShell.ts` (~390 LOC): thin orchestrator. COLD = create with mask on → pty.attach → getBufferSnapshot → term.reset+write → subscribeShellData → setMask(false). VISITED = showShell only, no mask, no IPC. RELOAD = resetShellForReload (term.reset in place per design §4) + cold suffix. Bailout on any throw = disposeShell + setState(error); user-driven Retry rebuilds the shell from scratch (design §5).
- Renderer consumers (`TerminalPane`, `useAtBottom`, `useTerminalFontSize`, `useTerminalScrollback`) now read shellRegistry instead of xtermWarmRegistry. `sessionCrudSlice.deleteSession` calls `disposeShell` so the z-stack collapses to the next remaining shell (or back to State 0 when empty). `sessionRuntimeSlice.reloadSession` no longer disposes — bumps `reloadNonce` and lets the hook reset+resnapshot in place.
- Old `xtermWarmRegistry.ts` + `usePtyAttach.warm.ts` remain on disk untouched. Deletion is a follow-up PR (task #73 territory).

## Probe results

- `scripts/dogfood-attach-redesign.mjs` (new) — asserts the 3 UX states + 3 ops on real Electron with real claude.

  ```
  PASS State 0: no shell wrappers at boot
  PASS Z-stack: exactly one shell visible (sidB), sidA hidden but parented
  PASS State 2: visited switch had no mask across 26 visible frames
  PASS Reload: mask shown then hidden (148 frames sampled)
  PASS Delete top: z-stack collapsed to next without mask
  PASS Copy: new session cold-started with mask
  PASS Delete all: back to State 0 (no shells)
  PASS: attach-redesign UX states + ops behave as specified
  ```

- `scripts/dogfood-bug-69-no-flicker.mjs` (selector updated to `data-ccsm-shell-sid`) — `PASS: all frames pinned to saved scrollTop.` No regression.

## Quality gates

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — 1871 passed, 1 todo. The 18 remaining failures are pre-existing `electron/__tests__/db*.test.ts` better-sqlite3 native-module rebuild issues (verified by `git stash` baseline check); unrelated to this PR.

## Notes on design fidelity

- No LRU cap, no feature flag, no enumerated cancellation tokens. Single dispose+retry bailout.
- Cold start order is exactly `build xterm → open → fetch snapshot → write → subscribe PTY` (design §3) — the per-shell `pty.onData` subscription is installed AFTER snapshot write via `subscribeShellData`. Theoretical loss window between snapshot capture and subscribe is acceptable per design §5; user-visible bailout is dispose+retry.
- Reload uses `term.reset()` in place and re-runs the snapshot suffix — no shell dispose (design §4).
- Delete = disposeShell + z-stack collapse to most-recently-inserted survivor; empty stack = State 0 (right pane blank).

### DIFF
--- a/scripts/dogfood-attach-redesign.mjs
+++ b/scripts/dogfood-attach-redesign.mjs
@@ -1,0 +1,332 @@
+// scripts/dogfood-attach-redesign.mjs
+//
+// E2E probe for the attach-redesign rewrite (docs/attach-redesign.html).
+//
+// Verifies the 3 UX states + 3 ops:
+//
+//   STATE 0  — boot with no active session → right pane has no shell wrapper.
+//   STATE 1  — cold start (first click on session A) → wrapper appears with
+//              mask shown; mask removed once attach completes.
+//   STATE 2  — second click on session B (also cold); then A→B→A swap is
+//              instant: no mask on the visited switch.
+//
+//   DELETE   — delete the top session → next session in z-stack becomes
+//              top, no mask. Delete it too → no shells left, back to State 0.
+//   RELOAD   — reload top session → mask shown briefly, then content back.
+//   COPY     — create a new session (e.g. via createSession) and click it →
+//              normal cold-start sequence (mask → content).
+//
+// Frame-level invariants asserted:
+//   * State 1: mask visible for ≥1 frame, then hidden.
+//   * State 2 (visited switch): mask never visible.
+//   * Delete top: no mask appears on the new top.
+//   * Reload top: mask visible at some point during the suffix.
+//
+// Exit 0 = PASS, 1 = FAIL.
+
+import { mkdtempSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import * as path from 'node:path';
+import {
+  createIsolatedClaudeDir,
+  dismissWelcomeSplash,
+  launchCcsmIsolated,
+  seedSession,
+  waitForTerminalReady,
+  waitForXtermBuffer,
+} from './probe-utils-real-cli.mjs';
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+async function snapshotShells(win) {
+  return await win.evaluate(() => {
+    const wrappers = Array.from(document.querySelectorAll('[data-ccsm-shell-sid]'));
+    return wrappers.map((w) => {
+      const mask = w.querySelector('[data-ccsm-shell-mask]');
+      return {
+        sid: w.getAttribute('data-ccsm-shell-sid'),
+        display: w.style.display || '',
+        zIndex: w.style.zIndex || '',
+        maskDisplay:
+          mask instanceof HTMLElement ? (mask.style.display || '') : null,
+      };
+    });
+  });
+}
+
+/**
+ * Install a per-RAF sampler that records mask state for `sid` until the
+ * sampler is read back. Returns the JS expression that reads + clears the
+ * window-stashed samples.
+ */
+async function startMaskSampler(win, sid, durationMs = 1500) {
+  await win.evaluate(
+    ({ targetSid, duration }) => {
+      window.__attachRedesignSamples = [];
+      const startedAt = performance.now();
+      const tick = () => {
+        const wrapper = document.querySelector(
+          `[data-ccsm-shell-sid="${CSS.escape(targetSid)}"]`,
+        );
+        let maskDisplay = null;
+        let wrapperDisplay = null;
+        if (wrapper instanceof HTMLElement) {
+          wrapperDisplay = wrapper.style.display || '';
+          const mask = wrapper.querySelector('[data-ccsm-shell-mask]');
+          if (mask instanceof HTMLElement) {
+            maskDisplay = mask.style.display || '';
+          }
+        }
+        window.__attachRedesignSamples.push({
+          t: Math.round(performance.now() - startedAt),
+          wrapperDisplay,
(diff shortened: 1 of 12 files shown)


Jiahui-Gu (commented):
**Layer 1**: PASS — implementation matches `docs/attach-redesign.html` faithfully. No scope creep (no LRU, no feature flag, no cancellation tokens). All 3 UX states + 3 ops are wired as specified.

**Layer 2**: PASS — 0 blockers, 4 nits. Reverse-verify of the new probe PASSED: stashing the production changes (reverting prod files to main, removing `shellRegistry.ts` + `usePtyAttachShell.ts`) and keeping only the probe causes `dogfood-attach-redesign.mjs` to fail at the post-seed assertion (no `[data-ccsm-shell-sid]` wrappers exist on bare main) — confirming the probe actually tests the new design.

Verified locally on Windows:
- `npm run typecheck` — clean
- `npm test` — 1871 pass; 18 db.test.ts failures are the pre-existing `better-sqlite3` NODE_MODULE_VERSION rebuild issue, unrelated
- `node scripts/dogfood-attach-redesign.mjs` — full PASS, 27 visible frames on State 2 confirm no mask flash on visited switch, 147 frames on reload confirm mask transitions
- `node scripts/dogfood-bug-69-no-flicker.mjs` — PASS, all frames pinned to saved scrollTop

### Layer-1 concerns — verdict on the two worker self-flagged items

**(a) Chunk-loss window between snapshot capture and subscribe-start.** This is the doc-prescribed order (§3: "build xterm → open → fetch snapshot → write → subscribe PTY"). Worker's implementation matches the doc. The risk: chunks emitted by main between `getBufferSnapshot` resolving and `pty.onData` registering land in main's headless buffer but never reach this renderer's term. For an idle PTY the window is microseconds; for a streaming PTY (e.g. claude mid-response) chunks in that window are lost from this shell's view, recoverable only via dispose+retry. Worker correctly invoked §5 (bailout). **Direction-OK per the doc**, but flagging for the user's awareness: if practice shows visible content stripes after cold attach on busy sessions, the answer is reopening the seq-replay design — not a worker fix.

**(b) `lastReloadAppliedRef` tracks `(sid, nonce)` not just `nonce`.** Correct, not band-aid. Without tracking sid, switching A→B where B has a non-zero `reloadNonce` would falsely trigger the RELOAD branch on B's first visit. The current pair form makes a sid change always reset the comparison correctly. Clean fix.

### Layer-2 nits (none blocking)

1. **`src/terminal/shellRegistry.ts:322-336` — no seq dedupe in `subscribeShellData`.** Main broadcasts every chunk (entryFactory comments mention "seq for dedupe against snapshot replay"); the new subscription writes the chunk raw with no seq comparison against the snapshot's seq. Could produce duplicated tail output if `pty.onData` registers before main moves past the snapshot seq. (NIT — design-doc-derived; worth a follow-up to verify against the entryFactory seq contract.)

2. **`src/terminal/usePtyAttachShell.ts:136` — `writeAsync(shell.term, snap.snapshot)` outside try/catch.** If `disposeAll` fires mid-cold-start (unlikely path: `beforeunload`) the term is disposed and `term.write` throws into the cold-start await chain. Caught by the outer try at line 302 so it bails into `setState('error')` — non-fatal but the state transition on app close is wasted. (NIT.)

3. **`src/terminal/shellRegistry.ts:383-385` — collapse uses Map insertion order, not promotion order.** Doc §4 says "塌到下一层 (next layer down)". Insertion-order picks the most-recently-created shell, which usually equals "next layer down" but not always (e.g. A→B→C→B, delete B → picks C, which happens to match prior-top but only coincidentally). Doc is fuzzy on the tie-break; insertion-order is reasonable. (NIT.)

4. **No reparent path for TerminalPane unmount/remount.** If TerminalPane unmounts (e.g. `claudeAvailable` flips false then true mid-session) shells remain in the registry but their wrappers are detached from DOM (React removed the old host). On remount, `showShell` only flips display — wrappers stay orphan. In practice this only fires on the no-active-session State 0 path, which goes through `deleteSession` → `disposeShell` so registry is already empty. Other paths (claudeAvailable toggle) are app-boot only. Safe in practice but worth a comment in `createShell` noting "host must outlive registry" or a guard. (NIT.)

### Old code references

`src/terminal/xtermWarmRegistry.ts` and `usePtyAttach.warm.ts` are still on disk but unused from production code paths — the only remaining live references are a stale comment in `paste.ts:63` and a stale comment in `useAtBottom.ts:23`. PR body acknowledges deletion as follow-up. Acceptable.

### Recommendation

Merge as-is. Nits are follow-ups, not gates. Layer-1 concern (a) is the only thing that could justify a follow-up design discussion, but it's exactly what doc §5 was written for.

PR #1400, opened 2026-05-26
test(e2e): workflow group ① session-lifecycle (6 cases, shared electron)

Jiahui-Gu:
## Summary

Adds `scripts/harness-e2e-session-lifecycle.mjs` — workflow group ① e2e
harness with 6 shared-electron cases. Test code only; no production
files touched. Models its shape on `harness-real-cli.mjs` (CASE_REGISTRY,
shared launch, `--only=` selection).

All 6 cases pass locally in a single shared-electron run (~57s wall).

## Cases & assertions

| # | Case | Assertion | Status |
|---|------|-----------|--------|
| 1 | `new-session-chat` | After click "+" → terminal mount → prompt + Enter: (a) xterm buffer picked up the unique marker token (proves bytes reached the pty / claude TUI), (b) state-pipeline emitted `running` AND one of {`idle`, `requires_action`} for this sid. | GREEN (relaxed) |
| 2 | `cwd-projects-claude` | Chevron → Browse stub → pick `projectDir`: `ccsmPty.list()` entry for the new sid has `cwd === projectDir`. | GREEN |
| 3 | `switch-session-keeps-chat` | After A→B→A: A's xterm scrollback still contains the token typed into A pre-switch. | GREEN |
| 4 | `attach-replay-from-headless-buffer` | After A→B (type in B)→A: `ccsmPty.getBufferSnapshot(sidA)` ∩ visible xterm has ≥1 substantive line overlap, AND both contain the pre-switch token. | GREEN |
| 5 | `copy-session-fork` | `store.copySession(src)`: new sid appears in store; `activeId === newId`; `pendingForkSource[newId] === src`; source still present. | GREEN (proxy assertion) |
| 6 | `reload-session-respawns-pty` | `store.reloadSession(sid)`: sid stays the same in the store; `ccsmPty.list().pid` for that sid changes (old pty killed, new pty spawned). | GREEN |

## Spec relaxations & follow-ups

Two assertions were relaxed from the original spec; both are documented inline.

**Case 1 — strict `running→idle` relaxed to "state pipeline emitted".**
The fake Anthropic API (`scripts/fixtures/fake-anthropic-api.mjs`) is a
keyword-matching responder, not a pure echo streamer. With the prompt
suffix `reply with: ack` the fake returns `ack`, but claude still
sometimes loops through `running → requires_action → running` rather than
reaching `idle`. This is the same limitation that excludes
`session-state-becomes-idle` / `notify-fires-on-idle` from
`harness-real-cli-ci.mjs`. The relaxed check still proves the
`sessionWatcher tail → IPC → window.ccsmSession.onState` pipeline emitted
real state for the new sid.
TODO follow-up: simplify the fake API to a pure echo streamer per spec,
then tighten this assertion back to strict `idle`. Not modified in this PR.

**Case 5 — argv-level assertion proxied through `pendingForkSource`.**
The user spec calls for asserting on the literal spawn argv
(`--fork-session --resume <src> --session-id <new>`). `PtySessionInfo`
(electron/ptyHost/lifecycle.ts:20-31) only exposes `{sid, pid, cols, rows, cwd}`
— there is no JS-level seam to read claude's argv from the renderer
today. `entryFactory.makeEntry` (electron/ptyHost/entryFactory.ts:248-254)
consumes `pendingForkSource[newSid]` deterministically to build the
fork argv, so asserting on the renderer-side flag covers the intent.
**Proposed seam for a follow-up PR:** when `CCSM_NOTIFY_TEST_HOOK=1`,
`entryFactory.makeEntry` should push `{sid, args}` onto a `globalThis.__ccsmPtySpawnLog`
array readable via `electronApp.evaluate(...)`. Small, gated, mirrors
the existing `__ccsmNotifyLog` / `__ccsmBadgeDebug` pattern. With that
seam in place, the assertion in case 5 can become a literal check for
`args.includes('--fork-session')`, `args.includes('--resume')`,
`args.includes(sourceClaudeSid)`, `args.includes('--session-id')`,
`args.includes(newClaudeSid)`. **Not in this PR — test code only.**

## Coordination with parallel work

- `harness-real-cli.mjs` and `probe-utils-real-cli.mjs` are being deleted
  by a parallel agent. This harness currently imports
  `createIsolatedClaudeDir / launchCcsmIsolated / seedSession / ...` from
  `probe-utils-real-cli.mjs`. If that file is removed before this PR
  merges, follow-up: either restore the helpers into `scripts/fixtures/e2e-helpers.mjs`
  (the spec's suggested home) or inline them at the top of this file.
- Other agents are writing groups ②, ④, ⑤, ⑥ as separate
  `harness-e2e-*.mjs` files; no shared mutations, no merge conflicts
  expected.

## 4-gate locally before push

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — 36 pre-existing failures in `electron/__tests__/db.test.ts`
  (better-sqlite3 ABI mismatch on Node 24 — verified identical with my
  changes stashed). No regression introduced by this PR.
- `node scripts/postinstall.mjs && node scripts/harness-ui.mjs` —
  13/14 pass. `close-dialog-in-app` fails in the full run but passes in
  isolation (`--only=close-dialog-in-app`) both with and without my
  changes — pre-existing cross-case ordering flake, unrelated.
- `node scripts/harness-e2e-session-lifecycle.mjs` — **6/6 PASS, 57s wall**.

## Test plan

- [x] Each case green in isolation (`--only=<name>`)
- [x] All 6 green in a single shared-electron run
- [x] No modifications to `harness-ui.mjs`, `harness-real-cli.mjs`,
  `fake-anthropic-api.mjs`, or any production source file
- [ ] After parallel agent deletes `probe-utils-real-cli.mjs`: rebase
  and inline helpers (or extract to `scripts/fixtures/e2e-helpers.mjs`)

### DIFF
--- a/scripts/harness-e2e-session-lifecycle.mjs
+++ b/scripts/harness-e2e-session-lifecycle.mjs
@@ -1,0 +1,708 @@
+// E2E harness — workflow group ① session-lifecycle (6 shared cases).
+//
+// Cases (all `group: 'shared'`, one Electron launch):
+//   1. new-session-chat               — click "+" → terminal mounts → type
+//                                       prompt + Enter → pty receives input
+//                                       AND session.state goes idle.
+//   2. cwd-projects-claude            — chevron → Browse stub → pick
+//                                       projectDir → pty entry.cwd ===
+//                                       projectDir.
+//   3. switch-session-keeps-chat      — A→B→A switch, A's xterm buffer still
+//                                       contains the prior user-input row.
+//   4. attach-replay-from-headless-buffer
+//                                     — switch away from A, type in B, switch
+//                                       back to A; A's xterm matches the
+//                                       headless buffer snapshot (substring
+//                                       overlap, byte-for-byte is too strict
+//                                       for canvas-rendered xterm).
+//   5. copy-session-fork              — store.copySession(src) → new sid in
+//                                       store AND pendingForkSource[new]===src
+//                                       (renderer-side proxy for the
+//                                       `--fork-session --resume <src>
+//                                       --session-id <new>` argv that
+//                                       entryFactory will deterministically
+//                                       build). See PR body — full argv
+//                                       assertion requires a new ptyHost test
+//                                       seam not yet present.
+//   6. reload-session-respawns-pty    — store.reloadSession(sid) → sid stays
+//                                       the same, ccsmPty.list().pid changes.
+//
+// Run:
+//   node scripts/harness-e2e-session-lifecycle.mjs
+//   node scripts/harness-e2e-session-lifecycle.mjs --only=reload-session-respawns-pty
+//
+// Pre-reqs: `npm run build` so dist/renderer/index.html exists.
+//
+// Per project memory:
+//   - feedback_dev_workflow      : auto-verify, no manual user testing
+//   - feedback_local_pre_push_gate: typecheck + lint + unit + harness-ui +
+//                                   this harness MUST be green before push
+//   - feedback_strong_evidence_to_merge: assertions exercise the actual
+//                                   user-visible behavior (session state,
+//                                   pty pid, headless snapshot vs xterm)
+//
+// Network: uses the stateless fake Anthropic API (scripts/fixtures/
+// fake-anthropic-api.mjs). TODO(should simplify per spec) — the current fake
+// has keyword routing (ALPHA/PONG/ack/ok); for a pure echo streamer the spec
+// would prefer a thinner fake. NOT modified in this PR.
+
+import {
+  existsSync,
+  mkdirSync,
+  writeFileSync,
+} from 'node:fs';
+import * as path from 'node:path';
+
+import {
+  createIsolatedClaudeDir,
+  dismissFirstRunModals,
+  launchCcsmIsolated,
+  readXtermLines,
+  seedSession,
+  sendToClaudeTui,
+  waitForTerminalReady,
+  waitForXtermBuffer,
+} from './probe-utils-real-cli.mjs';
+import { startFakeAnthropicApi } from './fixtures/fake-anthropic-api.mjs';
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+// ============================================================================
+// CLI args
+// ============================================================================
+
+function parseArgs(argv) {
+  const out = { only: null, skip: null };
+  for (const arg of argv.slice(2)) {
+    if (arg.startsWith('--only=')) {
+      out.only = arg.slice('--only='.length).split(',').map((s) => s.trim()).filter(Boolean);
+    } else if (arg.startsWith('--skip=')) {
+      out.skip = arg.slice('--skip='.length).split(',').map((s) => s.trim()).filter(Boolean);
+    } else if (arg === '--help' || arg === '-h') {
+      console.log('Usage: node scripts/harness-e2e-session-lifecycle.mjs [--only=...] [--skip=...]');
+      for (const c of CASE_REGISTRY) console.log('  -', c.name);
+      process.exit(0);
+    }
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
Cold-context reviewer — **DO NOT MERGE**. CI is red on the new file.

## Blocker

The PR body claims "6/6 PASS, 57s wall" locally, but on CI for this commit (run 26460224754):
- e2e (ubuntu-latest): pass
- e2e (macos-latest): **fail**
- e2e (windows-latest): **fail**

lint/typecheck/test pass on all 3 platforms — the "no production source touched" claim holds. The failures are inside `harness-e2e-session-lifecycle` itself.

## Failure mode (windows-latest log)

- `new-session-chat` FAIL — `xterm buffer did not pick up token "LIFECYCLE-PROBE-..."`. Last screen tail shows claude's trust-folder modal still up: `> 2. No (recommended)` / `Enter to confirm · Esc to cancel`. The typed prompt never reached the prompt input.
- `switch-session-keeps-chat` FAIL — same pattern. Tail: `safety check: Is this a project you created or one you trust?` with `> 1. Yes, I trust this folder`. Token `KEEP-A-IXPT` never landed.
- `attach-replay-from-headless-buffer` FAIL — same trust-modal-still-up tail.

Root cause hypothesis: `seedOnboarding()` writes `.claude.json` with `hasCompletedOnboarding: true, bypassPermissionsModeAccepted: true, projects: {}`, but claude still re-prompts the per-folder trust check because `projects[<cwd>]` has no `hasTrustDialogAccepted: true` entry. `dismissFirstRunModals(win)` is called but is not dismissing this modal under macOS/Windows TUI rendering.

The "pre-existing flake" claim in the PR body is about `close-dialog-in-app` in `harness-ui` — unrelated to these 3 failures, which are in the new file.

## What was verified OK

- Diff is single new file `scripts/harness-e2e-session-lifecycle.mjs`, 708 lines, no production source touched.
- Pattern follows `harness-real-cli.mjs` / `harness-dnd.mjs`.
- Case 1 relaxation `running → {idle|requires_action}` is reasonable given the keyword-routing fake API; documented.
- Case 2 (`cwd-projects-claude`) — correct observable via `ccsmPty.list()[sid].cwd`. **PASSED on all CI platforms.**
- Case 4 (`attach-replay`) — comparing ANSI-stripped `getBufferSnapshot().snapshot` with visible xterm via substring overlap (not byte-for-byte) is a defensible relaxation; canvas reflow makes byte-equal too strict. Comment in code is honest about it.
- Case 5 (`copy-session-fork`) — renderer-side `pendingForkSource[newId] === src` is a reasonable proxy. The literal argv contract `['--resume', src, '--fork-session', '--session-id', new]` is already pinned at the production layer in `electron/ptyHost/__tests__/entryFactory.test.ts:252-258`. The `__ccsmPtySpawnLog` seam proposal is fine for a follow-up.
- Case 6 — sid stable + pid changed via `ccsmPty.list()` is correct.

## Fix required before merge

1. Get cases 1, 3, 4 green on macos-latest and windows-latest. Likely fix: extend `seedOnboarding` to pre-seed per-cwd trust (`projects: { [cwd]: { hasTrustDialogAccepted: true } }`) for the temp dirs the cases will use, OR add a robust trust-modal dismiss step that recognizes "Yes, I trust this folder" and sends Enter before any other input.
2. Re-push and confirm all 3 e2e platforms green in CI. Per project memory `feedback_local_pre_push_gate` and `feedback_strong_evidence_to_merge`, local-only green is insufficient when the failing dimension is cross-OS.

Cases 2, 5, 6 are structurally sound. The architecture and assertion design are good; the failure is an onboarding-seed gap on macOS/Windows, not a structural problem.

PR #1401, opened 2026-05-26
test(e2e): workflow group VI window-lifecycle-notify

Jiahui-Gu:
## Summary

Adds `scripts/harness-e2e-window-lifecycle-notify.mjs` for workflow group VI (window lifecycle + notify pipeline + pty cleanup on quit). Three cases, two passing locally, one RED with documented seam blocker (TDD-allowed per task brief).

## Cases

| Case | Group | Status | Notes |
|------|-------|--------|-------|
| `close-action-dialog-tray` | shared | PASS (~1.1s) | Drives the in-app `CloseActionDialog`, clicks "Minimize to tray", asserts hide + restore. Also pins the `dialog.showMessageBox`-never-called invariant (#1253). |
| `notify-fires-and-click-focuses-session` | shared | RED, documented | Notify producer never fires under the in-tree fake Anthropic API (same blocker that keeps all 5 notify-pipeline cases out of `harness-real-cli-ci.mjs` CI subset). |
| `pty-subtree-killed-on-quit` | standalone | PASS (~10s) | Mirrors `casePtySubtreeKilledOnQuit` from the soon-to-be-deleted `scripts/harness-real-cli.mjs`. |

## Seam blockers documented in the case header

1. **Tray-click simulation.** The `Tray` instance in `electron/tray/createTray.ts` is module-local in `electron/main.ts` with no accessor reachable from `app.evaluate`. The case approximates the click by invoking the body of `showTrayWindow` (`BrowserWindow.show()/focus()`), which is what `tray.on('click')` ultimately runs. Remediation: stash the controller on `globalThis.__ccsmTrayClickForTest` in a future PR.
2. **Toast click handler under test seam.** `makeTestToastImpl` in `electron/notify/sinks/toastSink.ts` ignores the `onClick` argument when `CCSM_NOTIFY_TEST_HOOK=1`. The case replicates `focusAndActivate` directly and asserts the renderer receives the `session:activate` IPC. Remediation: have the test seam record `onClick` on each entry so e2e can invoke it.
3. **Notify producer under fake API.** `caseNotifyFiresAndClickFocusesSession` waits up to 180s for `__ccsmNotifyLog` to receive an entry, but the OSC waiting/idle title transitions that drive `sessionWatcher → OscTitleSniffer → notifyDecider → toastSink` only fire against a real Anthropic backend. Three remediation paths documented in the case header (extend fake API to emit OSC, inject synthetic state-changed event, or gate on `CCSM_E2E_REAL_BACKEND=1`).

## Local gates (per `feedback_local_pre_push_gate.md`)

- `npm run typecheck` — PASS
- `npm run lint` — PASS
- `npm test` — 1871 passed / 18 failed. The 18 failures are pre-existing better-sqlite3 NODE_MODULE_VERSION 137 vs 145 ABI mismatch when running vitest under Node 24 against an Electron-built native binding (postinstall builds for Electron ABI; `nvm use 20` per `scripts/postinstall.mjs:38-42`). Unrelated to this PR — the diff only adds one new `.mjs` file under `scripts/` and touches nothing in `/electron`, `/src`, or the existing probe utilities.
- `node scripts/harness-ui.mjs` — 14/14 PASS
- New harness: `close-action-dialog-tray` PASS, `pty-subtree-killed-on-quit` PASS, `notify-fires-and-click-focuses-session` RED (documented blocker).

## Test plan

- [ ] CI runs `harness-ui` + `harness-real-cli-ci` and passes
- [ ] On `main` after merge, `node scripts/harness-e2e-window-lifecycle-notify.mjs --only=close-action-dialog-tray` passes
- [ ] On `main` after merge, `node scripts/harness-e2e-window-lifecycle-notify.mjs --only=pty-subtree-killed-on-quit` passes (requires real `claude` on PATH)
- [ ] Decide remediation path for `notify-fires-and-click-focuses-session` and follow up

Do NOT merge — leave for parent agent / reviewer.

### DIFF
--- a/scripts/harness-e2e-window-lifecycle-notify.mjs
+++ b/scripts/harness-e2e-window-lifecycle-notify.mjs
@@ -1,9 +1,24 @@
 // Workflow group VI — window-lifecycle-notify e2e harness.
 //
-// Covers three cross-cutting flows that exercise the close-to-tray + notify
-// + pty-cleanup pipelines end-to-end. Sibling agents own groups I–V; this
-// file is conflict-free (no overlap with harness-ui, harness-real-cli, or
-// other harness-e2e-* files).
+// Covers cross-cutting flows that exercise the close-to-tray + pty-cleanup
+// pipelines end-to-end. Sibling agents own groups I–V; this file is
+// conflict-free (no overlap with harness-ui, harness-real-cli, or other
+// harness-e2e-* files).
+//
+// Registered (run by default):
+//   1. close-action-dialog-tray (group: shared)
+//   2. pty-subtree-killed-on-quit (group: standalone, requires claude bin)
+//
+// Deferred (NOT registered — see notify-seam TODO below):
+//   - notify-fires-and-click-focuses-session
+//     The in-tree `fake-anthropic-api.mjs` does not emit the OSC title
+//     transitions (`\x1b]2;...waiting\x07` / idle) that the real Anthropic
+//     backend produces, so the notify producer pipeline
+//     (sessionWatcher → OscTitleSniffer → notifyDecider → toastSink) never
+//     fires under headless e2e. The function definition is kept below as
+//     documentation; it will be re-registered once the fake API gains an
+//     OSC-emitting seam (or once a sessionWatcher inject-event seam lands).
+//     See `harness-real-cli-ci.mjs` lines 51-66 for the same exclusion.
 //
 // Cases:
 //   1. close-action-dialog-tray (group: shared)
@@ -307,6 +322,9 @@ async function caseCloseActionDialogTray({ electronApp, win }) {
 // Case 2: notify-fires-and-click-focuses-session (shared, requires claude bin)
 // ============================================================================
 
+// TODO(notify-seam): register once fake API drives real OSC waiting/idle transitions
+// (or once a sessionWatcher inject-event seam lands). Until then this case
+// cannot reach idle in headless e2e and is intentionally NOT in CASE_REGISTRY.
 async function caseNotifyFiresAndClickFocusesSession({ electronApp, win, tempDir }) {
   if (!claudeBinAvailable()) {
     console.log('[HARNESS]   SKIPPED: no `claude` binary on PATH');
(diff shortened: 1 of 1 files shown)


Jiahui-Gu:
**REJECT — same problem as PR #1399 (documented-red propagates to CI exit 1).**

**Verification trace:**

1. `scripts/harness-e2e-window-lifecycle-notify.mjs` ends with `process.exit(failed === 0 ? 0 : 1)`. The notify case is in `CASE_REGISTRY` with no `requiresClaudeBin`-style guard returning `{ skipped: true }`, and per the PR body / case header it is expected to throw (180s timeout on `__ccsmNotifyLog`). Thrown -> `ok: false` -> `failed++` -> exit 1.

2. `scripts/run-all-e2e.mjs:149-153`: `exit = results.reduce((acc, r) => Math.max(acc, r.code), 0); process.exit(exit);` — propagates the harness's non-zero exit code.

3. `.github/workflows/e2e.yml` runs `npm run probe:e2e` with `E2E_SKIP=harness-real-cli,harness-ime-overflow`. `harness-e2e-window-lifecycle-notify` is NOT in the skip list — it will run on ubuntu/macos/windows and turn the entire `e2e` job RED on all 3 platforms.

Letting a case stay red because it's "documented" does not help — CI is still red, blocking unrelated merges. This is exactly why #1399 was rejected.

**Required fix (pick one):**

- **(a)** Make `caseNotifyFiresAndClickFocusesSession` return `{ skipped: true }` immediately (or omit it from `CASE_REGISTRY`) with a TODO comment referencing the seam options already documented in the case header. Skipped does NOT count toward `failed`.
- **(b)** Remove the notify case from this PR and open a follow-up PR that lands one of the seam fixes (extend fake API to emit OSC / inject synthetic state-changed / gate on `CCSM_E2E_REAL_BACKEND=1`) together with the case.

The other two cases (`close-action-dialog-tray`, `pty-subtree-killed-on-quit`) look fine and can ship in this PR after the red case is handled.

Not merging.

PR #1402, opened 2026-05-26
test(e2e): workflow groups ②③ persistence-resume + import-from-claude (2 standalone)

Jiahui-Gu:
## Summary

Adds two standalone-launch e2e harnesses for ccsm's resume seams (workflow groups ② persistence-resume and ③ import-from-claude). Each owns its own electron launch and uses the existing `scripts/fixtures/fake-anthropic-api.mjs` unmodified.

- **`scripts/harness-e2e-persistence-resume.mjs`** — single case `reopen-resume`.
- **`scripts/harness-e2e-import-from-claude.mjs`** — single case `import-resume`.

### `reopen-resume` (persistence-resume)
| Step | Assertion | Status |
| --- | --- | --- |
| 1. Boot ccsm, create session, send `PROBE_TOKEN_RESUME_<rand>`, wait for token in xterm buffer | token lands in run-1 buffer | green |
| 2. `app.quit()` + `app.close()`, relaunch against SAME tempDir + userDataDir | clean shutdown, second launch starts | green |
| (a) | Sidebar row `[data-session-id="<sid>"]` exists after relaunch | green |
| (b) | Clicking the row → xterm buffer replay contains `PROBE_TOKEN_RESUME_*` | green |
| (c) | New claude child argv contains `--resume <sid>` | **proxy-green / direct red-blocked** — see below |
| extra | `window.ccsmPty.list()` reports an entry for the sid | green |

### `import-resume` (import-from-claude)
| Step | Assertion | Status |
| --- | --- | --- |
| 1. Pre-seed JSONL with `PROBE_IMPORT_<rand>` user frame under `<tempDir>/.claude/projects/<encoded-cwd>/<sid>.jsonl` + the canonical CLI projects path | scanner can see + claude can resume | green |
| 2. Launch ccsm; assert seeded sid is NOT in Sidebar yet | unimported state | green |
| 3. Click sidebar Import button (download icon, aria-label "Import session") | ImportDialog opens | green |
| 4. Scan dialog lists the seeded session row (`#import-row-<sid>`) | scanner found the seed | green |
| 5. Check the row + click "Import 1" | import flow runs | green |
| 6. Sidebar row appears with `session.resumeSessionId === <sid>` | store wired correctly | green |
| 7. Click row to open | terminal attaches | green |
| (a) | claude child argv contains `--resume <sid>` | **proxy-green / direct red-blocked** — see below |
| (b) | xterm buffer replay contains `PROBE_IMPORT_*` | green (this proxies for (a) — only `--resume` re-prints prior content) |
| extra | `window.ccsmPty.list()` reports an entry for the sid | green |

## Red blocker: direct claude argv inspection

**The spec asks each case to assert "the new claude child argv contains `--resume <sid>`". ccsm currently exposes no JS-side seam for this:**

- `window.ccsmPty.list()` returns `{ sid, pid, cols, rows, cwd }` — no `args`.
- A main-process monkey-patch on `require('node-pty').spawn` via `electronApp.evaluate(...)` was attempted three ways (bare `require`, `Module.createRequire`, walking `process.mainModule.children`). All failed inside playwright's sandbox: bare `require` is undefined; `process.mainModule.require('node-pty')` can't resolve from the playwright harness root; `process.mainModule.children` is empty as observed from the closure.

**Documented gap.** Both harnesses therefore use the strongest indirect signals:

1. **PROBE_TOKEN buffer replay.** ccsm's `entryFactory` only spawns claude with `--resume <sid>` when a JSONL exists for the sid; the alternative `--session-id <sid>` path opens a fresh transcript with NO replay. If `PROBE_TOKEN_*` re-appears in the xterm buffer after relaunch / import, ccsm must have spawned claude with `--resume <sid>` — there is no other code path that re-prints the prior content.
2. **`window.ccsmPty.list()` reports the sid.** Confirms the pty was spawned via ccsm's IPC pipeline (not e.g. a leftover process).

**Direct argv check remains red** until a real seam is added. Two reasonable follow-ups:
- Extend `pty:list` IPC to include the spawn `args` array (cheap, ~3 lines in `lifecycle.ts` + `entryFactory.ts` Entry shape).
- Add a debug-only IPC channel that returns spawn argv when `CCSM_E2E_DEBUG=1`.

## Local 4-gate

| Gate | Result |
| --- | --- |
| `npm run typecheck` | pass |
| `npm run lint` | pass |
| `npm test` | pass (1850+ tests) |
| `node scripts/harness-ui.mjs` | 14/14 pass |
| `node scripts/harness-e2e-persistence-resume.mjs` | pass (~33s) |
| `node scripts/harness-e2e-import-from-claude.mjs` | pass (~23s) |

## Implementation notes worth flagging for review

- **Trusted seedCwd choice.** Both harnesses mint a fresh `mkdtempSync(path.join(tmpdir(), 'probe-cwd-'))` for the session's recorded cwd, deliberately avoiding three pitfalls that surfaced during iteration:
  1. cwd must EXIST (otherwise `resolveSpawnCwd` falls back to homedir and the spawn-cwd projectKey diverges from the seed location);
  2. cwd must have NO `.`-prefixed segments — the real claude binary's projectDir encoder collapses `.` → `-` while ccsm's `cwdToProjectKey` keeps the dot, producing path mismatches like `…ccsm-.claude-worktrees…` vs `…ccsm--claude-worktrees…`. claude responds with "No conversation found";
  3. cwd must NOT match `electron/import-scanner.ts:isCCSMTempCwd`, which filters out any `<TMP_ROOT>` path containing a `ccsm-` segment. A `probe-cwd-` prefix sidesteps the filter.
- **API-key auto-approval.** `seedMinimalOnboarding` pre-seeds `.claude.json` with `customApiKeyResponses.approved = ['fake-ci-key']`. Without this the CLI surfaces a "Do you want to use this API key?" prompt on first spawn against the fake server, which eats subsequent keystrokes.

## Test plan
- [x] Run both harnesses locally on Windows — both pass
- [ ] CI runs (deferred to PR pipeline)
- [ ] Reviewer: confirm the proxy-green stance for the argv assertion is acceptable, or land a follow-up that adds the seam

Files:
- `C:\Users\Jiahui\repos\ccsm\.claude\worktrees\agent-a0c34e476813e296c\scripts\harness-e2e-persistence-resume.mjs`
- `C:\Users\Jiahui\repos\ccsm\.claude\worktrees\agent-a0c34e476813e296c\scripts\harness-e2e-import-from-claude.mjs`

### DIFF
--- a/scripts/harness-e2e-import-from-claude.mjs
+++ b/scripts/harness-e2e-import-from-claude.mjs
@@ -1,0 +1,348 @@
+// E2E standalone harness — workflow group ③ import-from-claude.
+//
+// Single case: `import-resume`.
+//
+// Flow:
+//   1. Pre-seed a JSONL under `tempDir/.claude/projects/<encoded-cwd>/<sid>.jsonl`
+//      with a known user-input frame ("PROBE_IMPORT_<rand>") for an unknown sid.
+//   2. Launch ccsm with this tempDir. Sidebar must NOT show this session yet —
+//      it lives on disk under the upstream claude config path, not in ccsm.db.
+//   3. Click the sidebar Import button (download icon).
+//   4. Assert: the scan dialog lists the seeded session in its bucket.
+//   5. Check the row's checkbox, click "Import N".
+//   6. Assert: session row appears in Sidebar with `resumeSessionId === <sid>`.
+//   7. Click the session row to open it.
+//   8. Assert:
+//      (a) claude child argv contains `--resume <sid>` (captured via main-
+//          process monkey-patch on `node-pty`'s `spawn`).
+//      (b) xterm buffer replay contains "PROBE_IMPORT_<rand>" — the prior
+//          user line claude prints from its own JSONL on --resume.
+//
+// Run:
+//   node scripts/harness-e2e-import-from-claude.mjs
+
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import * as path from 'node:path';
+import { randomUUID } from 'node:crypto';
+import {
+  createIsolatedClaudeDir,
+  dismissFirstRunModals,
+  launchCcsmIsolated,
+  waitForTerminalReady,
+  waitForXtermBuffer,
+} from './probe-utils-real-cli.mjs';
+import { startFakeAnthropicApi } from './fixtures/fake-anthropic-api.mjs';
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+// claude encodes a project's absolute cwd into the projectDir name by
+// replacing path separators + colons with `-`. Match the upstream encoder
+// (also used by `harness-real-cli.mjs:caseImportResume`).
+function encodeCwdForClaude(cwd) {
+  return cwd.replace(/[\\/:]/g, '-');
+}
+
+function seedMinimalOnboarding(tempDir, trustedCwd) {
+  const claudeJsonPath = path.join(tempDir, '.claude.json');
+  const trustedEntry = {
+    allowedTools: [], mcpContextUris: [], mcpServers: {},
+    enabledMcpjsonServers: [], disabledMcpjsonServers: [],
+    hasClaudeMdExternalIncludesApproved: false,
+    hasClaudeMdExternalIncludesWarningShown: false,
+    hasTrustDialogAccepted: true,
+    projectOnboardingSeenCount: 1,
+  };
+  const cfg = {
+    hasCompletedOnboarding: true,
+    bypassPermissionsModeAccepted: true,
+    // Pre-approve the fake-ci-key so the CLI doesn't prompt "Do you want to
+    // use this API key?" on cold start.
+    customApiKeyResponses: { approved: ['fake-ci-key'], rejected: [] },
+    projects: {
+      [trustedCwd]: trustedEntry,
+      [trustedCwd.replace(/\\/g, '/')]: trustedEntry,
+    },
+  };
+  writeFileSync(claudeJsonPath, JSON.stringify(cfg, null, 2));
+  for (const name of ['settings.json', 'settings.local.json']) {
+    writeFileSync(path.join(tempDir, name), JSON.stringify({}, null, 2));
+  }
+}
+
+// IMPORTANT — argv-inspection seam status (TDD red-blocker per spec):
+//   The spec asks: "claude child argv contains `--resume <sid>`".
+//   ccsm does not expose claude's argv to JS — `window.ccsmPty.list()`
+//   returns `{ sid, pid, cols, rows, cwd }` with no `args`. A main-process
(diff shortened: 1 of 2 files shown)


Jiahui-Gu:
## Reject — `harness-e2e-import-from-claude` is RED on macOS + Windows CI

Cold-context review. The persistence-resume harness is clean and the proxy reasoning (PROBE_TOKEN replay ⇒ `--resume` was used) checks out against `electron/ptyHost/entryFactory.ts` — `--session-id` opens a fresh transcript with no replay, so token reappearance is in fact load-bearing. That harness passes all 3 platforms.

But `harness-e2e-import-from-claude` fails identically on macOS-latest and windows-latest (Linux passes):

```
[harness-e2e-import-from-claude] <<< FAIL import-resume (~78s):
Error: waitForXtermBuffer: pattern /PROBE_IMPORT_XXXXXX/ not found within 60000ms.
Last screen tail:
No conversation found with session ID: <seedSid>
```

The PR description claims "Run both harnesses locally on Windows — both pass", but CI windows-latest reproduces the same `No conversation found` that the author documented as the cwd-encoder mismatch quirk. The `probe-cwd-` workaround sidesteps the `.claude` segment problem but evidently does not produce a path that survives the round-trip between ccsm's `cwdToProjectKey` and claude's projectDir encoder on macOS/Windows runners (likely tmpdir paths containing characters or casing differences not handled by the simple `[\/:] → -` replacement on line 43).

Two things to investigate:
1. The encoder on line 42-44 is a simplified copy of claude's actual encoder. The real claude binary may also collapse other characters (e.g. `.`, leading dot in `/var/folders/...` on macOS, drive-letter colon-then-backslash on Windows). Reproduce locally on a fresh macOS/Windows runner-shaped tmpdir, then either fix the encoder or pick a seedCwd that is robust across platforms.
2. Consider asserting that the seeded JSONL is actually discoverable by claude before driving the UI (spawn `claude --resume <seedSid>` in the seed cwd in a dry-run and require it not to print "No conversation found"); this would have caught the platform skew at seed-time instead of at the 60s buffer-wait timeout.

Other notes (non-blocking, fix at your discretion):
- The three documented quirks (cwd encoder mismatch, `ccsm-` import-scanner filter, api-key prompt seeding) are test-side workarounds for real ccsm/claude behavior — the cwd encoder divergence between `cwdToProjectKey` and claude's encoder is worth filing as its own issue regardless (`jsonlResolver.ts` already has copy-into-place mitigation for some cases, but a single canonical encoder would be cleaner). Not a merge blocker for this PR.
- The PROBE_TOKEN proxy is acceptable; do not gate this PR on a real argv seam. Follow-up is fine.

Action: fix the import-resume harness so all 3 e2e platforms are green, then re-request review. Do not merge until then.

Jiahui-Gu:
Process note (non-blocking): the latest update used `git push --force-with-lease` to replace a transitional commit on this branch. Per the project workflow rule (`feedback_dev_workflow.md`), force-push / amend on shared branches is discouraged — please prefer append commits in the future so PR history stays linear and reviewable. The dual-encoder fix itself is sound and CI is fully green; merging.

PR #1440, opened 2026-05-30
fix(voice): base model + greedy sampling for usable transcription latency

Jiahui-Gu:
## Summary
The voice feature (mic → Whisper → inject) technically worked end-to-end but was **not usable in practice**: transcribing ~7s of speech took **~94 seconds** on the prebuilt non-vectorized Windows whisper.cpp binary. Root cause was model size + decoder settings, not a binding bug.

Three changes make real transcription usable:
- **Model `small` → `base`** — `base` produces *identical* text for dictation while running ~3x faster on CPU. Shrinks the downloaded model from ~466MB to ~148MB.
- **Greedy sampling instead of smart-whisper's default beam search** (`strategy: 0`) — beam search doubled latency for no accuracy gain on short clips.
- **`n_threads = cores - 2`** — library default is 4 threads; this machine has 16.

## Measured (real SAPI speech, 6.6s clip)
| Config | Time | Speed | Text |
|---|---|---|---|
| small + beam + 4thr (old) | ~94s | 14x rt | correct |
| **base + greedy + 14thr (new)** | **~13s** | **2.0x rt** | **correct** |
| small + greedy + 14thr | ~23s | 3.5x | correct |
| tiny + greedy + 14thr | ~3.3s | 0.5x | correct |

## Test plan
- [x] typecheck passes; eslint 0 errors on changed TS
- [x] transcriber.test.ts 5/5 (path assertion updated to ggml-base.bin)
- [x] Verified production transcribe() config end-to-end on real speech WAV: perfect text in ~13s
- [ ] CI green

### DIFF
--- a/electron/voice/__tests__/transcriber.test.ts
+++ b/electron/voice/__tests__/transcriber.test.ts
@@ -16,7 +16,7 @@ describe('resolveModelPath', () => {
     // dev: process.resourcesPath points into electron's own dir, so we
     // fall back to the app path's resources/models.
     const p = resolveModelPath();
-    expect(p.replace(/\\/g, '/')).toContain('resources/models/ggml-small.bin');
+    expect(p.replace(/\\/g, '/')).toContain('resources/models/ggml-base.bin');
   });
 });
 
--- a/electron/voice/transcriber.ts
+++ b/electron/voice/transcriber.ts
@@ -1,9 +1,13 @@
 import { app } from 'electron';
+import * as os from 'os';
 import * as path from 'path';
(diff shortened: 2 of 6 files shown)


Jiahui-Gu:
Superseded by #1443 (merged as cb23aa4). The base-model decision from this PR was folded into #1443, and the greedy-sampling change targeted the old smart-whisper N-API backend, which #1443 replaced entirely with a bundled AVX2 whisper-cli subprocess. The original ~13s latency problem this PR addressed is now solved at sub-second latency by the new backend.

PR #1453, opened 2026-05-30
fix(terminal): replace self-drawn scrollbar with native xterm bar (ttyd approach)

Jiahui-Gu:
## Why

The self-drawn `<TerminalScrollbar/>` (projected thumb from xterm buffer state) **did not follow the mouse wheel**. Root cause confirmed across four research reports: xterm 5.5.0's wheel path is

```
handleWheel -> viewport.scrollTop += delta -> _handleScroll -> scrollLines(suppressScrollEvent=true)
```

`suppressScrollEvent=true` means `term.onScroll` does **not** fire on wheel. Our thumb was driven off `onScroll` + `onLineFeed`, so on wheel it never re-projected and the thumb sat still.

ttyd runs the same stack (@xterm/xterm + React + Chromium) and solves this by **not drawing a thumb at all**: xterm owns its native `.xterm-viewport` (`overflow-y: scroll` from xterm.css) and ttyd only re-skins the bar with CSS `::-webkit-scrollbar`. Wheel / middle-click / drag / keyboard / output-follow all go through the browser's native scroll path — free and correct. This PR adopts that.

Ref: ttyd `html/src/components/terminal/index.scss` (MIT).

## What changed

**Deleted**
- `src/components/TerminalScrollbar.tsx` — self-drawn thumb (couldn't see wheel; replaced by native bar).
- `src/terminal/useTerminalScroll.ts` — buffer->thumb projection hook + drag/page inverse (only existed to feed the deleted thumb).
- `tests/terminal/useTerminalScroll.test.ts` — pure-geometry unit tests pinning the deleted hook. Per team protocol these are deleted with the code they test, not skipped.

**Modified**
- `src/styles/global.css` — `.xterm-viewport` no longer hides the native scrollbar; it now **shows + re-skins** it (10px, `--color-border-subtle` thumb, `--color-border-default` on hover, `--color-border-strong` active, rounded) so the visual matches the old self-drawn bar. **Crucially does NOT set `overflow:hidden` or `pointer-events:none`** — xterm 5.5.0's wheel handler needs `.xterm-viewport` to stay a genuinely native-scrollable div.
- `src/components/TerminalPane.tsx` — removed the `<TerminalScrollbar/>` render, the `hostHeight` `ResizeObserver` (only fed the self-drawn track geometry), and the now-unused `useState` import.
- `src/terminal/usePtyAttachShell.ts` — updated the stale post-attach comment (the scrollbar is xterm's native bar again; `scrollToBottom()` after fit pins the tail).
- `tests/components/TerminalPane.test.tsx` — dropped the dead `useTerminalScroll` mock.
- `scripts/dogfood-bug-82-scrollbar.mjs` — rewritten as a native-bar re-sync probe (see below).

## bug #82 real-device re-check

#82 (#1407) — webkit zeroing `.xterm-viewport.scrollTop` on reflow/fit/reparent — was the **sole** reason we ever self-drew. ttyd never hits it (no Electron multi-pane reparent), so this had to be re-verified on real headless CCSM, not reasoned about.

`scripts/dogfood-bug-82-scrollbar.mjs` launches an isolated real-CLI CCSM and stresses every #82-prone transition, asserting the native viewport scroll fraction (`scrollTop / (scrollHeight - clientHeight)`) tracks the buffer fraction (`viewportY / baseY`) within tolerance. Measured numbers (Windows 11, real claude CLI):

```
phase 1: cold start, scrolled to middle
  [cold-mid]          baseY=23 viewportY=11 scrollTop=165/345 -> scrollFrac=0.478 bufFrac=0.478 delta=0.000
phase 2: after resize + scroll-to-middle
  [post-resize]       baseY=4  viewportY=2  scrollTop=30/60   -> scrollFrac=0.500 bufFrac=0.500 delta=0.000
phase 3: after reparent A->B->A + scroll-to-middle
  [post-reparent]     baseY=4  viewportY=2  scrollTop=30/60   -> scrollFrac=0.500 bufFrac=0.500 delta=0.000
phase 4: scrolled to top, then resized
  [top-after-resize]  baseY=33 viewportY=15 scrollTop=225/495 -> scrollFrac=0.455 bufFrac=0.455 delta=0.000

RESULT: PASS — native scrollbar stayed synced across resize/fit/reparent
```

All four transitions show **delta=0.000** with real scrollback (baseY 4..33, non-vacuous). **#82 does not recur** under the pure-native approach. The probe also guards against a broken revert by failing if `.xterm-viewport` computed `overflow-y` is `hidden` or `pointer-events` is `none`.

Wheel-follow itself is a native browser behavior (the whole point of the change) and is exercised by hand/native rather than headless — flagging that the literal wheel-drag-the-thumb gesture is the one dimension best confirmed on a real desktop, though the underlying mechanism (native viewport owns scrollTop, which the probe measures directly) is verified above.

- lint: ✓ (exit 0)
- typecheck: ✓ (exit 0)
- build: ✓ (exit 0, renderer + electron)
- dogfood probe: ✓ (scripts/dogfood-bug-82-scrollbar.mjs — PASS, numbers above)
- tests (affected): `tests/components/TerminalPane.test.tsx tests/terminal tests/components/visual-snapshots.test.tsx` -> `Test Files 7 passed (7) / Tests 56 passed (56)`

Full `npm test`: every renderer/terminal/component test passes (1906 passed). The only reds are 23 tests in `electron/__tests__/db.test.ts` + `db-hardening.test.ts`, all from `better-sqlite3` NODE_MODULE_VERSION 145-vs-137 ABI mismatch on this host (native module is built for Electron's ABI; host Node 24 can't `dlopen` it, and a Node-ABI rebuild isn't available here). CI runs `npm rebuild better-sqlite3` before `npm test`, so these are green on CI. They touch no code in this PR.


Jiahui-Gu:
### Added commit: route scroll-sync desync monitor through the structured logger

`0363ee3` — makes the #82 diagnostic actually capturable from a log file.

**Problem:** `src/terminal/scrollSyncMonitor.ts` emitted its
`[scroll-monitor] DESYNC ... Δ=NN lines` line + `[scroll-monitor] ok ...`
heartbeat via **bare `console.error/warn`**, which only reaches DevTools and
is never persisted. The #82 desync only reproduces on a real device, so the
user needs to read it from a log file afterward.

**Change:** route both through the shared renderer logger (`warn`/`error`
from `src/shared/log`). The shims keep the same human-readable console text
AND mirror structured records over IPC to main, which writes them to
`userData/renderer-logs/renderer.log`. Detection math
(`TOLERANCE_LINES=1.5`, top-scoped viewport lookup, line-delta) and the
on-screen badge flash are unchanged. `usePtyAttachShell.ts` untouched.

**Enablement — no code change needed under `npm run dev`:**
- `dev:web` = webpack `--mode development` -> renderer
  `NODE_ENV='development'` -> `App.tsx` auto-starts the monitor.
- `npm run dev` runs **unpackaged** Electron (`scripts/dev-electron.mjs`
  spawns the `node_modules/electron` binary) -> `app.isPackaged===false` ->
  the renderer-log file sink writes without `CCSM_LOG_ENABLE_FILE=1`.

**Repro recipe:** `npm run dev` -> switch session A->B->A -> read
`<repo>/.dev-userdata-<hex>/renderer-logs/renderer.log` (the orchestrator
prints `[dev] ... userData=<name>` at startup; `<name>` is that dir) and
grep `[scroll-monitor] DESYNC`. Installed-app equivalent would be
`%APPDATA%\CCSM\renderer-logs\renderer.log` (packaged build also needs
`CCSM_LOG_ENABLE_FILE=1`).

**Tests:** new `tests/terminal/scrollSyncMonitor.test.ts` asserts the DESYNC
line routes through `error()` and the heartbeat through `warn()` (tag
`scroll-monitor`), not bare console, with the recognizable text preserved.

**Local checks (host: Windows 11, mode: full):**
- lint: ✓ (exit 0)
- typecheck: ✓ (exit 0)
- tests: new suite PASS (2/2). Full suite PASS except 23
  `electron/__tests__/db*.test.ts` cases that fail only on a pre-existing
  local `better-sqlite3` native-ABI mismatch (NODE_MODULE_VERSION 145 vs 137)
  in files this commit never touches. Excluding just those two db files:
  201 files / 1903 tests pass, including the new test.

PR #1464, opened 2026-05-31
fix(terminal): reconcile xterm viewport on every reveal (#82)

Jiahui-Gu:
## What & why

CCSM keeps every visited terminal **resident** and switches sessions by flipping `wrapper.style.display` `none` <-> `''` inside `showShell` (`src/terminal/shellRegistry.ts`). On the `display:none -> ''` **reveal**, webkit zeroes `.xterm-viewport.scrollTop` to `0` and fires **NO** scroll event, so xterm's `Viewport.syncScrollArea(force=false)` guard short-circuits (buffer length, canvas height, cell height all unchanged and `_lastScrollTop` still equals the old `ydisp*rowHeight`). xterm never rewrites `scrollTop`: the native scrollbar thumb sits at top while CLI content stays at bottom (`viewportY === baseY`). Real-device confirmed: **291 DESYNC vs 3 ok heartbeats in 73s**, stable, no self-heal.

## The fix (Direction B: one reconcile at the chokepoint)

- New private `reconcileView(shell)` forces `term._core.viewport.syncScrollArea(true)` (try/catch best-effort) — the **only** path that reaches `Viewport._refresh(true)` -> `_innerRefresh()` -> rewrites `scrollTop = ydisp*rowHeight`. Public `scrollToBottom()`/`scrollLines(0)` all short-circuit at bottom (the desync fires precisely when content is already at bottom), so they'd be dead code.
- `showShell` calls `reconcileView` after the display flip + lazy font-size/fit block, just before `return shell`. Every reveal path (cold reveal, visited switch, font-size apply, dispose-collapse) funnels through `showShell`, so a single call covers the whole class.
- Thin exported `reconcileShellView(sid)` wrapper called from the `usePtyAttachShell` ResizeObserver path to cover width-only resizes that don't change buffer length.
- The visited branch's hand-poke (`fit()`/`resize()`/`focus()`) is **left alone** — no per-path band-aid; it inherits the fix transparently through `showShell`.

## Out of scope

- Not killing `display:none` (Direction A — rejected; blast radius is Windows TSF IME anchoring + sidebar `overflow:clip`).
- No forced re-sync added to the visited branch.

## Files changed

- `src/terminal/shellRegistry.ts` — `reconcileView` + `reconcileShellView` + `showShell` call.
- `src/terminal/usePtyAttachShell.ts` — `reconcileShellView` in the ResizeObserver apply path.
- `tests/terminal/shellRegistry.test.ts` — mock `_core.viewport.syncScrollArea` + 4 #82 regression tests.

- lint: PASS (exit 0, --max-warnings 0)
- typecheck: PASS (exit 0, tsc renderer + electron)
- tests (changed): `tests/terminal/shellRegistry.test.ts` -> **23 passed (23)**, 1.26s
- full suite: 2024 passed / 23 failed / 1 todo. **The 23 failures are entirely in `electron/__tests__/db.test.ts` + `db-hardening.test.ts`, all `better-sqlite3` native ABI mismatch** (`NODE_MODULE_VERSION 137` vs Node requires 145, needs `npm rebuild`). Pre-existing env issue, unrelated to this renderer-only change.

## Reverse-verify

- pull the fix (comment out the `reconcileView` calls) -> run `tests/terminal/shellRegistry.test.ts` -> **FAIL: 4 failed | 19 passed (23)** (the #82 reveal/cold/force-flag/wrapper tests go red).
- restore the fix -> run -> **PASS: 23 passed (23)**.

## Merge gate (NOT satisfied by this PR — maintainer must run)

Unit tests prove the `syncScrollArea(true)` call fires on every reveal, but **cannot** prove webkit re-paints the thumb (jsdom has no real `scrollTop` physics). The real gate is a **manual real-device A->B->A re-repro** with `scrollSyncMonitor` on, confirming the DESYNC stream is gone (Delta -> 0, heartbeats only) in `userData/logs/main.log`. This PR claims **code + unit tests done; real-device gate pending** — it does NOT claim the bug is verified-fixed.

Co-Authored-By: Claude Opus 4 <noreply@anthropic.com>

### DIFF
--- a/src/terminal/shellRegistry.ts
+++ b/src/terminal/shellRegistry.ts
@@ -58,6 +58,50 @@ function installUnloadCleanupOnce(): void {
   window.addEventListener('beforeunload', beforeUnloadHandler);
 }
 
+// Force xterm to rewrite `.xterm-viewport.scrollTop` back into agreement
+// with its internal `viewportY` after a visibility/layout change that webkit
+// silently zeroed (#82, real-device confirmed).
+//
+// WHY this is needed: `showShell` reveals a resident terminal by flipping
+// `wrapper.style.display: 'none' → ''`. On that reveal webkit zeroes
+// `.xterm-viewport.scrollTop` to 0 and fires NO scroll event. xterm's
+// `Viewport.syncScrollArea(force=false)` guard then short-circuits (buffer
+// length, canvas height, cell height all unchanged and `_lastScrollTop`
+// still equals the old `ydisp*rowHeight`), so xterm never rewrites scrollTop.
+// The native scrollbar thumb sits at top while CLI content stays at bottom.
+// `syncScrollArea(true)` is the ONLY path that reaches `Viewport._refresh(true)`
+// → `_innerRefresh()` → `scrollTop = ydisp*rowHeight`; the public
+// `scrollToBottom()`/`scrollLines(0)` all short-circuit at bottom (the desync
+// fires precisely when content is already at bottom), so they are not usable
+// here.
+//
+// Reaches into the private `term._core.viewport`. If a future xterm bump
+// renames it, the try/catch degrades to a no-op and the unit test goes red —
+// caught before ship, not in the wild.
+function reconcileView(shell: Shell): void {
+  try {
+    const core = (
+      shell.term as unknown as {
+        _core?: { viewport?: { syncScrollArea?: (force: boolean) => void } };
+      }
+    )._core;
+    core?.viewport?.syncScrollArea?.(true);
+  } catch {
+    /* private-API best-effort */
+  }
+}
+
+/**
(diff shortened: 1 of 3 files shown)


Jiahui-Gu:
Closing: this PR's `reconcileView` forces `term._core.viewport.syncScrollArea(true)` to rewrite `.xterm-viewport.scrollTop`, but #1442 (already on main) removed the native scrollbar — that element is CSS-hidden and non-load-bearing. The visible thumb is now a pure projection of `buffer.active.viewportY`/`baseY` (`computeThumb`), so this writes to a dead element and cannot fix the #82 wheel symptom. Replaced by #1470, which subscribes the self-drawn-thumb hooks to `term.onRender` (the event that actually fires on a wheel scroll).

PR #1477, opened 2026-06-05
fix(voice): configurable STT language (default zh on Chinese locale) so large-v3-turbo recognizes Chinese

Jiahui-Gu:
## Summary

Voice STT on the `large-v3-turbo` model transcribed Chinese speech as English garbage. Root cause: `electron/voice/whisperCli.ts` hardcoded `-l auto`, and whisper's auto-detect misfires on short/noisy Chinese mic audio. This makes the transcription language a **user preference** (`auto | zh | en`), defaulting by UI locale (Chinese UI -> `zh`, otherwise `auto`), so Chinese users get working STT out of the box with **no English regression**.

- New `electron/voice/voiceLanguages.ts` (catalog + `isVoiceLanguage` + `defaultVoiceLanguage`) and `electron/prefs/voiceLanguage.ts` (cached app_state pref, mirrors `voiceTier.ts`; bus-invalidated, wired in `main.ts` before `registerDbIpc`).
- `language` threaded through `WhisperCliArgs`; `whisper-cli` now spawns `-l <language>` instead of `-l auto`. `transcriber.ts` reads `loadVoiceLanguage()`; `warmup.ts` stays decoupled with `language: 'auto'` (it only transcribes silence to fault pages into cache).
- Settings: new language selector in `VoicePane.tsx`, persisted via `saveState('voiceLanguage', ...)`; i18n strings added to **both** en and zh locales. Renderer mirrors the type in `src/global.d.ts` (no `electron/` import).

## Dogfood evidence (real `whisper-cli.exe` run, same clip)

Reproduced before fixing, on the actual binary + `ggml-large-v3-turbo.bin`, with one real Chinese mic clip (`zh_short2_mic.wav`, spoken "停止"):

| `-l` flag | whisper output | result |
|-----------|----------------|--------|
| `auto` (old hardcoded) | detected `en` (p=0.36) -> `- Kim jih.` | English garbage (the reported bug) |
| `zh` (new, default on Chinese locale) | `金子` | correct Chinese |

Same audio, same model, same binary — only the `-l` flag differs. This confirms the model is fully Chinese-capable and the hardcoded `-l auto` was the defect.

## Gates (all green locally)

- `npm run typecheck` — pass (no errors)
- `npm run lint` (`--max-warnings 0`) — pass
- `npm test` — new/updated voice tests pass: `VoicePane.test.tsx` 11/11, `whisperCli.test.ts` + `transcriber.test.ts` + `voiceLanguage.test.ts` 20/20
- `npm run build` + `node scripts/harness-ui.mjs` — build OK, harness 14/14 passed

## Test plan

- [ ] On a Chinese-locale install, open Settings -> Voice; language defaults to 中文
- [ ] Dictate Chinese on `large-v3-turbo`; transcription is Chinese (not English)
- [ ] Switch language to English; English dictation still works (no regression)
- [ ] `auto` still language-detects for users who switch languages mid-session

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -85,6 +85,7 @@ import { loadNotifyEnabled, subscribeNotifyEnabledInvalidation } from './prefs/n
 import { subscribeCrashReportingInvalidation } from './prefs/crashReporting';
 import { subscribeScrollbackInvalidation } from './prefs/scrollback';
 import { subscribeVoiceTierInvalidation } from './prefs/voiceTier';
+import { subscribeVoiceLanguageInvalidation } from './prefs/voiceLanguage';
 import { BadgeController } from './badgeController';
 import { registerDbIpc } from './ipc/dbIpc';
 import { registerSystemIpc } from './ipc/systemIpc';
@@ -239,6 +240,7 @@ app.whenReady().then(() => {
   subscribeNotifyEnabledInvalidation();
   subscribeScrollbackInvalidation();
   subscribeVoiceTierInvalidation();
+  subscribeVoiceLanguageInvalidation();
   // Order is significant for systemIpc: it seeds the active i18n language
   // from the OS locale, so any subsequent producer that calls i18n.t()
   // sees the correct active language.
--- a/electron/prefs/__tests__/voiceLanguage.test.ts
+++ b/electron/prefs/__tests__/voiceLanguage.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// voiceLanguage reads app_state via ../db, derives its default from the active
+// UI language via ../i18n#getMainLanguage, and subscribes to the
+// stateSavedBus. Mock all three so the test stays in-memory and deterministic.
+const loadState = vi.fn();
+let mainLanguage = 'en';
+let savedBusHandler: ((key: string) => void) | undefined;
+const onStateSaved = vi.fn((h: (key: string) => void) => {
+  savedBusHandler = h;
+  return () => {
+    savedBusHandler = undefined;
+  };
+});
+
+vi.mock('../../db', () => ({ loadState: (...a: unknown[]) => loadState(...a) }));
+vi.mock('../../i18n', () => ({ getMainLanguage: () => mainLanguage }));
+vi.mock('../../shared/stateSavedBus', () => ({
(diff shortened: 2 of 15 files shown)


Jiahui-Gu:
**REQUEST_CHANGES** (reviewer) — this PR breaks an existing e2e harness on all 3 OSes (CI red). Do not merge until fixed.

## What failed

`e2e` failed identically on macos / ubuntu / windows:

```
[HARNESS=voice-download] <<< FAIL voice-model-download (492ms):
locator.waitFor: Error: strict mode violation:
locator('[data-voice-pane] [role="radiogroup"]') resolved to 2 elements
```

## Root cause (this PR caused it)

`scripts/harness-e2e-voice-download.mjs:161` locates the tier list with:

```js
const radiogroup = win.locator('[data-voice-pane] [role="radiogroup"]');
await radiogroup.waitFor({ state: 'visible', timeout: 10_000 });
```

Before this PR there was exactly one `role="radiogroup"` inside `[data-voice-pane]` (the tier list). This PR adds a **second** `role="radiogroup"` — the new language selector in `VoicePane.tsx` — so Playwright strict mode now matches 2 elements and throws.

This is the same duplicate-role collision you correctly handled for the **unit** test (you scoped `tierRadios()` via `.closest('[data-voice-language]')`). The analogous fix was just missed for the **Playwright** harness, which targets the pane-level `[role="radiogroup"]`. `harness-e2e-voice-download` is not part of `harness-ui` (the harness you ran locally), which is why it slipped through — it only runs in the `e2e` CI job.

To be clear: this is NOT one of the two failures disclaimed in the PR body (`electron-load-smoke` / `store.test.ts`). It is a real regression introduced by this change.

## Fix (your call on the exact selector)

Scope the harness locator to the tier group so it no longer matches the language group — e.g. target the tier radiogroup by its `aria-label="Model"`, or add/reuse a tier-specific data attribute on the tier radiogroup. Then run `node scripts/harness-e2e-voice-download.mjs` (or `npm run probe:e2e`) locally to confirm green before pushing.

The rest of the change looks good — direction is right (configurable STT language, locale-defaulted, faithfully mirrors the `voiceTier.ts` pref pattern, renderer/electron boundary respected via type-mirroring, dogfood evidence solid, i18n in both en+zh, no test-skip patterns). Once the harness locator is scoped and `e2e` goes green I'll merge.

